PR #24, opened 2013-05-09
Support Windows earlier than Vista

orgads:
GetFinalPathNameByHandle was introduced in Vista

Since p can now only be NULL or ret, no need to
strdup it


jrosdahl:
Closing this due to inactivity. Please reopen if wanted.

PR #42, opened 2014-12-02
Support syntax CCACHE_COMPILERCHECK=hash:<random_hash_string>

jjgod:
This can be faster than running a command directly, because the command to run might just produce the same result for every ccache invocation. By reading it from config file or environment variable we can just run that command once and store the result in the string following `"hash:"` and reusing it.

For example when we want to use clang revision for clang built from SVN trunk, we can just call `clang --version` once and store the revision number like:

    export COMPILERCHECK=hash:218707

(Assuming 218707 is the current build revision.)

### DIFF
--- a/MANUAL.txt
+++ b/MANUAL.txt
@@ -281,6 +281,9 @@ setting key. Boolean options are indicated with ``[boolean]''
     compiler's source has not changed, or if the compiler only has changes that
     don't affect code generation). You should only use the *none* setting if
     you know what you are doing.
+*string:value*::
+    Use *value* as the string to calculate hash from. This can be the compiler
+    revision number you retrieved earlier and set here via environment variable.
 _a command string_::
     Hash the standard output and standard error output of the specified
     command. The string will be split on whitespace to find out the command and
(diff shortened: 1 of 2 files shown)


jrosdahl:
Any progress on updating the pull request?

jjgod:
Sorry, I totally forgot about this after the vacation, I will do this right away after getting back to work tomorrow.

jjgod:
Updated, please see the above commit, 
https://github.com/jjgod/ccache/commit/c1ae899f2355039beb0490a6dde3469422f07ab4

jrosdahl:
Looks good. Some minor things:

1. I suggest changing "random_hash_string" to "value" in the documentation.
2. You may want to correct your commit message to say "CCACHE_COMPILERCHECK" instead of "COMPILERCHECK" and "string:" instead of "hash:".

jrosdahl:
jjgod: Any progress on this pull request? Do you want me to perform the small suggestions I had?

jjgod:
Sorry about the delay, it's updated now.

jrosdahl:
Thanks, applied. I also wrote a test case for the new functionality.

PR #43, opened 2014-12-03
[RFC PATCH] Added hash abstractions

Nefelim4ag:
Added hash abstraction for simplify at compile time change of hash algorithm

created struct hstate; for allow in future change hash algorithm by config.
All functions just allocate enough memory for handle any hash's state what ccache supported;

Signed-off-by: Timofey Titovets <nefelim4ag@gmail.com>

### DIFF
--- a/ccache.c
+++ b/ccache.c
@@ -388,12 +388,12 @@ get_path_in_cache(const char *name, const char *suffix)
  * also updated. Takes over ownership of path.
  */
 static void
-remember_include_file(char *path, struct mdfour *cpp_hash)
+remember_include_file(char *path, struct hstate *cpp_hash)
 {
 #ifdef _WIN32
 	DWORD attributes;
 #endif
-	struct mdfour fhash;
+	struct hstate fhash;
 	struct stat st;
 	char *source = NULL;
 	size_t size;
@@ -459,7 +459,7 @@ remember_include_file(char *path, struct mdfour *cpp_hash)
 			goto failure;
 		}
 		hash_result_as_bytes(&fhash, pch_hash.hash);
-		pch_hash.size = fhash.totalN;
+		pch_hash.size = hash_get_len(&fhash);
 		hash_delimiter(cpp_hash, "pch_hash");
 		hash_buffer(cpp_hash, pch_hash.hash, sizeof(pch_hash.hash));
 	}
@@ -487,7 +487,7 @@ remember_include_file(char *path, struct mdfour *cpp_hash)
 
 		h = x_malloc(sizeof(*h));
 		hash_result_as_bytes(&fhash, h->hash);
-		h->size = fhash.totalN;
+		h->size = hash_get_len(&fhash);
 		hashtable_insert(included_files, path, h);
 	} else {
 		free(path);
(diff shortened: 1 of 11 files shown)


jrosdahl:
Hi,

I think selection of the hash algorithm (the cryptographical one; the one used for hash tables does not need to be switchable IMO) should be done in runtime, not compile time. I don't think that the proposed change will make it easier do implement runtime switching. What do you think?

Nefelim4ag:
(Oh, i've check my indents, i will fix it if we decide what this patch set really needed, ugly patchset sorry)

Okay, this is my point of view and  i can miss something, fix me if i'm wrong.

1. Compile time switching not very usable (I don't want allow maintainers to decide which algorithm i must use and also it add problem with sharing ccache).
I think what performance overhead is not noticeable in runtime (startup) switching case
/* But it must be benchmarked for easy decide how big "not noticeable"... */

2. Okay, for prevention confusing files with same names, but generated by different hash algorithm, i think that just adding hash name, to file name, something like:
md4-017caffaaaed838f9620cefb786d4c-506638.d
Is enough.

From this point:
Switch in runtime is a just "startup time" switching, ie:
ccache process started -> read config -> get hash name
After passing this hash name to if-else-if code (or switch case) in hashing callbacks.
Like:
hash_result_as_bytes(struct hstate *hstate, void *out)
{
    if (hashname == MD4) // or (long long unsigned)"md4" {
        struct mdfour *md = (struct mdfour *) hstate;
        hash_buffer(hstate, NULL, 0);
        mdfour_result(md, out);
    } else if (hashname == CITYHASH ) {
        ....
    }
}

Hm, really, yes, switchable hash algorithm for hash table are unnecessary =\

jrosdahl:
> Okay, for prevention confusing files with same names, but generated by different hash algorithm, i think that just adding hash name, to file name, something like: md4-017caffaaaed838f9620cefb786d4c-506638.d Is enough.

I don't think that including hashname in filenames would be necessary or very useful.

> After passing this hash name to if-else-if code (or switch case) in hashing callbacks.

I would instead suggest a design like this (just a sketch):

    struct hash_algorithm {
      void* (*hash_start)(void);
      void (*hash_start)(void* state);
      void (*hash_buffer)(void* state, const void *s, size_t len);
      ...
    };

    static struct hash_algorithm md4 = { md4_hash_start, md4_hash_buffer, ... };

    /* Initialization: */
    struct hash_algorithm* chosen_algorithm;
    if (str_eq(conf->hash_algorithm, "md4")) {
      chosen_algorithm = &md5;
    } else if (...) {
      ...
    }

    /* Then in the rest of the program: */
    void* state = chosen_algorithm->hash_start();
    chosen_algorithm->hash_buffer(state, buffer, sizeof(buffer));
    ...

Nefelim4ag:
Hm.. looks interesting and new for me, i will try to implement it, like you suggest

jrosdahl:
Any news on this pull request?

itensionanders:
I don't think there's any need to change hash algorithm, since MD4 hits a sweet spot between risk and speed. Did some further experiments with a "native" (ehrm) version of it by using openssl in ffe8809.

But the speedup was marginal at best, even though the included algorithm is "no attempt to be fast". So there is no imminent/real need to allow for different hash algorithms - or even their implementations ?

jrosdahl:
Closing due to inactivity. I also agree with @itensionanders that the motivation for supporting several algorithms isn't very high.

PR #46, opened 2015-02-07
Fix the error list in Xcode: don't delete clang diagnostics after compiler failures

ryb-ableton:
This branch fixes a bug that prevented Xcode from displaying errors and warnings as separate entries in the issue navigator. Clang's serialized diagnostics were accidentally deleted after compilation failures, causing Xcode to fall back on showing unparsed compiler output.


jrosdahl:
Applied, thanks!

PR #47, opened 2015-03-19
Added support for CUDA nvcc compiler

seeraven:
The CUDA nvcc compiler does not work well when feeding it with preprocessed input. The preprocessor output works well for generating the hash. But when a compiler run is necessary, the nvcc compiler must process the input files again. This is accomplished by setting the run_second_cpp flag to true.
In addition, support for the -optf/--options-file was added, that allows to specify additional compiler arguments (e.g., defines) read from a file. Obviously, these must be included in the hash as well.


jrosdahl:
Thanks, applied on master (with a few minor changes).

PR #51, opened 2015-09-07
Minor improvements to Windows support

petrochenkov:
The main motivation here is ability to build LLVM in MSYS2 with ccache.
See for example https://github.com/rust-lang/rust/issues/26556


jrosdahl:
Thanks, applied.

PR #53, opened 2015-10-30
Pass -stdlib= only to preprocessor

mattkretz:
Clang warns about an unused -stdlib=libc++ argument if the argument is
passed to the second clang invocation. The -stdlib= argument is only
necessary on preprocessing (to set the system include paths) and on
linking (which is not cached anyway)

Signed-off-by: Matthias Kretz <kretz@kde.org>


jrosdahl:
Applied, thanks.

PR #55, opened 2015-11-27
Add initial travis config

misery:
Hi there!

I added an initial configuration for travis-ci. I disabled tests for non-gcc compilers as they fail at the moment.
I think mingw32 should be added, too. Maybe appveyor.com and coveralls.io could be added. ;-)

OpenSSL uses mingw32:
https://github.com/openssl/openssl/blob/master/.travis.yml

You need to create a free travis account and enable ccache repository on your profile page: https://travis-ci.org/profile/

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,34 @@
+language: c
+
+addons:
+    apt:
+        packages:
+            - zlib1g-dev
+
+os:
+    - linux
+    - osx
+
+compiler:
+    - clang
+    - gcc
+
+matrix:
+    include:
+        - os: linux
+          compiler: clang
+          env: CFLAGS="-fsanitize=address -g" ASAN_OPTIONS="detect_leaks=0"
+    exclude:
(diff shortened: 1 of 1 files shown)


TheOneRing:
As ccache can only be compiled using msys on windows, I think the best would be to cross compile ccache. As travis runs ubuntu this shouldn't be that hard.

White-Tiger:
I'm fine either way... Travis CI cross-compile, AppVeyor with MSYS or simply both.
I just hope we'll get a working ccache on Windows someday... *(maybe even pre-compiled binaries :P?)*

When I tried to compile ccache with MSYS2, I had to revert 737c96f and 4fdb958 and had to manually undef `HAVE_MKSTEMP` *(it compiled with it, but couldn't create temp files. Well.. they were just temporarily which means that the file got deleted once the last file descriptor was closed IIRC)*
So a MinGW-w64 Travis CI setup should help to find those issues

PR #56, opened 2015-11-30
Allow not including system headers in manifest

itensionanders:
If CCACHE_SLOPPINESS includes "no_system_headers", ccache will not add any
system headers to the manifest's list of include files to always check.
This allows ccache to only check non-system headers, but will also cause
it to return stale cache hits if such system headers have been changed.


jrosdahl:
Applied, thanks!

PR #57, opened 2015-11-30
Add feature to ignore headers in manifest

itensionanders:
Certain headers are never changed, or perhaps not so important (timestamps).
Avoid repeatedly hashing these in the manifest, to make direct mode faster.
One can either use the full path to the include file, or perhaps a directory
of headers to ignore all the headers under that directory (e.g. compiler)


jrosdahl:
```
if (strncmp(path, ignore, ignore_len) == 0) {
	goto ignore;
}
```

So `ignore_headers_in_manifest` specifies a list of prefixes, and header file paths that match any of the prefixes are ignored? This means that the ignore entry "foo" will match "foo/bar.h" like intended but also "foo.h".

Would it make sense to change the logic so that a header file is ignored if its path matches an ignore entry *exactly* or if an ignore entry matches a *directory* in the path?

I.e., something like

```
if (strncmp(path, ignore, ignore_len) == 0
    && (ignore[ignore_len-1] == '/'
        || path[ignore_len] == '/'
        || path[ignore_len] == '\0')) {
	goto ignore;
}
```

?

itensionanders:
Yes, that makes sense. I guess it would be mostly used to match directories anyway ?
Ultimately one could go overboard with globs or regexps, but that seems overkill now.

jrosdahl:
OK, merged and tweaked. Thanks!

PR #30, opened 2013-06-20
Memcached ccache

tardyp:
This is an attempt at adding memcached backend for ccache.

The goal is to optimize cache hits in a big build cluster, without having to rely on single point of nfs failure.

I still need to make real life tests beside unit tests. This PR is for early feedback.

PR #58, opened 2015-12-07
Memcached

afbjorklund:
### Support for memcached as a distributed cache for ccache

This is the rebased and squashed version of memcached-ccache, memcached-review and memcached-only. It adds two new options, memcached_conf and memcached_only (and a read_only_memcached).

Based on tardyp:memcached-ccache code from 2013 (Pull Request #30)

See http://memcached.org for the cache server (you will also need libmemcached)
Some getting started instructions can be found here: [GETTING_STARTED.md](https://gist.github.com/afbjorklund/2bd0056027f8308d4cd8)

Update: The above instructions will install everything but on localhost only.
Here is how to continue after that, by using [MULTIPLE_SERVERS.md](https://gist.github.com/afbjorklund/164ee5a653a6040f888fb332ddac6ec0)


jrosdahl:
Looks promising from a cursory glance!

I've started looking at the patch series now but I can't promise to be very responsive during christmas...

> [PATCH 01/16] Add function lost in merge (imerge bug?)
> 
> Seems like the merge in f49770a4b142e82a3cf1f5e1da158f69e0447972
> was faulty, in that it dropped the safe_create_wronly function ?

I don't remember, of course, but my guess is no. I think that I simply inlined the function since it only was used in one place and because the master branch at the time had started using `mkstemp` instead of relying on the home-grown `tmp_string()` name being unique. (The major point of git-imerge is to carefully take things step by step, so I doubt that I trusted git-imerge to do something without verifying the result.)

> [PATCH 02/16] Add function to write to a file

I think that `safe_write` needs to be even safer to earn its name. :-) More specifically, it should handle EINTR (and EAGAIN while at it) like `copy_fd` does.

On the other hand:

> [PATCH 04/16] Add memcached support, as a secondary cache

Using `write_file` to populate the ccache from memcached won't fly for several reasons:

* If the system crashes in the middle of a copy, the cached file will likely be corrupt.
* The files won't be compressed even if compression is requested in the ccache configuration.
* Cache statistics won't get updated properly.

I think that we need a function that more or less does what `put_file_in_cache` already does but with a data buffer as the input.

> \+/* blob format for storing:
> \+
> \+    char magic[4]; # 'CCH1', might change for other version of ccache
> \+ # ccache will erase the blob in memcached if wrong magic

Have you considered adding the magic to the key (as part of the string or as part of the hash) instead of the value? Then it wouldn't be necessary to verify/prune results with the wrong format. ccache does similar things for the object hash already; see for instance the first `if` statement in `calculate_object_hash`.

> +memcached_suite() {

Since memcached support is opt-in, the memcached part of the test suite also needs to be opt-in in some way.

itensionanders:
Thanks for reviewing. Will take a look at the write functions, and rebase + squash the late bug fixes...

When it comes to magic, I think that was only for the memcached storage. It does use the main hash.
Currently we have three types of memcached entries, so it can be useful to verify that they are "sane".
We probably already need a second version of it, to be able to include cov and dwo entries in the cache. But they are supposed to be compressed before uploading, so copying on download should be OK.

itensionanders:
Did the rebase to "master" and the cleanup with regards to squashing bug fixes and commit messages.
The comments about write_file/put_file_in_cache and memcached_suite still needs addressing, though.

itensionanders:
Have rebased to master, and addressed the cppcheck style issues.
Also made the testsuites opt-in, as in Makefile created by configure.

itensionanders:
@jrosdahl, all issues are addressed. Added a "put_data_in_cache".
Made sure to update stats, and made write_file use a temporary file.

jrosdahl:
Looks good. Thanks! I have pushed the commits to a new `dev/memcached` branch along with some tweaks.

I think that it would be a good idea to ask on the mailing list for other testers/comments before merging it to trunk.

itensionanders:
Sounds good! I wrote some "getting started" instructions (above), how to build it.
What we need now is some better benchmarks to compare local/remote cache...

Maybe use the same code as in https://ccache.samba.org/performance.html,
or possibly some other similar newer code, on a handful of configured servers ?

itensionanders:
Added your commits here too, just so that we all test the same code.

itensionanders:
@jrosdahl: added some fixes needed for large objects (>1M) on multiple servers (>1)
Maybe you can add that to the "dev/memcached branch" as well, for consistency ?

jrosdahl:
> @jrosdahl: added some fixes needed for large objects (>1M) on multiple servers (>1)
> Maybe you can add that to the "dev/memcached branch" as well, for consistency ?

Done!

itensionanders:
@jrosdahl: I found a problem with the previous fix. Turned out that I was comparing values with strcmp, but they were (not guaranteed to be) NUL-terminated... Have switched to memcmp, now it works OK.

jrosdahl:
OK, pushed to dev/memcached.

itensionanders:
@jrosdahl: updated the branch here (itensionanders/memcached), and it has some bug fixes too

Can we get those to the dev/memcached branch as well, please ? Probably cherry-pick/rebase ?

gjasny:
On RHEL6 with locally built libmemcached-1.0.18 I have to add ``-pthread`` to LDFLAGS because of those symbols not found otherwise:

```
nm ~/sw/lib/libmemcached.so|grep pthread                     
                 U pthread_mutex_lock@@GLIBC_2.2.5
                 U pthread_mutex_unlock@@GLIBC_2.2.5
                 U pthread_once
```

jrosdahl:
> @jrosdahl: updated the branch here (itensionanders/memcached), and it has some bug fixes too
>
> Can we get those to the dev/memcached branch as well, please ? Probably cherry-pick/rebase ?

Sure, merged and pushed now.

gjasny:
I observe some behavior I cannot explain. This are my statistics without memcached:

```
cold 156.15s user 24.13s system 717% cpu 25.131 total
warm 26.83s user 7.72s system 322% cpu 10.705 total
```
These are the one with memcached_only:
```
cold 126.68s user 20.84s system 607% cpu 24.291 total
warm 127.32s user 21.12s system 595% cpu 24.932 total
``` 
This is my configuration:
```
ccache -p
(default) base_dir = 
(environment) cache_dir = /uhome_old/ccache
(default) cache_dir_levels = 2
(default) compiler = 
(default) compiler_check = mtime
(default) compression = false
(default) compression_level = 6
(default) cpp_extension = 
(default) direct_mode = true
(default) disable = false
(default) extra_files_to_hash = 
(default) hard_link = false
(environment) hash_dir = true
(default) ignore_headers_in_manifest = 
(default) keep_comments_cpp = false
(environment) log_file = /uhome/gjasny/tmp/ccache.log
(default) max_files = 0
(/uhome_old/ccache/ccache.conf) max_size = 214.7G
(environment) memcached_conf = --SERVER=docker02.drs.expertcity.com
(environment) memcached_only = true
(default) path = 
(default) prefix_command = 
(default) prefix_command_cpp = 
(default) read_only = false
(default) read_only_direct = false
(default) read_only_memcached = false
(default) recache = false
(default) run_second_cpp = false
(environment) sloppiness = file_macro, time_macros
(default) stats = true
(default) temporary_dir = 
(environment) umask = 002
(default) unify = false
```
This is the memcached config before warm run:
```
STAT cmd_get 2196
STAT cmd_set 1682
STAT cmd_flush 0
STAT cmd_touch 0
STAT get_hits 815
STAT get_misses 1381

STAT curr_items 185
STAT total_items 1682
STAT expired_unfetched 0
STAT evicted_unfetched 805
STAT evictions 885
```
This is after the warm run:
```
STAT cmd_get 2328
STAT cmd_set 1782
STAT cmd_flush 0
STAT cmd_touch 0
STAT get_hits 885
STAT get_misses 1443

STAT curr_items 184
STAT total_items 1782
STAT expired_unfetched 0
STAT evicted_unfetched 857
STAT evictions 941
```

Is the memcached functionality just a a storage backend or does it affect caching and/or sloppyness.
Maybe you could add some more statistics to shed some light on this.

PS: I also have a CCACHE_LOG file but would like to share it only privately.

Thanks,
Gregor

gjasny:
Those are the logs for one file for two subsequent runs:
```
[2016-04-17T12:58:02.975502 30481] Command line: ccache /opt/rh/devtoolset-4/root/usr/bin/g++ -I../3rdParty/GoogleMock/include -I../3rdParty/GoogleMock -I../3rdParty/GoogleMock/gtest/include -I../3rdParty/GoogleMock/gtest -Wall -Werror -fdiagnostics-show-option -pthread -Wformat=2 -fstack-protector -g -ggdb -D_DEBUG -O0 -Wall -Werror -fdiagnostics-show-option -pthread -Wformat=2 -fstack-protector -Wall -Wshadow -DGTEST_HAS_PTHREAD=1 -fexceptions -Wno-error -Wextra -Wno-unused-parameter -Wno-missing-field-initializers -std=c++14 -MMD -MT 3rdParty/GoogleMock/CMakeFiles/gmock.dir/src/gmock-all.cc.o -MF 3rdParty/GoogleMock/CMakeFiles/gmock.dir/src/gmock-all.cc.o.d -o 3rdParty/GoogleMock/CMakeFiles/gmock.dir/src/gmock-all.cc.o -c ../3rdParty/GoogleMock/src/gmock-all.cc
[2016-04-17T12:58:02.975546 30481] Hostname: drsbuild05.drs.expertcity.com  
[2016-04-17T12:58:02.975583 30481] Working directory: /uhome/gjasny/SuperBuild/rtp/_build64_d
[2016-04-17T12:58:02.975722 30481] Source file: ../3rdParty/GoogleMock/src/gmock-all.cc  
[2016-04-17T12:58:02.975735 30481] Dependency file: 3rdParty/GoogleMock/CMakeFiles/gmock.dir/src/gmock-all.cc.o.d
[2016-04-17T12:58:02.975744 30481] Object file: 3rdParty/GoogleMock/CMakeFiles/gmock.dir/src/gmock-all.cc.o
[2016-04-17T12:58:02.975769 30481] Trying direct lookup
[2016-04-17T12:58:02.975846 30481] Looking for object file hash in /uhome_old/ccache/4/8/a439e0d8accba9e04ce36a593f9475-3172.manifest
[2016-04-17T12:58:03.005107 30481] Got object file hash from manifest
[2016-04-17T12:58:03.005165 30481] Getting 985137d4f6787827d46cf8792c915d28-3621028 from memcached
[2016-04-17T12:58:03.007608 30481] memcached_mget aca7c94816fb5e4f9a60cb479850a5e3-1024000 1024000
[2016-04-17T12:58:03.007645 30481] memcached_mget 6c4ed4406596cd6b866e7c0e06ee3192-627248 627248
[2016-04-17T12:58:03.023899 30481] Failed to get all 2 chunks
[2016-04-17T12:58:03.023930 30481] Failed to get key from memcached 985137d4f6787827d46cf8792c915d28-3621028: SUCCESS
[2016-04-17T12:58:03.024070 30481] Running preprocessor
[2016-04-17T12:58:03.024083 30481] Executing /opt/rh/devtoolset-4/root/usr/bin/g++ -Wall -Werror -fdiagnostics-show-option -pthread -Wformat=2 -fstack-protector -g -ggdb -O0 -Wall -Werror -fdiagnostics-show-option -pthread -Wformat=2 -fstack-protector -Wall -Wshadow -fexceptions -Wno-error -Wextra -Wno-unused-parameter -Wno-missing-field-initializers -std=c++14 -I../3rdParty/GoogleMock/include -I../3rdParty/GoogleMock -I../3rdParty/GoogleMock/gtest/include -I../3rdParty/GoogleMock/gtest -D_DEBUG -DGTEST_HAS_PTHREAD=1 -MMD -MT 3rdParty/GoogleMock/CMakeFiles/gmock.dir/src/gmock-all.cc.o -MF 3rdParty/GoogleMock/CMakeFiles/gmock.dir/src/gmock-all.cc.o.d -E ../3rdParty/GoogleMock/src/gmock-all.cc
[2016-04-17T12:58:03.386028 30481] Got object file hash from preprocessor
[2016-04-17T12:58:03.386053 30481] Preprocessor created 3rdParty/GoogleMock/CMakeFiles/gmock.dir/src/gmock-all.cc.o.d
[2016-04-17T12:58:03.386079 30481] Already searched for 985137d4f6787827d46cf8792c915d28-3621028
[2016-04-17T12:58:03.386159 30481] Running real compiler
[2016-04-17T12:58:03.386170 30481] Executing /opt/rh/devtoolset-4/root/usr/bin/g++ -Wall -Werror -fdiagnostics-show-option -pthread -Wformat=2 -fstack-protector -g -ggdb -O0 -Wall -Werror -fdiagnostics-show-option -pthread -Wformat=2 -fstack-protector -Wall -Wshadow -fexceptions -Wno-error -Wextra -Wno-unused-parameter -Wno-missing-field-initializers -std=c++14 -c -o 3rdParty/GoogleMock/CMakeFiles/gmock.dir/src/gmock-all.cc.o /uhome_old/ccache/tmp/gmock-all.stdout.drsbuild05.drs.expertcity.com.30481.NEfINO.ii
[2016-04-17T12:58:06.113010 30481] Unlink /uhome_old/ccache/tmp//uhome_old/ccache/9/8/5137d4f6787827d46cf8792c915d28-3621028.o.tmp.stdout.drsbuild05.drs.expertcity.com.30481.XXXXXX.drsbuild05.drs.expertcity.com.30481.05rD4E
[2016-04-17T12:58:06.113180 30481] Unlink /uhome_old/ccache/tmp//uhome_old/ccache/9/8/5137d4f6787827d46cf8792c915d28-3621028.o.tmp.stderr.drsbuild05.drs.expertcity.com.30481.XXXXXX.drsbuild05.drs.expertcity.com.30481.Dq6ylv.2
[2016-04-17T12:58:06.113220 30481] Unlink /uhome_old/ccache/tmp//uhome_old/ccache/9/8/5137d4f6787827d46cf8792c915d28-3621028.o.tmp.stderr.drsbuild05.drs.expertcity.com.30481.XXXXXX.drsbuild05.drs.expertcity.com.30481.Dq6ylv
[2016-04-17T12:58:06.116781 30481] memcached_mset aca7c94816fb5e4f9a60cb479850a5e3-1024000 1024000
[2016-04-17T12:58:06.129085 30481] memcached_mset 6c4ed4406596cd6b866e7c0e06ee3192-627248 627248
[2016-04-17T12:58:06.134679 30481] memcached_set 985137d4f6787827d46cf8792c915d28-3621028 60 (1651248)
[2016-04-17T12:58:06.134855 30481] Storing 985137d4f6787827d46cf8792c915d28-3621028 in memcached
[2016-04-17T12:58:06.141660 30481] Added object file hash to /uhome_old/ccache/4/8/a439e0d8accba9e04ce36a593f9475-3172.manifest
[2016-04-17T12:58:06.141707 30481] Storing 48a439e0d8accba9e04ce36a593f9475-3172 in memcached
[2016-04-17T12:58:06.142855 30481] Acquired lock /uhome_old/ccache/9/stats.lock
[2016-04-17T12:58:06.142953 30481] Releasing lock /uhome_old/ccache/9/stats.lock
[2016-04-17T12:58:06.142962 30481] Unlink /uhome_old/ccache/9/stats.lock
[2016-04-17T12:58:06.142975 30481] Result: cache miss                     

subsequent run

[2016-04-17T13:09:34.608369 31398] Command line: ccache /opt/rh/devtoolset-4/root/usr/bin/g++ -I../3rdParty/GoogleMock/include -I../3rdParty/GoogleMock -I../3rdParty/GoogleMock/gtest/include -I../3rdParty/GoogleMock/gtest -Wall -Werror -fdiagnostics-show-option -pthread -Wformat=2 -fstack-protector -g -ggdb -D_DEBUG -O0 -Wall -Werror -fdiagnostics-show-option -pthread -Wformat=2 -fstack-protector -Wall -Wshadow -DGTEST_HAS_PTHREAD=1 -fexceptions -Wno-error -Wextra -Wno-unused-parameter -Wno-missing-field-initializers -std=c++14 -MMD -MT 3rdParty/GoogleMock/CMakeFiles/gmock.dir/src/gmock-all.cc.o -MF 3rdParty/GoogleMock/CMakeFiles/gmock.dir/src/gmock-all.cc.o.d -o 3rdParty/GoogleMock/CMakeFiles/gmock.dir/src/gmock-all.cc.o -c ../3rdParty/GoogleMock/src/gmock-all.cc
[2016-04-17T13:09:34.608409 31398] Hostname: drsbuild05.drs.expertcity.com
[2016-04-17T13:09:34.608451 31398] Working directory: /uhome/gjasny/SuperBuild/rtp/_build64_d
[2016-04-17T13:09:34.608601 31398] Source file: ../3rdParty/GoogleMock/src/gmock-all.cc
[2016-04-17T13:09:34.608615 31398] Dependency file: 3rdParty/GoogleMock/CMakeFiles/gmock.dir/src/gmock-all.cc.o.d
[2016-04-17T13:09:34.608628 31398] Object file: 3rdParty/GoogleMock/CMakeFiles/gmock.dir/src/gmock-all.cc.o
[2016-04-17T13:09:34.608655 31398] Trying direct lookup
[2016-04-17T13:09:34.608726 31398] Looking for object file hash in /uhome_old/ccache/4/8/a439e0d8accba9e04ce36a593f9475-3172.manifest
[2016-04-17T13:09:34.638049 31398] Got object file hash from manifest
[2016-04-17T13:09:34.638090 31398] Getting 985137d4f6787827d46cf8792c915d28-3621028 from memcached
[2016-04-17T13:09:34.640241 31398] memcached_mget aca7c94816fb5e4f9a60cb479850a5e3-1024000 1024000
[2016-04-17T13:09:34.640284 31398] memcached_mget 6c4ed4406596cd6b866e7c0e06ee3192-627248 627248
[2016-04-17T13:09:34.640562 31398] Failed to get key in memcached: NOT FOUND
[2016-04-17T13:09:34.640594 31398] Failed to get key from memcached 985137d4f6787827d46cf8792c915d28-3621028: SUCCESS
[2016-04-17T13:09:34.640783 31398] Running preprocessor
[2016-04-17T13:09:34.640797 31398] Executing /opt/rh/devtoolset-4/root/usr/bin/g++ -Wall -Werror -fdiagnostics-show-option -pthread -Wformat=2 -fstack-protector -g -ggdb -O0 -Wall -Werror -fdiagnostics-show-option -pthread -Wformat=2 -fstack-protector -Wall -Wshadow -fexceptions -Wno-error -Wextra -Wno-unused-parameter -Wno-missing-field-initializers -std=c++14 -I../3rdParty/GoogleMock/include -I../3rdParty/GoogleMock -I../3rdParty/GoogleMock/gtest/include -I../3rdParty/GoogleMock/gtest -D_DEBUG -DGTEST_HAS_PTHREAD=1 -MMD -MT 3rdParty/GoogleMock/CMakeFiles/gmock.dir/src/gmock-all.cc.o -MF 3rdParty/GoogleMock/CMakeFiles/gmock.dir/src/gmock-all.cc.o.d -E ../3rdParty/GoogleMock/src/gmock-all.cc
[2016-04-17T13:09:35.030586 31398] Got object file hash from preprocessor
[2016-04-17T13:09:35.030626 31398] Preprocessor created 3rdParty/GoogleMock/CMakeFiles/gmock.dir/src/gmock-all.cc.o.d
[2016-04-17T13:09:35.030654 31398] Already searched for 985137d4f6787827d46cf8792c915d28-3621028
[2016-04-17T13:09:35.030746 31398] Running real compiler
[2016-04-17T13:09:35.030758 31398] Executing /opt/rh/devtoolset-4/root/usr/bin/g++ -Wall -Werror -fdiagnostics-show-option -pthread -Wformat=2 -fstack-protector -g -ggdb -O0 -Wall -Werror -fdiagnostics-show-option -pthread -Wformat=2 -fstack-protector -Wall -Wshadow -fexceptions -Wno-error -Wextra -Wno-unused-parameter -Wno-missing-field-initializers -std=c++14 -c -o 3rdParty/GoogleMock/CMakeFiles/gmock.dir/src/gmock-all.cc.o /uhome_old/ccache/tmp/gmock-all.stdout.drsbuild05.drs.expertcity.com.31398.OSIbWj.ii
[2016-04-17T13:09:37.575799 31398] Unlink /uhome_old/ccache/tmp//uhome_old/ccache/9/8/5137d4f6787827d46cf8792c915d28-3621028.o.tmp.stdout.drsbuild05.drs.expertcity.com.31398.XXXXXX.drsbuild05.drs.expertcity.com.31398.cF0Kwy  
[2016-04-17T13:09:37.575944 31398] Unlink /uhome_old/ccache/tmp//uhome_old/ccache/9/8/5137d4f6787827d46cf8792c915d28-3621028.o.tmp.stderr.drsbuild05.drs.expertcity.com.31398.XXXXXX.drsbuild05.drs.expertcity.com.31398.GdDk7M.2  
[2016-04-17T13:09:37.575979 31398] Unlink /uhome_old/ccache/tmp//uhome_old/ccache/9/8/5137d4f6787827d46cf8792c915d28-3621028.o.tmp.stderr.drsbuild05.drs.expertcity.com.31398.XXXXXX.drsbuild05.drs.expertcity.com.31398.GdDk7M
[2016-04-17T13:09:37.579082 31398] memcached_mset aca7c94816fb5e4f9a60cb479850a5e3-1024000 1024000
[2016-04-17T13:09:37.590789 31398] memcached_mset 6c4ed4406596cd6b866e7c0e06ee3192-627248 627248
[2016-04-17T13:09:37.596378 31398] memcached_set 985137d4f6787827d46cf8792c915d28-3621028 60 (1651248)  
[2016-04-17T13:09:37.596558 31398] Storing 985137d4f6787827d46cf8792c915d28-3621028 in memcached  
[2016-04-17T13:09:37.603474 31398] Added object file hash to /uhome_old/ccache/4/8/a439e0d8accba9e04ce36a593f9475-3172.manifest  
[2016-04-17T13:09:37.603511 31398] Storing 48a439e0d8accba9e04ce36a593f9475-3172 in memcached
[2016-04-17T13:09:37.604666 31398] Acquired lock /uhome_old/ccache/9/stats.lock
[2016-04-17T13:09:37.604760 31398] Releasing lock /uhome_old/ccache/9/stats.lock  
[2016-04-17T13:09:37.604769 31398] Unlink /uhome_old/ccache/9/stats.lock  
[2016-04-17T13:09:37.604783 31398] Result: cache miss                     
```

Will check tomorrow if memcached really has 20g of storage capacity.

gjasny:
My error was that I started with the dev/memcached branch which stored incompatible or broken elements into the memcache. I dropped the cache and switched to itensionanders/memcached. Now compilation is even faster than with a local cache.

itensionanders:
@jrosdahl: Thank you, looks like it's in perfect synch now!

@gjasny: Think that -lpthread entered in a later libmemcached...
It's not needed with the old libmemcached (1.0.8) in Ubuntu 14.04
But it does seem to be needed with the latest libmemcached-1.0.18 ?
So most likely a check is needed in configure.ac, for `AX_PTHREAD`:

```
# The name that we can dlopen(3).
dlname='libmemcached.so.11'

# Names of this library.
library_names='libmemcached.so.11.0.0 libmemcached.so.11 libmemcached.so'

# The name of the static archive.
old_library='libmemcached.a'

# Linker flags that can not go in dependency_libs.
inherited_linker_flags=' -pthread'

# Libraries that this one depends upon.
dependency_libs=' -ldl -lsasl2'

# Names of additional weak libraries provided by this library
weak_library_names=''

# Version information for libmemcached.
current=11
age=0
revision=0

```

itensionanders:
Would be nice to add a memcached target for Travis, but it only has Ubuntu 12.04 (with 0.44)
The environment for 14.04 (with 1.0.8) is still in Beta. Maybe we should build it from source ?

https://docs.travis-ci.com/user/ci-environment/

As a side note, we have SASL in e049385eff9d62b51a4d861d09a52067dda74985, if needed.

itensionanders:
@gjasny: If you want a discussion you can try the [mailing list](https://lists.samba.org/mailman/listinfo/ccache/). But there's *both* a storage solution (memcached_only), and a sharing solution (memcached_conf) - that still uses disk storage.

> Is the memcached functionality just a a storage backend or does it affect caching and/or sloppyness.
Maybe you could add some more statistics to shed some light on this.

itensionanders:
@jrosdahl : Finally added a memcached test run, using the Trusty beta Build Environment in Travis...

yoava333:
@itensionanders I tried to compile this dev/memcached on a 32bit machine, and I got format string errors on trying to print size_t. I've fixed it and here's the patch
```
diff --git a/memccached.c b/memccached.c
index b6ce09e..f23acdb 100644
--- a/memccached.c
+++ b/memccached.c
@@ -87,7 +87,7 @@ static memcached_return_t memccached_big_set(memcached_st *ptr,
        mdfour_result(&md, (unsigned char *) subkey);
        *((uint32_t *) (subkey + 16)) = htonl(n);
        s = format_hash_as_string((const unsigned char *) subkey, n);
-       cc_log("memcached_mset %s %ld", s, n);
+       cc_log("memcached_mset %s %zu", s, n);
        ret = memcached_set(ptr, s, strlen(s), value + x, n,
                            expiration, flags);
        free(s);
@@ -101,7 +101,7 @@ static memcached_return_t memccached_big_set(memcached_st *ptr,
        p += 20;
    }

-   cc_log("memcached_set %.*s %ld (%ld)", (int) key_length, key, buflen,
+   cc_log("memcached_set %.*s %zu (%zu)", (int) key_length, key, buflen,
           value_length);
    ret = memcached_set(ptr, key, key_length, buf, buflen,
                        expiration, flags);
@@ -234,7 +234,7 @@ static char *memccached_big_get(memcached_st *ptr,
            return NULL;
        }
    }
-   cc_log("memcached_get %.*s %ld (%ld)", (int) key_length, key, *value_length,
+   cc_log("memcached_get %.*s %zu (%zu)", (int) key_length, key, *value_length,
           buflen);
    for (i = 0; i < numkeys; i++) {
```

Jimilian:
Maybe somebody already has `.spec` file for RPM or RPM itself for this version of ccache? :pray:

itensionanders:
@yoava333 : added, thanks!

itensionanders:
@Jimilian : there is no such package yet, we have been building from git (see GETTING_STARTED.md).
The instructions are for Ubuntu, but could be adopted to Fedora ? Their packaging is a bit weird, though.

In that it enables itself by default, and then use CCACHE_HASHDIR to negate any CCACHE_BASEDIR:
http://pkgs.fedoraproject.org/cgit/rpms/ccache.git/tree/ccache.spec

Will see if I can't make some CentOS RPMS available too, with the upcoming multi-node instructions...

Basically I think you replace "apt-get install libmemcached-dev" with "yum install libmemcached-devel" ?

Jimilian:
@itensionanders , yes, I figured out how to replace names :) I need RPM to add it to [amazon image](http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/AMIs.html). But I'm not familiar with RPM techniques. Anyway thanks for `.spec` file from Fedora, I will try to adopt it.

Jimilian:
@itensionanders We decided to put your version of ccache in Docker instead of AMI. So, problem was solved. 

But can you clarify also one thing. If ccache doesn't have item in local storage, it goes to memcache. Does it save results locally afterwards? (I checked code, but didn't find answer)
I'm expecting something like:
`ccache -> local_storage -> memcache -> local_storage -> ccache`
And if it wants to push data, similar:
`ccache -> memcache -> local storage.`

itensionanders:
The files will be downloaded from memcached into local storage, and they will be put into local storage before uploading to memcached. There is the "memcached-only" option, to avoid the local storage...

gjasny:
Another issue I noticed was that in the memcached_only configuration
compression was not working as expected. Cannot remember the details right
now.

itensionanders:
I have a "mostly" working of compression and manifest (for direct mode). Was trying to get statistics in too (storing them in memcached instead of in "stats" files), but maybe I should push what I have instead.

jrosdahl:
I would like to release a 3.3 version relatively soon. Do you think we should aim for having memcached support in 3.3 or 3.4?

Jimilian:
@jrosdahl From my point of view, it would be awesome to have memcached support in next release. We are using ccache version based on memcache ~ 2 months and it's quite stable, we didn't have any issues.

itensionanders:
I'll take a look at updating the branch, and adding coverage support as well.

itensionanders:
Now to add manifests to memcached-only, and there's some fixes needed to the python helpers -
apparently `python-memcache` does **not** support ketama hashing, which is very unfortunate

https://github.com/mau21mau/python-ketama

This means that you wouldn't be able to find the values, if you use more than one server...
Due to the fact that we *always* set `MEMCACHED_DISTRIBUTION_CONSISTENT` in ccache

itensionanders:
@jrosdahl : I would like it included in 3.3, hopefully (being opt-in) it wouldn't break anything ?

jrosdahl:
Sure, let's aim for that.

jrosdahl:
>> @jrosdahl : I would like it included in 3.3, hopefully (being opt-in) it wouldn't break anything ?

> Sure, let's aim for that.

On the other hand, I don't see a problem with pushing memcached support to 3.4. It's a big feature enough to warrant a feature release alone. And we can release it as soon as we feel that memcached support is ready; we don't have to wait for a long time (as was the case between 3.1 and 3.2 and between 3.2 and the upcoming 3.3).

One thing I would like to see before including memcached support in a ccache release is the possibility of having memcached support as a runtime plugin. The reason for that is to improve prebuilt ccache versions in e.g. Debian, Fedora, etc – we would of course like distributions to include ccache memcached support, but it wouldn't be good to drag in memcached via dependencies from the ccache package. Solution: Put memcached support in a separate ccache-memcached package that provides the plugin.

itensionanders:
I think that the idea if having plugins is a good one, and would also come in handy for other storage formats like Couchbase or MySQL perhaps (the "memcached only" option is more intrusive than a plugin) and could be combined with the regular file storage (currently in the itensionanders/external branch)

So I think you can remove the "3.3" milestone from this feature, and we'll aim for `master` instead...

jrosdahl:
> @jrosdahl : You can remove the "3.3" milestone from this feature, and we'll aim for master instead...

OK.

afbjorklund:
On second thought, I think that adding a "plugin system" is probably *more* overhead than just linking with libmemcached. I would rather see that the memcached storage became an integral part and default, and then individuals could opt out with `--disable-memcached` rather than the opposite (we have today).

jrosdahl:
> On second thought, I think that adding a "plugin system" is probably more overhead than just linking with libmemcached.

For clarification, I'm not suggesting a generic plugin system, only dynamic loading of the memcached parts. Sure, it's more overhead if you mean more code, but it won't be much extra code.

> I would rather see that the memcached storage became an integral part and default, and then individuals could opt out with --disable-memcached rather than the opposite (we have today).

I would say that most users can't opt out with --disable-memcached since most users use a prepackaged version.

That said, I too see a value in having it as an integral part of ccache – not for code reasons but for documentation reasons.

amacneil:
What's the current status of this branch? Is there anything that could be done to help get it merged into master?

FWIW have been running the `dev/memcached` branch on our CI server for the past 2 months with no issues.

afbjorklund:
I think that this branch could be updated and then merged, now that 3.3 has branched off into maintenance mode. Had some ideas about how to update it, looks like we need to extend the cache format to support more files/extensions - as there are more and more of them added as of lately...
But at least the `.gcno` should be supported (#104), maybe also `.dwo` and possibly even `.su`.

I have changed companies, and haven't had so much time to work on this memcached feature.
But I think that both it and the LZ4 compression are nice improvements on the current ccache...
If we can get the distributed cache integrated out-of-the-box, I think ccache 4.x might be in order ?
The decision is of course up to Joel as the maintainer, will just do a rebase against the "master".

afbjorklund:
Now, to rewrite all the code and redo all the tests. This is going to take a while, it seems.

jrosdahl:
> I have changed companies, and haven't had so much time to work on this memcached feature.

And I became a father about a month ago, so I haven't had much time nor interest for ccache since then.

> If we can get the distributed cache integrated out-of-the-box, I think ccache 4.x might be in order ?

Sounds reasonable.

jrosdahl:
> Had some ideas about how to update it, looks like we need to extend the cache format to support more files/extensions - as there are more and more of them added as of lately...

Yes please. As mentioned on [the mailing list last year](http://www.mail-archive.com/ccache@lists.samba.org/msg01185.html), I don't think that adding more and more file types as files of their own in the cache makes sense. (We could make an exception for the object files if we want to keep the hard link feature.)

afbjorklund:
Originally the format stored each file as a separate entry, but then there could be network overhead for each (since they might be stored on different servers). For other formats* where the overhead wasn't key, I just went with simple "one file, one entry". Like you say, it might be a good idea to do this always.

\* e.g. MySQL
```
CREATE TABLE IF NOT EXISTS ccache
               (file VARCHAR(52) PRIMARY KEY, data LONGBLOB, time DATETIME)
               ENGINE=MyISAM
```

For the short run, I think I will just modify the format (in a new "magic" version). Instead of always having 4 suffixes in a fixed order (no matter if present or not), it would also include the actual suffix and make the list dynamic (terminated by a sentinel). That way, any new files (.gcno/.su) could be added too.

So instead of this:
```
    uint32_t obj_len; # network endian
    char obj[obj_len];
    uint32_t stderr_len; # network endian
    char stderr[stderr_len];
    uint32_t dia_len; # network endian
    char dia[dia_len];
    uint32_t dep_len; # network endian
    char dep[dep_len];
```

There would be 1-4 (or more) entries of:
```
    uint32_t suffix_len; # network endian
    char suffix[suffix_len];
    uint32_t len; # network endian
    char data[len];
```
* `.o`
* `.stderr` (if any)
* `.dia` (if any)
* `.d` (if any)

And then two zeroes, to end the list ?

----

We avoid the hardlink feature, since it you do something like strip debug symbols you might end up modifying the cache files. So using hardlinks requires a lot of discipline of (each and every) cache user.

Another reason is compression, which is probably a better feature than hard link (it also indirectly provides the files with a checksum, which is a nice feature to have). Like it says in the MANUAL:

> Another thing to keep in mind is that if the resulting object file is modified in any way, this corrupts the cached object file as well. Hard links are never made for compressed cache files.

If two files per lookup is OK (like today .o and .stderr), then maybe the main cache should consider adding the content-addressable-storage suggested in #98 instead. Again with a special case for .o ?

Jimilian:
@afbjorklund if you are going to re-work this great PR, can you add support for "fire-and-forget" mode? Or point me to place there I can inject it (I didn't find implementation of `memccached_set` function)? ;)
"fire-and-forget" should speed-up cache missing a little bit. (Some details: http://stackoverflow.com/a/38260425/868947)

afbjorklund:
@Jimilian : thank you for the tip, will look into these "quiet" operations.
We are using `moxi` (from Couchbase) for proxying, so it is of interest.

If you want to look at it yourself, then all the api-interfacing code is in:
https://github.com/ccache/ccache/blob/dev/memcached/memccached.c
And I think you are right about setq, but still need to handle get timeouts:
https://github.com/memcached/memcached/wiki/BinaryProtocolRevamped

> The get command gets a single key. The getq command will not send a response on a cache miss. 

> set: Quiet mutations only return responses on failure. Success is considered the general case and is suppressed when in quiet mode, but errors should not be allowed to go unnoticed.

Also, while we are talking about it, some of the function names are horrible.
(`raw_get` ? `new_get` ? _eww_. and the memory handling must be fixed too)

bdrewery:
This patch is needed to fix a crash in memccached_raw_set() due to update_manifest_file() getting a NULL manifest_name value.  This happens with Clang 3.8 on FreeBSD.

```
Core was generated by `/usr/local/bin/ccache cc -O2 -pipe -DSHELL -I. -I/root/git/freebsd/bin/sh -g -st'.
Program terminated with signal SIGSEGV, Segmentation fault.
#0  strlen (str=0x0) at /usr/src/lib/libc/string/strlen.c:100
100             va = (*lp - mask01);
(gdb) bt
#0  strlen (str=0x0) at /usr/src/lib/libc/string/strlen.c:100
#1  0x0000000000418f02 in memccached_raw_set (key=0x0, data=0x802332000 "\037\213\b", len=2512) at memccached.c:255
#2  0x0000000000404136 in update_manifest_file () at ccache.c:1104
#3  0x000000000040b163 in to_memcached (args=0x802221280) at ccache.c:1640
#4  0x00000000004089f5 in ccache (argc=35, argv=0x7fffffffdfb8) at ccache.c:3815
#5  0x0000000000407e5f in ccache_main (argc=35, argv=0x7fffffffdfb8) at ccache.c:3977
#6  0x0000000000403362 in main (argc=35, argv=0x7fffffffdfb8) at main.c:25
```

It looks like in db705bf3d8f7aa3d2231b44a6570b536c4e2126d `manifest_name` was converted to a global, but one local usage was missed.

```
--- ccache.c.orig	2017-03-09 09:08:38.803365000 -0800
+++ ccache.c	2017-03-09 09:08:42.059821000 -0800
@@ -2153,7 +2153,7 @@ calculate_object_hash(struct args *args,
 			conf->direct_mode = false;
 			return NULL;
 		}
-		char *manifest_name = hash_result(hash);
+		manifest_name = hash_result(hash);
 		manifest_path = get_path_in_cache(manifest_name, ".manifest");
 		/* Check if the manifest file is there. */
 		struct stat st;
```

afbjorklund:
@bdrewery : Yeah, sorry about that - I introduced the bug when "upgrading" to C99 but forgot to push the fixed branch. There were other tricky stuff like that, that got lost in the Great Comment Format diff.

afbjorklund:
Apparently I was also rewriting the cache format (to allow for new things like coverage) and fixing the python programs (put things in the wrong bucket, due to missing libketama support) at the same time.

bdrewery:
Is this the right place for feedback and questions, or the lists?

afbjorklund:
@bdrewery : Think this is mostly about the code itself, and general discussion is better on the list. But I haven't been paying much attention to either lately, so any channel is good for getting the ball rolling...

afbjorklund:
Changed our company name (again), so I think I'm going to stick with private gmail instead.
It's still by me and all still available under the GNU General Public License, no matter what.

cleeland:
Is this waiting for some sort of milestone prior to acceptance into master?

jrosdahl:
> Is this waiting for some sort of milestone prior to acceptance into master?

No, it's waiting for spare time and motivation, both of which I'm quite short of at the moment.

cleeland:
Understood!  I have motivation, so I will try to carve out some time.

afbjorklund:
@cleeland : Glad to hear! Let me know if you need any help with the code...
I still have some local branches that I need to clean up and push for review.

Even though NFS/SQL might be the easiest to deploy in large organizations,
I _still_ think that the moxi/memcached solution has the most technical merits:

* https://www.slideshare.net/northscale/moxi-memcached-proxy

* https://www.couchbase.com/memcached (Community Edition)

afbjorklund:
Might have to start over, from the `dev/memcached` branch. This one is getting unwieldy...

afbjorklund:
Updated the branch (merged with master), just to see what the differences were.
Seems like there is nothing remaining, that is not already in dev/memcached.

Should switch to that branch instead, then we also get these nice fixes too:
99ac65e Install libmemcached-tools when testing memcached support in travis-ci
f772427 test: Fix race condition when killing memcached
4e6c15b Only activate memcached_only functionality if memcached_conf is set
67fc1b4 test: Add test cases for memcached_only functionality
11dbd97 Ensure that parent directories exist for CACHEDIR.TAG creation
aa1309d Remove unnecessary MEMSTAT variable
a8e0c6f Restart memcached before each test case to keep tests isolated
02f3e28 Use str_eq instead of strlen for checking for empty string
61bfc91 test: Add note about missing clear of memcached cache
4b5478e Clean up after merge
ec1162f test: Remove too verbose printouts at test failure
a1b3fb3 Only enable memcached tests if ccache has memcached support
a02dd17 test: Note that memcached_only functionality needs testing
0c3cc28 memcached: Fix bug when getting object from memcached in from_fscache
31f91ee test: Fix and improve memstat detection and memcached shutdown
c5c73f0 test: Silence exit message from memcached
baff697 test: Enable memcached suites if memcached is available
And get rid of _some_ of the extra merges, from keeping the branch so long...

afbjorklund:
Switched to `dev/memcached` branch, and updated to `v3.4.2` (in #257)

martinm82:
@afbjorklund what do you think about adding support to a service discovery system like consul? We have successfully rolled out ccache with memcached support on our macOS/iOS build farm and the only thing that we do not like is the fact that we need to maintain on build agents a list of all active memcached servers. 
Adding/removing memcached servers from the cluster results in keeping this configuration up-to-date and one thing that we thought we could use is a service discovery like Consul.

Each memcached server would de-/register at Consul (using the key/value storage) and whenever a build would be triggered a list of all active memcached servers would be obtained before a build and used to feed ccache.

This would allow us adding and removing dynamically memcached servers without the need to use for example provisioning tools like Ansible/Salt to keep this list always up-to-date. 

We first thought we could add this part into the CI system directly and use the consul client to gather a list of servers via the key/value storage API (groups should be as well implemented so that you can differentiate between different distributed caches) but on the other side it would as well make sense to integrate that directly to ccache.

What's your opinion about this idea? We actually consider to implement this and in case of interest contribute back to the project.

jonnyyu:
@martinm82 How do you cluster memcached servers? IMO memcached server doesn't support cluster, so you need a proxy to route requests to different servers? Maybe that part should have the intelligence to discover new servers?

In our case, we use Couchbase to create a memcached cluster. And CouchBase has a memcached proxy server names 'moxi'. You can deploy it to each build machine to let it behave as a memcached server. moxi will detect the new server when new servers added to Couchbase cluster. 

See https://www.slideshare.net/northscale/moxi-memcached-proxy for more details about moxi.

afbjorklund:
I second the use of "moxi", and have updated [the example](https://gist.github.com/afbjorklund/164ee5a653a6040f888fb332ddac6ec0) to use it. Installed the old version (2.5.0), since the moxi-server package is sadly no longer available as stand-alone and the couchbase-server refuses to install properly (and is 10x bigger). The moxi installation _can_ be made to read from a central configuration, which makes it easier to roll out changes - but the example just uses a simple 11311 -> 11211 local proxy setting. By keeping a local server, we can also avoid some of the network/connection overhead from starting and stopping a lot of processes (i.e. like ccache is doing)

```
Usage:
  moxi [FLAGS] URL1[,URL2[,URLn]]
  moxi [FLAGS] -z url=URL1[,URL2[,URLn]]
  moxi [FLAGS] -z LOCAL_PORT=MCHOST[:MCPORT][,MCHOST2[:MCPORT2][,*]]

The -z parameter specifies a 'cluster configuration' that tells
moxi which servers to communicate with.

The URL approach allows you to specify the cluster configuration
via dynamic HTTP/REST-based lookup.  Example:
  moxi http://127.0.0.1:8091/pools/default/bucketsStreaming/default

You can also specify multiple cluster configuration URL's for higher
availability, using comma-separated URL's (no whitespace).  A URL list
as the last parameter is also assumed to be a -z cluster configuration.

The MCHOST:MCPORT approach allows you to specify the cluster
configuration for libmemcached/ketama hashing, where moxi listens
on the given LOCAL_PORT and forwards to downstream memcached servers
running at MCHOST:MCPORT.  More than one MCHOST:MCPORT can be listed,
separated by commas.  Example:
  moxi -z 11211=mc_server1:11211,mc_server2:11211

The -z cluster configuration can be also specified in a config file:
  moxi [FLAGS] -z /path/to/absolute/configFile
  moxi [FLAGS] -z ./path/to/relative/configFile

The optional FLAGS are...
```

It has always been tricky to build moxi from source (the binary is available for older ubuntu and centos), there is a lot of stuff missing from https://github.com/couchbase/moxi so you need to use the Couchbase build system: https://github.com/couchbase/tlm. I tried to make a moxi-only standalone version, but it wasn't worth it keeping it up-to-date... If you want to avoid _some_ of the download overhead (800M -> 128M), I made some custom repo manifests and instructions for it: see **[moxi.md](https://gist.github.com/afbjorklund/4155bdd4c4e2ed2e1489062662b57d32)**. Tries to avoid the stuff that is only used by couchbase, but not by moxi. I have tested it on Ubuntu 16.04

afbjorklund:
@martinm82 : but we should support everything that libmemcached, and if it is something relatively simple (e.g. like SASL, see 51883abb7c76d7c96f0b8bc5710ad564d5f8cf19) then we can consider adding it to ccache directly rather than _requring_ moxi.

martinm82:
I had a quick look at moxi and it seems quite nice but I have to admit that it puts more complexity into our setup, especially the building part and system which acts as a proxy but can have potentially issues. Also I am not sure how moxi actually knows about new memcached servers? Does it contain as well some central service where memcached servers can register? 

We run our memcached servers within Docker containers and together with registrator (https://github.com/gliderlabs/registrator) we can easily register/deregister them in a service discovery like Consul.
Asking Consul for a list of available memcached servers and passing it to ccache is therefore quite easy.

With this setup we use directly memcached servers (so no potential new issues with a new system), can add/remove servers easily and the setup is as well easier because it is based on Docker.

I am not saying that moxi is not suitable at all, just think that the deployment with Consul and registrator is a bit easier.

afbjorklund:
Currently ccache just reads the libmemcached configuration, so if that works for you then do continue to use it... Just set the CCACHE_MEMCACHED_CONF before each invocation (possibly with a new list of servers).

Note that when you add and remove servers, you will lose part of the cache... This is due how to the algorithm works, as in libketama versus vbucket. You will need to migrate/refill it yourself.

afbjorklund:
For more information about **vbuckets**, see:
https://blog.couchbase.com/scaling-memcached-vbuckets/

And for more about **ketama** ("consistent") see:
https://www.metabrew.com/article/libketama-consistent-hashing-algo-memcached-clients

----

If you are really really interested, then look into:
https://codeascraft.com/2017/11/30/how-etsy-caches/

There's some interesting tools there, like "mctop".
https://codeascraft.com/2012/12/13/mctop-a-tool-for-analyzing-memcache-get-traffic/

afbjorklund:
@martinm82 : moxi just queries a central CB configuration url, with a HTTP/REST query
https://docs.couchbase.com/server/5.1/rest-api/rest-cluster-intro.html (`/pools`)

Playing around with consul*, it looks like a one-liner something like this would do the trick:
`curl -s localhost:8500/v1/catalog/nodes | jq -r '.[] | "--SERVER="+.Address'`

Would output a string:
`--SERVER=127.0.0.1`

* https://www.consul.io/discovery.html
* https://stedolan.github.io/jq/tutorial/

Could be an alternative to just providing a static list of servers or a local configuration file...
But I don't think ccache itself has a natural place to store such a list between different runs ?

So most likely you would prime an environment variable or a configuration file at build start.
And then use the same list of servers for all subsequent individual compiles within the build.

afbjorklund:
@jonnyyu : technically it is merged to `dev/memcached` and isn't really a PR anymore at this point, but I guess it was not deemed stable enough for `master` (even though it is not even enabled by default...)

bdrewery:
For what it's worth, I've been using the patch, and new branch, for over a year and a half with good results.

jonnyyu:
Thanks @afbjorklund. How can we improve the stability of this branch to meet the merge criteria?

My concerns are:
- features and bug fixes take more time and effort to arrive memcached branch.
- features and bug fixes doesn’t concaider memcached scenario in design and test.

cfs-pure:
> Thanks @afbjorklund. How can we improve the stability of this branch to meet the merge criteria?
> 
> My concerns are:
> 
> * features and bug fixes take more time and effort to arrive memcached branch.
> * features and bug fixes doesn’t consider memcached scenario in design and test.

I concur, what is required or left to do in order to merge dev/memcached back into master?

afbjorklund:
I'll try to get the list of requirements, before it can be merged and made the default (i.e. --disable to opt out).

jrosdahl:
This pull request is an old bad conscience of mine. :slightly_frowning_face:

There are several reasons why I haven't taken this to master yet. I have been a bit short of spare time and motivation for ccache the last couple of years, and I have considered other daily (well, more like monthly) ccache tasks to be more important, like addressing bug reports and tending to other pull requests. But first and foremost, I'm hesitant to merge the memcached work in its current form –  not because the implementation isn't stable but because of the maintenance costs due to increased complexity:

* There is no backend abstration. The standard file-based backend is hardcoded, and so is the memcached backend. If/when we want to add support for other backends, e.g. Redis or a shared filesystem, then the implementation will suffer even more from too little abstraction and too much complexity and code duplication.
* One effect of the lack of abstraction is that when we make changes to what is stored in the backends, for instance when implementing #218 ("Idea: Introduce aggregated result files"), all backends may have to be refactored substantially.
* I would like it to be possible to have storage backends as plugins. The main reason for this is to make it 
possible to create binary ccache packages that don't depend on more or less esotheric libraries. For instance, as a Debian ccache package maintainer I would like to be able to create a ccache-memcached package in addition to the standard ccache package. (Note that this doesn't necessarily mean that we have to create a super-generic backend plugin interface, only that it should be possible to load the backend code dynamically.)

If we would land the current memcached backend solution on master and refine the implementation later, we would be more constrained by backward compatibility restrictions for at least the configuration keys. For instance, if we add a Redis backend, how should ccache know which backend to use if both memcached_conf and redis_conf are set? Maybe both? If so, in which order? Same question for memcached_only/redis_only, etc. We need to come up with a more future-proof way of configuring the backend(s).

I'm sorry if this makes some of you disappointed.

Here's what I would like to see:

* Implement #218.
* Refactor the current file-based code into a backend interface.
* Introduce a way of doing backend-specific configuration.
* Devise a good way of configuring which backends to use and in which order.
* Implement the memcached backend as a plugin.

What do you think? Opinions are welcome.

afbjorklund:
It makes sense, but adds even _more_ work before it can be merged... It's similar to how the ccache configuration system is often one of the biggest "hurdles" to adding new functionality, even though it might be a better user experience in the end. Anyway, I suppose we could using something like the [Apache module system](https://httpd.apache.org/docs/current/mod/) and let the user decide how they want to actually build it...

https://httpd.apache.org/docs/current/install.html#configure
> Also at this point, you can specify which features you want included in Apache by enabling and disabling modules. Apache comes with a wide range of modules included by default. They will be compiled as shared objects (DSOs) which can be loaded or unloaded at runtime. You can also choose to compile modules statically by using the option --enable-module=static.

As we discussed earlier, something similar could be used for compression too. Currently it is very wired to zlib, and adding something like lz4 is still just another special case (similar to replacing/extending file storage with memcached storage). It makes for a lot of code branching and conditional compilation, and hard to change at runtime (different ccache).

Will see if I can revive some of the old code, and see what the API would look like for such a backend abstraction. I did multiple implementations, but only kept one of them - due to the added complexity... The web (http) and database (sql) versions also had _other_ issues (such as being slow), but I guess they could make for sample implementations.

afbjorklund:
> The main reason for this is to make it possible to create binary ccache packages that don't depend on more or less esotheric libraries. For instance, as a Debian ccache package maintainer I would like to be able to create a ccache-memcached package in addition to the standard ccache package.

Not sure that adding `libmemcached-dev` to the list of dependencies qualifies here, but OK.

cleeland:
> Here's what I would like to see:
>
> Implement #218.

Does this really need to be fully implemented as a pre-requisite?  Or would it be sufficient to create an abstraction that synthesizes access to the constituents parts of the file and leave the choice of single/multi-file to the abstraction?

jrosdahl:
@afbjorklund writes:

> It makes sense, but adds even more work before it can be merged...

Yes, I'm painfully aware...

> It's similar to how the ccache configuration system is often one of the biggest "hurdles" to adding new functionality, even though it might be a better user experience in the end.

Not sure if I understand this. In what way is the configuration system a hurdle? Should we change something?

> Anyway, I suppose we could using something like the Apache module system and let the user decide how they want to actually build it...

I'm not very familiar with the Apache module system, but I think that it would be overkill to be that flexible. I don't see a need for compiling backends statically into the main binary, for instance. Do you?

> As we discussed earlier, something similar could be used for compression too. Currently it is very wired to zlib, and adding something like lz4 is still just another special case (similar to replacing/extending file storage with memcached storage). It makes for a lot of code branching and conditional compilation, and hard to change at runtime (different ccache).

I'm less interested in having support for different compression algorithms. I think that there is a clear difference between compression and storage: We can choose a single compression algorithm and ccache will work fine for everybody, but I don't think that we can make a single opininated choice for storage backends.

> Will see if I can revive some of the old code, and see what the API would look like for such a backend abstraction.

Thank you.

> I did multiple implementations, but only kept one of them - due to the added complexity...

Right. Even if we only ever will write one storage backend (memcached), I think that the code will benefit from a clean API instead of being sort of hardcoded.

> Not sure that adding libmemcached-dev to the list of dependencies qualifies here, but OK.

Note that I'm talking about runtime dependencies, not buildtime dependencies. It is an installation dependency on **libmemcached11** (not the **-dev** package) I want to avoid for the main ccache package (and linking statically with libmemcached is not a good option). The envisioned ccache-memcached package would only contain the memcached backend plugin, not the main program.

jrosdahl:
@cleeland writes:

> Does this really need to be fully implemented as a pre-requisite? Or would it be sufficient to create an abstraction that synthesizes access to the constituents parts of the file and leave the choice of single/multi-file to the abstraction?

My tentative idea about the backend API is that it more or less only should contain simple key-value storage primitives. I don't want backends to be aware of which parts a result comprises or how/if they are compressed, etc.

The current code stores several entries per result (.o, .stderr, .d, ...) on disk, so using a simple key-value API both for that and for memcached operations would mean that ccache would make several network calls per result, which isn't good. Compare this with @afbjorklund's implementation which [combines all entries into one](https://github.com/ccache/ccache/blob/f7712082106bef085515f86ae47e7dbbd2a48b57/src/memccached.c#L70) to avoid multiple memcached operations. This is why I thought that having #218 in place first would make things easier.

Does this make sense?

afbjorklund:
We used the Apache runtime for the Ganglia modules, which is why I came to think of it...

e.g. https://github.com/ganglia/ganglia-modules-linux/blob/master/io/Makefile.am

Then again it also uses `automake` (while ccache does not), so maybe that also helped.

https://apr.apache.org/docs/apr/1.6/group__apr__dso.html

jonnyyu:
Thanks for discussing the possibilities to evolve this branch.

In the meantime, could you do a favor to merge master to this branch?
The branch is lag behind quite a few changes. I tried to resolve but there're too many conflicts areas I am not confident to resolve. thanks!

jrosdahl:
> In the meantime, could you do a favor to merge master to this branch?

I understand that you and others are using the dev/memcached branch in production, which makes me uncomfortable since the branch was meant for **dev**eloping stuff, not to hold a production-ready and supported semi-permanent branch. I want to have freedom to make backward incompatible changes on any dev/* branch or just delete them and start over. At the same time I don't want to do you a disservice, but I also don't want to spend time on maintaining this branch since I have decided that I want to see another direction of the solution as described earlier. This solution will most likely be incompatible with the dev/memcached branch, and it will be done iteratively on master and/or other branches, while this branch will be deleted or left as it is until we are done cherry-picking code from the branch.

jimklimov:
Hi, I found this work from earlier mailing list archives :) and wanted to ask if this or newer direction of shared ccache database index would fit our use-case: we have a build farm with many recyclable nodes (their local storage tends to disappear), and a shared NFS server for persistent data, including a `.ccache` directory used by the many build nodes.

Sometimes, the NFS access seems like a bottleneck (e.g. local work on the server, such as `find` listing of files in the index is from few times to an order of magnitude faster than from a build worker over NFS).

I wonder if memcached running on each node (the NFS server as well as builders) could be a faster solution than filesystem queries over NFS to find if an object with needed checksum exists, while *writes* to the `.ccache` directory would still be backed by the same NFS share so newly compiled and cached object files do persist across worker recyclings?

Hope the question makes sense :)

Also I wonder if this removal of (just part of) filesystem-bound work, would help avoid the issue we have sometimes, when a worker is abruptly killed, and it leaves `stats.lock(....lock)` files that effectively block and radically slow down subsequent `ccache` clients because these lock files never disappear, and subsequent `ccache` programs loop over the dozens of added `.lock` extensions until they give up... this is very annoying to say the least ;)

afbjorklund:
@jimklimov : yes, this was/is the idea of the memcached backend. there is no current support for a mixed memcached/nfs store such as the one you are mentioning, instead other solutions such as couchbase were recommended for persistent cache. But it should be possible to do a home-grown solution, either by synching the local cache to external backup storage or by using a memcached-download script.

It also runs in a mixed mode, where files are still stored locally on each server - but pushed/pulled to the remote for usage by other servers. Instead of the alternative, without local disk - here called "memcached only" - where each server talks directly to the memcached cluster... It is also recommended to use a local proxy, to save each and every ccache from having to talk with servers on the network directly.

Reference:

* https://www.couchbase.com/memcached-replacement

* https://www.slideshare.net/northscale/moxi-memcached-proxy

There's no good docs on how to set up a cluster, just some [sample Vagrantfile](https://gist.github.com/afbjorklund/164ee5a653a6040f888fb332ddac6ec0#file-multiple_servers-md) on multiple servers...

jrosdahl:
Closing this PR now as it won't be merged in its current form. It will hopefully reappear in another form in the future.

afbjorklund:
As noted above, this `memcached` branch was a slightly older (3.4.3) version than the `dev/memcached`, which was updated to commit `f771208` based on ccache `v3.4.3-48-g554195a` (v3.5~35) - before it was deleted.

Due to the new storage backend, the new memcached structure will probably **not** be compatible with the old memcached structure used here - but more likely to be based on the "aggregated" storage format of #411.

afbjorklund:
I moved the code that _used_ to be in this branch, into a separate ccache fork and into a separate binary:
(when requiring a special branch and a special ./configure, it was harder to build and harder to install)

https://github.com/afbjorklund/memccache

I think that the C++ version is better served by the new API, and the http and redis backends in PR #676
With couchbase no longer supporting memcached, I think those other alternatives are easier to support...

The old code (in C) will follow `3.7-maint`.

PR #60, opened 2015-12-20
Add option for preprocessing directives only

itensionanders:
Option to handle directives but do not expand macros while preprocessing.
It is about 30% faster to generate the partially preprocessed source than
it is to perform full preprocessing.  However, it comes at a slight cost:
the time taken to perform the initial compilation via ccache is increased.

See https://gcc.gnu.org/ml/gcc-patches/2007-02/msg01098.html (Ollie Wild)


itensionanders:
Most of the time it is probably more efficient to try to get direct hits instead of indirect hits, like with the other variants/patches that I experimented with. But this also seemed interesting, so gave it a shot...

It might be interesting in the context of promoting CCACHE_CPP2 to the regular/default way to do things, as was suggested for better warnings/errors. As in: don't expand macros etc in the context.

Mostly that was for clang (no -fdirectives_only), but anyway...

https://lists.samba.org/archive/ccache/2013q4/001161.html

jrosdahl:
>  However, it comes at a slight cost: the time taken to perform the initial compilation via ccache is increased.

Do you have any numbers to share?

itensionanders:
It defers all macro processing to the compilation, so it is proportional to the number of macros (etc) that you have. They also need more space for storing, so there is a larger preprocessed intermediate file.
However, the above implementation had a *serious* flaw. You need to also pass the `-fdirectives-only` to the compilation. Otherwise you get compilation errors with `-fpreprocessed`. Not sure how I missed...

So maybe one should have just added the compilation flag to the regular CFLAGS instead, and that this whole change was for nothing. Definitely needs some more looking into, before adding to ccache. And like I mentioned, it is optimizing a rather rare case anyway (i.e. preprocessed hits). Still is an interesting path to investigate later, for instance for distcc... But maybe it is not ready for prime time just yet.

itensionanders:
The compiler flag is still OK to use. This patch, however, is bogus and the configuration breaks ccache.

PR #61, opened 2015-12-20
Add prefix option for precompiler too

itensionanders:
To avoid overload the local CPU with precompiling, it can be useful
to let distcc know (take a local slot) when ccache is precompiling.
Those slots are kept in a separate locking queue (--localslots_cpp)
from the remote compilations, in order to use many more remote slots.


jrosdahl:
Sounds reasonable. Please rebase to latest master.

itensionanders:
Note that this feature is currently broken in distcc, and needs some patching to work as intended:
https://github.com/distcc/distcc/pull/161

But this "cpp prefix" feature is generic, and could possibly be used for something other than distcc ?
e.g. Facebook's [warp](https://code.facebook.com/posts/476987592402291/under-the-hood-warp-a-fast-c-and-c-preprocessor/)

itensionanders:
@jrosdahl: Rebased.

jrosdahl:
Applied, thanks!

PR #63, opened 2016-01-30
Always run the unit tests during check

itensionanders:
Should probably fix the "test.sh" script for clang with ELF, so that "make test" can be run instead ?
Currently there are some problems, like the name of the preprocessed file being recorded instead.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,6 +26,7 @@ script:
     - ./autogen.sh
     - ./configure
     - make
+    - make quicktest
     - if [ "$CC" = "gcc" ]; then
           ./test.sh;
       fi


jrosdahl:
Applied, thanks!

> Should probably fix the "test.sh" script for clang with ELF, so that "make test" can be run instead ?

What does "clang with ELF" mean?

itensionanders:
"clang with elf" means using the clang compiler with the elf object format, e.g. on Linux:
`SUITE: "base", TEST: "BASIC" - Files differ: reference_test1.o != test1.o`

> reference_test1.o: ELF 64-bit LSB  relocatable, x86-64, version 1 (GNU/Linux), not stripped
>     1: 0000000000000000      0 FILE    LOCAL  DEFAULT      ABS test1.c
> 
> test1.o:           ELF 64-bit LSB  relocatable, x86-64, version 1 (GNU/Linux), not stripped
>     1: 0000000000000000      0 FILE    LOCAL  DEFAULT      ABS /home/anders/ccache/testdir.5067/.ccache/tmp/test1.stdout.anders-ThinkPad.5663.hQiayq.i
> 
> 
>

itensionanders:
The recommendation when using the clang compiler is to always set CCACHE_CPP2

itensionanders:
Regular "test" fixed in PR #66.

PR #64, opened 2016-01-30
Debuginfo and -fdebug-prefix-map

itensionanders:
This is relevant in the light of global activation of ccache and global setting of CCACHE_HASHDIR:
https://bugzilla.redhat.com/show_bug.cgi?id=759592

By relocating the gcc debuginfo (similar to relocating source with CCACHE_BASEDIR),
it should be possible to still get cache hits even when generating debugging information...

### DIFF
--- a/MANUAL.txt
+++ b/MANUAL.txt
@@ -376,7 +376,7 @@ WRAPPERS>>.
     working directory in the debug info set incorrectly. This option is off by
     default as the incorrect setting of this debug info rarely causes problems.
     If you strike problems with GDB not using the correct directory then enable
-    this option.
+    this option. The option *only* applies when generating debug info (-g).
 
 *ignore_headers_in_manifest* (*CCACHE_IGNOREHEADERS*)::
 
--- a/MANUAL.txt
+++ b/MANUAL.txt
@@ -242,6 +242,10 @@ setting key. Boolean options are indicated with ``[boolean]''
     directory. If set to the empty string (which is the default), no rewriting
     is done. See also the discussion under
     <<_compiling_in_different_directories,COMPILING IN DIFFERENT DIRECTORIES>>.
+    If using GCC*, you might want to look into the *-fdebug-prefix-map* option
+    for relocating debug info to a common prefix (mapping prefix with old=new).
+
+    * or newer versions of Clang, see: http://reviews.llvm.org/rL250094
 
 *cache_dir* (*CCACHE_DIR*)::
 
--- a/NEWS.txt
+++ b/NEWS.txt
@@ -28,6 +28,9 @@ New features and improvements
 - If hard-linking is enabled and but fails (e.g. due to cross-device linking),
   ccache now falls back to copying instead of running the compiler.
 
+- Add support for relocating debuginfo cwd using `-f-debug-prefix-map`.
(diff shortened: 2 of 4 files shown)


jrosdahl:
>This is relevant in the light of global activation of ccache and global setting of CCACHE_HASHDIR:
> https://bugzilla.redhat.com/show_bug.cgi?id=759592

(Yeah, I'm still annoyed about that.)

> By relocating the gcc debuginfo (similar to relocating source with CCACHE_BASEDIR),
> it should be possible to still get cache hits even when generating debugging information...

I like the hack. I wondering, though, whether anybody uses CCACHE_HASHDIR for other reasons, e.g. to make sure that `__FILE__` expands correctly? But perhaps it's unlikely enough to warrant changing semantics of CCACHE_HASHDIR.

Could you include at suitable item to include in NEWS.txt as well?

> `new = x_strndup(sep + 1, strlen(sep + 1));`

Simpler: `new = x_strdup(sep + 1);`

itensionanders:
We were seeing the other side of the problem. If you have a shared cache, you can get hits from other people! So even if you set the base dir to get cache hits, the debuginfo will point "somewhere else".
Like if you have one developer compiling `/home/jill/project/foo.c`, you will get a foo.o file with debuginfo pointing to `/home/jack/project/foo.c`. This can lead to "interesting situations" if the other developer has slightly edited that file (and made it world readable), or just file not found.

I think both the manual and the commit log are pretty clear that CCACHE_HASHDIR was only introduced for debuginfo, so the only downside of "removing" it would be cache misses for compiles without any -g. But it was made a separate edit (ff8a91e) for that reason, so the old semantics could be left in place if desired. I suppose you need to set CCACHE_BASEDIR too, in ordet to get any hits anyway. Will make an entry for the NEWS file, and fix style. Perhaps a better example too, like in the test:

    CCACHE_HASHDIR=1 CCACHE_BASEDIR=$PWD ccache cc -g -fdebug-prefix-map=$PWD=. foo.c

jrosdahl:
> I think both the manual and the commit log are pretty clear that CCACHE_HASHDIR was only introduced for debuginfo,

Sure, but "introduced for" ≠ "used for".

>  so the only downside of "removing" it would be cache misses for compiles without any -g.

Sorry to be splitting hairs, but I don't think that conclusion holds since it's unknown whether there are any users who set CCACHE_HASHDIR to get other effects.

But: This doesn't means I'm against the proposal – I think it's great, I just think we need to understand the implications, for instance to know what to write in the manual and news file.

> Will make an entry for the NEWS file, and fix style. Perhaps a better example too, like in the test:

Awesome, thanks!

itensionanders:
Hmm, if the user relies on hash_dir recording the pwd for some other strange reason than just part of the debuginfo it is probably not really valid to relocate it with the debuginfo (-f-debug-prefix-map) either ? So I would rather see that it was documented as only affecting the debuginfo. Maybe we need some *other* parameter caching the cwd for other reasons ? I can only thing of coverage, but it does it too...

itensionanders:
Rebased, added NEWS and fixed the failing (unrelated?) unit test

jrosdahl:
Thanks, merged!

PR #66, opened 2016-02-02
Fix test for clang (also on Linux ELF)

itensionanders:
The test suite didn't run on clang, due to the elf object format and due to bit rot. Make it.


jrosdahl:
Applied, thanks!

PR #69, opened 2016-02-15
Add appveyor, Windows ci, support.

TheOneRing:
See https://ci.appveyor.com/project/TheOneRing/ccache

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,28 @@
+version: '{build}'
+
+build_script:
+    - ps: |    
+        function CONVERT-MSYS-PATH([string] $path ) {
+            $path = $path.Replace( "`\", "/" )
+            if ( $path.Length -gt 1 -and $path[1] -eq ":" ) {
+                $path = "/{0}/{1}" -f ([string]$path[0]).ToLower() , $path.Substring(3)
+            }
+            return $path
+        }
+        function MSYS-EXEC([string] $command) {
+            Write-Host "bash:", $command
+            & "C:\msys64\usr\bin\bash.exe" "--login" "-c" ("cd {0} && $command" -f (CONVERT-MSYS-PATH $env:APPVEYOR_BUILD_FOLDER ))
+        }
+        MSYS-EXEC "./autogen.sh"
+        MSYS-EXEC "./configure --with-bundled-zlib"
+        MSYS-EXEC "make -j$env:NUMBER_OF_PROCESSORS"
+        if (Test-Path "$env:APPVEYOR_BUILD_FOLDER\ccache.exe") {
+            Push-AppveyorArtifact "$env:APPVEYOR_BUILD_FOLDER\ccache.exe"
+        } else {
+            Exit 1
+        }
+
+environment:
+    matrix:
+    - MSYSTEM: MINGW64
+    - MSYSTEM: MINGW32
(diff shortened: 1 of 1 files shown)


itensionanders:
Fix for unit test can be cherry-picked from b4d227af6e6e5e2dc14bc6765f328da5d5b97ded.
Alternatively (to Appveyor), one could maybe set up Travis cross-compiling with mingw32  ?

itensionanders:
See also http://www.mail-archive.com/ccache@lists.samba.org/msg01273.html
Avoid the missing signal support and formats is easy enough, but the tests fail...

Added the code at 42fb31fd3301f79a2b16a5b90d163a37549b9554 (itensionanders/mingw) if you want to play with it ?

White-Tiger:
Those changes in:
`test/framework.c` : https://github.com/jrosdahl/ccache/commit/42fb31fd3301f79a2b16a5b90d163a37549b9554#diff-2c3cfe373cd1b61a79e8a64b2abd09aeL183
`test/test_conf.c`

wouldn't be needed if MinGW is allowed to use GCC's printf functions instead of MSVC.
`-D__USE_MINGW_ANSI_STDIO=1`
(which would be default for c99, c++11 etc.)

TheOneRing:
The appveyor build can also provide binaries. https://ci.appveyor.com/project/TheOneRing/ccache/build/20/job/dh2ty4468xgtb8ay/artifacts

White-Tiger:
Well... Travis CI can also provide binaries.. though they'll usually upload as Github releases or to some other server.

itensionanders:
I'm thinking that AppVeyor might be the best option after all, since that saves having to install "wine".
The build part (probably) works fine, as seen in 38033d8. But the "make test" part would be trouble...

Anyway, the main thing is actually making it work on Windows again ? There's some dust and bit rot.

itensionanders:
@White-Tiger : sure, but probably overkill for those uses... Better to have it work out-of-the-box ?
That `%zu` can probably be avoided on **all** platforms, we don't have _that_ many configs do we ?

itensionanders:
I don't think you need to create binaries, not from the CI tests at least.
Pushed fixed unit tests to: itensionanders/ccache@mingw (477caff)

TheOneRing:
and broken again
https://ci.appveyor.com/project/TheOneRing/ccache/build/22

itensionanders:
Build and test fixed in d2c6751b27c808a878ab7a7ec6c9acb00b437e8c (from #73)

TheOneRing:
It now compiles :)
I guess its now only about tests and the install target.

itensionanders:
@TheOneRing: there seems to be a couple of failing tests still ?
`
FAILED: 4 tests
C:/projects/ccache/test/test_argument_processing.c:166: Failed assertion:
C:/projects/ccache/test/test_argument_processing.c:354: Failed assertion:
C:/projects/ccache/test/test_conf.c:341: Failed assertion:
C:/projects/ccache/test/test_conf.c:353: Failed assertion:
`
Perhaps only for the 32-bit target though, as the 64-bit looks OK ?

And the usage of "git" and "asciidoc" seems to be silently failing:
`bash.exe : configure: WARNING: unable to determine ccache version`
`bash.exe : make: *** No rule to make target '/c/projects/ccache/ccache.1', needed by 'install'.  Stop.`
Maybe it should do so with slightly more grace. Or avoid man perhaps.

jrosdahl:
What's the status of this pull request?

itensionanders:
I was a bit skeptic about the need to do "make install" and publish artifacts, but it would be nice to actually run the Windows test binaries after producing them... As in compiling was doable with Travis, but running was much harder (with WINE) than doing it in Appveyor ? (Starting bash, the other way around...) Either way, we did get some basic Win32 functionality going again. Would be nice to keep it that way ?

jrosdahl:
Closing this as superseded by #247.

Issue #68, opened 2016-02-10
Stale NFS Handle


nick-schultz:
We are sharing the ccache dir on an NFS share.  During high load, we are seeing consistent "Stale NFS file handle" errors to the stats file.

Any suggestions as to what can be done to avoid this problem? 

````
[2016-02-10T14:16:20.237532 12551] Trying direct lookup

[2016-02-10T14:16:20.238577 12551] Looking for object file hash in /nfs/compile_cache/a/5/c2da3efd19dcc3ef79042ca6df7c02-26442.manifest
[2016-02-10T14:16:20.706020 12551] Got object file hash from manifest

[2016-02-10T14:16:20.711564 12551] Copying /nfs/compile_cache/1/c/87c3236c570c5e35767209c40fcdc7-7031461.o to /out_dir/BubbleInsertionTest.obj via /out_dir/BubbleInsertionTest.obj.fmyciw202.12551.Jq5p9b (uncompressed)

[2016-02-10T14:16:20.724023 12551] Created from cache: /nfs/compile_cache/1/c/87c3236c570c5e35767209c40fcdc7-7031461.o -> /out_dir/BubbleInsertionTest.obj

[2016-02-10T14:16:20.726563 12551] Copying /nfs/compile_cache/1/c/87c3236c570c5e35767209c40fcdc7-7031461.d to /out_dir/BubbleInsertionTest.d via /out_dir/BubbleInsertionTest.d.fmyciw202.12551.kcTxOq (uncompressed)

[2016-02-10T14:16:20.734324 12551] Created from cache: /nfs/compile_cache/1/c/87c3236c570c5e35767209c40fcdc7-7031461.d -> /out_dir/BubbleInsertionTest.d

[2016-02-10T14:16:20.737888 12551] Succeeded getting cached result
[2016-02-10T14:16:20.738724 12551] Acquired lock /nfs/compile_cache/1/stats.lock
[2016-02-10T14:16:20.750046 12551] Releasing lock /nfs/compile_cache/1/stats.lock
[2016-02-10T14:16:20.750053 12551] Unlink /nfs/compile_cache/1/stats.lock
[2016-02-10T14:16:20.751956 12551] Result: cache hit (direct)             
[2016-02-10T14:16:20.751964 12551] Cleaning up cache directory /nfs/compile_cache/1
******[2016-02-10T14:16:56.050417 12551] FATAL: lstat /nfs/compile_cache/1/stats failed: Stale NFS file handle*****
```

[...]

PR #70, opened 2016-02-17
Handle "Stale NFS File Handle" cases better.

nick-schultz:
This is in response to #68 

I came across an earlier commit, 3049c10, that was resolving a similar issue.  I believe adding ESTALE as a condition to ignore may be appropriate as well. 

edit, I meant 8d91b7c


itensionanders:
I think you mean 8d91b7ceb32f52b8bec3594f375c48fe92fe39df, but anyway (looks ok).

jrosdahl:
Applied on 3.2-maint and merged to master. Thanks!

PR #71, opened 2016-02-17
Add the default sloppiness to the pth test for pch

itensionanders:
Please add this fix, to make the clang adress sanitizer (asan) tests more stable...

Example of unrelated fail: https://travis-ci.org/jrosdahl/ccache/jobs/109353650


itensionanders:
Already merged, as b4d227a, so this cherry-pick is not needed.

PR #72, opened 2016-02-18
MinGW build for Windows

itensionanders:
Mostly a quick hack to get the MinGW build working, and make the "quicktest" pass with Wine...


jrosdahl:
Applied, thanks!

PR #73, opened 2016-03-04
Fix the MinGW issues

itensionanders:


itensionanders:
Apparently Travis doesn't add depends of packages.

itensionanders:
"Disallowing packages: mingw-w64-i686-dev, mingw-w64-common"

TheOneRing:
could you sue mingw-w64?
gcc-mingw-w64-x86-64 
mingw-w64 is today more often used than mingw32 and it supports in general a better subset of windows apis.

itensionanders:
@TheOneRing: Travis is currently using Ubuntu 12.04.
And the support for 64-bit in Wine is poorer than 32-bit.

Otherwise, I agree fully.

itensionanders:
But it should still work with MinGW-w64, so there is no reason why Appveyor couldn't test those "modern" Windows targets (both 32-bit and 64-bit) and Travis could test "legacy" MinGW and generic Windows ?

PR #75, opened 2016-03-25
Fix clang plugin hashing.

mkolomeytsev:
This matters when compiling in different directories and clang plugin
is located in them. make_relative_path is redundant in that case.

https://lists.samba.org/archive/ccache/2016q1/001425.html


jrosdahl:
Applied on 3.2-maint and merged to master. Thanks!

PR #77, opened 2016-03-28
Fix warnings from uncrustify

itensionanders:
Seems like 6fa07483c5d11a35d49de26b724b7528073ab95e left the "uncrustify" target in a bad state:

`uncrustify.cfg:1228 Expected 'True' or 'False' for mod_full_brace_if_chain, got add`

Also fix this old warning, by disabling uncrustify for SUITE(name):

`test/main.c:27 Detected a macro that ends with a semicolon. Expect failures if used.`

### DIFF
--- a/test/main.c
+++ b/test/main.c
@@ -24,9 +24,11 @@
 #include "getopt_long.h"
 #endif
 
+/* *INDENT-OFF* disable uncrustify */
 #define SUITE(name) unsigned suite_ ## name(unsigned);
 #include "test/suites.h"
 #undef SUITE
+/* *INDENT-ON* enable uncrustify */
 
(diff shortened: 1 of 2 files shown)


jrosdahl:
Interesting, the uncrustify version I have installed (0.59) didn't warn about it. Applied, thanks!

itensionanders:
I used the system version of uncrustify from Ubuntu 14.04 LTS (i.e. 0.59-2).
But it doesn't seem to have any special patches or any special config in it ?

http://packages.ubuntu.com/trusty/uncrustify
https://github.com/uncrustify/uncrustify/commit/20b21c2 (uncrustify-0.59~44)

PR #78, opened 2016-03-30
Convert absolute paths to relative paths inside the dependency file

mihais:
When using DIRECT MODE and the -MD flag to generate a dependency file
the toolchain header files are used with their absolute paths as dependencies.
If the same toolchain is used but from a different location the previously
cached .d file is copied to the current build environment. Because it contains
absolute paths that could disappear or where the current user does not have
access the compilation will fail complaining that some header files cannot
be found.
This mostly happens when the cache is shared among multiple users or multiple
clones of a source tree and when the same toolchain is used but from different
locations.
This patch adds support for this kind of usage by replacing the absolute paths
with relative ones based on the BASEDIR value.

https://lists.samba.org/archive/ccache/2014q2/001183.html
https://lists.samba.org/archive/ccache/2014q2/001184.html

### DIFF
--- a/ccache.c
+++ b/ccache.c
@@ -177,6 +177,9 @@ time_t time_of_compilation;
  */
 static struct hashtable *included_files;
 
+/* uses absolute path for some include files */
+static bool has_absolute_include_headers = false;
+
 /* List of headers to ignore */
 static char **ignore_headers;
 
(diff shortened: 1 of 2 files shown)


jrosdahl:
Applied, thanks!

PR #80, opened 2016-04-02
Increase buffer size for IO operations.

mkolomeytsev:
https://lists.samba.org/archive/ccache/2016q2/001429.html

### DIFF
--- a/ccache.h
+++ b/ccache.h
@@ -74,6 +74,9 @@ enum stats {
 #define str_eq(s1, s2) (strcmp((s1), (s2)) == 0)
 #define str_startswith(s, p) (strncmp((s), (p), strlen((p))) == 0)
 
+/* buffer size for IO operations */
+#define READ_BUF_SZ (1<<18)
+
 /* ------------------------------------------------------------------------- */
 /* args.c */
 
--- a/hash.c
+++ b/hash.c
@@ -102,7 +102,7 @@ hash_int(struct mdfour *md, int x)
 bool
 hash_fd(struct mdfour *md, int fd)
 {
-	char buf[16384];
+	char buf[READ_BUF_SZ];
 	ssize_t n;
 
 	while ((n = read(fd, buf, sizeof(buf))) != 0) {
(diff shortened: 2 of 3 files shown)


ccache.h
```diff
@@ -74,6 +74,9 @@ enum stats {
 #define str_eq(s1, s2) (strcmp((s1), (s2)) == 0)
 #define str_startswith(s, p) (strncmp((s), (p), strlen((p))) == 0)
 
+/* buffer size for IO operations */
+#define READ_BUF_SZ (1<<18)
```
itensionanders: Maybe use different constants, for hash (old 16K) and copy (old 10K) ?
And maybe also configurable, for different systems / target platforms...

jrosdahl:
@itensionanders: You mentioned some experiments you made in https://lists.samba.org/archive/ccache/2016q2/001431.html. What's your stance on this pull request? Do you think that we should just increase the buffer size to e.g. 256 KiB as suggested by @mkolomeytsev? Or just increase to16 KiB in copy_fd and copy_file? Or do something more fancy?

itensionanders:
Some benchmark would be nice ? I don't have any figures ready, sorry.  But 256K "sounds" big ?
I think the main reason for going from 10K to 16K was to get an even number of 4K pages...

jrosdahl:
> Some benchmark would be nice ? I don't have any figures ready, sorry. But 256K "sounds" big ?

Yes, it sounds big to me too. I'm surprised that it makes such difference for Michael on OSX.

> I think the main reason for going from 10K to 16K was to get an even number of 4K pages...

Yep, that's uncontroversial.

I've now done some benchmarks myself now. On my 64-bit Linux system, I see no performance difference when going from 10 KiB to 16 KiB. However, going from 16 KiB to 32 KiB gives a significant performance increase (≈ 10%). Buffer sizes between 64 KiB and 2 MiB perform the same as 32 KiB for me. 4 MiB is worse again.

I think we can conclude that different systems behave differently, but going for multiple of 4 KiB and at least 32 KiB sounds good.

Related: By using `read` instead of `gzread` for reading uncompressed files increases performance with another ≈30%! I had hoped that `gzread` would have very little overhead for uncompressed files, but apparently not. See also #81.

I'll apply this PR but change the size to 64 KiB for now. I'll also look into refactoring the copy_*/hash_fd functions in spirit of #81.

PR #81, opened 2016-04-22
Introduce copy_uncompressed_file, for direct copy

afbjorklund:
For copying compressed files etc, fall back to the original copy version.
(this is mostly a refactoring, for future features and also optimizations)

### DIFF
--- a/ccache.c
+++ b/ccache.c
@@ -988,7 +988,7 @@ put_file_in_cache(const char *source, const char *dest)
 		}
 	}
 	if (!do_link) {
-		ret = copy_file(
+		ret = copy_uncompressed_file(
 		  source, dest, conf->compression ? conf->compression_level : 0);
 		if (ret != 0) {
 			cc_log("Failed to copy %s to %s: %s", source, dest, strerror(errno));
@@ -1009,11 +1009,14 @@ static void
 get_file_from_cache(const char *source, const char *dest)
 {
 	int ret;
-	bool do_link = conf->hard_link && !file_is_compressed(source);
+	bool is_compressed = file_is_compressed(source);
+	bool do_link = conf->hard_link && !is_compressed;
 
 	if (do_link) {
 		x_unlink(dest);
 		ret = link(source, dest);
+	} else if (!is_compressed) {
+		ret = copy_uncompressed_file(source, dest, 0);
 	} else {
 		ret = copy_file(source, dest, 0);
 	}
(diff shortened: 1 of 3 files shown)


jrosdahl:
I like the idea but not the code duplication. :) Have you checked whether copy_uncompressed_file is measurable faster than copy_file for uncompressed files? Is the potential speedup worth the extra maintenance cost, albeit small?

itensionanders:
Yes and no. It was faster (for the actual file copy), but it didn't make much difference. Basically I just wanted a version without the gzip complexity, especially when adding more compressions to the mix.

To make it really go faster, you would have to resort to a lot of non-portable tricks. And even then you would have to have pretty large files. But some things like preallocating (ftrunc) can be very useful.

See https://github.com/itensionanders/ccache/tree/uncompressed-fast for the stuff involved ?

The new lz4f was at https://github.com/itensionanders/ccache/commit/06d1081fe9b37741a75af57a250d55f70cd99f13#diff-6d7e72ea1ca2b7d0258e568a155084bd
(but should probably be split out to a separate file, much like zlib handles the gzip streams ?)

Another use case was compressed->compressed copy, like done in the "external" branch...

itensionanders:
Oops, turned out that I missed some error handling when "translating" from gzio to stdio...

jrosdahl:
> Oops, turned out that I missed some error handling when "translating" from gzio to stdio...

Which means? :)

itensionanders:
You could get silent read errors, updated the branch with those checks added though.

Think the test build failed due to a timeout, actually.

jrosdahl:
OK, but what does it mean from a pull request point of view? Should I drop it?

itensionanders:
I think the name of the new function is awful, so it could probably need some more thought...

But what I wanted to accomplish compared to copy_file was:

1. Add an option to *not* decompress the source, in case it was e.g. a manifest being copied it got "corrupted" by being unpacked
2. Allow for more performant copies by doing "straight" binary copies, like with other buffer sizes or special opeating system features

It coukd definitely need some tests and benchmarks written, either way.

jrosdahl:
Regarding ef2c268: I tried out `posix_fadvise` and it gave me neither better nor better performance. However, `fallocate` gave significantly worse performance (≈15%) on my system.

afbjorklund:
I think this code will show up at other places if needed, and the "fast" track really wasn't.

PR #82, opened 2016-04-22
Make the cleanup limit into a configurable variable

itensionanders:
When doing manual cleanups, it is useful to be able to set the
cleanup limit multiple rather than hardcoding it as 80% always.


jrosdahl:
Sounds good to me. However, I think that we should strive not to add more features to the 3.2 series. Would it be OK to bring this to master instead?

itensionanders:
I'm thinking I actually started this on master, and then chose 3.2 to make it "easier" :-)
Basically I was looking through the old 3.1 stuff, to update... (original was 0405827f2f2185186a95ebc82b9dd8bd6c901f22)

jrosdahl:
@itensionanders: Could you please rebase this to master?

jrosdahl:
Ported and applied on master, now.

PR #83, opened 2016-04-22
Count the number of cleanups performed, in stats

itensionanders:
When running a busy ccache, it is useful to keep track of the number
of cleanups performed to know when you need to increase the cache size.

The number shown is actually the number of subdirectories, so when
you do a full cleanup the counter usually increases by 16 (not just 1).

### DIFF
--- a/cleanup.c
+++ b/cleanup.c
@@ -130,11 +130,12 @@ delete_sibling_file(const char *base, const char *extension)
 
 /* sort the files we've found and delete the oldest ones until we are
    below the thresholds */
-static void
+static bool
 sort_and_clean(void)
 {
 	unsigned i;
 	char *last_base = x_strdup("");
+	bool clean = false;
 
 	if (num_files > 1) {
 		/* Sort in ascending mtime order. */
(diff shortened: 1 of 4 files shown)


cleanup.c
```diff
@@ -130,11 +130,12 @@ delete_sibling_file(const char *base, const char *extension)
 
 /* sort the files we've found and delete the oldest ones until we are
    below the thresholds */
-static void
+static bool
 sort_and_clean(void)
 {
 	unsigned i;
 	char *last_base = x_strdup("");
+	bool clean = false;
```
jrosdahl: I suggest the name `cleaned` instead.

jrosdahl:
Looks good. Apply to master instead?

cleanup.c
itensionanders: Sounds good, think the name stuck even after converting it from `int` to `bool`

itensionanders:
Master is OK too.

jrosdahl:
Applied to master with a follow-up commit for the variable name change. Thanks!

PR #85, opened 2016-04-28
Convert -iprefix/-B/-L paths to relative paths

vijunag:
The patch fixes the issue mentioned here https://github.com/jrosdahl/ccache/issues/84.

### DIFF
--- a/ccache.c
+++ b/ccache.c
@@ -2451,10 +2451,18 @@ cc_process_args(struct args *args, struct args **preprocessor_args,
 			continue;
 		}
 		if (str_startswith(argv[i], "-iprefix")) {
-			char *relpath = make_relative_path(x_strdup(argv[i+1]));
-			char *option = format("%s", relpath);
-			args_add(cpp_args, "-iprefix");
-			args_add(cpp_args, option);
+			char *relpath = NULL;
+			char *option = NULL;
+			if (strlen(argv[i]) > strlen("-iprefix")) {
+				relpath = make_relative_path(x_strdup(argv[i] + strlen("-iprefix")));
+				option = format("-iprefix%s", relpath);
+				args_add(cpp_args, option);
+			} else {
+				relpath = make_relative_path(x_strdup(argv[i+1]));
+				option = format("%s", relpath);
+				args_add(cpp_args, "-iprefix");
+				args_add(cpp_args, option);
+			}
 			free(relpath);
 			free(option);
 			i++;
--- a/ccache.c
+++ b/ccache.c
@@ -2441,6 +2441,31 @@ cc_process_args(struct args *args, struct args **preprocessor_args,
 			free(option);
 			continue;
 		}
+                if (str_startswith(argv[i], "-B")) {
+			char *relpath = make_relative_path(x_strdup(argv[i] + 2));
+			char *option = format("-B%s", relpath);
+			args_add(stripped_args, option);
+			free(relpath);
+			free(option);
+			continue;
+                }
+                if (str_startswith(argv[i], "-iprefix")) {
+			char *relpath = make_relative_path(x_strdup(argv[i+1]));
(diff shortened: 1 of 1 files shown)


ccache.c
```diff
@@ -2441,6 +2441,31 @@ cc_process_args(struct args *args, struct args **preprocessor_args,
 			free(option);
 			continue;
 		}
+                if (str_startswith(argv[i], "-B")) {
```
jrosdahl: Please use tabs to indent the code, see HACKING.txt. Tip: You can use `make uncrustify` to make the code use ccache's code conventions.

ccache.c
```diff
@@ -2441,6 +2441,31 @@ cc_process_args(struct args *args, struct args **preprocessor_args,
...
+			char *relpath = make_relative_path(x_strdup(argv[i] + 2));
+			char *option = format("-B%s", relpath);
+			args_add(stripped_args, option);
+			free(relpath);
+			free(option);
+			continue;
+                }
+                if (str_startswith(argv[i], "-iprefix")) {
+			char *relpath = make_relative_path(x_strdup(argv[i+1]));
```
jrosdahl: Looks like this won't work well for

```
-iprefixfoo.h
```

? (Yes, that's a valid case, see http://www.mail-archive.com/ccache@lists.samba.org/msg01235.html.)

ccache.c
```diff
@@ -2441,6 +2441,31 @@ cc_process_args(struct args *args, struct args **preprocessor_args,
...
+			char *option = format("-iprefix %s", relpath);
+			args_add(stripped_args, option);
+			free(relpath);
+			free(option);
+			i++;
+			continue;
+                }
+                if (str_startswith(argv[i], "-L")) {
```
jrosdahl: Please reuse code between the `-B` and `-L` cases since they are identical except for `B` and `L`.

ccache.c
jrosdahl: Please use tabs to indent the code, see HACKING.txt. Tip: You can use make uncrustify to make the code use ccache's code conventions.

jrosdahl:
The tests fail after your change – it would nice if you could make sure that they stay working.

jrosdahl:
By the way, the patch has author "vijay@localhost.localdomain". Is that intentional?

ccache.c
vijunag: I RTFM'ed GCC and clearly there is a space between iprefix and the directory as mentioned here 
https://gcc.gnu.org/onlinedocs/cpp/Invocation.html
> 
-iprefix prefix
    Specify prefix as the prefix for subsequent -iwithprefix options. If the prefix represents a directory, you should include the final ‘/’. 
>
Also,  the link you've provided is broken. Can you please paste the compiler invocation here ?

ccache.c
vijunag: Done

ccache.c
jrosdahl: > I RTFM'ed GCC and clearly there is a space between iprefix and the directory as mentioned here 

Yes, but GCC is more forgiving than what's documented, so ccache needs to be as well.

> Also, the link you've provided is broken.

Bummer. Here's the link without the extraneous ".": http://www.mail-archive.com/ccache@lists.samba.org/msg01235.html

> Can you please paste the compiler invocation flags here ?

Sorry, I don't understand what you're asking for.

vijunag:
-iprefixfoo.h support has been added.

jrosdahl:
I've had a closer look at this and think that it would be better to instead improve the compopt framework (see compopt.c) to rewrite paths to *all* options that have properties TAKES_CONCAT_ARG  and TAKES_PATH. Adding code for specific options like suggested in this pull request quickly gets a bit messy and duplicated.

jrosdahl:
I took at shot at a more generic approach in 5fdb774.

PR #88, opened 2016-05-22
preprocessor mode: enable using identical header in different paths

jbasila:
Please see the following thread: http://www.mail-archive.com/ccache@lists.samba.org/msg01329.html

This fix will improve the ccache hit as it makes the preprocessor mode to be come insensitive to identical head files path changes.

### DIFF
--- a/test.sh
+++ b/test.sh
@@ -1247,8 +1247,8 @@ EOF
     checkstat 'cache miss' 1
     $CCACHE $COMPILER -c `pwd`/file.c
     checkstat 'cache hit (direct)' 1
-    checkstat 'cache hit (preprocessed)' 0
-    checkstat 'cache miss' 2
+    checkstat 'cache hit (preprocessed)' 1
+    checkstat 'cache miss' 1
 
     testname="__FILE__ in include file"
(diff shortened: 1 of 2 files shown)


test.sh
```diff
@@ -1458,12 +1458,12 @@ EOF
     checkstat 'cache miss' 1
     CPATH=subdir2 $CCACHE $COMPILER -c foo.c
     checkstat 'cache hit (direct)' 1
-    checkstat 'cache hit (preprocessed)' 0
-    checkstat 'cache miss' 2 # subdir2 is part of the preprocessor output
+    checkstat 'cache hit (preprocessed)' 1
+    checkstat 'cache miss' 1 # subdir2 is part of the preprocessor output
```
jrosdahl: This comment is now a bit misleading, right? Perhaps just remove it?

jbasila:
I removed the comment.

jrosdahl:
Applied now, thanks!

jrosdahl:
For future reference: This feature has been reverted. See #134 and 7cad4ac.

PR #89, opened 2016-05-26
Make it compile in mingw-w64 on Windows...

melven:
requires to define _POSIX in mingw-w64, e.g.
> CC="gcc -D_POSIX" ./configure

### DIFF
--- a/ccache.h
+++ b/ccache.h
@@ -257,6 +257,12 @@ typedef int (*COMPAR_FN_T)(const void *, const void *);
 #undef HAVE_MKSTEMP
 #endif
 
+/* mkstemp() on some versions of mingw sets _O_TEMPORARY, so
+   override */
+#ifdef __MINGW32__
+#undef HAVE_MKSTEMP
+#endif
+
 #ifdef _WIN32
 char *win32argvtos(char *prefix, char **argv);
 char *win32getshell(char *path);
--- a/system.h
+++ b/system.h
@@ -48,6 +48,19 @@
 #include <time.h>
 #include <unistd.h>
 #include <utime.h>
+#ifdef __MINGW32__
+#include <pthread.h>
+#include <windows.h>
+#include <winsock.h>
+#ifndef ESTALE
+#define ESTALE WSAESTALE
+#endif // ESTALE
+#ifndef WNOHANG
+#define WNOHANG 1
+#endif // WNOHANG
+#define HAVE_GETFINALPATHNAMEBYHANDLEW
+#endif // WIN32
+
 
 extern char **environ;
 
--- a/system.h
+++ b/system.h
@@ -49,6 +49,7 @@
 #include <unistd.h>
 #include <utime.h>
 #ifdef __MINGW32__
+/* use pthread for some signal stuff as not all modern posix functions are available */
 #include <pthread.h>
 #include <windows.h>
 #include <winsock.h>
@@ -59,7 +60,7 @@
 #define WNOHANG 1
 #endif // WNOHANG
 #define HAVE_GETFINALPATHNAMEBYHANDLEW
-#endif // WIN32
+#endif // __MINGW32__
 
 
 extern char **environ;
--- a/ccache.c
+++ b/ccache.c
@@ -338,22 +338,22 @@ temp_dir()
 void
 block_signals(void)
 {
-#ifndef WIN32
+#ifndef _WIN32
 	sigprocmask(SIG_BLOCK, &fatal_signal_set, NULL);
 #else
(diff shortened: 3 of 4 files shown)


jrosdahl:
Thanks for the patch.

However, it seems to break the i586-mingw32msvc-gcc build of ccache: https://travis-ci.org/jrosdahl/ccache/jobs/133046727

Why use `WIN32` instead of `_WIN32_?

Please follow the ccache coding conventions, e.g. use tabs for indentation.

TheOneRing:
While _WIN32 is a system define WIN32 is a custom define that needs to be defined manually. So please use _WIN32.
As mingw is probably the only native compiler that will ever be supported on windows, __MINGW32__ might also be replace by _WIN32 ?

jrosdahl:
@melven: Any updates regarding this pull request?

itensionanders:
@jrosdahl: this doesn't compile for me either.

Besides the WIN32 confusion mentioned above, there doesn't seem to be a `sigset_t` available ?
And hard-coding values from autoconf compatibility checks isn't really an improvement, either...

`#define HAVE_GETFINALPATHNAMEBYHANDLEW`

itensionanders:
Apparently you need to define _POSIX, to get _sigset (mentioned in the commit message)
But there's still some other implicit functions, like waitpid/kill/sigemptyset/sigaddset ?

melven:
Sorry, I had some other things to do; I will try to improve the patch soon...

Thanks for the feedback!

ccache.h
```diff
@@ -257,6 +257,12 @@ typedef int (*COMPAR_FN_T)(const void *, const void *);
 #undef HAVE_MKSTEMP
 #endif
 
+/* mkstemp() on some versions of mingw sets _O_TEMPORARY, so
```
itensionanders: Noticed this too, 18c773058637b16a2b522225bd9230ab3fb5a099

system.h
```diff
@@ -48,6 +48,20 @@
...
+#include <pthread.h>
+#include <windows.h>
+#include <winsock.h>
+#ifndef ESTALE
+#define ESTALE WSAESTALE
+#endif // ESTALE
+#ifndef WNOHANG
+#define WNOHANG 1
+#endif // WNOHANG
+#define HAVE_GETFINALPATHNAMEBYHANDLEW
```
itensionanders: Why is this needed ? Shouldn't it be found by configure (in config.h)

itensionanders:
I think that you need to make sure that configure adds the `-D_POSIX` if it is needed.
It would also be nice if it could cope with the ancient MinGW in Travis (Ubuntu 12.04)

melven:
Sorry for my slow response - I identified several problems with my build system/this merge request:

1. As I don't have autoconf/etc I used the a configure/config.h.in from the latest release (which does not yet contain the check for GetFinalPathNameByHandleW) - my fault!
2. The mkfstemp thing is (now) already fixed in 18c7730, thanks for that.
3. Compiling with -D_POSIX and using undeclared functions (kill, etc...) is quite ugly and I'm not sure if it works as expected (do the tests cover this? it should only get triggered when trying to close/kill ccache while compiling, right?)

So I would close this for now.

Further remarks:

Remaining problems for me on windows:

Minor fix for running .sh files (as compiler wrappers):
```
diff --git a/execute.c b/execute.c
index d7c75b8..0c20933 100644
--- a/execute.c
+++ b/execute.c
@@ -128,7 +128,8 @@ void add_exe_ext_if_no_to_fullpath(char *full_path_win_ext, size_t max_size,
        if (!ext || (!str_eq(".exe", ext)
                     && !str_eq(".bat", ext)
                     && !str_eq(".EXE", ext)
-                    && !str_eq(".BAT", ext))) {
+                    && !str_eq(".BAT", ext)
+                    && !str_eq(".sh",  ext))) {
                snprintf(full_path_win_ext, max_size, "%s.exe", path);
        } else {
                snprintf(full_path_win_ext, max_size, "%s", path);
```

The test *compilercheck=command* fails.

When I comment it out, I obtain the error: *the tmp directory should be empty* for suite base.

In the ccache.log there are random errors with unlinking files (*no such file or directory* for unlink/x_unlink). I suppose this is a timing issue on windows...

itensionanders:
Oh, I forgot how horrible that environment was. Like you mention, even `printf` is broken with the regular MinGW (actually MSVCRT but anyway). And that compilercheck=command failure is actually from the stray `^M` in the output, due to it being handled in "binary". Getting the .bat files working in MSYS also looks rather painful, with the chronic confusion of the path separator - is it `'/'` or the more insane `'\\'` ?

So MinGW sorta/kinda compiles and passes the unit tests (with #119), but getting it to *work* on Windows...

PR #90, opened 2016-05-26
Fortran support

melven:
first commit works for old Fortran 77.
second commit tries direct mode for Fortran 90/95 and newer...

Special things in (modern) Fortran:
* preprocessing doesn't resolve module dependencies; module files are required at compile time;
  so I disabled usage of indirect mode there (assuming ccache doesn't care about file dependencies after preprocessing...)
* the compiler may output an arbitrary number of module files per source file; these are currently not cached, but recreated using the -fsyntax-only flag on a cache hit.

Needs some more testing...

### DIFF
--- a/ccache.c
+++ b/ccache.c
@@ -213,6 +213,9 @@ static bool direct_i_file;
 /* the name of the cpp stderr file */
 static char *cpp_stderr;
 
+/* the name of the custom fortran module directory (default: pwd dirname) */
+static char *fortran_module_dir = NULL;
+
 /*
  * Full path to the statistics file in the subdirectory where the cached result
  * belongs (<cache_dir>/<x>/stats).
@@ -1138,6 +1141,12 @@ to_cache(struct args *args)
 	 */
 	x_unsetenv("DEPENDENCIES_OUTPUT");
 
+  /* for Fortran explicitly specify target module directory */
+  if (fortran_module_dir) {
+    args_add(args, "-J");
+    args_add(args, fortran_module_dir);
+  }
+
 	if (conf->run_second_cpp) {
 		args_add(args, input_file);
 	} else {
@@ -1147,6 +1156,9 @@ to_cache(struct args *args)
 	cc_log("Running real compiler");
 	status = execute(args->argv, tmp_stdout_fd, tmp_stderr_fd, &compiler_pid);
 	args_pop(args, 3);
+	if (fortran_module_dir) {
+    args_pop(args, 2);
+	}
 
 	if (x_stat(tmp_stdout, &st) != 0) {
 		/* The stdout file was removed - cleanup in progress? Better bail out. */
@@ -1914,14 +1926,21 @@ calculate_object_hash(struct args *args, struct mdfour *hash, int direct_mode)
  * then this function exits with the correct status code, otherwise it returns.
  */
 static void
-from_cache(enum fromcache_call_mode mode, bool put_object_in_manifest)
+from_cache(enum fromcache_call_mode mode, bool put_object_in_manifest, struct args* args)
 {
 	struct stat st;
 	bool produce_dep_file = false;
+  char *tmp_stdout, *tmp_stderr;
+  int tmp_stdout_fd, tmp_stderr_fd, status;
 
 	/* the user might be disabling cache hits */
 	if (conf->recache) {
 		return;
+	}
+
+	/* Fortran 90/95 code might depend on module files, so allow only direct mode then */
+	if (fortran_module_dir && mode != FROMCACHE_DIRECT_MODE ) {
+    return;
 	}
 
 	if (stat(cached_obj, &st) != 0) {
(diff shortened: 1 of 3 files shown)


jrosdahl:
Thanks for you effort! However, I'm not convinced that it would be a good idea to add Fortran support to ccache. I'm primarily concerned of the increased maintanance burden.

melven:
I understand. Still it would be nice to have better support for Fortran in compilers and tools (such as ccache). There is a lot of software that make C/C++ development easier, but really only few (none?) that really work for modern Fortran.

As a compromise: Could you reconsider the first commit with Fortran 77 support? This would already speed up things like building LAPACK a lot. And I don't see a real difference for the maintenance here.

jrosdahl:
> As a compromise: Could you reconsider the first commit with Fortran 77 support? This would already speed up things like building LAPACK a lot. And I don't see a real difference for the maintenance here.

Sure, applied in c0875cf.

melven:
Thanks

boegel:
I ran into this error earlier today while compiling CP2K:

```
    /tmp//CP2K/2.4.0/goolf-1.4.10-ipi/cp2k-2.4.0/makefiles/../src/input_constants.F:16.6:

  USE f77_blas
      1
Fatal Error: Can't open module file 'f77_blas.mod' for reading at (1): No such file or directory
make[2]: *** [input_constants.o] Error 1
```

I took me a while to figure out why this was failing all of a sudden, since it worked fine before, but then I realised I had `ccache` in place for `gfortran` (`ccache` works so well that you easily forget about having it in place... :+1:). 
I remembered someone mentioned an issue related to Fortran modules and `ccache`. A post on the ccache mailing list (https://lists.samba.org/archive/ccache/2016q4/001476.html) pointed me here.

@jrosdahl I understand and support your decision not to take the Fortran support in `ccache` too far, let me be clear on that.

I was just wondering whether this could be handled better, to avoid a failing compilation when `ccache` is wrapping `gfortran` and Fortran modules come into play...

Can we avoid caching the result when Fortran modules were being generated?
A cache miss is way more desirable than a compilation failure imho...

Thoughts @melven? Can this be done without imposing a maintenance burden on `ccache`?

I'm aware of `f90cache`, but it seems like it is not being updated which means it's unusable with GCC 5.x and newer (cfr. https://github.com/ccache/ccache/issues/131#issuecomment-249509674).

melven:
Hi @boegel 

I'm sorry my changes cause problems.

The intention of merging only the Fortran 77 stuff in c0875cf was to avoid complex problems with ".mod" files in "modern" Fortran. However, gfortran seems to silently accept Fortran 90/95 and newer code.
In particular setting the language explicitly ("gcc -x f77" and "gcc -x f95") doesn't make any difference (which I didn't know/check): it always accepts Fortran 95+.

So I don't see any usable way to check that a file really contains Fortran 77 code.
@jrosdahl: That means to **revert c0875cf** as ccache shouldn't cause problems for code accepted by the compiler, right?
(The only possibility I figured out is grepping "gfortran -fdump-fortran-original" for "attributes: (MODULE" or grepping the preprocessed code for module lines, which doesn't seem suitable)

Regards
Melven

PS @ boegel:
Fyi if you are interested in working Fortran support with modules:
The problem shouldn't occur with *both* commits suggested above (the second one recreates ".mod" files by calling "gfortran -fsyntax-only"). This could be a start for a more sophisticated approach.
But you would have to convince jrosdahl about the increased complexity / maintenance...

boegel:
@melven Your suggestions try to predict whether any `.mod` files *would* be generated (meaning the compilation result shouldn't be cached), but how about just letting `ccache` *check* whether any `.mod` files were generated and then decide whether or not to actually cache the compilation result, is that an option?
It's OK if that would lead to false positives, the worse thing that can happen is that a compilation result where `.mod` files are suddenly detected is not being cached, but that's not the end of the world (well, a compilation failure isn't either, but you get the idea ;-)).

Thanks for the suggestion of patching the official `ccache` release with the other stuff you have here, I'll take a look at that.

melven:
This would also work, but requires to overwrite "-J" with a temporary directory and to detect and copy the generated files afterwards...

jrosdahl:
@boegel: Could you open a separate issue for this problem?

> This would also work, but requires to overwrite "-J" with a temporary directory and to detect and copy the generated files afterwards...

@melven: Do you think that this is the way forward or should we revert c0875cf as you mentioned earlier?

fcasson:
I have interest in the Fortran use case. From what I understand, the reason this is hard is that you need the dependency tree to be provided across all the source files that may interact via modules. This information should be available from cmake. builds. Has anyone looked into trying to get dependency graph out of cmake and pass it to ccache for the Fortran use case?

I also saw this tool: https://perso.univ-rennes1.fr/edouard.canot/f90cache/ but it also needs the dependency tree given to it. I'd prefer to use ccache as it looks a lot better maintained.

melven:
> I have interest in the Fortran use case. From what I understand, the reason this is hard is that you need the dependency tree to be provided across all the source files that may interact via modules. This information should be available from cmake builds. Has anyone looked into trying to get dependency graph out of cmake and pass it to ccache for the Fortran use case?
> 
> I also saw this tool: https://perso.univ-rennes1.fr/edouard.canot/f90cache/ but it also needs the dependency tree given to it. I'd prefer to use ccache as it looks a lot better maintained.

Hi,

your assumption is correct concerning the dependency tree.
I'm not as familiar with the CMake-Format of that dependency file.
One could also obtain this information from the compiler and store it in ccache - that would be independent of ccache.
Unfortunately, I don't have enough time to look into this currently. But we are also still interested in this feature (supporting Fortran).

This topic is closely related to C++20 modules, see https://github.com/ccache/ccache/issues/1252.<br>
So there will probably be more interest in C++20 modules and we can later adjust that implementation to make it work for Fortran, too...

PR #92, opened 2016-05-28
Better parsing of g* Options, and only use last argument

nolange:
Commonly some base configuration like CFLAGS="-O2 -g3 ..." will be
modified for builds by appending "-g0" to disable debug information.
This commit only en/disables options based on the last argument
that modifies debuginfo generation.

Also some more arguments are supported.

PS. Some more work, but it would be usefull to "normalize" the commandline,
so some permutations of the identical arguments wont evade the cache.
Eg. take only the last argument when possible (like with the -g options) and then
define a normalized order.

### DIFF
--- a/ccache.c
+++ b/ccache.c
@@ -2125,6 +2125,8 @@ cc_process_args(struct args *args, struct args **preprocessor_args,
 	char **argv;
 	bool result = true;
 	bool found_color_diagnostics = false;
+	int deb_level = 0;
+	const char *deb_argument = NULL;
 
 	expanded_args = args_copy(args);
 	stripped_args = args_init(0, NULL);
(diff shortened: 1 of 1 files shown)


ccache.c
```diff
@@ -2125,6 +2125,8 @@ cc_process_args(struct args *args, struct args **preprocessor_args,
 	char **argv;
 	bool result = true;
 	bool found_color_diagnostics = false;
+	int glevel = 0;
+	const char *gArgument = NULL;
```
jrosdahl: Please use variable names `like_this` instead of `likeThis`, just like the rest of the code.

ccache.c
```diff
@@ -2329,21 +2331,31 @@ cc_process_args(struct args *args, struct args **preprocessor_args,
...
-			if (str_eq(argv[i], "-g3")) {
-				/*
-				 * Fix for bug 7190 ("commandline macros (-D)
-				 * have non-zero lineno when using -g3").
-				 */
-				cc_log("%s used; not compiling preprocessed code", argv[i]);
-				conf->run_second_cpp = true;
+			const char *pLevel = argv[i] + 2;
+			int foundlevel = -1;
+			if (str_startswith(argv[i], "-ggdb"))
```
jrosdahl: Please add curly braces around `if` bodies even if it only contains one statement.

jrosdahl:
Looks like a good idea!

nolange:
Fixed the naming and coding conventions, hopefully thoroughly

PR #103, opened 2016-06-21
Fix some cppcheck style suggestions

scop:


itensionanders:
This change looks wrong. Anyway, I get different output from cppcheck:

```
[lockfile.c:134]: (information) Unmatched suppression: nullPointer
[manifest.c:458]: (error) Uninitialized variable: index
[manifest.c:477]: (error) Uninitialized variable: fi
[manifest.c:477]: (error) Uninitialized variable: index
[util.c:1085]: (style) The function 'GetFileNameFromHandle' is never used.
```

So maybe there are other things to fix, before "make cppcheck" is clean ?

jrosdahl:
Applied to 3.2-maint, thanks.

PR #104, opened 2016-06-21
Adding coverage support to memcached.

ikudriavtsev:


Jimilian:
@itensionanders, can you take a look at this PR?

itensionanders:
@Jimilian : looking good, would be nice if it could be made backwards compatible by allowing both sizes of values (with different magic) ? should probably add the dwo files too, while we're extending the structure (i.e. 6 values). but having coverage support is a very good idea, so we should add it I think. have some other changes, so will try to upload them tomorrow - and hopefully include this too.

jrosdahl:
@Jimilian: Any comments to @itensionanders's questions?

Jimilian:
@jrosdahl it's even not my changes - we need to ask @ikudriavtsev. But we are using this solution in "production" - and it looks good.

itensionanders:
I think it wouldn't be too hard to merge this with a "new" format (two extra fields). It's a bit icky to keep streting the "api" like that, but I suppose 4 or 6 doesn't matter - needs to be rewritten either way. And I think the new test needs to be made conditional on memcached, like the other ones are already. But the coverage feature is needed for us too, so it should be included in the main memcached branch (soon)

ikudriavtsev:
Hey @itensionanders , I didn't really get what you meant by "allowing both sizes of values (with different magic)"...
As for the dwo files - I think it's better going one step at a time, first coverage, then dwo.

itensionanders:
I was  thinking that the old format could be kept around, by introducing a new format with more fields. But maybe that is overkill, and we should invalidate the cache while introducing it in mainline.
Either way it is a good feature, and should be merged into `dev/memcached` after fixing the minors. It looks like there will be some more changes done (in #58) so it should be included with those...

jrosdahl:
Closing this PR due to timeout.

PR #107, opened 2016-07-01
Multiple -arch options support

ivaigult:
This pull request adds multiple "-arch" options support.

According to http://www.unix.com/man-page/osx/1/gcc/, -arch is APPLE only option for specifying target architecture. In case of specifying "-arch" multiple times a "fat" (all in one) binary will be created.

The idea of changes is as follows: 

1. For the preprocessor mode: run preprocessor multiple times (once per architecture) and then add each output to resulting hash
2. Direct mode remains the same, except minor changes with included_files variable initialization

### DIFF
--- a/ccache.c
+++ b/ccache.c
@@ -106,6 +106,11 @@ static char *output_dia = NULL;
  *  up). Contains pathname if not NULL. */
 static char *output_dwo = NULL;
 
+/* Vector for storing -arch options */
+#define MAX_ARCH_ARGS (10)
+static size_t arch_args_size = 0;
+static char* arch_args[MAX_ARCH_ARGS] = {NULL};
+
 /*
  * Name (represented as a struct file_hash) of the file containing the cached
  * object code.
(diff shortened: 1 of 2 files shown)


itensionanders:
This is a bit hard to test on other platforms, but for those who want to follow along in the code (that also has a bit of readability issues with whitespace etc) here is the wrapper code that the "Apple gcc" is running:
http://opensource.apple.com/source/gcc/gcc-5646/driverdriver.c

ivaigult:
@itensionanders, I am not really sure that it makes sense to test the changes on other platforms, except regression testing of course. I personally tried my changes on our OSX product in Jenkins and was satisfied with the results :)

itensionanders:
We do run a travis test* for OS X, but that depends on some (darwin-specific) tests for -arch flags being added to test.sh of course... (hint, hint) For some reason the variable is called `$HOST_OS_APPLE`.
* https://travis-ci.org/ccache/ccache/jobs/141590399

itensionanders:
@ivaigult : otherwise the approach looks fine (run cpp multiple times, cache the fat binary). But it would be nice to have test suite explicilty for arch support, that would test with e.g. `i386` and `x86_64` ?

ivaigult:
@itensionanders : Finally, it's there.

ivaigult:
@jrosdahl : Is it looking good for you? Will it be merged?

jrosdahl:
> Is it looking good for you? Will it be merged?

I haven't had time to look at it yet.

itensionanders:
@ivaigult : the test looks reasonable, can you rebase (-i) your branch into fewer commits please ?

jrosdahl:
Applied now with some cleanups. Thanks!

PR #114, opened 2016-07-14
support assembler

paboldin:
Support compiling files through the assembler:
$ gcc -S abc.c
$ gcc -c abc.s


jrosdahl:
Applied along with a test case. Thanks!

Issue #113, opened 2016-07-14
"-D not hashed" test failure with gcc 6.1.1


scop:
This is with ccache 3.2.6 on Fedora 24 and development, gcc 6.1.1:

```
CC='gcc' ./test.sh
compiler: /usr/lib64/ccache/gcc
version: gcc (GCC) 6.1.1 20160621 (Red Hat 6.1.1-3) Copyright (C) 2016 Free Software Foundation, Inc. This is free software; see the source for copying conditions. There is NO warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
test dir: testdir.8818
starting testsuite base
SUITE: "base", TEST: "-D not hashed" - Expected "cache hit (preprocessed)" to be 14, got 13
cache directory                     /builddir/build/BUILD/ccache-3.2.6/testdir.8818/.ccache
primary config                      /builddir/build/BUILD/ccache-3.2.6/testdir.8818/.ccache/ccache.conf
secondary config      (readonly)    /etc/ccache.conf
cache hit (direct)                     0
cache hit (preprocessed)              13
cache miss                            40
called for link                        2
called for preprocessing               2
multiple source files                  1
compiler produced stdout               1
couldn't find the compiler             1
bad compiler arguments                 1
unsupported source language            2
unsupported compiler option            1
output to a non-regular file           1
no input file                          1
files in cache                         3
cache size                          12.3 kB
max cache size                       5.0 GB
TEST FAILED
Test data and log file have been left in testdir.8818
Makefile:103: recipe for target 'test' failed
```

[Tarball of testdir.8818 contents](https://github.com/ccache/ccache/files/363209/testdir.8818.tar.gz)

itensionanders:
Hmm, I was thinking this to be fixed (#96 and a21800173c6d981a3b2d40c57934335a0b08f69f)

itensionanders:
Sure *looks* like the same issue:

```
$ gcc --version
gcc (GCC) 6.1.1 20160621 (Red Hat 6.1.1-3)
Copyright (C) 2016 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.

$ gcc -DNOT_AFFECTING=1 -E test1.c
# 1 "test1.c"
# 1 "<built-in>"
# 1 "<command-line>"
[...]

Issue #96, opened 2016-06-11
Test suite failure with gcc 6


jrosdahl:
From https://bugzilla.samba.org/show_bug.cgi?id=11709:

```
Test suite fails when building with gcc 6, Fedora development:

CC='gcc' ./test.sh
compiler: /usr/lib64/ccache/gcc
version: gcc (GCC) 6.0.0 20160129 (Red Hat 6.0.0-0.7) Copyright (C) 2016 Free Software Foundation, Inc. This is free software; see the source for copying conditions. There is NO warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
test dir: testdir.15690
starting testsuite base
SUITE: "base", TEST: "-D not hashed" - Expected "cache hit (preprocessed)" to be 14, got 13
cache directory                     /builddir/build/BUILD/ccache-3.2.4/testdir.15690/.ccache
primary config                      /builddir/build/BUILD/ccache-3.2.4/testdir.15690/.ccache/ccache.conf
secondary config      (readonly)    /etc/ccache.conf
cache hit (direct)                     0
cache hit (preprocessed)              13
cache miss                            40
called for link                        2
called for preprocessing               2
multiple source files                  1
compiler produced stdout               1
couldn't find the compiler             1
bad compiler arguments                 1
unsupported source language            2
unsupported compiler option            1
output to a non-regular file           1
no input file                          1
files in cache                         3
cache size                          12.3 kB
max cache size                       5.0 GB
TEST FAILED
Test data and log file have been left in testdir.15690
Makefile:82: recipe for target 'test' failed
```

itensionanders:
GCC 6 uses some new kind of magic number for the <command line>, making the comparison fail:

```
@@ -1,7 +1,8 @@
 # 1 "test1.c"
 # 1 "<built-in>"
 # 1 "<command-line>"
+# 31 "<command-line>"
 # 1 "/usr/include/stdc-predef.h" 1 3 4
-# 1 "<command-line>" 2
+# 32 "<command-line>" 2
 # 1 "test1.c"
 int foo10(int x) { return x; }
```

Anyway, definining macros (-DNOT_AFFECTING=1) sure does affect the output.

As in: making it normal again. But if you just run gcc, you get those fancy 31/32 ?

itensionanders:
Apparently the line number bug only affects gcc (without macros) and `.c`, not g++ and `.cc`

[...]

PR #115, opened 2016-07-15
Buggy cpp in gcc version 6

itensionanders:
This is an updated/corrected fix, for #96.


itensionanders:
Fixes #113.

jrosdahl:
Applied. Thanks!

PR #118, opened 2016-07-19
Add support for more compression types

afbjorklund:
This adds support for compression _type_, in addition to the compression _level_...
While doing so, it also makes the default level compatible by changing it to -1.

Included (and bundled) is support for the **[lz4f](http://lz4.org)** format, which is faster than gzip.
Also evaluated were [lzop](lzop.org) and [snappy](http://google.github.io/snappy/), but only the detection remains here.


jrosdahl:
Sounds like a good idea. Should we perhaps introduce support for different compression types as plugins for the same reason as discussed in #58?

itensionanders:
Yeah, probably ? Another reason for doing that, is that it could keep the "api" between different implementations sane - instead if the current copy/paste between different code forks ?

itensionanders:
And again, like #80, this could do with a more standardized benchmark for evaluating ?

jrosdahl:
Yes, would be nice.

afbjorklund:
Also thinking about just deleting the alternative compression (i.e. lzop and snappy)
And avoid the complexity about the plugin system, when it is not really needed...

The stand-alone lzop (minilzop and minilzo) is not really as stable, and not as fast.

http://www.lzop.org/

And the snappy both has all the C++ complexity, or the unofficial C port uncertainty.

http://google.github.io/snappy/

So think I will just drop their detection in the rebase, and avoid using any others.
Better to add the in-memory compression of manifests (using zlib) in that case...

afbjorklund:
Ouch, that rebase was _hard_. All the code changed (to C99) and with it all the comments (to //)...
Missed more than a few spots, but hopefully the tests caught most of it so that it's not broken now.

jrosdahl:
> Ouch, that rebase was hard.

Sorry about that.

A tip for the future: I have found [git-imerge](https://github.com/mhagger/git-imerge) to be very handy in situations like this.

afbjorklund:
Another major rebase, for 3.5 this time. Saved the old stuff as `compression-3.3`

afbjorklund:
Hmm, seems like Travis have upgraded to Clang version 5.0 ? (from 3.8)

afbjorklund:
Darnit, the Travis clang static analyzer is not the same as the regular one in clang 5.0.0
```
src/util.c:318:4: warning: Potential leak of memory pointed to by 'p'
                        cc_log("unknown size: %ld", (long) info.contentSize);
                        ^~~~~~
```
Time to clean up.

PR #119, opened 2016-07-20
Fix the MinGW tests

afbjorklund:
The MinGW target had some test failures, mostly due to Travis only compiling but not running them...

Was tested with MinGW and Wine: `./configure --host=i686-w64-mingw32 && make quicktest`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,7 +27,7 @@ matrix:
           before_install:
               - sudo apt-get -qq update
               - sudo apt-get install -y gcc-mingw-w64-x86-64
-          env: HOST="--host=x86_64-w64-mingw32" TEST="test/main.exe"
+          env: HOST="--host=x86_64-w64-mingw32" TEST="unittest/run.exe"
         - os: linux
           compiler: clang
           env: CFLAGS="-fsanitize=undefined" ASAN_OPTIONS="detect_leaks=0"
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,8 +8,6 @@ addons:
             - gperf
             - elfutils
             - zlib1g-dev
-            - mingw32
-            - mingw32-binutils
             - clang    # scan-build
 
 os:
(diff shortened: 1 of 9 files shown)


conf.c
```diff
@@ -481,8 +481,14 @@ conf_set_value_in_file(const char *path, const char *key, const char *value,
 		return false;
 	}
 
+#if defined (__MINGW32__) && (__MINGW32_MAJOR_VERSION < 5)
```
jrosdahl: Perhaps fix (i.e., move `#if ...` to) create_tmp_file instead?

util.c
```diff
@@ -1658,6 +1662,9 @@ x_readlink(const char *path)
 }
 #endif
 
+/* Avoid breaking the api */
```
jrosdahl: I would prefer to update the signature of `read_file` instead of relying on a global.

conf.c
itensionanders: Possibly. The whole thing is rather sad, really. Not as sad as printf, but sad.

util.c
itensionanders: Yeah, that would be a more reasonable solution. Perhaps refactor read_file/read_text_file ?

itensionanders:
It seems like `usleep` doesn't work on newer MinGW, it now requires fancy `nanosleep` instead.

itensionanders:
Also `printf`on MinGW (as in: not -w64) is *seriously* broken, and there are failures from hash command.

afbjorklund:
@jrosdahl : With the renewed interest in Windows, I updated to _64-bit_ MinGW and Wine.
That fixed some of the concerns regarding C99 or printf, also addressed your earlier review.

The cross-compilation and the quicktest runs now:
`PASSED: 480 assertions, 95 tests, 10 suites`

* https://travis-ci.org/ccache/ccache/jobs/216464039

util.c
```diff
@@ -1550,7 +1554,7 @@ read_file(const char *path, size_t size_hint, char **data, size_t *size)
 	}
 	size_hint = (size_hint < 1024) ? 1024 : size_hint;
 
-	int fd = open(path, O_RDONLY | O_BINARY);
+	int fd = open(path, O_RDONLY | (binary ? O_BINARY : O_TEXT));
```
jrosdahl: Looks like `binary` is always passed as false, so perhaps just hardcode `O_TEXT` instead?

Also, why do the mingw tests need `O_TEXT` instead of `O_BINARY`?

conf.c
jrosdahl: I don't want to merge it as is; we should try to keep `#ifdef`s in utility functions only.

afbjorklund:
Rebased, but didn't change anything

PR #125, opened 2016-08-25
Add missing ./autogen.sh to compilation steps

gosuwachu:


jrosdahl:
That list of commands is under the "Building code from a release archive" section, so it's intentional that `./autogen.sh` is not present. Under "Building code from the source code repository" there is this text at the end:

```
Run `./autogen.sh` and then follow the steps mentioned under "Installation"
above.
```

Do you have any suggestion on how to improve the structure of the document so that users will understand both use cases (building from a release archive and building from a source code repository)?

gosuwachu:
I think it is clear what is the difference, but I haven't noticed that there are two different sections. I just scanned the file for something that looked like build instructions.

If you look at other projects like (https://github.com/swig/swig) they use autoconf as well and have `./configure` checked into the repository. Maybe you could do the same, and then leave just one set of instructions.

jrosdahl:
> If you look at other projects like (https://github.com/swig/swig) they use autoconf as well and have ./configure checked into the repository. Maybe you could do the same, and then leave just one set of instructions.

Sorry, that's the wrong way to do it, so that won't happen. :)

After some more thinking, I suggest splitting the instructions into two separate instruction files:let INSTALL.md cover the "build from repo" case and let the "build from release archive case" instructions live in a less prominent place in the repo but be named INSTALL in the release archive.

gosuwachu:
Not sure why it is wrong if it works for other people - given swig is a popular project, this must mean people don't find problems with checked in ./configure :) Can you say what are your concerns?

jrosdahl:
> Not sure why it is wrong if it works for other people - given swig is a popular project, this must mean people don't find problems with checked in ./configure :) Can you say what are your concerns?

Sure, it works in the same way as having checked in pre-built object files or other generated files. My opinion is that source control is for storing source code (or other source files), not generated files.

jrosdahl:
I split the installation instructions in 2a2cf41. Please have a look.

Closing this one.

gosuwachu:
Thanks, looks good!

PR #126, opened 2016-08-27
[WIP] Re-enable ccache when -fmodules is detected, and add support for -ivfsoverlay option

gosuwachu:
Support for `-fmodules` has been disabled as a fix to this bug report: https://bugzilla.samba.org/show_bug.cgi?id=11048

I have tried to reproduce the issue on versions v3.2.1-3.2.5, and error reported in above bug report was caused by incorrectly handling `-isysroot` option in conjunction with `-fmodules`. `-isysroot` was not passed to the compiler when compiling preprocessed file, and this caused OS X module version to be used instead of iOS version. Workaround for this issue on v.3.2.1 is to define `CCACHE_CPP2`.

In latest ccache v3.3 `-isysroot` is being passed when compiling preprocessed output, so all I had to do is re-enable `-fmodules` and add support for `-ivfsoverlay`.

`-ivfsoverlay` overlays virtual file system on top of real file system. Used to map header file to different location.


gosuwachu:
I have just noticed preprocessor output doesn't change when the content of a module changes, and ccache uses cached version incorrectly. Work continues...

jrosdahl:
Thanks for working on this.

gosuwachu:
Preprocessor output when using `-fmodules` doesn't contain included headers - all include that have modules defined for them are replaced with `@import` syntax. This makes it impossible to use preprocessed file get list of included header files.

However dependency file does mention them. I see ccache parses preprocessor output to get list of included headers instead of dependency file - is there are reason to do that?

gosuwachu:
I have added support for parsing dependency files. I think ccache can be a little bit more intelligent when it detects `-fmodules` option, and automatically switch to use dependency file, or at least fail if dependency file is not being generated (when options `-MMD/-MD` are not present).

This is still work in progress, but seems to work for me. Feedback will be appreciated.

gosuwachu:
I still need to add more tests, but it it nearly there.

piotrwach:
Closed this PR, and re-opened new one here: #130

Issue #127, opened 2016-08-31
ccache no longer supports building of single -archs


markusthegeek:
When I run the following lines with ccache 3.3, I get the same armv7 binary two times instead of one for armv7 and one for arm64 with ccache 3.2.7. The same happens when building for simulator using two lines for i386 and x86_64.

It is great that ccache 3.3 is now supporting to build with multiple -arch flags in one line but old builds should not break.

run one time with ${target} = armv7 and one time with ${target} = arm64:
```ccache $(xcrun -sdk iphoneos -find clang) -x c++ -std=c++14 -O2 -arch ${target} -isysroot $PLATFORMPATH/$platform.platform/Developer/SDKs/$platform$SDKVERSION.sdk -miphoneos-version-min=$IPHONEOS_DEPLOYMENT_TARGET```

**Comment by jrosdahl** [comment]
@ivaigult: Could you have a look at this?

**Comment by timblechmann** [comment]
actually a serious regression:

first run:

```
[2016-09-01T08:41:00.687278 55207] === CCACHE 3.3 STARTED =========================================
[2016-09-01T08:41:00.687278 55207] Config: (default) base_dir = 
[2016-09-01T08:41:00.687278 55207] Config: (default) cache_dir = /Users/tim/.ccache
[2016-09-01T08:41:00.687278 55207] Config: (default) cache_dir_levels = 2
[2016-09-01T08:41:00.687278 55207] Config: (default) compiler = 
[2016-09-01T08:41:00.687278 55207] Config: (default) compiler_check = mtime
[2016-09-01T08:41:00.687278 55207] Config: (/Users/tim/.ccache/ccache.conf) compression = true
[2016-09-01T08:41:00.687278 55207] Config: (/Users/tim/.ccache/ccache.conf) compression_level = 3
[2016-09-01T08:41:00.687278 55207] Config: (default) cpp_extension = 
[2016-09-01T08:41:00.687278 55207] Config: (default) direct_mode = true
[2016-09-01T08:41:00.687278 55207] Config: (default) disable = false
[2016-09-01T08:41:00.687278 55207] Config: (default) extra_files_to_hash = 
[2016-09-01T08:41:00.687278 55207] Config: (default) hard_link = false
[2016-09-01T08:41:00.687278 55207] Config: (default) hash_dir = true
[2016-09-01T08:41:00.687278 55207] Config: (default) ignore_headers_in_manifest = 
[...]

PR #128, opened 2016-09-01
Fix for "ccache no longer supports building of single -archs #127"

ivaigult:
Hashing of -arch options was missed. Now it's there. Tests are updated.

### DIFF
--- a/ccache.c
+++ b/ccache.c
@@ -1740,6 +1740,12 @@ calculate_object_hash(struct args *args, struct mdfour *hash, int direct_mode)
 		free(gcda_name);
 	}
 
+	// Adding -arch to hash since cpp output is affected
+	for(size_t i = 0; i < arch_args_size; ++i) {
+		hash_delimiter(hash, "-arch");
+		hash_string(hash, arch_args[i]);
+	}
+
 	struct file_hash *object_hash = NULL;
 	if (direct_mode) {
 		// Hash environment variables that affect the preprocessor output.
--- a/test.sh
+++ b/test.sh
@@ -1119,19 +1119,62 @@ SUITE_multi_arch_PROBE() {
     fi
 }
 
+SUITE_multi_arch_SETUP() {
+    generate_code 1 test1.c
+    unset CCACHE_NODIRECT
+}
+
 SUITE_multi_arch() {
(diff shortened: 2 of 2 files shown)


timblechmann:
thanks for the quick fix ... will this justify a `3.3.1` hotfix release?

ivaigult:
@timblechmann I don't know if it will justify a hotfix release, but you may use `export CCACHE_NODIRECT=1` as workaround on the current release.

test.sh
```diff
@@ -1119,19 +1119,62 @@ SUITE_multi_arch_PROBE() {
...
+    $CCACHE_COMPILE -arch i386 -arch x86_64 -c test1.c
+    expect_stat 'cache hit (direct)' 1
+    expect_stat 'cache miss' 3
+    
     $CCACHE_COMPILE -arch i386 -arch x86_64 -c test1.c
+    expect_stat 'cache hit (direct)' 2
+    expect_stat 'cache miss' 3
+
+    #The same for preprocessor mode
+    clear_cache
```
jrosdahl: Please begin a new test here instead of explicitly clearing the cache.

test.sh
```diff
@@ -1119,19 +1119,62 @@ SUITE_multi_arch_PROBE() {
...
 }
 
+SUITE_multi_arch_SETUP() {
+    generate_code 1 test1.c
+    unset CCACHE_NODIRECT
+}
+
 SUITE_multi_arch() {
-    # -------------------------------------------------------------------------
+# -------------------------------------------------------------------------
```
jrosdahl: Revert this change?

ccache.c
```diff
@@ -1740,6 +1740,12 @@ calculate_object_hash(struct args *args, struct mdfour *hash, int direct_mode)
 		free(gcda_name);
 	}
 
+	// Adding -arch to hash since cpp output is affected
+	for(size_t i = 0; i < arch_args_size; ++i) {
```
jrosdahl: `for(` -> `for (`

jrosdahl:
Thanks, looks good.

jrosdahl:
Timed out waiting. No problem, I'll just do the tweaks myself. :-)

Thanks again for looking into this!

PR #130, opened 2016-09-15
Clang modules support

piotrwach:
This has been achieved by not passing '-fmodules' flag when generating preprocessor output, so that ccache can parse it to get list of included headers. Also, to get Clang modules working correctly, you have to set `run_second_cpp` to `true`, or define environment variable `CCACHE_CPP2`.

This is continuation of previous PR: #126

### DIFF
--- a/ccache.c
+++ b/ccache.c
@@ -2172,6 +2172,9 @@ cc_process_args(struct args *args, struct args **preprocessor_args,
 	// false. If run_second_cpp is true, they will be passed to the compiler as
 	// well.
 	struct args *cpp_args = args_init(0, NULL);
+	// clang_module_args contains clang module related arguments that should not
+	// be passed to preprocessor.
+	struct args *clang_module_args = args_init(0, NULL);
 	// dep_args contains dependency options like -MD. They only passed to the
 	// preprocessor, never to the compiler.
 	struct args *dep_args = args_init(0, NULL);
@@ -2318,6 +2321,23 @@ cc_process_args(struct args *args, struct args **preprocessor_args,
 		// We must have -c.
 		if (str_eq(argv[i], "-c")) {
 			found_c_opt = true;
+			continue;
+		}
+
+		// clang modules cause preprocessor output to not include header files,
+		// instead headers that have a module defined for them are replaced with
+		// '@import <module name>;', this causes preprocessor output to be useless
+		// for determining included headers. To workaround this problem, don't pass
+		// '-fmodules' when generating preprocessor output.
+		if (str_eq(argv[i], "-fmodules")) {
+			if (!conf->run_second_cpp) {
+				cc_log("Clang modules are not supported when run_second_cpp = false");
+				stats_update(STATS_UNSUPPORTED);
+				result = false;
+				goto out;
+			} else {
+				args_add(clang_module_args, argv[i]);
+			}
 			continue;
 		}
 
@@ -3034,6 +3054,8 @@ cc_process_args(struct args *args, struct args **preprocessor_args,
 	}
 
 	*compiler_args = args_copy(stripped_args);
+	args_extend(*compiler_args, clang_module_args);
+
 	if (conf->run_second_cpp) {
 		args_extend(*compiler_args, cpp_args);
 	} else if (explicit_language) {
--- a/ccache.c
+++ b/ccache.c
@@ -2329,7 +2329,7 @@ cc_process_args(struct args *args, struct args **preprocessor_args,
 		// '@import <module name>;', this causes preprocessor output to be useless
 		// for determining included headers. To workaround this problem, don't pass
 		// '-fmodules' when generating preprocessor output.
-		if (str_eq(argv[i], "-fmodules")) {
+		if (str_eq(argv[i], "-fmodules") || str_eq(argv[i], "-fcxx-modules")) {
 			if (!conf->run_second_cpp) {
 				cc_log("Clang modules are not supported when run_second_cpp = false");
 				stats_update(STATS_UNSUPPORTED);
--- a/ccache.c
+++ b/ccache.c
@@ -2332,7 +2332,7 @@ cc_process_args(struct args *args, struct args **preprocessor_args,
 		if (str_eq(argv[i], "-fmodules") || str_eq(argv[i], "-fcxx-modules")) {
 			if (!conf->run_second_cpp) {
 				cc_log("Clang modules are not supported when run_second_cpp = false");
-				stats_update(STATS_UNSUPPORTED);
+				stats_update(STATS_UNSUPPORTED_OPTION);
 				result = false;
 				goto out;
 			} else {
--- a/compopt.c
+++ b/compopt.c
@@ -56,7 +56,6 @@ static const struct compopt compopts[] = {
 	{"-arch",           TAKES_ARG},
 	{"-aux-info",       TAKES_ARG},
 	{"-b",              TAKES_ARG},
-	{"-fmodules",       TOO_HARD},
 	{"-fno-working-directory", AFFECTS_CPP},
 	{"-fplugin=libcc1plugin", TOO_HARD}, // interaction with GDB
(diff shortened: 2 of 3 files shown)


piotrwach:
I had to reimplement this after I realized that relaying on dependency headers to get list of include files doesn't always work. However, I think this change is now ready to go.

jrosdahl:
Thanks. Looks good from a distance. However, I don't have bandwidth to review this properly at the moment.

One question, though: Why do clang modules need to be tested on Darwin/macOS?

gosuwachu:
This is just the only environment where I have tested it. It should work with clang on any OS.

shepting:
I'm looking forward to when this lands.

jgavris:
👍  Me too

afbjorklund:
Probably the reason why it's not tested much on other platforms,
is that they don't have much demand for Objective-C (or modules).

Also, it will have some problems with the test dependencies:
```
Running test suite clang_modules.While building module 'Test1' imported from test1.m:1:
In file included from <module-includes>:1:
./test1.h:1:9: fatal error: 'Foundation/Foundation.h' file not found
#import <Foundation/Foundation.h>
        ^
test1.m:1:9: fatal error: could not build module 'Test1'
#import "test1.h"
 ~~~~~~~^~~~~~~~~
2 errors generated.
```
Once upon a time there was [GNUstep](http://www.gnustep.org/), wonder if it still works...

But basically this has the same generic problem as Solaris or Windows,
it is hard to test on the regular platform (i.e. Linux). Maybe _doable_, though ?

afbjorklund:
Yeah, apparently it is still available (`libgnustep-base-dev`).
But you pretty much _need_ to be using "gnustep-make"... :-(

http://www.gnustep.org/developers/documentation.html

``` make
#
# An example GNUmakefile
#

# Include the common variables defined by the Makefile Package
include $(GNUSTEP_MAKEFILES)/common.make

# Build a simple Objective-C program
TOOL_NAME = simple

# The Objective-C files to compile
simple_OBJC_FILES = simple.m

-include GNUmakefile.preamble

# Include in the rules for making GNUstep command-line programs
include $(GNUSTEP_MAKEFILES)/tool.make

-include GNUmakefile.postamble
```

Otherwise you need to mess with flags for gnustep-config/objc...
Instead of `GNUSTEP_MAKEFILES=/usr/share/GNUstep/Makefiles`

And that leads a long maze of little things like:
```/usr/include/GNUstep/Foundation/Foundation.h:31:9: fatal error: 'objc/objc.h' file not found```

Now, how to get `test.sh` to configure cross-platform and use makefiles...

**Comment by piotrwach** [comment]
I will update this PR to remove dependencies on Objective-C, as clang modules can be used for C++ as well.

**Comment by afbjorklund** [comment]
It was even worse than that, apparently the packaged GNUstep only supports GCC...
So one needs to build a custom installation, in order to be able to compile with Clang.

Maybe the test could just use a private framework, instead of dragging in Foundation ?
```#import "Foo/Foo.h"```

**Comment by afbjorklund** [comment]
Rewriting for C++ sounds like a good alternative.

**Comment by afbjorklund** [comment]
Weird, it will still use `@import` syntax even in C++.

```
# 1 "test1.cpp" 2
@import Test1; /* clang -E: implicit import for "./test1.h" */
```

> At present, there is no C or C++ syntax for import declarations. Clang will track the modules proposal in the C++ committee.

https://clang.llvm.org/docs/Modules.html

**Comment by jgavris** [comment]
I was under the impression it's a best practice to avoid using the explicit module import statement...progressive enhancemment. 

> Includes as imports
> 
> The primary user-level feature of modules is the import operation, which provides access to the API of software libraries. However, today’s programs make extensive use of #include, and it is unrealistic to assume that all of this code will change overnight. Instead, modules automatically translate #include directives into the corresponding module import. For example, the include directive
> 
> `#include <stdio.h>`
>
> will be automatically mapped to an import of the module std.io.

**Comment by afbjorklund** [comment]
@jgavris : that was the preprocessor (cpp) output, if it wasn't apparent...

``` C++
#include "test1.h"
int main() { return 0; }
```

``` shell
$ clang -fmodules -E test1.cpp
# 1 "test1.cpp"
# 1 "<built-in>" 1
# 1 "<built-in>" 3
# 325 "<built-in>" 3
# 1 "<command line>" 1
# 1 "<built-in>" 2
# 1 "test1.cpp" 2
@import Test1; /* clang -E: implicit import for "./test1.h" */
int main() { return 0; }

```

afbjorklund:
The poor old preprocessor is really out in the cold, the `.i` files is not getting much love these days.
(i.e using CCACHE_CPP2 was the first step, saying "we will build directly from sources thank you")

Now with \@ imports, it's not even going to be \# included but handled by a more sentient being instead.
Maybe it is slightly more natural to C++ or Objective-C, but it really is a different world from good old C...

The whole thing reminds me of D:
https://dlang.org/pretod.html

C++ is just hacking it, as usual. :-)
Now, back to Objective-C again...

gosuwachu:
I think clang modules support should be put behind 'sloppiness' flag, as ccache won't take into account changes in module maps that can potentially affect the output.

afbjorklund:
Apparently modules didn't make it into C++17, so I'm not sure about using it for C++ anymore.
Having it only for Objective-C makes sense, but would be nice with something not Foundation.

Agree with the sloppiness, it seems that the \@ import feature doesn't include any dependencies ?
It doesn't even have a timestamp, so it's pretty much impossible to know if it has changed or not.

Or do we need to start parsing modulemaps ?

piotrwach:
Modules work for both Objective-C and C++, so I think we should use C++ in tests, as it is easier to test on Linux.

We use modules for both C++ and Objective-C in our mobile app to optimize build times. In some cases we have seen built time improvements to up to 50%.

It is a shame that clang doesn't output path to module map when running preprocessor: 
`@import Test1; /* clang -E: implicit import for "./test1.h" */`

I have experimented with generating dependency files (`-MD`), paring them to find module maps, and add them to the hash. However, this required some significant changes.

afbjorklund:
I suppose I would need to look into what the C++ compiler actually _does_,
when it happens upon those `@module` statements...

Maybe this is a similar use case to the selective ignore that we added, in:
`ignore_headers_in_manifest` (https://github.com/ccache/ccache/pull/57)


Or if it actually does something clever, like keeping a semi-digested token cache ?
Still looks very out-of-place (in C++) to me, but maybe they'll change the syntax...

afbjorklund:
Ah, it sure did (cache them). Ok, then I see the interest for the .pcm (i.e. just like for the .pch)
`-fimplicit-module-maps -fmodules-cache-path=/tmp/org.llvm.clang.anders/ModuleCache`

sethfowler:
@piotrwach @afbjorklund it looks like this fell off the radar. What would it take to get it moving again?

afbjorklund:
Not sure it was ever on the main radar, since it was mostly useful for Mac/clang users.
Then again, if `-fmodules` (and friends) don't *hurt* anything for others - then why not...

@piotrwach : have you looked into the failing tests ?

gosuwachu:
I am not involved in the project that is using ccache with modules support anymore, and have limited amount of time. However, I may find some time to work on it next month. If you want it sooner, feel free to make the necessary changes mentioned in this PR. It should be max few hours worth of work.

afbjorklund:
Seems like the feature has some way to go before merge, then...

But we have seen from the .pch, that these things can get tricky.

linlinyao1:
Hi there, wondering if any progress is made in last a few months?

afbjorklund:
If support for modules is still important (like: is anyone using it, or is it just on macs), then I suppose we would need some more test cases and examples to make sure it actually works and doesn't hurt others ?

jrosdahl:
Is anybody interested in continuing this work? Otherwise I see no other alternative than to close the pull request.

jrosdahl:
Closing this now since nobody seems to be interested in working on the PR.

PR #132, opened 2016-09-19
Convert argument to '--sysroot' option to relative path

Mathiasdm:
Previously, only the argument to '--sysroot=' was converted.
This changeset adds support for '--sysroot'.


jrosdahl:
Applied, thanks!

PR #135, opened 2016-09-29
Support caching of .su file generated by GCC flag -fstack-usage

nordlow:


ccache.c
```diff
@@ -1898,6 +1957,10 @@ from_cache(enum fromcache_call_mode mode, bool put_object_in_manifest)
 		// The compiler won't generate notes if there is no code
 		get_file_from_cache(cached_cov, output_cov);
 	}
+	if (generating_stackusage && stat(cached_su, &st) == 0 && st.st_size > 0) {
+		// The compiler won't generate notes if there is no code
```
afbjorklund: I'm wondering if this applies to stackusage the way it did for coverage... (notes)

ccache.c
```diff
@@ -1242,6 +1277,23 @@ to_cache(struct args *args)
 		}
 	}
 
+	if (generating_stackusage) {
+		// GCC won't generate notes if there is no code.
```
afbjorklund: I'm wondering if this applies to stackusage the way it did for coverage... (notes)

ccache.c
```diff
@@ -1053,6 +1064,21 @@ to_cache(struct args *args)
 		tmp_cov = NULL;
 	}
 
+	char *tmp_su;
+	if (generating_stackusage) {
+                char *tmp_aux;
+                // GCC has some funny rule about max extension length.
```
afbjorklund: I'm wondering if this applies to stackusage the way it did for coverage... (gcno)

Really should dig down the GCC source code that generates these someday...

nordlow:
Ok. What to do?

afbjorklund:
Well, a test case would be nice ? (they're in the `test.sh` file)

I'm just looking at the https://gcc.gnu.org/onlinedocs/gcc/Developer-Options.html list for this :-)
But it looks it does use the _same_ auxname, but could be good to verify in the gcc source code.

nordlow:
Which test is most similar to check the existance of a created `.su` file?

Should add a separate test section or append to an existing?

afbjorklund:
Probably the ones for `-ftest-coverage` ? But you'll need to check for gcc, as well:
`
clang: error: unknown argument: '-fstack-usage'
`

afbjorklund:
Something like this perhaps: 339d9c351085d2eeabc550eab8fe04e01462f417 (squash into yours)

nordlow:
Ok, thanks! I'll check it out.

afbjorklund:
I meant that you could simply include the tests in your commit, if you want:
https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History#Squashing-Commits

nordlow:
Thanks! Is there a way to run only a single test?

nordlow:
> But you'll need to check for gcc, as well:
>  clang: error: unknown argument: '-fstack-usage' 
> 

Do you have an existing logic in ccache.c that already does these for other flags that are supported on GCC but not on Clang?

afbjorklund:
You can run a single test, with `./test.sh direct` (as shown by `make -n test`)
For the check, I only meant not failing the tests... (see $COMPILER_TYPE_GCC)

nordlow:
I've updated this PR with the test code you proposed. Are we ready to merge?

nordlow:
Ping.

afbjorklund:
Yeah, I think so. What do you think, @jrosdahl ?

Did have some minor remark whether the "notes" comment still applies (to stack usage),
the way that it did do coverage notes (a.k.a `.gcno`). It _looked_ a bit like copy/paste ?

jrosdahl:
Thanks for the patch, and thanks @afbjorklund for reviewing it!

Applied on master.

nordlow:
I don't see it merged on this master, is it your master @jrosdahl ?

jrosdahl:
Pushed now.

nordlow:
Thanks.

nordlow:
Caching of `-fstack-usage` is not included in 3.3.4. Have you forgotten to merge it?

jrosdahl:
> Caching of -fstack-usage is not included in 3.3.4. Why?

3.3.4 is a bug fix release cut from the 3.3-maint branch. -fstack-usage support is not a bug fix and is thus targeted for the next feature release (3.4), which is what master is for. See also the notes about branches on https://ccache.samba.org/repo.html.

> My company needs it!

Frankly, I don't understand what you want to accomplish with that statement.

Answer one: That's not how open source projects driven by volunteers work. Volunteers do what they like to do, and saying that your company needs something doesn't help your cause.

Answer two: Since you wrote the patch, your company already has access to it, right? Or are you only allowed to run official releases of open source software in your company?

nordlow:
I agree on both your points. I was in a bad mood today when I wrote that post. I'll use master for now. Thanks anyway for your interest.

jrosdahl:
OK, no problem!

PR #139, opened 2016-10-14
Only run the preprocessor once if ccache is called recursively

mostynb:
By passing the preprocessor hash to prefix_command, we can avoid recalculating the hash if ccache is called recursively (via prefix_command).

Example use-case: use ccache with a local cache dir, and CCACHE_PREFIX set to a shell script which unsets CCACHE_PREFIX and calls ccache again wth a large cache dir located on an NFS server.

This is just a prototype (eg no documentation yet), but I would like some feedback before continuing.  Some things still need to be worked out:
- Is it OK to force preprocessor mode when called recursively? (At the least, we should probably log when doing so if direct mode was specified.)
- Needs testing with run_second_cpp

### DIFF
--- a/ccache.c
+++ b/ccache.c
@@ -1405,10 +1405,8 @@ get_object_name_from_cpp(struct args *args, struct mdfour *hash)
 }
 
 static void
-update_cached_result_globals(struct file_hash *hash)
+update_cached_result_globals_common(const char *object_name)
 {
-	char *object_name = format_hash_as_string(hash->hash, hash->size);
-	cached_obj_hash = hash;
 	cached_obj = get_path_in_cache(object_name, ".o");
 	cached_stderr = get_path_in_cache(object_name, ".stderr");
 	cached_dep = get_path_in_cache(object_name, ".d");
@@ -1422,6 +1420,14 @@ update_cached_result_globals(struct file_hash *hash)
 	}
 
 	stats_file = format("%s/%c/stats", conf->cache_dir, object_name[0]);
+}
+
+static void
+update_cached_result_globals(struct file_hash *hash)
+{
+	char *object_name = format_hash_as_string(hash->hash, hash->size);
+	cached_obj_hash = hash;
+	update_cached_result_globals_common(object_name);
 	free(object_name);
 }
 
@@ -3221,6 +3227,15 @@ ccache(int argc, char *argv[])
 
 	cc_log("Object file: %s", output_obj);
 
+	const char *specified_object_hash = getenv("CCACHE_HASH");
+	if (specified_object_hash) {
+		update_cached_result_globals_common(specified_object_hash);
+		cc_log("CCACHE_HASH=%s was specified, checking if we have it", specified_object_hash);
+
+		// If we can return from cache at this point then do so.
+		from_cache(FROMCACHE_CPP_MODE, false);
+	}
+
 	struct mdfour common_hash;
 	hash_start(&common_hash);
 	calculate_common_hash(preprocessor_args, &common_hash);
@@ -3294,6 +3309,10 @@ ccache(int argc, char *argv[])
 	}
 
 	add_prefix(compiler_args, conf->prefix_command);
+	char *hash_name = format_hash_as_string(object_hash->hash, object_hash->size);
+	cc_log("Setting CCACHE_HASH=%s in case ccache is called recursively", hash_name);
+	setenv("CCACHE_HASH", hash_name, 1);
+	free(hash_name);
 
 	// Run real compiler, sending output to cache.
 	to_cache(compiler_args);


afbjorklund:
This sounds like something that could be accomplished by the secondary cache that we talked about earlier, if still wanting to hold on to NFS for storage (instead of going for something like memcached).

Maybe should get that code updated some day.

mostynb:
I can see a few references to secondary caches in the mailing list, but nothing recent, or that appears to have landed- is there a patchset/branch/pull request somewhere that you can point me to?  I suspect that my patch probably has the benefit of being comparatively simple.

Memcached doesn't sound like a good fit for my use-case: a team of developer pcs on a LAN,  where the devs work on a large codebase with enormous libraries which require a long time and a lot of memory to create and link.  Devs don't want to set aside a large chunk of memory for distributed caching- they need it for linking + IDE + debugger etc, and the pcs join and leave the (icecc) network at unpredictable times.  From a management point of view in this scenario an NFS cache is much simpler (though I have no experience with memcached, so perhaps I am wrong).

We currently use ccache+icecc, but due to the size of the code and the amount of churn + long-lived release branches, our caches are not as large or hot as we would like.  A tiered setup using this patch could be quite efficient:
1) check local ccache
2) check a (large) shared NFS ccache
3) ask icecc to compile

mostynb:
Hmm, problem: if CCACHE_CPP2=no then the prefix command contains a random temp file name, and so the second ccache will need to run the preprocessor again anyway.  I need to think this through some more.

mostynb:
Setting CCACHE_SLOPPINESS=file_macro for the second ccache skips hashing the file name, avoiding the problem mentioned above.  I'm not yet sure if time_macros should be added too.

Another, smaller, issue: what do if CCACHE_COMPRESS is enabled?  If the secondary ccache has this set, then the object files will be decompressed to be provided to the primary ccache, and if CCACHE_COMPRESS is enabled there also then the files will be re-compressed.  It might be nice to just return the compressed files to the primary ccache.

afbjorklund:
Yeah, there's some newer code and possibly a Pull Request coming as well. It does leave the files as compressed in the secondary cache. It works like you have outlined above:

> 1) check local ccache
> 2) check a (large) shared NFS ccache

The old code (for 3.1 and 3.2) can be seen at b028afdafba90b913038e3e0f71075ef9994c412 and 28d4068c794840987722688a92efff1664a3b9f5, if curious. It has been rewritten (and simplified) since. Will see about a cleanup and a public submission...

PS. The "mem" in memcached is at the server side.

mostynb:
Perhaps I can help out with the new/unpublished version of the secondary cache feature?

mostynb:
@afbjorklund: *ping*

afbjorklund:
@mostynb : sorry for the late reply, I was hoping to get some more traction on the memcached version and/or to clean up the NFS version for a public release. but neither happened, will keep trying though...

jrosdahl:
@afbjorklund: What's your current view on this pull request compared to your external and memcached branches?

afbjorklund:
@jrosdahl : I prefer the memcached storage, as originally sugged in #30 (now merged into #58) on technical merits of being the better implementation - over keeping the cache in files on a shared storage. Not everyone has access to NFS (or similar file storage), and it was not intended for such "hammering" of small files (requiring a lot of RAM to operate smoothly) and isn't as easy to distribute and to proxy (e.g. `moxi`).

However, it (memcached) does come with some downsides:
1) The branch, which by now is fairly large, first needs to be integrated upstream and released
2) The memcached servers needs to be kept running, and perhaps even be backed up to disk

When using NFS as a cache, it's a fairly small patch to add support for a "secondary" storage. And "someone else" is already making sure that _those_ servers are up and running (with all their overhead)
But I still think that it should be integrated into the code base, so that the user can choose primary and secondary (just like you can now on the `memcached` branch, with setting `memcached_only = false`)

Sadly, the code never got cleaned up for submission. But I will see what I can do about that next week.
By extension, one could consider _more_ storage backends ? For instance, I did try doing one in SQL...

``` SQL
CREATE TABLE IF NOT EXISTS ccache
               (file VARCHAR(52) PRIMARY KEY, data LONGBLOB, time DATETIME)
```

mostynb:
I finally have a good excuse to return to this topic...

One possible use case for this functionality would be in a distributed fashion in conjunction with something like icecc: cache is run on the client, on cache miss the request is sent to icecc, which in turn checks its own ccache (avoiding recalculating the hash) before scheduling a compile job if this lookup also misses.

@afbjorklund wrt the memcached branch- are the cached objects stored in memory, or just the hashes?  The codebase I'm working with (Chromium) is very large and has an insane amount of churn, and probably wouldn't scale well enough if the objects themselves are stored in memory.  I just noticed that you're based in GBG also, in case you're available to chat over a beer sometime...

afbjorklund:
@mostynb : The entire cache is stored in memcached, including the objects. In order to overcome the 1M limit of memcached, they are split into "small pieces" and then reassembled. But you can also have a multi-level cache, where the most recently used objects are kept in your local memory cache and the others are forwarded to a potentially larger remote cache - that can even be disk-backed if needed.

A face-to-face meeting sounds great, send me a PM and we can schedule something.

jrosdahl:
Have you drunk any beer yet?

mostynb:
Well yes, but not together yet :)  I will try to get organised this week.

The good news is that some related projects have appeared in the meantime, and I think it's worth investigating more ambitious changes.  For example, if ccache was cleanly split into a frontend (input hashing) and backend (LRU storage), then it would be easy to plug in any of several backends (the existing local fs storage, memcached, bazel-remote, etc).

Marking this issue closed, since I would not bother implementing this idea by itself.

afbjorklund:
Well, I still think it's a good idea and the current solution with ccache/distcc exists and is working...
For memcached, there is now an "extstore" available to store the keys in ram and the values on ssd.

* https://github.com/memcached/memcached/wiki/Extstore

afbjorklund:
@mostynb : we discussed making the storage and compression pluggable, but it added complexity too...
afaict, that "[bazel-remote](https://docs.bazel.build/versions/master/remote-caching.html)" is a simple HTTP service ? (unfortunately they removed the [Hazelcast](https://hazelcast.org/) cache)

We will open a new story about the multi-level storage, even the memcached backend uses that.

mostynb:
I looked at memcached's experimental extstore setup, but IIUC it falls short for ccache's use case in that storage is still lost when memcached restarts.  You could setup a secondary system to pre-warm memcached (eg netflix does this[*]), but it sounds annoying.

[*] https://medium.com/netflix-techblog/evolution-of-application-data-caching-from-ram-to-ssd-a33d6fa7a690

Yes, bazel-remote is a simple HTTP service, fairly well stress tested apparently.  See also https://github.com/mozilla/sccache for a similar setup with multiple backends.

afbjorklund:
I guess couchbase is still the easiest solution, if you want persistence for your memcached.

PR #145, opened 2016-11-01
Support NVCC '-ccbin' argument

### LINKED ISSUE #144: Support nvcc (CUDA) -ccbin/--compiler-bindir option
The `-ccbin` argument specifies the path to the host compiler or the directory in which it resides. Currently, ccache treats the argument as a source file and bails out with "unsupported language."

colesbury:
Fixes #144

### DIFF
--- a/ccache.c
+++ b/ccache.c
@@ -1770,6 +1770,16 @@ calculate_object_hash(struct args *args, struct mdfour *hash, int direct_mode)
 			continue;
 		}
 
+		if ((str_eq(args->argv[i], "-ccbin")
+		     || str_eq(args->argv[i], "--compiler-bindir"))
+		     && i + 1 < args->argc
+		     && x_stat(args->argv[i+1], &st) == 0) {
+			hash_delimiter(hash, "ccbin");
+			hash_compiler(hash, &st, args->argv[i+1], false);
+			i++;
+			continue;
+		}
+
 		// All other arguments are included in the hash.
 		hash_delimiter(hash, "arg");
 		hash_string(hash, args->argv[i]);
--- a/compopt.c
+++ b/compopt.c
@@ -30,6 +30,7 @@ struct compopt {
 };
 
 static const struct compopt compopts[] = {
+	{"--compiler-bindir", TAKES_ARG | TAKES_PATH},
 	{"--param",         TAKES_ARG},
 	{"--save-temps",    TOO_HARD},
 	{"--serialize-diagnostics", TAKES_ARG | TAKES_PATH},
@@ -56,6 +57,7 @@ static const struct compopt compopts[] = {
 	{"-arch",           TAKES_ARG},
 	{"-aux-info",       TAKES_ARG},
 	{"-b",              TAKES_ARG},
+	{"-ccbin",          TAKES_ARG | TAKES_PATH},
 	{"-fmodules",       TOO_HARD},
 	{"-fno-working-directory", AFFECTS_CPP},
 	{"-fplugin=libcc1plugin", TOO_HARD}, // interaction with GDB


jrosdahl:
Thanks, looks reasonable. However, I see an edge case: If `-ccbin` specifies a directory and `compiler_check` is `content` then the content of the directory node is hashed. I guess it works, but is it the wanted behavior?

How does NVCC find the compiler to run if `-ccbin` specifies a directory?

jeremad:
If this help in any way, it works like a charm on my big complicated CUDA project :+1:

lukeyeager:
I'm using this branch for Caffe2 builds, both locally and on TravisCI: https://github.com/caffe2/caffe2/pull/735

I have requested CMake support for ccache+nvcc here, in anticipation of this being merged.
https://gitlab.kitware.com/cmake/cmake/issues/16953

lukeyeager:
CMake now supports `-DCMAKE_CUDA_COMPILER_LAUNCHER=ccache`. It will be in v3.9.
https://gitlab.kitware.com/cmake/cmake/merge_requests/949

Thanks @bradking!

bradking:
> It will be in v3.9.

Actually it will be in CMake 3.10 (and is already available in nightly binaries).  We're already in the release candidate cycle for CMake 3.9 so it's too late for new features in general.

colesbury:
@lukeyeager, do you know how `nvcc` chooses the compiler if `-ccbin` is a directory?

lukeyeager:
Here's the `--help` information for that flag with the CUDA 8.0 compiler:
```
--compiler-bindir <path>                   (-ccbin)                          
        Specify the directory in which the host compiler executable resides.  The
        host compiler executable name can be also specified to ensure that the correct
        host compiler is selected.  In addition, driver prefix options ('--input-drive-prefix',
        '--dependency-drive-prefix', or '--drive-prefix') may need to be specified,
        if nvcc is executed in a Cygwin shell or a MinGW shell on Windows.
```
I'm not sure what that means, exactly. Let me go find one of the compiler guys ...

lukeyeager:
When you pass a directory to `-ccbin`, `nvcc` is going to look for a binary matching the name of the standard compiler on your platform in that directory (`gcc` on linux, probably `clang` on osx, and whatever the name of the binary is for the msvc compiler on windows). I believe that without the `-ccbin` flag, nvcc looks for the same binary, but looks for it in `$PATH`.

But, as @bas-aarts pointed out, the compiler is often dynamically linked to libraries outside of that directory, and may even call out to other backend executables, etc. So just getting a checksum of the binary isn't strictly safe. How does `ccache` typically handle this? It might be a good enough solution to just use the last modified date of the directory.

bas-aarts:
Assuming no-one intentionally messes with their compiler installation, a checksum on the version string should suffice.

afbjorklund:
Sounds like this flag could be related to the $CCACHE_PATH, then ?

You can use a custom command to $CCACHE_COMPILERCHECK,
if you feel that mtime (or content, if paranoid) is not enough a check.

lukeyeager:
Seems like this strategy should be sufficient:
```sh
if [ -f "$ccbin" ]; then
    # File: md5sum
    ccache_hash="$(md5sum "$ccbin" | awk '{print $1}')"
elif [ -d "$ccbin" ]; then
    # Directory: mtime
    ccache_hash="$(stat -c %Y "$ccbin")"
else
    echo "'$ccbin' is not a file or directory"
    exit 1
fi
```

colesbury:
@jrosdahl, what do you think the best strategy is? Should we:

1. keep the behavior in the PR (hash the directory node)
2. guess the which compiler NVCC will use (gcc/clang/msvc)
3. bail out if -ccbin refers to a directory

jrosdahl:
> @jrosdahl, what do you think the best strategy is? Should we:
>
> 1. keep the behavior in the PR (hash the directory node)
> 2. guess the which compiler NVCC will use (gcc/clang/msvc)
> 3. bail out if -ccbin refers to a directory

I'd say 2 or 3. Would it be hard to guess the compiler?

seeraven:
@jrosdahl: Yes that's right.

jrosdahl:
Applied the remaining parts now, thanks!

Also, I took a shot at option 2 discussed previously ("guess the which compiler NVCC will use (gcc/clang/msvc)") in 9c051cc60a13a70e0ddbd278662cd2ad37a18184. @colesbury and/or @seeraven: I would appreciate it if you could review my commit.

jrosdahl:
Closing this now. (A review of my commit would still be appreciated...)

Issue #142, opened 2016-10-27
"compiler produced stdout" stat undocumented.


phraemer:
Hi,

In the stats there is a line "compiler produced stdout"
I can find no explanation in the documentation or elsewhere of what it means exactly and what should be done to avoid this.

thanks!
James

afbjorklund:
Normally the compiler produces output in the provided object file (and friends, like dependency files), and warnings/errors on stderr. If there is anything on stdout, then something is "broken" inside the compiler.

phraemer:
I see. I my case I'm using `distcc` in pump mode and it's normal for pump to output to stdout.

e.g.
```bash
ccache pump distcc clang -v -c main.c -o main.o
```

This will output the following to stdout...
```
__________Using distcc-pump from /usr/local/bin/../Cellar/distcc/3.2rc1/bin
__________Using        1 distcc server in pump mode
__________Shutting down distcc-pump include server
```

ccache will then report "compiler produced stdout", fallback to the compiler.

Is this really a problem with pump or should ccache accept output to stdout, or at least be configurable?

afbjorklund:
I think there is an integration problem... You also really want the "cpp2" mode (since pump mode doesn't handle preprocessed source code very well), but since that is now the default (since 3.3) it _should_ work.

i.e. we should improve on the current "state of affairs" (from distcc's doc):

> Use with ccache
> ---------------
> 
> The best way to use is to set up a similar masquerade directory for
> ccache and put it on the PATH before distcc.
> 
> NOTE: This use of ccache is incompatible with use of distcc's "pump" mode.
> 
> (If you're using "pump" mode, it might be possible to use ccache on
> the distcc server machines. But we haven't tried that setup.)
> 

Using distcc in non-pump (i.e. normal) mode should work OK with ccache 3.1-3.2, once you get rid of some of all the bugs with configuration in those old distcc versions... Hopefully distcc 3.3 will be better.

If you decide to use ccache only on the distcc servers, then you really want those to share a cache...

----
Ultimately, one would like both the cache server and the include server to be started up *once*. Then _all_ the compilations for the current build would be performed against those, instead of stopping/starting.

Basically the cache server (manifest memcached) only stores the timestamps and checksums of files, whereas the include server (distcc include_server.py) also stores the compressed contents of them.

But that's all work-in-progress...

afbjorklund:
[...]

PR #146, opened 2016-11-01
Allow ccache to work with distcc's "pump" wrapper

afbjorklund:
The recommended way to use distcc is to start the include_server.py just once, but make it work with the `distcc-pump` wrapper too... It outputs some noise to standard output, that needs to be filtered out.

```
For example:
      eval `pump --startup`
      make -j80
      pump --shutdown
```

In order for the local machine to not get overrun by all the preprocessing, it is important to set `CCACHE_PREFIX_CPP=distcc` and patch bugs about setting --localslots_cpp (distcc/distcc#161).

### DIFF
--- a/ccache.c
+++ b/ccache.c
@@ -700,7 +700,7 @@ make_relative_path(char *path)
 // - Stores the paths and hashes of included files in the global variable
 //   included_files.
 static bool
-process_preprocessed_file(struct mdfour *hash, const char *path)
+process_preprocessed_file(struct mdfour *hash, const char *path, bool pump)
 {
 	char *data;
 	size_t size;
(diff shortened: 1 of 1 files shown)


afbjorklund:
Would fix issues reported in #142, by not writing to stdout anymore.

jrosdahl:
Looks good, although it's an unfortunate hack.

I think that pump should print such progress messages to `UNCACHED_ERR_FD` just like distcc does.

afbjorklund:
Yeah, fixing it in distcc would probably be the better option - but it is hard to do retroactively (a lot of broken `distcc-pump` out there), and there's not much development/maintenance ongoing in distcc (sadly).

Here is the culprit: https://github.com/distcc/distcc/blob/master/pump.in

afbjorklund:
Note that using pump mode is now *required*, when using the cpp2 option (not using the preprocessor).

jrosdahl:
Applied on master, thanks!

PR #147, opened 2016-11-03
Add test for Clang static analyzer

afbjorklund:
Run CI with Clang static analyzer:

http://clang-analyzer.llvm.org/


jrosdahl:
Applied on master, thanks!

PR #155, opened 2017-01-26
ccache.c : add --show-stats-regularly=SEC to watch the stats grow

jimklimov:
Helps verify if the build setup is correct and ccache is indeed involved.

### DIFF
--- a/ccache.c
+++ b/ccache.c
@@ -62,6 +62,7 @@ static const char USAGE_TEXT[] =
   "    -o, --set-config=K=V  set configuration key K to value V\n"
   "    -p, --print-config    print current configuration options\n"
   "    -s, --show-stats      show statistics summary\n"
+  "    -S NUM, --show-stats-regularly=NUM      refresh stats summary every NUM sec\n"
   "    -z, --zero-stats      zero statistics counters\n"
   "\n"
   "    -h, --help            print this help text\n"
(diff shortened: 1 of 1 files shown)


jrosdahl:
Thanks for the suggestion! However, I'm not interested in including such functionality in ccache since it's already easy (and better) to use `watch` (assuming you're using some kind of Unix derivative), e.g. like this:

    watch -n1 -d ccache -s

jimklimov:
I am not sure that the `watch` solution is a good fit (better that nothing, although a shell loop can get similar effect on any platform) as it means that a `ccache` process will be forked and spawned every second (with forking being an expensive procedure that we want to avoid, especially on systems with more security checks enabled) - so a single-process solution still has its benefits IMHO.

By the way, is there any inherent problem in proposed approach - such as that this persistent watching process might lock the cache database and block other `ccache` processes from doing their work? Or on the other hand that it would get stale/inconsistent data without locking a database?

afbjorklund:
Maybe a better long-term option would be to refactor ccache into a program and a library, then other things (such as stats reporting, for instance) could use this library to connect to the stats database ?

jimklimov:
Refactor - maybe it has benefits, like making a daemon that uses the DB and clients that talk to it (in parallel if need be, and remotely maybe). But this is beneficial mostly if there is an issue with DB-locking (or conflict otherwise) in the first place - and I don't know about ccache internals enough to claim either way :)

afbjorklund:
Well, it is up to @jrosdahl as the maintainer anyway. I am using Ganglia to monitor, and it could use a library. For now I just made a simple wrapper in python (forking `ccache`). But that works too, I suppose.

afbjorklund:
I also thought about doing a machine-readable output, but ended up parsing the human-readable output instead (out of laziness). My main concern is that I really wants stats _per build job_, so have to parse log...

jrosdahl:
(Sorry for the delayed follow-up, my ccache backlog is still a bit large.)

> By the way, is there any inherent problem in proposed approach - such as that this persistent watching process might lock the cache database and block other ccache processes from doing their work? Or on the other hand that it would get stale/inconsistent data without locking a database?

No, the approach looks fine, it's just that I wanted to avoid feature creep (although it's very small in this case). I just didn't see why one would like to use `--show-stats-regularly` instead of `watch`.

> with forking being an expensive procedure that we want to avoid, especially on systems with more security checks enabled

...but this sounds a bit surprising. If it's noticeably expensive to fork a ccache process to print stats on such systems, isn't it then too expensive to use ccache for compilations?

jimklimov:
Hi, found this branch lying around in my workspace, and found the PR not merged :)

Just to clarify: all forking is expensive, variably on different OSes (stuff like RBAC and SELinux add a lot to computation time to allow or deny making a new process, set up labels or cgroups, etc. for example) as opposed to spawning threads in an existing process.

On a diverse CI farm, I do see old Linux systems perform a `configure` script run like twice as fast as a modern and better secured OS - even if the modern host runs the build with local storage, and the oldie uses the same storage over NFS and takes a hit due to that and still comes out faster. Still, with all the scripting involved there and maybe thousands of forks along the way, it adds up to a lot, measurably.

PR #157, opened 2017-02-15
Removed time_of_compilation check wrt SLOPPY_FILE_STAT_MATCHES

svpv:
The whole code seems to be a thinko.  For a hit, neither ctime
nor mtime should be greater than or equal to time_of_compilation.
The code only seems to work because time_of_compilation is 0
at this stage (i.e. has not been initialized).

While at it, I also introduce an optimization: when sizes do
not match, it's a good chance to bail out early; there is no
point in further hashing the file.

### DIFF
--- a/ccache.c
+++ b/ccache.c
@@ -1869,12 +1869,6 @@ from_cache(enum fromcache_call_mode mode, bool put_object_in_manifest)
 		return;
 	}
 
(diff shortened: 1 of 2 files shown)


jrosdahl:
Applied on 3.3-maint, thanks!

PR #160, opened 2017-03-01
POC add ignore_implicit_pch sloppiness flag

RedBeard0531:
I'm filing this to start a conversation rather than to get it merged in, so there are no docs or test yet.

I've been trying out ccache with pch-enabled compilations and there seem to be two posibilities, both with strong enough downsides that they don't seem viable:

* Use sloppiness=pch_defines when compiling the pch. This causes compiles that use the pch file to use the cache even if #defines in the headers have changed. This is a non-starter for any project using config headers.
* Don't cache the pch file but cache the compilations that use it. This is problematic because both clang and gcc embed timestamps in the pch file so the hashes are never the same, so the cache becomes dead every time you recompile the pch file.

This patch proposes a third option very similar to the second. The difference is that when the ignore_implicit_pch flag is used, ccache will not go looking for pch files, and will pretend that the compiler is just using the normal headers. This relies on the build system to ensure that the pch files accurately reflect the headers before calling into ccache (which is why it is sloppiness-guarded), but the build system wouldn't really work if it didn't do that.

I've tested locally with gcc 6.3.1 and clang 3.9.1. Both will output the full preprocessed file when using -E as long as you *don't* include -fpch-preprocess. This means that ccache will be able to use the non-direct mode. However I haven't tested older compilers. I have tried builds where I change #defines in the config header (which is included in the pch) and with this flag, ccache seems to be caching correctly and rebuilding only when it is getting an actual new value.

Does anyone know what downsides this approach has, other than relying on the build system to keep the pch file up to date? I would assume that ccache is seeking out implicit pch files and including them in the hash for a good reason, but I don't know why. Did some older compilers not include the real output with -E when using a pch file?

Thanks for considering this. If you like this approach either feel free to take it and run with it, or I can put in the work to document, test, etc.

### DIFF
--- a/ccache.c
+++ b/ccache.c
@@ -2103,7 +2103,7 @@ detect_pch(const char *option, const char *arg, bool *found_pch)
 			cc_log("Detected use of precompiled header: %s", arg);
 			pch_file = x_strdup(arg);
 		}
-	} else {
+	} else if (!(conf->sloppiness & SLOPPY_IGNORE_IMPLICIT_PCH)) {
 		char *gchpath = format("%s.gch", arg);
 		if (stat(gchpath, &st) == 0) {
 			cc_log("Detected use of precompiled header: %s", gchpath);
--- a/ccache.h
+++ b/ccache.h
@@ -69,6 +69,7 @@ enum stats {
 // Allow us to not include any system headers in the manifest include files,
 // similar to -MM versus -M for dependencies.
 #define SLOPPY_NO_SYSTEM_HEADERS 64
+#define SLOPPY_IGNORE_IMPLICIT_PCH 128
 
 #define str_eq(s1, s2) (strcmp((s1), (s2)) == 0)
 #define str_startswith(s, prefix) \
--- a/conf.c
+++ b/conf.c
@@ -116,6 +116,8 @@ parse_sloppiness(const char *str, void *result, char **errmsg)
 			*value |= SLOPPY_PCH_DEFINES;
 		} else if (str_eq(word, "time_macros")) {
 			*value |= SLOPPY_TIME_MACROS;
+		} else if (str_eq(word, "ignore_implicit_pch")) {
+			*value |= SLOPPY_IGNORE_IMPLICIT_PCH;
 		} else {
 			*errmsg = format("unknown sloppiness: \"%s\"", word);
 			free(p);
--- a/conf.c
+++ b/conf.c
@@ -641,6 +641,9 @@ conf_print_items(struct conf *conf,
 	if (conf->sloppiness & SLOPPY_NO_SYSTEM_HEADERS) {
 		reformat(&s, "%sno_system_headers, ", s);
 	}
+	if (conf->sloppiness & SLOPPY_IGNORE_IMPLICIT_PCH) {
+		reformat(&s, "%signore_implicit_pch, ", s);
+	}
 	if (conf->sloppiness) {
 		// Strip last ", ".
 		s[strlen(s) - 2] = '\0';


jrosdahl:
> I'm filing this to start a conversation rather than to get it merged in, so there are no docs or test yet.

Thanks!

> I would assume that ccache is seeking out implicit pch files and including them in the hash for a good reason, but I don't know why. Did some older compilers not include the real output with -E when using a pch file?

Are you talking about

https://github.com/ccache/ccache/blob/76a9959f3d3e17acd9f344e416ed6929d303083f/ccache.c#L636-L646? If so, that code is run to save a hash of the pch content for later usage when ccache tries direct mode lookup.

My memory has unfortunately faded a bit regarding precompiled headers... But if I apply your changes to latest master, I get the following test suite error:

```
Running test suite pch......
FAILED

Test suite:     pch
Test case:      Use .gch, no -fpch-preprocess, -include, no sloppiness
Failure reason: Expected "can't use precompiled header" to be 1, actual 

ccache -s:
cache directory                     /home/joel/code/ccache/ccache/testdir.29280/.ccache
primary config                      /home/joel/code/ccache/ccache/testdir.29280/ccache.conf
secondary config      (readonly)    
stats zero time                     Wed Jan 31 21:44:10 2018
cache hit (direct)                     0
cache miss                             0
cache hit rate                      0.00 %
preprocessor error                     1
cleanups performed                     0
files in cache                         0
cache size                           0.0 kB
```

jrosdahl:
@RedBeard0531: Any news on test and docs?

llunak:
>     * Don't cache the pch file but cache the compilations that use it. This is problematic because both clang and gcc embed timestamps in the pch file so the hashes are never the same, so the cache becomes dead every time you recompile the pch file.

This is quite easy to work around. Clang has -Xclang -fno-pch-timestamp and for GCC it's possible to use CCACHE_PCH_EXTSUM and instead of making a checksum of the PCH do a checksum of the result of a preprocessing run on the source used to create the PCH. That's how LibreOffice handles it.

Still, that requires tweaks to the build, and this looks like a nice simple way of handling the problem.

jrosdahl:
@llunak: Since you're interested in precompiled headers, would you like to continue @RedBeard0531's work on this?

llunak:
I've been actually looking into this. The basic problem (for me?) is that GCC reports incomplete -MD dependencies if a PCH is used, the first dependency is always the PCH and anything that comes from it is skipped. Which breaks the depend mode, and I have already switched to the depend mode. Clang would work, but there a much simpler solution is -fno-pch-timestamps, the GCC workaround is more complicated, but this feature would help only with the preprocessing mode.

Which for me puts this into the theoretically-nice-but-not-useful-to-me category. I still could become interested/bored enough to do this one day, so it'd be nice if you kept this open even without activity if it's not a problem, but no promises.

jrosdahl:
> Which for me puts this into the theoretically-nice-but-not-useful-to-me category. I still could become interested/bored enough to do this one day, so it'd be nice if you kept this open even without activity if it's not a problem, but no promises.

Sure, let's keep it open for a while longer then.

llunak:
I've decided I'm not going to implement this, so I suggest closing this.

The reasoning for my decision:
* GCC: It would work only in non-direct mode, which is the least-performant one. And for the timestamp issue there is the CCACHE_PCH_EXTSUM workaround described above. So, frankly, I couldn't be bothered.
* Clang: -fno-pch-timestamps is a much simpler solution. Moreover, with Clang the PCH actually may matter. If the PCH is created with the -building-pch-with-obj flag, the PCH will be flagged that there will be a dedicated .o file that will contain shared implementations from the PCH, so that the .o files created using the PCH do not need to emit duplicates (in the style of how MSVC PCH have an extra .obj file). There is no extra compiler option, it is all controlled by the flag in the PCH. Which means that ignoring the PCH could cause undefined references if this setting changed (as much as that's unlikely to happen). So again, IMO just not worth the effort.

jrosdahl:
> I've decided I'm not going to implement this, so I suggest closing this.

OK. Thanks a lot for explaining your reasoning!

Issue #122, opened 2016-07-23
Fix and re-enable mingw build


jrosdahl:
I disabled the i586-mingw32msvc build in ddf5ca4d59f2e0ccbc85b53b3692d9a0f34b5f6b since GCC 4.2 apparently is too old for proper C99 support, which is required since e527cccb.

@itensionanders: Since you have been working on the mingw build before, do you have any ideas on what would be the best approach? Would it be possible to go back to using gcc-mingw-w64-i686? Or should we make some ugly workaround like (if possible) ignoring the specific error?

itensionanders:
We can require MinGW-w64, it is not the default on (non-beta) Travis but the old MinGW is horribly broken in other ways (as seen in #119). We probably also need some **real** Windows tests (#69), if we want that OS to be a supported platform ? So I can go with either disabling MinGW or requiring newer "stuff". But having native Windows support is *not* important to me personally. Supporting Linux/Unix is OK. I used [MSYS](http://www.mingw.org/wiki/msys) anyway...

* http://mingw.org
* http://mingw-w64.org

jrosdahl:
I have no personal interest in Windows support either, so let's keep it disabled then until somebody who cares enough shows up. :-)

itensionanders:
Note that I don't have any objections to keeping the *code* compiling with MinGW in the name if portability - that should be doable, possibly after upgrading due to the new GCC requirements...

The problem is keeping the "quicktest" and "test" running, on a foreign system like Windows ? That would probably need "real" Windows (not Wine) and a maintainer plus something like Appveyor

itensionanders:
@jrosdahl : but if you just want the tests running again, you can look at ec41946803b604706e1371c65b15be372373da7e (mingw-w64)

mati865:
While it's a bit outdated you can take a look at https://github.com/Alexpux/MINGW-packages/tree/master/mingw-w64-ccache

afbjorklund:
The latest changes in #119 will make the MinGW build [run again](https://travis-ci.org/ccache/ccache/builds/216464035), using the Travis VM infrastructure...
So it will run MinGW twice as slow as the other container-based builds, but still half the time of ASAN.

The tests pass on Linux, but there's still work to do to make them pass on Windows.
There's some compatibility with e.g. symlinks, that needs fixing for MSYS2/Cygwin.

Will adress those problems separately (Windows *run-time* support)

This one would be possible to fix now (Windows *compile-time* support)

[...]

Issue #62, opened 2015-12-24
cygwin path handling


kaidokert:
I have to build some projects under cygwin now and then, with toolchain from here : https://launchpad.net/gcc-arm-embedded
The peculiarity of these toolchain builds for windows is that they accept relative paths just fine, and everything passed in in 'cygpath mixed format' , i.e  e:/dir/file.c, but absolute cygwin paths like /cygdrive/e/dir/file.c are not recognized.

ccache currently passes in preprocessed files as absolute pathnames, and obviously this will fail, the toolchain will not find the file. 

I wrote a quick horrible hack ( which i'm not proud of, but i needed to get this working ) to run 'cygpath -m' on the input filenames, which currently works for me for this particular project:
https://github.com/kaidokert/ccache/commit/24a2ba801a4d72389d420b3d582c06b4a11b111b

This would fail in more complex situations as there are are i think other places in ccache code where it would pass in absolute paths to temporary files ( precompiled headers maybe ? ) 

Is it worth doing this properly via a config parameter, or is this just too messy/broken config to deal with ? The hack is fine for me and saves me hours of compile time

itensionanders:
You need to make your changes conditional (for Cygwin), for them to get merged.
i.e. flip your `#if 0` to `#ifndef __CYGWIN__` and do something @@ for dev.mk ?

kaidokert:
yes, but this cannot be compile-time only thing, as cygwin "native" toolchains would not accept this either. so it has to be runtime- config, via a flag and/or config option.
I should probably not use popen() either and do proper exec() as the rest of the code.

this requires a bit more code and testing than I had time to throw together, I'll revisit when I get a chance.

jrosdahl:
I'm closing this due to inactivity. Feel free to suggest reopening it if there are more ideas on what to do.

PR #162, opened 2017-03-11
Visual C/C++ compatibility

jd-gascuel:
Hi Joel,

I started a fork to try to make changes to ccache, so we can compile and use it for windows.
In the following you will see:
- A bunch of small cleanups to make Visual compiler happy. Mostly bad includes and missing definition on windows. All them should be completly transparent, and I think will improve the main branch quality.
- There is a few 32/64 bits issues that might deserve a better correction. I just casted them down to 32 bits...
- I added a `compiler_is_msvc()`  at the top of `ccache.c`, and started a table of options specific to the Visual compilers.
- Plus a few management of specific options.

So it starts to work, and knows how to compile itself.

**Known issues:**
- *cl* compiler do write part of the error messages into stdout, and other parts to stderr. This cannot be avoided... So I just made a concatenation of stdour+stderr->stderr. Not very clean, and a better approach might be: 1) to change the `execute()` function to have only one file, or 2) have ccache to can both stdout and stderr outputs. Note both solution won't keep sync between lines send to stdout and stderr, which might be confusing... but is truly microsoft's fault.
- *cl* options can be `/option` or `-option` : I did not deal with that properly yet. But this should not be too hard.
- I should investigate on `@file` support.
- I have not a very clear view of what option shall be and shall not be in the `compopts_msvc` table.
 
**Futur ?**
The main question here is: how to contribute that to ccache ?
- Are you happy to have a windows branch in the official ccache repo ? and to merge it to the default trunk once it works great ?
- Is there a way to can other annex files (*cl* generates `.sbr` files *with complete symbolic information*) into the cache, so we can restore them too ?

Best regards,

**Note**: I used Visual 2015 to compile, because earlier version will shock on variable definition inside `for()`.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -114,12 +114,12 @@ before_build:
   - set l=
 
   # ==== MinGW gcc compiler
-  - if "%COMPILER%-%PLATFORM%" EQU "gcc-x64" set PATH=C:\mingw-w64\i686-6.3.0-posix-dwarf-rt_v5-rev1\bin;%PATH%
+  - if "%COMPILER%-%PLATFORM%" EQU "gcc-x64" set PATH=C:\msys64\mingw64\bin;%PATH%
   - if "%COMPILER%-%PLATFORM%" EQU "gcc-x64" (
       set CC=gcc&
       echo "Using gcc 64 bits."
     )
-  - if "%COMPILER%-%PLATFORM%" EQU "gcc-x86" set PATH=C:\MinGW\bin;%PATH%
+  - if "%COMPILER%-%PLATFORM%" EQU "gcc-x86" set PATH=C:\msys64\mingw32\bin;%PATH%
   - if "%COMPILER%-%PLATFORM%" EQU "gcc-x86" (
       set CC=gcc&
       echo "Using gcc 32 bits."
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -140,7 +140,7 @@ test_script:
   - make PLATFORM=%PLATFORM% test
 
 after_test:
-  - 7z a ccache-$(version).zip ccache.exe AUTHORS.txt LICENSE.txt MANUAL.txt GPL*.txt NEWS.txt
+  - 7z a ccache-%APPVEYOR_BUILD_VERSION%.zip ccache.exe AUTHORS.txt LICENSE.txt MANUAL.txt GPL*.txt NEWS.txt
 
 artifacts:
   - path: config.log
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -120,7 +120,7 @@ test_script:
   - make PLATFORM=%PLATFORM% test
 
 after_test:
-  - 7z a ccache-{version}.zip ccache.exe AUTHORS.txt LICENSE.txt MANUAL.txt GPL-1.0.txt NEWS.txt
+  - 7z a ccache-$(version).zip ccache.exe AUTHORS.txt LICENSE.txt MANUAL.txt GPL*.txt NEWS.txt
 
 artifacts:
   - path: config.log
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -139,7 +139,9 @@ test_script:
   - ccache -help
   - ccache -p
   - ccache.exe -Czs
-  # Launch unit
+  # Allow cross-compiler final tests (compiled with visual, tested for gc)
+  - if %COMPILER% NEQ gcc et PATH=C:\msys64\mingw64\bin;%PATH%
+  # Launch internal and final tests
   - make PLATFORM=%PLATFORM% test
 
 after_test:
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -140,7 +140,7 @@ test_script:
   - ccache -p
   - ccache.exe -Czs
   # Allow cross-compiler final tests (compiled with visual, tested for gc)
-  - if %COMPILER% NEQ gcc set "PATH=C:\msys64\mingw64\bin;%PATH%"
+  - if "%COMPILER%" NEQ "gcc" set "PATH=C:\msys64\mingw64\bin;%PATH%"
   # Launch internal and final tests
   - make PLATFORM=%PLATFORM% test
 
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -42,6 +42,7 @@ branches:
 #   - if (...) blocks require & line separator.
 #   - nested () in if (...) block require to be escaped with ^
 #   - quoted var cannot be quoted again when using it.
+#   - split too long blocks in several if ... ( ... )
 install:
   # Have MSYS2 in path, including pacman
   - set PATH=C:\MSYS64\usr\bin;%PATH%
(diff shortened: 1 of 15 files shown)


afbjorklund:
What's with all the whitespace changes ? Seem to be unnecessary, and against `uncrustify`...

This is going to be very hard to review and test, since it most of it will now require Windows.
(the nice part about the MinGW version, was that you could compile/test it from elsewhere)

So there's a chance that this will end up with the Mac stuff, as being hard for others to test ?

----

But otherwise it sounds reasonable, making ccache more "portable" (or maybe less, but)

Will see what @jrosdahl says, when it comes to the project scope and maintainability...

afbjorklund:
Seems that there is an awful lot of hacks needed to support this platform and compiler.
And the whitespace issues have only gotten worse ? Maybe it is not feasible to do this...

jd-gascuel:
Dear Joel, Dear Anders, 

Sorry for the whitespace mess:  I did not know about the uncrustify stuff. I added it in my pre-commit hook, so this should append again.

I now have a running version on AppVeyor (https://ci.appveyor.com/project/jd-gascuel/ccache <https://ci.appveyor.com/project/jd-gascuel/ccache>):
- It is automatically build with your autogen / configure / make suite
- it is compiled with Visual 2013 and Visual 2015 compilers, in either 32 and 64 bits versions.
- it pass the internal unit tests
- one can download a compiled alpha version (eg. https://ci.appveyor.com/api/buildjobs/86tv41vdbkr29u9a/artifacts/ccache.exe <https://ci.appveyor.com/api/buildjobs/86tv41vdbkr29u9a/artifacts/ccache.exe>)

To do so:
- I had to fix 2 bugs from the original ccache sources, when compiled in WIN32 environment. I guess you could cherry-pick them:
  o make_relative_path(char *path) does a free(path+1):
     https://github.com/jd-gascuel/ccache/commit/3160e9e3928d33ade688510f856cae653b08ed07 <https://github.com/jd-gascuel/ccache/commit/3160e9e3928d33ade688510f856cae653b08ed07>
  o hash_command_output() makes a double free of win32args == cmd
     https://github.com/jd-gascuel/ccache/commit/936a295336bddab9fb1d77183ae2545df2d26560 <https://github.com/jd-gascuel/ccache/commit/936a295336bddab9fb1d77183ae2545df2d26560>

- I had to upgrade config.guess and config.sub to the latest version. The old version (2012-07-31) did not 
  recognise my MSYS2 build image. This seems to make Travis shock: I’ll have a look there.

- I made quiet a few changes in configure.ac, Makefile.in and dev.mk.in. The main challenge there was:
  o $(CC) was used both for compilation and link. On Windows it is two separate executables.
  o Use of standard OBJEXT to manage .o vs. .obj, and of new LIBEXT to manage .a vs. .lib
  o Use different compile options (eg. -nologo -MD -O3 -W3). And link options too.
  o Visual have a different syntax to compile objects: “cl -Fofoo.obj -c foo.c” instead of “gcc -o foo.o -c foo.c”
  o Visual have a different syntax to force include (“-FIconfig.h” instead of “—include config.h”)
 … and mainly the fact this was my first autoconf project, my first MSYS project and my first AppVeyor project too.

Things I would like to do now:
- Make sure Travis works again !
- Make it run for another compiler (mingw32 and mingw64 ?) because some config stuff do not use the proper predicate yet.
- Make test.sh to run with cl.exe, if easy.
- Add a few unit tests for WIN32 specific paths and options.

Questions:
- Any advice on how to do simpler ?
- Any advice on things I should do differently to better suite your taste ?
- How do you prefer I manage the configuration stuff: in configure.ac (eg. settings a LINK variable) or in Makefile (using ifeq()) ?

Thanks again for your time,
--
JD dot Gascuel at free dot fr

afbjorklund:
No problem, the biggest problem with your whitespace changes is that it makes the rest of your changes unreviewable. It's a little better after uncrustify, but you also need to avoid unix2dos (adding the `\r`, ^M)

The bug fixes and some of the portability fixes look reasonable, but haven't looked into the details yet.
Still not convinced it is worth it, just to support Visual compiler or Windows as a development platform.

jd-gascuel:
This one is a bug in the original code:
- Windows do not have inodes. And `stat(path, struct stat*)` always returns zero in `struct stat.st_ino` (at least on MSYS2 platforms).
- `get_ cwd()`  checks inode number equality to test if the PWD environment variable is the current directory (not the case when the application change directory after being launched from the shell), and use that path instead.

Proposed correction to use _Windows_ specific `GetFileInformationByHandle()` instead...

nordlow:
Our company would benefit greatly from being able to use ccache also with M$'s compilers.  Please get this in!

jd-gascuel:
### Status
- We currently only manages _cl_ specific options `/c`, `/Fo...`, `/E`, `/O...` , their `- ` equivalent and the `@file`  syntax. Each have internal test to make sure they work fine.
This is sufficient for basic testing, but obviously we need more !
- We cannot allow unknown options with leading `/` to be parsed on _linux_ platform, because there is no mean to distinguish between an absolute path `/Foo/Bar.obj` and setting the output as `o/Bar.obj`. Note this only impact running _cl_ on a _linux_ platform... hence should not be of any concern.
- Internal unit test pass on all platforms (travis and AppVeyor).
- Global test (ie. `test.sh`) pass on travis (unchanged here).
- But not on Appveyor (ie. when compiled on _windows_ platforms):
   - it is not (yet ?) compatible with cl.exe computer. In particular we have problems with binary result comparaisons.
   - it should work when using `gcc` on _windows, but currently do not.

afbjorklund:
Still think this needs to be heavily rebased/squash, in order to be readable/reviewable.

If we could get the bugs isolated to separate small issues, and the compatibility shims in a separate file or something, then we could start discussing Windows-as-a-development-platform again (_perhaps_).
Last time I wanted to keep it compatible with MinGW32 and [Wine](https://www.winehq.org/) so I can test it, but it is a bit painful to try and stay compatible with that old 32-bit stuff. Kinda like keeping the Travis tests going on Ubuntu 12.04.

The DOS stuff does make me cringe, but as long as it is "hidden" in the `#ifdef _WIN32` ...

afbjorklund:
Oh, and if it helps anyone (on Windows) I just did a Dockerfile for testing with Linux...
Should be much less bloated than the old Vagrant stuff used for memcached/distccd.

afbjorklund:
Maybe we can support Cygwin (#62) too, in addition to MinGW (#122) - and this Visual Studio ?
Note that Cygwin is a "Unix" while MinGW is a "Windows", even though both run on Windows...

jrosdahl:
Hi @jd-gascuel,

Sorry for the delayed response. My bandwidth for ccache is very low at the moment.

Looks like you have lots of energy to add ccache support for Visual C/C++. Great! :-) However, I'm sorry to say that I'm not very interested in reviewing this pull request in its present form since it contains too many changes that would have negative implications on non-Windows ccache.

I think that it would be wise to take a step back and discuss how to proceed.

Different people naturally have different goals, interests and priorities for a project like ccache. I think that it would be great if ccache could work well for any compiler on any system, but the flip side is that it would be a maintanence hell to support both many compilers and many operating systems, at least without clean abstractions and a larger group of maintainers, neither of which ccache currently has.

For reference, here are some of the interests I have personally as the ccache maintainer (in priority order):

1. ccache should work with GCC on Linux.
2. ccache should be maintainable.
3. ccache should work with GCC-like compilers on Linux.
4. ccache should work with GCC-like compilers on Unix-like systems.
5. ccache should work with other compilers on Unix-like systems.
6. ccache should work with GCC-like compilers on Windows.
7. ccache should work with other compilers on Windows.

My ambition level is somewhere between number 3 and 4 above at the moment.

The key thing here is _maintainability_. As long as I'm the ccache maintainer, I'm forced by time and interest constraints to be restrictive in accepting changes that make it harder to maintain ccache. It's easy to accept new features that work now, but it's a totally different story to keep them working.

For example, that's why the present ccache support for GCC on Windows is more often than not in a broken state – once every other year somebody notices that the Windows support doesn't quite work and makes a race to improve the situation. But Windows support is a second-class ccache citizen; there are no official ccache releases for Windows and the releases are not tested on Windows. Since I'm not personally interested in Windows support, that's just how it has been.

Unless something changes, new support for a non-GCC compiler for Windows will share the same fate: it will maybe work fine now, but it needs somebody who cares for it to keep it alive.

In other words, I won't be the one who maintains Windows support for ccache. So what to do? I see a couple of different ways forward:

1. Maintain diversified ccache support in the same codebase. This requires finding and adding people who want to work on and maintain Windows functionality to the ccache maintainers group. Or:
2. Don't keep Windows support in the main ccache codebase. Create a sibling project for Windows support, run by others than me. Or perhaps just:
3. Help out with e.g. https://github.com/inorton/cclash instead.

I don't think the first option is a good idea. It will only create tension and conflicts of interest. I would like to keep the ccache codebase as simple as possible, without too many ifdefs and special cases for different platforms and compilers. Even if there are people who would like to help out with Windows support *now*, will there be in a year?

I think that a sibling project would be a much better solution than keeping everything in the same code base:

* Create a sibling project with another name (wccache? winccache?), run by people who want to work on and maintain Windows support, make releases and binary builds, etc.
* The project could be a "downstream ccache" project in the sense that it would incorporate new ccache changes and make sure they work fine with supported compilers. Or it could just be a fork that diverges from the current ccache and then takes on in whatever direction it wants.
* The project could be a long-running fork of the ccache repo or not. Whatever works.
* I would still be happy to accept clean patches to the main ccache codebase that make the life easier in the downstream project.

Comments? Other ideas on how to make things work in practice?

jd-gascuel:
### Status
Now global tests do pass with _Visual compiled_ version of ccache. This means that:
 - _ccache_ still do not support fully the _cl_ compiler, as we need support of specific options for this compiler, which will be the next step.
 - but _ccache_, once compiled with _Visual_ compilers, pass the internal **and** global tests (when `test.sh` is given the `CC=gcc` argument). This may seems weird, but it proves that _ccache_ is correctly handling windows path, process management and shell launch.
 - The _AppVeyor_ build system can be used to automatically run the same toolchain and directly see the red/green status of any commit or pull request.

### Futur
- I totally agree with @jrosdahl and @afbjorklund : this PR is a total mess and cannot be integrated like this. I definitely need an active squash/rebase/rework/uncrustify/eol process before anyone can do anything...
 - **but** it allowed me to make a forecast of what might be needed to have a workable solution, and to launch this discussion.
 - I am very interested by your proposal to integrate the stuff that are of general interest for _linux_ and _windows_, and/or that makes life easier for possible futur of a _windows_ version.
In particular the current situation seems a bit confusing, because you currently have partial support here and there...
To have a better view of how to process, I need to know wether you like to merge
   - changes to `configure.ac` to detect the platform we are compiled on (linux/MinGW/MSYS/...) ?
   - changes to `Makefile.in` and `dev.mk.in` to use the defined flags and compile/link .obj and .exe ?
   - changes to `util.c`, `test_util.c` and various `make_relative_path()` functions so they really support various stupidities of path on _Windows_ (ie. improving your `#ifdef _WIN32` stuff already there) ?

What would you like to have first ? Any advice here ? 
All the best,

PR #164, opened 2017-03-17
Support using multiple -fdebug-prefix-map

### LINKED ISSUE #163: Multiple -fdebug-prefix-map options are not considered correctly when deciding to hash the current directory
The OpenEmbedded build system passes multiple -fdebug-prefix-map= options to the compiler in order to map the source directory and sysroots separately.

Unfortunately, ccache appears to only store the last of these arguments. The previous ones are overwritten (and the local copies leak):

```
		if (str_startswith(argv[i], "-fdebug-prefix-map=")) {
			debug_prefix_map = x_strdup(argv[i] + 19);
			args_add(stripped_args, argv[i]);
			continue;
		}
```
The arguments still make it through to the compiler, but only the last is considered when deciding which directory to put into the hash when conf->hash_dir is enabled. This means that the current directory always ends up in the hash even when it has no effect on the generated files.

afbjorklund:
Previously only the last one was actually being applied to the hash.
Note that the compiler will actually only *use* one of all the paths.

Closes #163.

### DIFF
--- a/ccache.c
+++ b/ccache.c
@@ -182,7 +182,10 @@ static bool generating_coverage;
 static bool generating_stackusage;
 
 // Relocating debuginfo in the format old=new.
-static char *debug_prefix_map = NULL;
+static char **debug_prefix_maps = NULL;
+
+// Size of debug_prefix_maps list.
+static size_t debug_prefix_maps_len = 0;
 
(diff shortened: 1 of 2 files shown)


jrosdahl:
Applied on 3.3-maint, thanks!

PR #165, opened 2017-03-19
Fix DIR_DELIM_CH for win32 and non-win32 platforms, values were mixed up

oded-ist:


afbjorklund:
LGTM

jd-gascuel:
I agree to !

afbjorklund:
I suppose it mostly shows the lack of the tests on the Windows platform... (#119)

jrosdahl:
> I suppose it mostly shows the lack of the tests on the Windows platform...

In this case, I'm to blame both for the mixup and for not writing a test case... Fixed in c5effc8a.

jrosdahl:
Thanks! Applied to 3.3-maint.

PR #167, opened 2017-03-25
Add docker target for running test with Docker

afbjorklund:
Add standard container definition for running the tests,
to make things easier for developers on other platforms.

@jrosdahl : You can change the standard distro used,
if you prefer Debian (or something) ? Doesn't matter.

Eventually we could use multiple Dockerfiles for this,
for instance to run tests with clang - or something ?

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,2 @@
+.git
+**/.vagrant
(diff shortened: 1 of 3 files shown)


jrosdahl:
Applied on master, thanks!

PR #169, opened 2017-03-28
Cleaning up the build system

jd-gascuel:
Hi Joel, Hi Anders,

I spend the day squashing and re-basing my fork (sorry, dropping the previous PR also drop all the discussions we had there... Sorry for that!).

Here is a first cleaned PR that I can propose you. The idea is just to have a common ground when building various branches/forks/mirror/whatsoever versions of ccache...
But this PR does not change anything at the behavior of the actual code.

Summary:
 - Upgrade config.guess and config.sub to 2017-03 version: this is needed to detect MSYS2 platform.
 - Upgrade configure.ac (and *.in) to pass clean variables about the detected platform/compiler
 - Check `HAVE_XXX_H` before including various files detected by configure.
 - Move `compiler_is_XXX()` predicates at the beginning of `ccache.c`
 - Make `input_file`, `output_object` and `make_relative_path()` so we can unit test them.

--> Please let me know if you are happy with that.

All the best,


afbjorklund:
Why did you have to rename `util.c` ? Did the name conflict with something on Windows ?

Otherwise the upgrade of autotools and OBJEXT/LIBEXT looked reasonable, but not
so sure about the compiler - and I know that @jrosdahl had some concerns about that.

Basically this will be impossible to test, with everything `ifdef`d in the Makefile like that...

afbjorklund:
But I can give it a try with `mingw-w64`, maybe some of it is enough to get #122 going again.

jd-gascuel:
Hello Anders,

1. There is two: `util.c` and `test/util.c` ... perfectly legal on linux, but just crashes the brain of the windows linker. But my choice of name is probably just wrong. One can also consider just merging them both.

2. The compiler detection is just to be used at compiling ccache. It is NOT aimed at changing its behavior. in any way.
Ideally a binary of ccache compiled with cl using MSYS2 should be able to work for cl, gcc, cc, llvm, etc.
I will provide an *AppVeyor* matrix that test just that...

afbjorklund:
@jd-gascuel : seems you have a typo in "psapi" (spelled as "Psapi"), which fails the build...
(I suppose it doesn't matter either way on a DOS file system, but better keep it as lowercase)

`/usr/bin/i686-w64-mingw32-ld: cannot find -lPsapi`
`/usr/i686-w64-mingw32/lib/libpsapi.a`

configure.ac
```diff
@@ -9,38 +9,61 @@ AC_CONFIG_HEADER(config.h)
 
 AC_CANONICAL_HOST
 
+AC_SUBST(ccache_os, unknown)
 case $host in
     *mingw32* | *mingw64* | *cygwin* | *wince* | *mingwce*)
-        windows_os=yes
+        ccache_os=mingw
```
afbjorklund: What is the difference between MinGW and Windows ?

configure.ac
```diff
@@ -9,38 +9,61 @@ AC_CONFIG_HEADER(config.h)
...
+        AC_DEFINE(CCACHE_MINGW, 1, Compiling on a MinGW platform)
         AC_DEFINE(_WIN32_WINNT,0x0600, Windows Vista or newer is required)
         ;;
+
+    *msys*)
+        ccache_os=msys
+        AC_DEFINE(CCACHE_MSYS, 1, Compiling on a MSYS2 platform)
```
afbjorklund: What is the difference between MSYS and Windows ?

configure.ac
```diff
@@ -9,38 +9,61 @@ AC_CONFIG_HEADER(config.h)
...
+        AC_DEFINE(_WIN32_WINNT,0x0600, Windows Vista or newer is required)
+        ;;
+
+    *apple*)
+        ccache_os=mac
+        AC_DEFINE(CCACHE_MAC, 1, Compiling on an Apple platform)
```
afbjorklund: This platform is normally called Darwin, after the `uname`

afbjorklund:
Ah, I see. Maybe test/util.c should be called test_util.c or something.

jd-gascuel:
... or something. Because there is already a `test_util.c` too...

configure.ac
jd-gascuel: The good question is: what is the difference between MinGW and MSYS2 ?
I found at least one:
- MSYS2: You should link with `Psapi.lib`
- MINGW: You should link with `-lpsapi`  (sorry for the mixed case typo)

jd-gascuel:
I have that in my branch: It just drops color under mingw or msys:

```C
static bool
color_output_possible(void)
{
#ifdef _WIN32
	return false;
#else
	const char *term_env = getenv("TERM");
	return isatty(STDERR_FILENO) && term_env && strcasecmp(term_env, "DUMB") != 0;
#endif
}
```

afbjorklund:
Well, the problem was the `/* #undef HAVE_UNSETENV */` in config.h
This causes x_unsetenv to set it to "" instead, and a check was missing:

``` diff
--- a/ccache.c
+++ b/ccache.c
@@ -1710,7 +1710,7 @@ calculate_common_hash(struct args *args, struct mdfour *hash)
        // Possibly hash GCC_COLORS (for color diagnostics).
        if (compiler_is_gcc(args)) {
                const char *gcc_colors = getenv("GCC_COLORS");
-               if (gcc_colors) {
+               if (gcc_colors && gcc_colors[0] != '\0') {
                        hash_delimiter(hash, "gcccolors");
                        hash_string(hash, gcc_colors);
                }

```

jd-gascuel:
...
And I have also that in my branch:
```C
// This is like unsetenv.
void x_unsetenv(const char *name)
{
#ifdef HAVE_UNSETENV
	unsetenv(name);
#elif defined(_WIN32)
	char *str = format("%s=", name);
	putenv(str);
	free(str);
#else
	putenv(x_strdup(name)); // Leak to environment.
#endif
}
```

afbjorklund:
Yeah, it seemed to choke on the suffix not fitting into the DOS conventions;

```
test/test_conf.c:345: Failed assertion:
  Suite:      conf
  Test:       conf_set_new_value
  Expression: conf_set_value_in_file("ccache.conf", "stats", "chocolate", &errmsg)
  Message:    rename ccache.conf.tmp.anders-ThinkPad.8.P1kAlM to ccache.conf: No such file or directory

test/test_conf.c:357: Failed assertion:
  Suite:      conf
  Test:       conf_set_existing_value
  Expression: conf_set_value_in_file("ccache.conf", "path", "vanilla", &errmsg)
  Message:    rename ccache.conf.tmp.anders-ThinkPad.8.iG2Kb7 to ccache.conf: No such file or directory

```
Probably `create_tmp_fd` needs to be taught about using 8+3 somehow.

jd-gascuel:
What is your settings for ` MinGW-w64/Wine (i686)` ?
Can I re-enable them in my branch, just to be sure to produce compatible stuff... ?

afbjorklund:
Basically: `./configure --host=i686-w64-mingw32 && make quicktest`
Thought about adding a Dockerfile for it ? (gcc-mingw-w64-i686 + wine)

afbjorklund:
And apparently MinGW has some own "mkstemp", that configure finds - but doesn't work.
`Binary file /usr/i686-w64-mingw32/lib/libmingwex.a matches`

That was the last two fails (switching to the "cheap and nasty" replacement works better).

afbjorklund:
@jd-gascuel : Were you able to use the `Dockerfile` ?

```make docker```

Should be easy to extend to "clang" and "mingw", too...
Probably move them into subdirectories or something ?

**Comment by afbjorklund** [comment]
Uploaded the passing code as 00633e3e48d6505b6fce1b40f1d81efb6c1e38d6...36c54c7b8a2e11faec1336ddf6da46b13614f47c, if it helps you somehow ?
We'll need to upgrade the Travis build to 16.04 LTS (trusty), for this to work...

``` yaml
        - os: linux
          compiler: i686-w64-mingw32-gcc
          sudo: required
          dist: trusty
          env: HOST="--host=i686-w64-mingw32" TEST="test/main.exe"
```

jd-gascuel:
Oups... I am lost here: how do I fetch and merge your uploaded patch into my PR ?
(git newbie question, sorry).

jd-gascuel:
Oups. I got
> The program 'i686-w64-mingw32-gcc' is currently not installed
Reverting.

afbjorklund:
I suppose mentioning the branch name (`afbjorklund/mingw-build`) would have helped too...

afbjorklund:
And yeah, adding the requirements would be needed too.

afbjorklund:
Apparently one can use `wine1.6-amd64` for running 64-bit binaries,
so there is no particular to *not* use `--host=x86_64-w64-mingw32` ?

`PASSED: 480 assertions, 95 tests, 10 suites`

afbjorklund:
Updated the branch for #122 , would be nice to combine forces for Appveyor with #69 ?

https://github.com/TheOneRing/ccache/blob/fa4945221ddc8f1b38df23fb113a193ef117dc4a/appveyor.yml

https://github.com/jd-gascuel/ccache/blob/e946e2b6e3987168c915933fb334cbcaf7661d3a/.appveyor.yml

I wonder where that "usleep" went, maybe I left it in Windows or something ? (`nanosleep`)

----

Will leave this PR as the request for supporting Visual Studio / Windows as a *development* platform ?

As @jrosdahl said (https://github.com/ccache/ccache/pull/162#issuecomment-289215585), not so sure...

The other two relied heavily on just using MinGW, so wasn't too concerned with supporting Windows.

Seems to be working with [Wine](https://www.winehq.org/), haven't tried on Win.

----

I'm OK with supporting Mac using `clang` in the main code.

Also OK with supporting Win - by using `MinGW` and `Wine`.

But am *not* so sure about supporting Xcode / Visual Studio ?

That would require me actually running tests on the real thing.

jd-gascuel:
Hi @afbjorklund ,

This PR is **not** about supporting *Windows*. It is just about having a **common ground** to make it easier to share code with a branch/fork/XXX that do support windows.

This is why I did not include any change specific for *MSYS2* or *MinGW*.

About your (*not for the PR* usleep() question), I can propose the simple answer I used in my branch:
- `usleep()`  is in micro seconds. But does nit exist in the *Windows* world.
- `nanosleep()` is not so much an option, because it lacks portability.
- On the other end, for decades *Windows* do have a `Sleep()`function, which is in milli-seconds.

The good point is tha `ccache` indeed use just milli-seconds resolution (ie. starting at 1ms, and doubling until done), so a quiet simple patch to `bool lockfile_acquire()`is just:
```C
#ifdef _WIN32
        // We are lucky: Windows have ms resolution, which is enough here.
        Sleep(to_sleep/1000);
#else
        usleep(to_sleep);
#endif
```
Regards,

afbjorklund:
Thanks for clarifying, but still the changes proposed only affects Windows as a platform...
Maybe it would be interesting to look into testing it in the cloud, but it's not cross-platform.

Your change with `Sleep` looks good, will take the platform discussion to the mailing list.
I can probably find _Mac OS X 10.9_ or _Windows 7_, but latest/greatest isn't always available.

jd-gascuel:
I am quiet new to _ccache_ development, but I found AppVeyor really useful and efficient. 

My current [configuration](https://github.com/jd-gascuel/ccache/blob/windows/.appveyor.yml) compiles on *gcc* 32 and 64 bits, *Visual 2013* 32 and 64 bits, and *Visual 2015* 32 and 64 bits. And for each of them, runs the full test suite with *gcc*, *g++* and *cc*. Most probably overkill in the long run, but interesting and useful during the hard-coding time.
As it is integrated with _github_, one can also directly see status of each commit or PR, just like for *Travis*... which is certainly great to help code maintainability.

afbjorklund:
Yeah, it's definitely useful (with Travis or Appveyor). But some things needs direct access to the actual platform in order to do development, and when running locally it is easier to set up Docker or Vagrant ?

Anyway, the new MSYS2 system is kinda neat (I like `pacman`) and upgrading to 64-bit was needed anyway because of C99 and because of critical bugs (printf ? _sheesh_) in the old 32-bit MinGW/MSYS environment.

Apparently there's some things to be done with symlinks: [MSYS2-dont-use-symlinks.patch](https://github.com/Alexpux/MSYS2-packages/blob/master/ccache/MSYS2-dont-use-symlinks.patch)

So in order to work on Windows, those should be addressed (with tests) and the rest fixed ?

afbjorklund:
Apparently MSYS2 does *not* compile for Windows by default (have to use `-mwin32` for that),
which is why the test case fails... It also doesn't define `__MINGW32__` (nor `__MINGW64__`) anymore.

I like you reported earlier, the `uname` is now a weird "MSYS_NT-6.4" instead of "MINGW32_NT-6.1"
I suppose that `test.sh` should accept either MINGW* or MSYS*, and in either case (upper/lower).

----

And we need a better way to detect compiler, than looking at the name of binary (as it fails)

So need to go beyond the first line of the version string, since that doesn't have any real info:

`gcc (Ubuntu 5.4.0-6ubuntu1~16.04.4) 5.4.0 20160609`

`clang version 3.8.0-2ubuntu4 (tags/RELEASE_380/final)`

Probably look for the macros instead.

``` cpp
#define __GNUC__ 5
#define __clang__ 1

#define __llvm__ 1
#define __MINGW32__ 1
#define __MINGW64__ 1
#define __MSYS__ 1
```

afbjorklund:
@jd-gascuel : Oh, and I found the old nanosleep code (44fbd43a5469d3d884a19bed171b324bfd9da275)
It can use that where available, and microseconds on Windows...

We only sleep for 1.0 second (and longer) anyway, so it doesn't really matter which resolution is used :-)
But think that we should default to usleep still, since nanosleep uses a lot of CPU on certain platforms.

afbjorklund:
I had misunderstood a couple of things. First that the "MSYS" prompt is now a **Cygwin** environment!
Second was that the two MinGW shells do not really "work" OK, unless you install their toolchains too...

i.e. if you don't populate `/mingw32` and `/mingw64`, then they will _also_ use the default (i.e. Cygwin)
But after installing `mingw-w64-i686-gcc` and `mingw-w64-x86_64-gcc` it's all normal again (i.e. MinGW)

* helpful: https://sourceforge.net/p/msys2/discussion/general/thread/dcf8f4d3/#8473/588e

Seems to be a _lot_ left to do with portability: https://github.com/TeaCI/tea-ci/wiki/Msys2-on-Wine

----

But it should be doable (just a couple of bugfixes and bit-rot), to support *both* MinGW and Cygwin...

The "issue" would be trying to support a more native Windows, with only Visual Studio or something ?
Like porting the build system from autotools to cmake, as [discussed](https://lists.samba.org/archive/ccache/2016q3/thread.html#1451). That would be a breaking change.

Running configure with bash and make with gmake sounds like a reasonable workaround meanwhile.

jd-gascuel:
@afbjorklund this is precisely the point of this pull request. With it, we do have a common ground on which I can propose more stuff and convergence.

I do currently have an AppVeyor image that:
- Compile ccache with _MinGW64_'s _gcc_, using _MSYS2_ emulation layer to have access to _linux like_ environment on windows.
- Compile ccache with the original _Visual 2013_ compiler on windows.
- Compile ccache with the original _Visual 2015_ compiler on windows.
- Run for each of them the _quick test_ suite
- Run for each of  them the long test.sh suite with various compilers: _gcc_, _g++_ and _cc_

I am working on having the long test.sh suite work also for _Visual_ compilers (whatever the compiler used to compile ccache). Once done I will be able to propose:
- a more stable ccache compilation, working better on windows platform with mixed and weird path syntax you get when using native windows environment or various emulation layers above it.
- a more versatile ccache, able to handle visual options too.

I do not think that using CMake will be an improvement here: I agree with @jrosdahl  that having both system is a loss of energy. And as you said, it will be a deep change and have impacts on a lots of people compiling ccache in many environment.
Plus this PR proves that _autoconf_ tools can be used for _windows native_ environment with success too.

Please fill free to ask any changes you thing necessary to process further !

afbjorklund:
There's still some things that could be done about readability, like fix your Git settings (core.autocrlf) to not add those carriage returns (`'\r'` or `^M`) even if you prefer your DOS line endings locally (that is fine).
Indenting your #include's is also not recommended, since it doesn't even work on all platforms/compilers. It's sad that a lot of standard headers (stdbool/inttypes) are missing, but I guess that is portability for ya...

But like I said, MinGW/Cygwin should be fixed already and adding Appveyor/Windows is up to @jrosdahl. Doesn't hurt if the changes are kept minimal anyway, even if it is on a separate branch or a separate fork ?
Adding support for Windows options (/) and Windows paths (C:\\) is always going to be lots of code, though. I don't think there's any _good_ way of getting around that, at least not without using something like MSYS...

I do prefer things portable, and thank you for your interest in ccache.

Just not sure if using "real" Windows (without Bash) is ever worth it. :-P

staticfloat:
Just to add a little +1 for this; I was unable to compile `ccache` on a `ppc64le` machine until I updated the `config.guess` and `config.sub` files to the same contents as those found in this pull request.  I can confirm that it compiled successfully after incorporating these changes.

jrosdahl:
Note: I've updated config.guess and config.sub to modern versions on master now.

jrosdahl:
@jd-gascuel: Closing this now due to inactivity. Please resubmit any proposed fixes as new pull requests if you like.

PR #176, opened 2017-05-24
Fix compilation with MinGW32

orgads:


orgads:
ping

orgads:
reping

orgads:
PING!

orgads:
Hello?

afbjorklund:
Need to get the MinGW support back up and running first (#119, #122). Otherwise it can happen again.

But upgrading to MSYS2 and MinGW-w64 is probably the best Windows way forward, in the long run...

orgads:
I am using mingw-w64. Compilation fails with it.

jrosdahl:
Applied on master in 464ba863e7d6489a06951d899cef46acae82867b. Thanks!

Regarding the long ping time: That's partly because I just have prioritized other things like family lately, and partly because I'm just not interested in Windows stuff. See also https://github.com/ccache/ccache/pull/162#issuecomment-289215585 for background on the state of ccache on Windows.

PR #179, opened 2017-06-14
Add comment about why SLOPPY_INCLUDE_FILE_[CM]TIME compares using >=t

torarnv:
If a file has the same modification or change time as the current
compile, all should be good.

Removes the need for setting the sloppiness to `include_file_ctime` in
the tests, which was done to work around the fact that backdating
files (modification time) would bump their change time, and likely
be the same as the one of the following compilation job.

### DIFF
--- a/ccache.c
+++ b/ccache.c
@@ -600,12 +600,16 @@ remember_include_file(char *path, struct mdfour *cpp_hash, bool system)
 		}
 	}
 
+	// The comparison using >= is intentional, due to a possible race
+	// between starting compilation and writing the include file.
+	// See https://github.com/ccache/ccache/blob/master/MANUAL.txt
 	if (!(conf->sloppiness & SLOPPY_INCLUDE_FILE_MTIME)
 	    && st.st_mtime >= time_of_compilation) {
 		cc_log("Include file %s too new", path);
 		goto failure;
 	}
 
+	// The same >= logic as above applies to the change time of the file.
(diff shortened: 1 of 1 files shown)


torarnv:
Huh? Why is it conflicting? I've rebased locally against ToT master just now, where is the conflicting change coming from?

torarnv:
@jrosdahl I presume the `>` logic should be fine?

jrosdahl:
> Huh? Why is it conflicting? I've rebased locally against ToT master just now, where is the conflicting change coming from?

It’s because your commits are based on 8c8d581 which is more than a year old. What is “ToT master”?

> I presume the > logic should be fine?

No, the `>=` logic is deliberate and has to do with file system mtime resolution and a race condition (see https://github.com/ccache/ccache/blob/master/MANUAL.txt#L972): the header’s mtime might be equal to `time_of_compilation` even if the header is written to _after the compilation has started_. (It would probably be a good idea to write this as a comment in the code as well...)

torarnv:
Ah, doh!! 😬 My git clone was from git.samba.org, lesson learned!

Thanks for the explanation! Does the race apply to the ctime as well?

torarnv:
Changed to a commit explaining what's going on

jrosdahl:
> Does the race apply to the ctime as well?

Yep.

jrosdahl:
Applied, thanks!

PR #180, opened 2017-06-15
Update log message and docs when using SLOPPY_FILE_STAT_MATCHES

torarnv:
The size is no longer part of this check.


jrosdahl:
Applied, thanks!

PR #181, opened 2017-06-15
Dont cache stale precompiled headers

torarnv:
If one of the included files in a precompiled header is touched,
changing only its mtime, clang will produce a fatal error when the
precompiled header is then used:

    file 'foo.h' has been modified since the precompiled header 'foo.pch' was built

We need to take this into account when producing precompiled headers,
so that we don't pick out stale objects from the cache. This works
fine for direct mode, but in preprocess mode we don't have enough
information to decide if the object is still valid, so we skip the
cache entirely in that mode. 

Is there a better way to skip the preprocess? I guess we can't skip it entirely, as we need to build up the list of include files and hash etc, so that we can cache the result of the real compilation?

### DIFF
--- a/ccache.c
+++ b/ccache.c
@@ -207,7 +207,7 @@ static char *cpp_stderr;
 char *stats_file = NULL;
 
 // Whether the output is a precompiled header.
-static bool output_is_precompiled_header = false;
+bool output_is_precompiled_header = false;
 
 // Profile generation / usage information.
 static char *profile_dir = NULL;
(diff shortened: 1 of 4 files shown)


torarnv:
Fixed issue I was seeing with manifest not being used. The cause was the ctime of test.h being the same as the compilation_time, causing the cached ctime/mtime to be -1. Only way to work around it seems to be to sleep 1 after touching in that particular test.

With that solves the test passes without wiping the manifest, as the new object will be added to the manifest with the correct mtime and be picked at the next compilation.

torarnv:
All green 🍏

torarnv:
Any thoughts @jrosdahl ? Happy to update the patch if needed 😄

torarnv:
Ping @jrosdahl

jrosdahl:
Applied on 3.3-maint, thanks!

PrideChung:
@jrosdahl When will 3.3.5 be released? We constantly run into this bug.

jrosdahl:
> @jrosdahl When will 3.3.5 be released? We constantly run into this bug.

This weekend, I hope.

PrideChung:
Awesome! Can't wait to give it a try.

Issue #182, opened 2017-06-16
always treating (set) boolean environment variables as true can be surprising


mostynb:
Common mistakes that can be made:
export CCACHE_DISABLE=false
export CCACHE_CPP2=false

I wonder if it would be worthwhile giving warnings if boolean environment variables have "negative" sounding values ("NO","no","false","FALSE","0" or some subset)?

Or perhaps we should just change the processing of these variables to be more strict and simultaneously more obvious?  eg only accept "true", "false" or "" and take the empty value to mean "use the default for this setting"?

ilor:
I have independently made the same mistake of assuming CCACHE_CPP2=0 would disable the feature

afbjorklund:
Well, changing to an error would break legacy and ccache doesn't really have a good place to output warnings...

But "0" or "false" should be OK

Must only use 1 or "" for true...

afbjorklund:
Patches ?

mostynb:
Patches coming soon, I wanted to discuss the proposed behaviour first.

Can we simply print warnings to stderr, but not cache them?  And if we want this to be an error, exit with an error code without doing any work?  IMO this is better than silently using an "invalid" configuration (assuming there's a useful error/warning message).

jrosdahl:
Here is a previous discussion on the same topic: https://bugzilla.samba.org/show_bug.cgi?id=11826

> Can we simply print warnings to stderr, but not cache them?

We can, but that would be problematic since ccache then no longer works transparently. For instance, autoconf configure scripts have been known to fail tests if the compiler emits anything to stderr even if the produced result actually works. (And printing to stdout is also a no-no.)

> [...] IMO this is better than silently using an "invalid" configuration

I would say that it's valid configuration, but unexpected if you haven't read and understood the documentation, which many of course haven't. I agree that it's unfortunate design choice (made fifteen years ago), but as indicated in the discussion I linked to above, I'm also hesitant to break backward compatibility.

> And if we want this to be an error, exit with an error code without doing any work?

Yes, that's the only reasonable option if we decide to break compatibility. If so, I think your suggestion or some variation sounds fine, in other words:

* Treat `CCACHE_FOO` set to `0`, `false`, `disable` or `no` (case insensitive) as an error and treat any other value as "enable foo".
* Treat `CCACHE_NOFOO` set to `0`, `false`, `disable` or `no` (case insensitive) as an error and treat any other value as "disable foo".

[...]

PR #186, opened 2017-06-23
Fail if boolean env vars are set with common negative strings

mostynb:
Boolean environment variables have strange semantics: if the variable
is set then the configuration setting is considered "true".  This means
users can easily be mistaken that eg CCACHE_DISABLE=false means that
the "disable" setting is set to false.

To avoid too much backwards-incompatibility, we now exit with an error if
some common negative-sounding (case-insensitive) values:
"0", "false", "disable", "no".

All other values (including the empty string) are still considered to mean
"true".

Resolves https://github.com/ccache/ccache/issues/182

### DIFF
--- a/test.sh
+++ b/test.sh
@@ -1084,6 +1084,20 @@ EOF
     expect_stat 'cache hit (preprocessed)' 0
     expect_stat 'cache miss' 0
     expect_stat 'unsupported code directive' 1
+
+    # -------------------------------------------------------------------------
+    TEST "Invalid boolean environment configuration options"
+
+    for invalid_val in 0 false FALSE disable no ; do
+        CCACHE_DISABLE=$invalid_val $CCACHE $COMPILER --version > /dev/null 2>&1
+        if [ $? -eq 0 ] ; then
+            test_failed "'$invalid_val' should be rejected for boolean env vars"
+        fi
+        CCACHE_NODISABLE=$invalid_val $CCACHE $COMPILER --version > /dev/null 2>&1
+        if [ $? -eq 0 ] ; then
+            test_failed "'$invalid_val' should be rejected for boolean env vars"
+        fi
+    done
 }
 
 # =============================================================================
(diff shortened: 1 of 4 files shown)


mostynb:
@jrosdahl, @afbjorklund: here's an initial patch to try.

afbjorklund:
@jrosdahl had some idea about adding the same for the negative variant (i.e. NO...), but anyway.

Most complaints that I hear about the config system is that it is way too complex, not that it is hard.

test.sh
```diff
@@ -1099,6 +1099,16 @@ EOF
     expect_stat 'cache hit (preprocessed)' 0
     expect_stat 'cache miss' 0
     expect_stat 'unsupported code directive' 1
+
+    # -------------------------------------------------------------------------
+    TEST "Invalid boolean environment configuration options"
+
+    for invalid_val in 0 false FALSE disable no ; do
+        CCACHE_DISABLE=$val $CCACHE $COMPILER -version > /dev/null 2>&1
```
afbjorklund: Seems like this test _always_ fails: (maybe you meant --version ?)
```
gcc: error: unrecognized command line option ‘-version’
gcc: fatal error: no input files
compilation terminated.
```

mostynb: Good catch, pushed amended version.

mostynb:
> @jrosdahl had some idea about adding the same for the negative variant (i.e. NO...), but anyway.

I think this was already handled, but added a test case for this too.

Also fixed another bug in the test case (use the correct variable inside the loop) and made this an error rather than a warning.

afbjorklund:
LGTM

The "strange semantics" are the same as getenv(3) and \#ifdef, but if it trips newcomers up...
I think this implemented compromise is reasonable, to keep both old and new users happy.

mostynb:
Rebased on master.

jrosdahl:
Applied, thanks!

jrosdahl:
BTW @afbjorklund:

> Most complaints that I hear about the config system is that it is way too complex, not that it is hard.

This is actually the first time I hear any complaints about the config system... Is there something we can do? Perhaps open an issue about it?

PR #187, opened 2017-06-23
Add some fallthrough commants to work around GCC compiler warnings

GeertBosch:
This make my GCC 7.1 happy.

### DIFF
--- a/murmurhashneutral2.c
+++ b/murmurhashneutral2.c
@@ -30,8 +30,8 @@ murmurhashneutral2(const void *key, int len, unsigned int seed)
 
 	switch (len)
 	{
-	case 3: h ^= ((unsigned int) data[2]) << 16;
-	case 2: h ^= ((unsigned int) data[1]) << 8;
+	case 3: h ^= ((unsigned int) data[2]) << 16; // Fallthrough.
+	case 2: h ^= ((unsigned int) data[1]) << 8; // Fallthrough.
 	case 1: h ^= ((unsigned int) data[0]);
 		h *= m;
 	};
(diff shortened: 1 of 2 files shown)


afbjorklund:
Yet more C++17 annoyances...

https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html#index-Wimplicit-fallthrough

Looks good to me, I suppose - although we probably need to start testing with some newer compiler too ?
Apparently we can use `gcc-7` (at least in trusty) from a PPA repository, so should be _possible_ in Travis.

ppa:ubuntu-toolchain-r/test

afbjorklund:
This also affects the bundled zlib, along with the bundled murmurhash2 :

https://github.com/afbjorklund/ccache/pull/1/commits/969cd90dc998eb4b197f15544ee0dee95f96a946

https://github.com/madler/zlib/issues/316

In order to catch such things, we need a Travis build `--with-bundled-zlib` 

----

And even though it is abandoned, the murmurhash _should_ go upstream:
https://github.com/aappleby/smhasher/blob/master/src/MurmurHash2.cpp

murmurhashneutral2.c
```diff
@@ -30,8 +30,8 @@ murmurhashneutral2(const void *key, int len, unsigned int seed)
 
 	switch (len)
 	{
-	case 3: h ^= ((unsigned int) data[2]) << 16;
-	case 2: h ^= ((unsigned int) data[1]) << 8;
+	case 3: h ^= ((unsigned int) data[2]) << 16; // Fallthrough.
+	case 2: h ^= ((unsigned int) data[1]) << 8; // Fallthrough.
```
tbeu: Should rather use C style blockc omments.

afbjorklund: Didn't we go for new C ? (C99)

jrosdahl:
Applied on 3.3-maint in 3c448a4566db97676b2de1f7a270d3f8daad013b and merged to master. Thanks!

PR #188, opened 2017-06-25
Improve support for run_second_cpp = false (cpp1)

### LINKED ISSUE #185: Support clang's -frewrite-includes flag with CCACHE_NOCPP2 by sending -D defines to both preprocess and compile
Currently we need to use CCACHE_CPP2 when building with clang due to -Werror tripping differently on preprocessed output. clang has a -frewrite-includes flag that causes it to only rewrite #include lines without doing the full preprocessing to handle just this use case (it is especially important with icecream/distcc where the equivalent of CCACHE_CPP2 isn't an option). Unfortunately this doesn't work with CCACHE_NOCPP2 because the -D flags are stripped out when running the compiler. Additionally, it requires passing -xc rather than -xcpp-output since it hasn't been fully preprocessed yet. It may be worth always passing the -D flags unless it is actually harmful on some compilers.

gcc has a similar -fdirectives-only flag, but it includes all -D flags from the command-line in the preprocessed output, so it works correctly with ccache today.

afbjorklund:
Closes #185.

### DIFF
--- a/MANUAL.txt
+++ b/MANUAL.txt
@@ -476,6 +476,12 @@ of the original source code. This makes cache misses slightly faster since the
 source code only has to be preprocessed once. The downside is that some
 compilers won't produce the same result (for instance diagnostics warnings)
 when compiling preprocessed source code.
++
+An alternative to *run_second_cpp* is to use *-fdirectives-only* (for GCC) or
+*-frewrite-includes* (for Clang). This will cause the compiler to leave the
+macros and other preprocessor information, and only process the *#include*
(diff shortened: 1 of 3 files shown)


RedBeard0531:
I tested this patch and it made a substantial improvement to our distributed build times with icecream. Any chance of this getting merged in for the next ccache release?

afbjorklund:
@jrosdahl : merge ?

jrosdahl:
Applied, thanks!

PR #190, opened 2017-06-30
Clang emits warnings for unused linker arguments, respect that!

### LINKED ISSUE #189: ccache with clang doesn't match "linker' input unused"
Steps to reproduce:

```
(root) [ezyang@devgpu014.ash3 ~/local/labs/ccache-bug] cat test.c
int test() { return 0; }
(root) [ezyang@devgpu014.ash3 ~/local/labs/ccache-bug] /home/ezyang/ccache/bin/ccache -C
Cleared cache
(root) [ezyang@devgpu014.ash3 ~/local/labs/ccache-bug] /home/ezyang/ccache/lib/clang -c test.c
(root) [ezyang@devgpu014.ash3 ~/local/labs/ccache-bug] /home/ezyang/ccache/lib/clang -c test.c -Wl,-rpath,\$ORIGIN
(root) [ezyang@devgpu014.ash3 ~/local/labs/ccache-bug] /home/ezyang/ccache/bin/ccache -C
Cleared cache
(root) [ezyang@devgpu014.ash3 ~/local/labs/ccache-bug] /home/ezyang/ccache/lib/clang -c test.c -Wl,-rpath,\$ORIGIN
clang: warning: -Wl,-rpath,$ORIGIN: 'linker' input unused
```

I repro'd on master at b313a68664474579e702ab6392ea96ee66d2226f. This bug was very confusing as I was trying to fix my build system to stop passing the linker arguments spuriously here, but ccache was not giving me different output even when I fixed it.

I don't think this is an issue with gcc as it doesn't seem to give a warning with unused linker inputs.

```
(root) [ezyang@devgpu014.ash3 ~/local/labs/ccache-bug] /usr/bin/gcc -c test.c -Wl,-rpath,\$ORIGIN 
(root) [ezyang@devgpu014.ash3 ~/local/labs/ccache-bug]
```

I know ccache is gcc first, and "similar enough" compilers second, but it would be nice if there were a special case to handle this behavior.

ezyang:
If ccache concludes the invocation with/without linker arguments
is the same as before, then it may show/fail to show a warning
when it should.  I know we're not supposed to rely on the
is clang check, but this solves it in normal cases.

Fixes #189.

Signed-off-by: Edward Z. Yang <ezyang@fb.com>

### DIFF
--- a/ccache.c
+++ b/ccache.c
@@ -1723,19 +1723,24 @@ calculate_object_hash(struct args *args, struct mdfour *hash, int direct_mode)
 		hash_int(hash, MANIFEST_VERSION);
 	}
 
+	// clang will emit warnings for unused linker flags, so we shouldn't
+	// skip those arguments
+	int is_clang = compiler_is_clang(args);
+
 	// First the arguments.
 	for (int i = 1; i < args->argc; i++) {
 		// -L doesn't affect compilation.
-		if (i < args->argc-1 && str_eq(args->argv[i], "-L")) {
+		if (i < args->argc-1 && str_eq(args->argv[i], "-L") && !is_clang) {
 			i++;
 			continue;
 		}
-		if (str_startswith(args->argv[i], "-L")) {
+		if (str_startswith(args->argv[i], "-L") && !is_clang) {
 			continue;
 		}
 
 		// -Wl,... doesn't affect compilation.
-		if (str_startswith(args->argv[i], "-Wl,")) {
+		if (str_startswith(args->argv[i], "-Wl,") && !is_clang) {
+			// ...but it affects warnings with clang
 			continue;
 		}
 


afbjorklund:
@ezyang : Looks good, although you might want to also add a proper test case for this ?
Apparently the `-l` arguments were always passed to the compiler, giving similar warnings:

```
clang: warning: -lbar: 'linker' input unused
clang: warning: argument unused during compilation: '-Lfoo'
```

It's possible to hash the `-L` _always_ (even for gcc), but probably wait until it too complains...
Just if avoiding the compiler-specific code is more important than ignoring unused parameters.

jrosdahl:
Applied on 3.3-maint in da4fa01e97938fd969fbe0cf616fdaeccd87d6f3 and merged to master. Thanks!

PR #193, opened 2017-08-27
Bugfixes and Tests for nvcc

### LINKED ISSUE #144: Support nvcc (CUDA) -ccbin/--compiler-bindir option

seeraven:
This pull request contains two bugfixes:
- nvcc can not use preprocessed input. If the option ```run_second_cpp``` is set to false, ccache falls back to the behaviour of ```run_second_cpp=true``` for the cuda compiler.
- Unhandled command line options with directories or files caused ccache to fail and fall back to calling nvcc directly without caching. The command line options ```--compiler-bindir/-cdir```, ```--output-directory/-odir``` and ```--libdevice-directory/-ldir``` are now handled by ccache. This should fix #144 .

In addition, tests for nvcc were added by adding two more travis jobs and extending the ```test.sh``` script.

### DIFF
--- a/ccache.c
+++ b/ccache.c
@@ -488,6 +488,15 @@ compiler_is_gcc(struct args *args)
 }
 
 static bool
+compiler_is_nvcc(struct args *args)
+{
+	char *name = basename(args->argv[0]);
+	bool result = strstr(name, "nvcc") != NULL;
+	free(name);
+	return result;
+}
+
+static bool
 compiler_is_pump(struct args *args)
 {
 	char *name = basename(args->argv[0]);
@@ -2140,7 +2149,8 @@ cc_process_args(struct args *args, struct args **preprocessor_args,
 		}
 
 		// Handle cuda "-optf" and "--options-file" argument.
-		if (str_eq(argv[i], "-optf") || str_eq(argv[i], "--options-file")) {
+		if ((str_eq(argv[i], "-optf") || str_eq(argv[i], "--options-file"))
+		    && compiler_is_nvcc(args)) {
 			if (i == argc - 1) {
 				cc_log("Expected argument after %s", argv[i]);
 				stats_update(STATS_ARGS);
@@ -2275,8 +2285,8 @@ cc_process_args(struct args *args, struct args **preprocessor_args,
 			continue;
 		}
 
-		// Alternate form of -o with no space.
-		if (str_startswith(argv[i], "-o")) {
+		// Alternate form of -o with no space. Nvcc does not support this.
+		if (str_startswith(argv[i], "-o") && !compiler_is_nvcc(args)) {
 			output_obj = make_relative_path(x_strdup(&argv[i][2]));
 			continue;
 		}
@@ -2790,6 +2800,11 @@ cc_process_args(struct args *args, struct args **preprocessor_args,
 		stats_update(STATS_SOURCELANG);
 		result = false;
 		goto out;
+	}
+
+	if (!conf->run_second_cpp && str_eq(actual_language, "cuda")) {
+		cc_log("Call cuda compiler with original input, not preprocessed input file.");
+		conf->run_second_cpp = true;
 	}
 
 	direct_i_file = language_is_preprocessed(actual_language);
--- a/test.sh
+++ b/test.sh
@@ -122,6 +122,18 @@ expect_equal_files() {
     fi
 }
 
+expect_different_files() {
+    if [ ! -e "$1" ]; then
+        test_failed "compare_files: $1 missing"
+    fi
+    if [ ! -e "$2" ]; then
+        test_failed "compare_files: $2 missing"
+    fi
+    if cmp -s "$1" "$2"; then
+        test_failed "compare_files:: $1 and $2 are identical"
+    fi
+}
+
 expect_equal_object_files() {
     if $HOST_OS_LINUX && $COMPILER_TYPE_CLANG; then
         if ! which eu-elfcmp >/dev/null 2>&1; then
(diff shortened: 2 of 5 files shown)


cbeck88:
We are now using this patch in tesla autopilot firmware builds.

Previously, we could not use `ccache` with any of our (numerous) cuda targets. Thank you!

test.sh
```diff
@@ -103,6 +103,18 @@ expect_equal_files() {
     fi
 }
 
+expect_different_files() {
+    if [ ! -e "$1" ]; then
+        test_failed "compare_files: $1 missing"
```
jrosdahl: Please don't use tabs in shell scripts.

ccache.c
```diff
@@ -2395,8 +2395,8 @@ cc_process_args(struct args *args, struct args **preprocessor_args,
 			continue;
 		}
 
-		// Alternate form of -o with no space.
-		if (str_startswith(argv[i], "-o")) {
+		// Alternate form of -o with no space. Ignore nvcc -odir.
+		if (str_startswith(argv[i], "-o") && !str_eq(argv[i], "-odir")) {
```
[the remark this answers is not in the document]
seeraven: Fixed it and added a test for it. Actually, the option ```-optf``` caused the same problem, so I fixed that as well. Since nvcc can't handle the ```-o``` option without space, that part is not ignored if nvcc was detected.

test.sh
seeraven: There were no tabs in it. Perhaps from the first commit? I've squashed everything and rebased it on master.

seeraven:
The following changes were made:
- Squashed the commit.
- Rebased on master.
- The ```-o``` option without a separating space is ignored for nvcc. This resolves the ```-odir``` bug when used with gcc.
- The already existing option ```-optf``` is also only parsed when used with nvcc.
- Added another test to ensure the commands ```gcc -optf test.c``` and ```gcc -odir test.c``` work as expected, that is they create the files ```ptf``` resp. ```dir```.
- Removed the combination "clang with cuda" from the ```.travis.yml``` file. Travis uses now clang 5.0 which nvcc does not support (even with CUDA 9). A configuration with clang 4.x could be added, but it would probably require rewriting the ```.travis.yml``` file and putting everything in the ```matrix``` section.

test.sh
jrosdahl: (There were lots of space→tab indent changes in the first commit which then were changed back in a subsequent commit.)

jrosdahl:
Thanks for the updates! Looks mostly good.

Some questions and comments:

1. The nvcc test suites are painfully slow. Would it be possible to speed them up? Otherwise I'm afraid they mostly will be seen as an annoyance... Isn't there a way of compiling a "hello world" type program that is more or less instantaneous?
2. `$compiler_location` in `test.sh` could point to a masquerading ccache (see the `uncached_compile` function), and if so, the "option --compiler-bindir" test will fail. Also, `$compiler_version` isn't supposed to be used by tests (hence the lowercase name). However, I've improved things on master now (see 2b045cbccd40b78c8a7db0be05e21f4db77cebe6), so please update to use `$REAL_COMPILER` instead.
3. I get the following error locally:
    ```Compiler:         gcc (/home/joel/bin/gcc)
    Compiler version: gcc (Ubuntu 5.4.0-6ubuntu1~16.04.5) 5.4.0 20160609
    CUDA compiler:    Cuda compilation tools, release 7.5, V7.5.17 (/usr/bin/nvcc)
    
    Running test suite nvcc.......sh: 1: cicc: not found
    cuobjdump fatal   : Could not open input file 'reference_test1.o'
    sh: 1: cicc: not found
    
    FAILED
    
    Test suite:     nvcc
    Test case:      option --libdevice-directory
    Failure reason: Expected "cache miss" to be 1, actual 0
    ```
    Any ideas? I'm using Ubuntu 16.04 and the nvidia-cuda-toolkit package contains `/usr/lib/nvidia-cuda-toolkit/bin/cicc`.

seeraven:
Update:

I've rebased on master and modified the test.sh script to incorporate your points 1 and 2. nvcc is painfully slow when compiling cuda code. I've modified to use the c++ mode wherever I can to gain speed. However, a few tests must remain using cuda code because that is what we need to test at least once.

Currently, travis is unable to complete the tests because of an error in the pch test suite. I've checked that the same error occurs on the last commit on master, so it wasn't introduced by my latest modifications in the nvcc test suite.

Regarding the last point, I've tested with a travis configuration to ensure supporting 7.5 using the official nvidia repos - that worked. I'll assume you installed the canonical package (because 7.5 is not provided for 16.04 by NVidia). I'll try to reproduce the error with the canonical package, I'll assume the directories used in the failed test are not available. In the updated test.sh I've added a check to ensure the directories used in the test exist.

seeraven:
Update:
The test.sh script now detects the correct paths for the canonical ubuntu package as well as for the official nvidia ubuntu packages. In addition, if any of the required paths is not found, the test of the --libdevice-directory option is skipped.
It's just a feeling, but I think nvcc of CUDA 8 is faster than the one of 7.5.

jrosdahl:
> I've rebased on master and modified the test.sh script to incorporate your points 1 and 2. nvcc is painfully slow when compiling cuda code. I've modified to use the c++ mode wherever I can to gain speed. However, a few tests must remain using cuda code because that is what we need to test at least once.

OK, thanks. For reference, the "different GPU architectures" test takes ≈10s on my system, which is slightly more than how much time the entire existing test suite takes... The full test suite takes ≈9s without nvcc tests and ≈47s with nvcc tests for me.

> Currently, travis is unable to complete the tests because of an error in the pch test suite. I've checked that the same error occurs on the last commit on master, so it wasn't introduced by my latest modifications in the nvcc test suite.

Yes, fixed in 958c4c2.

> The test.sh script now detects the correct paths for the canonical ubuntu package as well as for the official nvidia ubuntu packages. In addition, if any of the required paths is not found, the test of the --libdevice-directory option is skipped.

Now all tests work if I disable my masquerading gcc, but with my default setup, the "option --libdevice-directory" test fails. But I was able to fix it like this:

```
diff --git a/test.sh b/test.sh
index 8d37791..41c1f28 100755
--- a/test.sh
+++ b/test.sh
@@ -3790,7 +3790,7 @@ SUITE_nvcc_ldir_SETUP() {
 }
 
 SUITE_nvcc_ldir() {
-    NVCC_OPTS_CUDA="-Wno-deprecated-gpu-targets -c"
+    NVCC_OPTS_CUDA="-Wno-deprecated-gpu-targets -c -ccbin $REAL_COMPILER"
     CCACHE_NVCC_CUDA="$CCACHE $REAL_NVCC $NVCC_OPTS_CUDA"
     CUOBJDUMP="$REAL_CUOBJDUMP -all -elf -symbols -ptx -sass"
     NVCC_DIR=$(dirname $REAL_NVCC)
```

jrosdahl:
Merged with the mentioned fix and some cleanups. Thanks!

jrosdahl:
Any ideas on how to speed up the tests are of course still welcome.

PR #202, opened 2017-11-05
Add -Wno-implicit-fallthrough to fix gcc7 builds

ldrumm:
gcc-7 has become extremely warny - even for idiomatic code (such as default fallthrough). This does
not fit well with `-Werror` as it causes the build to fail.

### DIFF
--- a/dev.mk.in
+++ b/dev.mk.in
@@ -1,6 +1,6 @@
 # GNU make syntax reigns in this file.
 
-all_cflags += -Werror
+all_cflags += -Werror -Wno-implicit-fallthrough
 all_cppflags += -MD -MP -MF .deps/$(subst .._,,$(subst /,_,$<)).d
 
 ASCIIDOC = asciidoc


afbjorklund:
Duplicate, #187

ldrumm:
@afbjorklund thanks, I didn't see that one in my (admittedly cursory) search.
I'd prefer to leave this one here to give maintainer choice as I believe they're both valid approaches, and let them close it
 
@Bu11etmagnet I think #187 is one of several possible right "ways to it". I believe in the end this is stylistic. It depends on how you view default fallthrough which I personally consider idiomatic C and is the intended behaviour of the code in question. Warnings are useless if fixing them means they make the code in question harder to read.

I personally think a comment describing what obvious code is doing is noise, and default fallthrough is a language rule.

    int a = 1; // assign the value 1 to the variable `a`

The above is self-evident code and the comment is just a distraction. I feel the same way about fallthrough, so I consider `-Wno-implicit-fallthrough` *one of several* right ways to do it.

afbjorklund:
@ldrumm : the comment is (slightly) less annoying, than the new language constructs for it:
`__attribute__ ((fallthrough));`
`[[fallthrough]];`
At least the syntactic comments parses everywhere, even though it's annoying for upstream...

i.e. parts of the patch is for murmurhash, which is an orphan

afbjorklund:
@Bu11etmagnet : We now need a special build just for GCC7, so probably more than a"one-time fix".
But I suppose this _always_ happens, as long as we insist on promoting warnings to errors (`-Werror`)...

afbjorklund:
It would be better to remove `-Werror`, than to remove default warnings.

@ldrumm : it _is_ more of a C language change (I blame C++17-envy), really...
Not sure who decided it was a good idea to change this post-fact, "warny" indeed.

Guess we will just treat this as other annoying lint warnings, and "fix" it.

afbjorklund:
OK, found out who: https://developers.redhat.com/blog/2017/03/10/wimplicit-fallthrough-in-gcc-7/

jrosdahl:
Fixed by #187.

PR #206, opened 2017-12-12
-fdebug-prefix-map is supported in clang 3.8 and newer

thughes:
It's not documented until newer versions of clang, but support was
added in 3.8: https://reviews.llvm.org/rL250094


afbjorklund:
:+1: LGTM

jrosdahl:
Applied on 3.3-maint in 6ef0357e066bbe05e76a10164ae268fffbabeb94 and merged to master. Thanks!

PR #208, opened 2017-12-19
Add fallthrough statements for GCC 7

pietern:
GCC 7 uses -Werror=implicit-fallthrough= by default.


pietern:
Dup of #187, nevermind.

PR #205, opened 2017-12-07
Fix leaked file descriptor

dahuebi:
A leaked file descriptor caused our setup to stall.

We use *ninja* + *ccache* + *distcc* + *avahi/zeroconf*.

PR #216, opened 2018-01-26
Fix broken UNCACHED_ERR_FD, it shouldn't be closed on exec

llunak:
Since that's exactly the point, to leak it to whatever ccache calls.

### DIFF
--- a/ccache.c
+++ b/ccache.c
@@ -3131,11 +3131,11 @@ static void
 setup_uncached_err(void)
 {
 	int uncached_fd = dup(2);
-	set_cloexec_flag(uncached_fd);
 	if (uncached_fd == -1) {
 		cc_log("dup(2) failed: %s", strerror(errno));
 		failed();
 	}
+	// Leak the file descriptor.
 
 	// Leak a pointer to the environment.
 	char *buf = format("UNCACHED_ERR_FD=%d", uncached_fd);


afbjorklund:
Introduced in e7899cec59193c9d7fc7c1d96d250433d130eaf6 (#205)

Can you come up with something that works for _both_ ninja and ccache ?

llunak:
Who, me? No. I don't even know what the Ninja problem is, nor I really know ccache. Only as far as I understand it, UNCACHED_ERR_FD is a fd that an invoked tool is supposed to write errors to, and if it's closed on exec, than the fd is invalid (or a random unrelated fd).

llunak:
> Can you come up with something that works for both ninja and ccache ?

FWIW, if the leaked fd is really a problem for a build system for some reason[*], then ccache could instead introduce some CCACHE_ERROR_PREFIX variable as a replacement for UNCACHED_ERR_FD, and invoked tools then could prefix all their messages with this text, and ccache could remove this prefix from the stderr the first time, and remove such lines completely when storing stderr in the cache.

[*] Not that I see how it could. The build system shouldn't even see a fd that's leaked to processes the build system never even sees.

afbjorklund:
Will see if we can get the real reason why `ninja` decided to hang without it... Otherwise, revert.

afbjorklund:
@dahuebi : do you have more input (or links), to what actually happened and why it was needed ?

jrosdahl:
> Fix broken UNCACHED_ERR_FD, it shouldn't be closed on exec

Oops, darn, damnit, etc. Not my brightest moment to accept #205 without thinking...

Applied, thanks! Also wrote a new test for UNCACHED_ERR_FD in 313e6f4c7b28ddff49bfaa406d8aa5ab55737bc6.

Issue #219, opened 2018-01-31
Keep use relative path to obj dir for gcda to increase hit when changing build path


tof1a:
https://github.com/ccache/ccache/blob/2b045cbccd40b78c8a7db0be05e21f4db77cebe6/ccache.c#L1576-L1579

When using coverage, a hash is done on the coverage path of the gcda file.
Today if no profile_dir is provided, the object dir is used, but ccahe covert is path in a real (absolute) path which do not allow to have a hit when we change build path (different user or server for instance).

SO, if we remove the else and the code inside which convert dir from obj dir when no profile_dir is provided, in order the relative path for obj dir is kept, then we will have a direct hit when we change the path where obj are generated.

jrosdahl:
The logic was introduced in 02d3f078bd2495b8db2264ae0b2c692b4c5ba1bd, seemingly on purpose. @afbjorklund: Do you remember details about this?

tof1a:
Indeed I see in the comment: "_Make sure to hash this path to avoid cache hits with another absolute path._", but what is the pb here ?

tof1a:
pullrequest #220 opened

afbjorklund:
@jrosdahl: Yeah sure.

When I first did the coverage support, I noticed the `.gcno` was using a relative path and thought I was OK. But later on, it was discovered that the `.gcda` was always using an absolute path in the `.o` file.
And this is why we are caching the path to the gcda ("data") - but notice this is **not** the gcno ("notes"). Anyway, the object file contains a hardcoded reference to where the output is created - at runtime.

So if I have two directories, each with their own identical source code, I will get _different_ results:

/tmp/foo/
test.c
test.gcno: test.c
test.o: /tmp/foo/test.gcda

/tmp/bar/
test.c
test.gcno: test.c
test.o: /tmp/bar/test.gcda

This means that if they were to share cache (same hash), they would write the wrong output file...

tof1a:
I have done more investigation based on your suggestions and here my outcomes:
ccache will reuse compiled object that should be identical. In case you are using gcov, object are compiled to generate coverage information at compilation time. In order to be portable, the path where those coverage information should be written is define with an absolute path for a generated object.
This native mechanism makes incompatible, the generated objects with ccache as the an object generated by a user in a specific path will not be reusable by an other user !
The idea then to solve the pb and make ccache working properly is to say to the compiler to not use an absolute path for all generated coverage object but to use a relative path (that was the goal of this change).
[...]

PR #220, opened 2018-02-01
issue #219

tof1a:
This will allow to hash the coverage file with relative path which will increase a lot ccache direct hit between build on different path
Then we will also have a more consistent mechanism, as today path provided with -fprofile-dir is not converted absolute, so we should do the same for the self generated one when not path is provided and object path is used.


afbjorklund:
Will need to write a test case for this, to make sure it is generating the correct .gcno **and** .gcda

jrosdahl:
Thanks for looking into this!

PR #227, opened 2018-02-15
Use correct path to version.c in .gitignore

mostynb:


jrosdahl:
`dev.mk` creates `src/version.c`, not `version.c`, so why should we ignore `version.c` (anywhere) instead?

mostynb:
Somehow, I ended up with a version.c in the root directory of the repository, but I am unable to reproduce this now.  I will close this now, and reopen if I can figure out how this occurred.  Sorry for the noise.

PR #230, opened 2018-02-27
Allow a pre-computed checksum file for precompiled headers

gjkloosterman:
Having large precompiled header files slows down direct mode off
CCache considerably. Allow CCache to check against a (much smaller)
pre-computed checksum file instead of the precompiled header file
itself. This checksum file will be used in the manifest instead of
the actual precompiled header file.

Note that to make this to work, the build system needs to keep a
checksum file in sync with the precompiled header.

The behavior can be activated by the `pch_external_checksum` option.


afbjorklund:
:+1: Neat idea!

doc/MANUAL.txt
```diff
@@ -451,6 +451,14 @@ might be incorrect.
     matching the compiler name in the normal *PATH* that isn't a symbolic link
     to ccache itself.
 
+*pch_external_checksum* (*CCACHE_PCH_EXTSUM*)::
```
jrosdahl: Please change the parenthesis content to the same format as the other boolean settings.

doc/MANUAL.txt
```diff
@@ -451,6 +451,14 @@ might be incorrect.
...
+
+    When this option is set, and ccache finds a precompiled header file,
+    ccache will look for a file with the extension ``.sum`` added
```
jrosdahl: The third and fourth backticks should be apostrophes; see [“Double quoted text”](http://asciidoc.org/asciidoc.css-embedded.html#X51) in the AsciiDoc User Guide.

doc/MANUAL.txt
```diff
@@ -451,6 +451,14 @@ might be incorrect.
...
+    (e.g. ``pre.h.gch.sum``), and if found, it will hash this file instead
```
jrosdahl: The third and fourth backticks should be apostrophes; see [“Double quoted text”](http://asciidoc.org/asciidoc.css-embedded.html#X51) in the AsciiDoc User Guide.

doc/MANUAL.txt
```diff
@@ -451,6 +451,14 @@ might be incorrect.
...
+    of the precompiled header itself.  To work around the performance
```
jrosdahl: Change `itself.  To` to `itself to`.

src/ccache.c
```diff
@@ -638,13 +638,28 @@ remember_include_file(char *path, struct mdfour *cpp_hash, bool system)
 
 	bool is_pch = is_precompiled_header(path);
 	if (is_pch) {
+		bool using_pch_sum = false;
+		if (conf->pch_external_checksum) {
+			// hash pch.sum instead of pch when it exists
+			// to prevent hashing a very large .pch file every time
+			char * pch_sum_path = format("%s.sum", path);
```
jrosdahl: `char * pch` → `char *pch`

src/ccache.c
```diff
@@ -638,13 +638,28 @@ remember_include_file(char *path, struct mdfour *cpp_hash, bool system)
...
+			if (x_stat(pch_sum_path, &st) == 0) {
+				char * old_path = path;
```
jrosdahl: `char * old` → `char *old`

gjkloosterman:
Thanks for the positive responses :-)  I've applied the review suggestions.

FYI, when testing the feature, I've found that .gch files are not fully deterministic: given the same inputs, compiler, and flags, gcc sometimes outputs a .gch with a different checksum.  We assume that given the same set of inputs, the gcc does not necessarily generates *identical*, but at least *equivalent*  .gch files.

To increase the chance of cache hits, we now are experimenting with a workaround on top of this external-pch-checks feature: instead of using the md5sum of the precompiled header as pre.h.gch.sum, we now create a checksum file containing the individual checksums of the files that were used to create the pre.h.gch.  The list of input files is obtained by parsing the pre.h.gch.d dependency output (we use "-MP -MD -MF").   Because CCache includes a checksum of the compiler in its hashes, we can ensure both the compiler and the source files are the same, and therefore the correctness of the .gch file.

Note that extracting the actual inputs from .d files can also be used to create some hybrid of "direct mode" and "run_second_cpp" that completely bypasses the first preprocessor run.  In short, on a direct cache miss, immediately run the actual compiler (with -MD).  Get the include files from the .d file and calculate their hashes.  Use these hashes for both the manifest file and to create the object hash.  This would be most effective when *not* using precompiled headers, and the penalty of running the preprocessor twice is considerably larger.  It will miss some features, we think this could be useful in combination with distcc. The best place to discuss this further would be the mailing list?  Or should I open an issue for it?

jrosdahl:
Applied to master, thanks!

jrosdahl:
> [...] Get the include files from the .d file and calculate their hashes. [...] The best place to discuss this further would be the mailing list? Or should I open an issue for it?

Yes, the mailing list will reach a larger audience. But an issue also works. Or both.

PR #232, opened 2018-02-28
The Dockerfile was moved to a subdirectory

afbjorklund:
Some breakage from the introduction of `misc`


jrosdahl:
Thanks, applied on 3.4-maint.

PR #236, opened 2018-03-10
build: Enable VPATH builds

peterbud:
Previously if build tree was not the same as source tree
the generated version.c landed in a build tree. Also the makefile
could not find the generated doc/ccache.1. Finally make distclean
was not removing the generated docs.

This is my first PR to this repo, I hope it matches the guidelines.

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -132,7 +132,7 @@ unittest/run$(EXEEXT): $(base_objs) $(test_objs) $(extra_libs)
 
 unittest/main.o: unittest/suites.h
 
-unittest/suites.h: $(test_suites) Makefile
+$(srcdir)/unittest/suites.h: $(test_suites) Makefile
 	sed -n 's/TEST_SUITE(\(.*\))/SUITE(\1)/p' $(test_suites) >$@
 
 .PHONY: check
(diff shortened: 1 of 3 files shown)


afbjorklund:
Thanks for doing this!

But the generated files should most likely go in `$(builddir)` rather than `$(srcdir)` ?
It should actually be theoretically possible to have the source directory "read only"

Unfortunately ccache only uses autoconf, not automake, or it would have this feature.
But we can make it work, also with the current makefiles, with some changes like these.

Thoughts, @jrosdahl ?

peterbud:
@afbjorklund , I agree that probably the best solution would be to move the generated files to the build tree. But for a first PR I thought let's start with a small change.  :)

afbjorklund:
You are right, and the proposed change was indeed only for "Developer mode".

afbjorklund:
Think we also need to fix the "test" targets:

`../unittest/main.c:26:20: fatal error: suites.h: No such file or directory`

afbjorklund:
And we should also include the "dist" target

afbjorklund:
@peterbud : I added #237 instead of this, to avoid generating stuff to the source directory...

jrosdahl:
@peterbud: Thanks for working on this! Does @afbjorklund's approach in #237 look good to you?

afbjorklund:
It should accomplish the same thing. It was just more work to do...
https://github.com/ccache/ccache/pull/236#issuecomment-372033585

peterbud:
I'll look into that and test it later when I'm home, but on first sight it looked cool. Quick turnaround!

jrosdahl:
Applied #237, so closing this one as a duplicate.

PR #238, opened 2018-03-11
win32: Remove set_cloexec_flag() from win32 builds

peterbud:
The function body was anyhow empty, so there is no point of this function in the win32 builds.

### DIFF
--- a/src/ccache.h
+++ b/src/ccache.h
@@ -193,11 +193,12 @@ int tmp_unlink(const char *path);
 int x_unlink(const char *path);
 #ifndef _WIN32
 char *x_readlink(const char *path);
+void set_cloexec_flag(int fd);
 #endif
 bool read_file(const char *path, size_t size_hint, char **data, size_t *size);
 char *read_text_file(const char *path, size_t size_hint);
 char *subst_env_in_string(const char *str, char **errmsg);
-void set_cloexec_flag(int fd);
+
 
 // ----------------------------------------------------------------------------
 // stats.c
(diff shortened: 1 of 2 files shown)


jrosdahl:
> The function body was anyhow empty, so there is no point of this function in the win32 builds.

The point is to avoid having to surround each call to `set_cloexec_flag` with `#ifndef _WIN32`. (There were originally three calls, as can be seen in e7899cec59193c9d7fc7c1d96d250433d130eaf6, but one of them was reverted; see #205 and #216.) I think that a better fix would be to remove the `#ifndef` in `init_log` instead, which should have been done in e7899cec but missed somehow. What do you think?

peterbud:
I don't know the history so well. There were two drivers when I looked into this:
* For a win32 build `gcc 7.3` complains for the empty function body:
```
../src/util.c: In function 'set_cloexec_flag':
../src/util.c:1665:22: error: unused parameter 'fd' [-Werror=unused-parameter]
 set_cloexec_flag(int fd)

```
* and when I have checked the code there was only two instances calling `set_cloexec_flag()` function, and one of them was already `#ifdef`-ed, so seemed to me logical to remove the function from win32 build

However I think what you mention is also makes sense, just for win32 we might decorate `fd` with `__attribute__((unused)) ` or something similar to silence `gcc `warning. Do you have any preference for that?

jrosdahl:
> However I think what you mention is also makes sense, just for win32 we might decorate `fd` with `__attribute__((unused))` or something similar to silence `gcc`warning. Do you have any preference for that?

I think I prefer the classic `(void)fd;` solution since that's what's used elsewhere in the code.

peterbud:
OK, then I close this PR!

PR #242, opened 2018-03-28
win32: Fix CCACHE_COMPILERCHECK=mtime test case

peterbud:
Previously if compiler command was an .sh script, win32 version automatically added an .EXE extension to the command line and then test case failed. With this change .sh files won't get appended by an EXE extension and test case successfully runs.

### DIFF
--- a/src/execute.c
+++ b/src/execute.c
@@ -115,9 +115,10 @@ win32getshell(char *path)
 void add_exe_ext_if_no_to_fullpath(char *full_path_win_ext, size_t max_size,
                                    const char *ext, const char *path) {
 	if (!ext || (!str_eq(".exe", ext)
-	             && !str_eq(".bat", ext)
-	             && !str_eq(".EXE", ext)
-	             && !str_eq(".BAT", ext))) {
+				&& !str_eq(".sh", ext)
+				&& !str_eq(".bat", ext)
+				&& !str_eq(".EXE", ext)
+				&& !str_eq(".BAT", ext))) {
 		snprintf(full_path_win_ext, max_size, "%s.exe", path);
 	} else {
 		snprintf(full_path_win_ext, max_size, "%s", path);


afbjorklund:
Running shell scripts on windows sounds a bit niche, but I suppose it can happen in a MinGW context.
Not sure what has happened with the whitespace, but otherwise it looks like it would be good to go...

We still don't have any Windows test environment, though ?

peterbud:
As `ccache` is a tool for compiling software, I believe running it under MSYS2/MINGW is probably the most common usage scenario for `ccache `on windows - at least I'm aware that few softwares are already using it in this way.

I have started working on making **all** testcases running successfully on MSYS2/MINGW environment.. I'm not there there yet, there are still few edge cases where tests are failing and changes are required to handle specifics of this OS, as you can guess mainly the directory separators and nonexistent symlinks are causing problems.

My plan is to submit PRs for each of those changes separately, unless you would like to digest them in once - let me know.

jrosdahl:
> My plan is to submit PRs for each of those changes separately, unless you would like to digest them in once - let me know.

One pull request per logical change sounds excellent.

> And about the whitespaces: As I have seen you are using TAB and not spaces to indent, right? Here spaces were used, which I have replaced with tabs/

Tabs for indentation ("scope level"), spaces for alignment ("aligning arguments and similar"). See also https://github.com/ccache/ccache/blob/master/CONTRIBUTING.md#code-style and the tip on uncrustify.

So it should be like

```
⟶if (!ext || (!str_eq(".exe", ext)
⟶.............&& !str_eq(".sh", ext)
⟶.............&& !str_eq(".bat", ext)
⟶.............&& !str_eq(".EXE", ext)
⟶.............&& !str_eq(".BAT", ext))) {
⟶⟶snprintf(full_path_win_ext, max_size, "%s.exe", path);
```

where `⟶` is a tab and `.` is a space.

(I personally like to both indent and align with spaces, but indentation was done with tabs when I took over the project.)

By the way, perhaps you want to run

    git config --global user.name "Peter Budai"

to fix your Git user name?

I've applied your patch to 3.4-maint now (with whitespace fixed). Will be merged to master eventually.

You can target 3.4-maint directly for bug fixes like these.

afbjorklund:
We might want to try to add an Appveyor build, if we want to support running _on_ Windows (i.e. MSYS) rather than just building _for_ Windows (MinGW). There was an older .appveyor.xml, that could be made "official" ?

But the main problem is that it is hard to find bugs specific to Windows (or Mac), since they are proprietary and none of the developers are running them. Much easier to find issues on Linux, or perhaps even BSD.

peterbud:
> We might want to try to add an Appveyor build, if we want to support running on Windows (MSYS) rather than just building for Windows (MinGW). There was an older .appveyor.xml, that could be made "official" ?

Agree, and I can help with that, has some experience with Appveyor.

> But the main problem is that it is hard to find bugs specific to Windows (or Mac), since they are proprietary and none of the developers are running them. Much easier to find issues on Linux, or perhaps even BSD.

I do run ccache on Windows. The MSYS2 enables to compile ccache to native Windows executable and still you can run it and use it with existing *nix like build tools and scripts. Also you can run (almost) all the tests to find bugs.

PR #246, opened 2018-04-01
Added remaining changes, to make it run on MSYS

### LINKED ISSUE #170: Windows: unittest hangs on Cygwin, test should support MSYS2
Apparently the test suite hangs on Cygwin, due to missing symlink support.

Need to expand on the `_WIN32` used for some test cases.

We should also support "MSYS" as a synonym for Cygwin, in the `uname -s`

This is used by the MSYS2 project: http://www.msys2.org/

afbjorklund:
Installed `msys2-x86_64-20161025.exe`, with `gcc` and `make` (using `pacman`).

After some fixes, it runs in Unix mode (Cygwin). Windows mode (MinGW) left to do.

Most issues were related to symlinks.

Closes #170 

```
unittest/run.exe
PASSED: 484 assertions, 95 tests, 10 suites
CC='gcc' /bin/sh ./test/run
Compiler version: gcc (GCC) 6.4.0

Running test suite base.......................................................
Running test suite nocpp2.......................................................
Running test suite cpp1.
Skipped test suite multi_arch [multiple -arch options not supported on MSYS_NT-6.1]
Skipped test suite serialize_diagnostics [--serialize-diagnostics not supported by compiler]
Running test suite debug_prefix_map..
Skipped test suite masquerading [symlinks not supported on MSYS_NT-6.1]
Running test suite hardlink.
Running test suite direct........................................
Running test suite basedir.....
Running test suite compression.
Running test suite readonly...
Running test suite readonly_direct..
Running test suite cleanup............
Running test suite pch...............
Running test suite upgrade.
Running test suite input_charset.
Skipped test suite nvcc [nvcc is not available]
Skipped test suite nvcc_direct [nvcc is not available]
Skipped test suite nvcc_ldir [nvcc is not available]
Skipped test suite nvcc_nocpp2 [nvcc is not available]
PASSED
```

### DIFF
--- a/test/suites/basedir.bash
+++ b/test/suites/basedir.bash
@@ -4,7 +4,7 @@ SUITE_basedir_SETUP() {
     mkdir -p dir1/src dir1/include
     cat <<EOF >dir1/src/test.c
 #include <stdarg.h>
-#include "test.h"
+#include <test.h>
 EOF
     cat <<EOF >dir1/include/test.h
 int test;
--- a/test/suites/basedir.bash
+++ b/test/suites/basedir.bash
@@ -45,6 +45,7 @@ SUITE_basedir() {
     expect_stat 'cache miss' 2
 
     # -------------------------------------------------------------------------
+ if ! $HOST_OS_WINDOWS && ! $HOST_OS_CYGWIN; then
     TEST "Path normalization"
 
     cd dir1
@@ -64,6 +65,7 @@ SUITE_basedir() {
     expect_stat 'cache hit (preprocessed)' 0
     expect_stat 'cache miss' 1
 
+ fi
     # -------------------------------------------------------------------------
     TEST "Rewriting in stderr"
 
--- a/test/suites/basedir.bash
+++ b/test/suites/basedir.bash
@@ -4,7 +4,7 @@ SUITE_basedir_SETUP() {
     mkdir -p dir1/src dir1/include
     cat <<EOF >dir1/src/test.c
 #include <stdarg.h>
-#include <test.h>
+#include "test.h"
 EOF
     cat <<EOF >dir1/include/test.h
 int test;
--- a/src/lockfile.c
+++ b/src/lockfile.c
@@ -42,7 +42,7 @@ lockfile_acquire(const char *path, unsigned staleness_limit)
 		free(my_content);
 		my_content = format("%s:%d:%d", hostname, (int)getpid(), (int)time(NULL));
 
-#ifdef _WIN32
+#if defined(_WIN32) || defined(__CYGWIN__)
 		int fd = open(lockfile, O_WRONLY|O_CREAT|O_EXCL|O_BINARY, 0666);
 		if (fd == -1) {
 			int saved_errno = errno;
--- a/test/run
+++ b/test/run
@@ -117,6 +117,8 @@ objdump_cmd() {
         xcrun dwarfdump -r0 $1
     elif $HOST_OS_FREEBSD; then
         objdump -W $1
+    elif $HOST_OS_WINDOWS || $HOST_OS_CYGWIN; then
+        strings $1 # for some reason objdump only shows the basename of the file, so fall back to brute force and ignorance
     else
         objdump -g $1
     fi
--- a/test/run
+++ b/test/run
@@ -326,6 +326,7 @@ HOST_OS_APPLE=false
 HOST_OS_LINUX=false
 HOST_OS_FREEBSD=false
 HOST_OS_WINDOWS=false
+HOST_OS_CYGWIN=false
 
 compiler_version="`$COMPILER --version 2>&1 | head -1`"
 case $compiler_version in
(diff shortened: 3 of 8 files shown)


afbjorklund:
@peterbud : I added some work-in-progress windows code to 217a26026a5e33a67cc46954a6c83fcb04c912ae

But there is still a lot of confusion about paths (`C:/foo/bar` versus `C:\foo\bar`)
plus a lot of basic file operations are plain failing, like rename/unlink of "stats" ?
So the Unix-based MSYS(2) is in a better shape than the Windows-based MinGW.
We are defaulting to the 32-bit version, simply since it works better with [Wine](https://www.winehq.org/)...

So I don't think we can say that the Windows ccache version works just yet ?

----

The MSYS(2) installer is found here: http://www.msys2.org/
It will install all three environments (MSYS/MinGW32/MinGW64)

Toolchain is found here: https://mingw-w64.org/doku.php
Normally you would just install the tools using `pacman`, though.

The old [Cygwin](http://www.cygwin.com/) and [MinGW ](http://www.mingw.org/)are now ancient and busted, so I think we'll drop those.

But the MSYS2 (on Windows) or MinGW-w64/Wine (on Linux) are quite useful still...

https://github.com/msys2/msys2/wiki/Porting

https://packages.ubuntu.com/xenial/mingw-w64

afbjorklund:
@peterbud : I put the code changes I had done up on a "windows" branch, do you have a similar one ?

The MSYS target and build should already be addressed ("travis"), but it doesn't actually _run_ anything
The idea is to use Travis to make sure that everything still **builds**, for the Windows (`_WIN32`) target...

We had an earlier effort to start a Appveyor build (#69), but it was never completed - lack of interest

afbjorklund:
This should be rebased to `3.4-maint`, on top of #245 when that is done...

test/suites/basedir.bash
```diff
@@ -4,7 +4,7 @@ SUITE_basedir_SETUP() {
     mkdir -p dir1/src dir1/include
     cat <<EOF >dir1/src/test.c
 #include <stdarg.h>
-#include <test.h>
+#include "test.h"
```
jrosdahl: Just so I understand, why is this needed? Isn't test.h included via the `-I` directive, not via `.`?

afbjorklund: WIll have to retry it on windows, see what the issue might have been (forgot)

afbjorklund: Doesn't seem to be needed anymore, if it ever was...

afbjorklund:
Passed the tests locally:
``` console
$ make test
unittest/run.exe
PASSED: 493 assertions, 99 tests, 10 suites
CC='gcc' /bin/sh ../test/run
Compiler version: gcc (GCC) 7.3.0

Running test suite base.......................................................
Running test suite nocpp2.......................................................
Running test suite cpp1.
Skipped test suite multi_arch [multiple -arch options not supported on MSYS_NT-6.1]
Skipped test suite serialize_diagnostics [--serialize-diagnostics not supported by compiler]
Running test suite debug_prefix_map..
Skipped test suite masquerading [symlinks not supported on MSYS_NT-6.1]
Running test suite hardlink.
Running test suite direct........................................
Running test suite basedir.....
Running test suite compression.
Running test suite readonly...
Running test suite readonly_direct..
Running test suite cleanup............
Running test suite pch.............
Running test suite upgrade.
Running test suite input_charset.
Skipped test suite nvcc [nvcc is not available]
Skipped test suite nvcc_direct [nvcc is not available]
Skipped test suite nvcc_ldir [nvcc is not available]
Skipped test suite nvcc_nocpp2 [nvcc is not available]
PASSED


```

afbjorklund:
@jrosdahl: this should be good to go (I think?), sorry I missed to follow up on that review commit...

As stated, this fixes the tests on MSYS2/MSYS (Cygwin mode), not on MSYS2/MinGW (Windows mode)
(See https://github.com/msys2/msys2/wiki/MSYS2-introduction for the difference between the modes)

peterbud:
Just to state the obvious: This is working for me as well on MSYS2 perfectly!

rkitover:
Hi, I have a related PR #311 

I checked out your branch, and the tests pass for me on latest msys2 both without and with the bundled zlib.

Some notes about the changes:

- Categorizing cygwin as "useless" may be a bit of a stretch, I believe there are far more cygwin users than there are msys2 users. It's very popular in corporate environments etc..
- There is no reason to disable symlinks on cygwin proper, they generally work transparently to POSIX regardless of what winsymlinks is set to in the CYGWIN env var.
- MSYS2 has symlink support, but not unprivileged last I checked and most people don't use this, so it would be fine to disable symlinks for MSYS/MSYS2 for the time being, although having an option to enable them might be nice.
- The mkstemp check I will try to improve in my PR, perhaps with defines, or perhaps with autoconf.

afbjorklund:
Symlinks is not a big deal, we only use them for the optional "masquerading" feature... Might as well leave them out, for both ?

rkitover:
Yeah if it doesn't hurt the functionality then it's no loss.

afbjorklund:
See #170 for details

PR #247, opened 2018-04-07
 CI: add Appveyor script

peterbud:
Currently not all test cases are running successfully, but we can monitor the build progress on Windows, see the results of unittests and test suites, and fix the remaining issues.

32-bit version: [Build process example output](https://ci.appveyor.com/project/peterbud/ccache/build/job/gk47ook208x3m25w)
Unitests: FAILED: 3 tests
Test suits: base/CCACHE_NOHASHDIR fails and aborts

64-bit version: [Build process example output](https://ci.appveyor.com/project/peterbud/ccache/build/job/wmk6chevonp4munb)
Unitests: fails pretty early at SUITE: argument_processing 
Test suits: base/CCACHE_NOHASHDIR fails and aborts


afbjorklund:
Looks like a good start! Can you add a MSYS (Cygwin) build too, using the regular test framework ?

``` yaml
script:
    - ./autogen.sh
    - ./configure $HOST
    - make
    - make ${TEST:-test}
```

That platform has some quirks (missing symlinks, PE executable format) that would be nice to test...

peterbud:
For hiding the clutter: it's pretty common to hide CI related stuff in a separate subfolder, all CI frameworks can handle if files are in a specific subfolder in the source tree, example [here](https://github.com/darktable-org/darktable/tree/master/.ci)

peterbud:
@afbjorklund : You can configure the location of the `appveyor.xml` when you enable Appveyor for a given Github repo:
image

That;s for example how I'm using it for darktable.

afbjorklund:
As for the CCACHE_NOHASHDIR and other Windows test failures, most have been fixed in 217a26026a5e33a67cc46954a6c83fcb04c912ae
They are rmostly elated to paths, either the wrong delimiter (`/` '\\`) is being used - or it needs "escaping".

I will clean that up into individual commits - after we have gotten the MinGW and MSYS merged (#246).
Also seems like my autoconf check for `GetFinalPathNameByHandle` doesn't work properly anymore...

```
checking for GetFinalPathNameByHandleW... no
```
From https://github.com/ccache/ccache/commit/a21da6fff6e7f796c42b1a6f784d42945395025b
The fallback code also had some stupid bug, that converted `C:\Users` to `sers`

afbjorklund:
Fixing Windows build is related to #122

peterbud:
@afbjorklund  I have added MSYS to the CI matrix, based on your suggestions I have corrected the script. Also I have realized that you had in an earlier (not yet merged) PR #246 a solution for fixing log printing: I have taken that here as well to have the same code base.

afbjorklund:
I think this will be a good addition to the Travis builds, once those have been cleaned up...
Currently travis will only _build_ the unittest, but not run it. But it might be able to in Wine.
I think some of the latest changes for MSYS/MinGW might actually have _broken_ Windows,
but I suppose that can be cleaned up by adding some kind of "vanilla" Win build target too ?

i.e. without bash, with all the legacy DOS weirdness. and then just execute `unittest/run.exe`

Though that is a much *bigger* change, than "just" fixing the Cygwin and Windows build targets.

peterbud:
The MINGW32/MINGW64 builds are native builds, and if you run `unittest/run.exe` from the bash is the the same if you would run that from command prompt or from powershell. Even I believe we can add powershell script to the CI if you want, but in order to do that we need to make a proper deployment in the CI script, as the compiled executable has dependencies (MSYS2 runtime DLL and zlib) which needs to be copied next to the executable, as they are not on the PATH if running from a simple command prompt/powershell.

Let me know if you want to add a powershell/cmd prompt test part.

Also based on my tests some of the test cases are broken on Windows, but not too many, so with thorough work we can fix them one-by-one.

afbjorklund:
Just thought I saw some additional issues, when running it with Wine ? (plus some Wine bugs)

Anyway, the priorities are something like the following:

1) Fix the Cygwin build, this should be pretty easy and goes into 3.4
Mostly it is about missing symlinks, that funky .exe extension and the PE executable format...

2) Try to address the issues running with MinGW (both 32 and 64)
This might require some additional investigation of DOS quirks like paths and batch files etc.

3) Perhaps look into supporting Windows as a real developer platform.
This is the final and least likely option, since it requires a lot of changes. But the goal of #169

peterbud:
I have updated this to resolve the conflict to master, but I'm not convinced that I did it correctly. What is your preference with this PR?

jrosdahl:
I'm going through old open PRs and have now come to this one.

> I have updated this to resolve the conflict to master, but I'm not convinced that I did it correctly.

No, the diff doesn't look right. Also, it seems be based on 3.4-maint, not master.

> What is your preference with this PR?

I don't know, frankly. I have now written a kind of brain dump on my view on Windows support here: #447.

peterbud:
Let's close it here. I'll open a new PR with an updated CI proposal

PR #253, opened 2018-04-12
Print compiler's stderr when compiling with memcached_only=true

mlopatkin:
In the memcached_only mode no files are created in the local cache. All
compilation results are stored in memcached backend but the routine
responsible for running a compiler and storing its outputs -
to_memcached - still used send_cached_stderr to present compiler's
stderr to the user. The latter function relied on the existence of
cached stderr in the local cache printing nothing.

The fix is to print the stderr data sent to memcached backend instead.

### DIFF
--- a/src/ccache.c
+++ b/src/ccache.c
@@ -1682,7 +1682,10 @@ to_memcached(struct args *args)
 	}
 
 	// Everything OK.
-	send_cached_stderr();
+	if (stderr_l) {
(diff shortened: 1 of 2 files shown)


afbjorklund:
Sounds reasonable, maybe the same test should be in base too

jrosdahl:
Applied on dev/memcached, thanks.

PR #259, opened 2018-04-24
Warning fixes, bug fix, meson port and msvc build

havardgraff:
I have been hacking a bit on ccache to be able to use on our windows-build.

### DIFF
--- a/config.h.meson
+++ b/config.h.meson
@@ -1,0 +1,139 @@
+#ifndef CCACHE_CONFIG_H
+#define CCACHE_CONFIG_H
+
+/* Define on Linux to activate all library features */
+#mesondefine _GNU_SOURCE
+
+const static char CCACHE_VERSION[] = @CCACHE_VERSION@;
+
+/* HEADERS */
+
+/* Define to 1 if you have the <sys/mman.h> header file. */
+#mesondefine HAVE_SYS_MMAN_H
+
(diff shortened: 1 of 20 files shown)


afbjorklund:
Seems like there is some overlap between this and #169 ?

Except that this one goes further, and adds a new buildsystem

jrosdahl:
Hi Havard, thanks for your interest in ccache.

I have applied the first two commits since they are isolated and reasonable. In the future, I would appreciate it if you would publish separate pull requests for different bug fixes and features.

I won't apply the Meson part — such a change should be discussed by the ccache community first. I'm not very interested in maintaining both an autotools and another build system, so the port should ideally be a complete drop-in for the current autotools solution and work satisfactory on all target systems of interest. I think that it would be a good idea to bring it up on [the mailing list](https://lists.samba.org/mailman/listinfo/ccache/). (Also, the Meson commit seems to contain both bug fixes and other changes, so it's hard to review it in a good way. Feel free to submit those fixes in separate pull requests.)

Regarding using another build system than autotools, see also the CMake discussion at https://www.mail-archive.com/ccache@lists.samba.org/msg01334.html.

jrosdahl:
@havardgraff: Closing this issue now. Please submit the remaining parts (standalone bug fixes and Meson conversion) as separate pull requests if you're still interested in driving this effort. As I mentioned above, it would be a good idea to discuss pros and cons of converting to Meson on the ccache mailing list.

PR #266, opened 2018-06-06
 Add SASL authentication support for memcached

afbjorklund:
Using couchbase for the memcached storage now _requires_ using SASL authentication:

* https://blog.couchbase.com/sasl-memcached-now-available/

* https://developer.couchbase.com/documentation/server/5.0/security/security-bucket-protection.html

Also add some docs and better error diagnostics, when we cannot connect to a server.


afbjorklund:
@jrosdahl: added this code a long time ago, but apparently only on the couchbase branch...

jrosdahl:
Applied, thanks.

PR #267, opened 2018-06-07
Add configure option to disable man pages

### LINKED ISSUE #264: Reduce deps for memcached
To `make install` the `memcached` branch it seems that you need to install 
```
apt-get install --no-install-recommends asciidoc xsltproc docbook-xml docbook-xsl
```
Is there any way to remove those deps with a configuration option (something like --no-docs ) ?

afbjorklund:
Building the man page requires asciidoc, if building from the source
repository. Add option to only install the program, but not the man.

When building from the release tarball, the man page is already built.
So we can still default to installing it, and requiring asciidoc/a2x.

Closes #264

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -101,11 +101,11 @@ ccache.1: doc/ccache.1
 	cp $< $@
 
 .PHONY: install
-install: ccache$(EXEEXT) ccache.1
+install: ccache$(EXEEXT) @disable_man@ccache.1
 	$(installcmd) -d $(DESTDIR)$(bindir)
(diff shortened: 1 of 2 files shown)


jrosdahl:
This was applied to master some months ago. Thanks!

cbeck88:
Are you sure this was actually applied to master? It doesn't appear that there is any such option in v3.4.3

afbjorklund:
This was rebased as: a59166b5a9815e18d83756fa957cd441667d992f

afbjorklund:
@cbeck88 : 3.4.3 is from `3.4-maint`, not from `master` (which is going to be 3.5)

cbeck88:
Thanks a bunch for explaining!

PR #270, opened 2018-06-10
Add a `-k/--get-config` option to print the value of a configuration key

gasche:
For a scripting application I needed to get the `cache_dir` value; this can be grepped from `--print-config` but it is fairly inconvenient. This PR adds a `-k KEY` or `--get-config=KEY` option that prints the value of a given configuration key (and a newline). The origin or the key are not printed.

To do this I had to add a field printer to the `confitem` struct, following the general structure of the per-type parsing functions. See individual commit messages for details.

### DIFF
--- a/src/ccache.c
+++ b/src/ccache.c
@@ -66,6 +66,7 @@ static const char USAGE_TEXT[] =
 	"    -C, --clear           clear the cache completely (except configuration)\n"
 	"    -F, --max-files=N     set maximum number of files in cache to N (use 0 for\n"
 	"                          no limit)\n"
+        "    -k, --get-config=K    get the value of the configuration key K\n"
 	"    -M, --max-size=SIZE   set maximum size of cache to SIZE (use 0 for no\n"
 	"                          limit); available suffixes: k, M, G, T (decimal) and\n"
 	"                          Ki, Mi, Gi, Ti (binary); default suffix: G\n"
@@ -3544,6 +3545,7 @@ ccache_main_options(int argc, char *argv[])
 		{"cleanup",       no_argument,       0, 'c'},
 		{"clear",         no_argument,       0, 'C'},
 		{"dump-manifest", required_argument, 0, DUMP_MANIFEST},
+		{"get-config",    required_argument, 0, 'k'},
 		{"hash-file",     required_argument, 0, HASH_FILE},
 		{"help",          no_argument,       0, 'h'},
 		{"max-files",     required_argument, 0, 'F'},
@@ -3557,7 +3559,8 @@ ccache_main_options(int argc, char *argv[])
 	};
 
 	int c;
-	while ((c = getopt_long(argc, argv, "cChF:M:o:psVz", options, NULL)) != -1) {
+	while ((c = getopt_long(argc, argv, "cCk:hF:M:o:psVz", options, NULL))
+               != -1) {
 		switch (c) {
 		case DUMP_MANIFEST:
 			manifest_dump(optarg, stdout);
@@ -3594,6 +3597,16 @@ ccache_main_options(int argc, char *argv[])
 		case 'h': // --help
 			fputs(USAGE_TEXT, stdout);
 			x_exit(0);
+
+		case 'k': // --get-config
+		{
+			initialize();
+			char *errmsg;
+			if (!conf_print_value(conf, optarg, stdout, &errmsg)) {
+				fatal("%s", errmsg);
+			}
+		}
+		break;
 
 		case 'F': // --max-files
 		{
--- a/src/conf.c
+++ b/src/conf.c
@@ -629,161 +629,65 @@ conf_set_value_in_file(const char *path, const char *key, const char *value,
 	return true;
 }
 
+static bool
+print_item(struct conf *conf, const char *key,
+           void (*printer)(const char *descr, const char *origin,
+                           void *context),
+           void *context)
+{
+	const struct conf_item *item = find_conf(key);
+	if (!item) {
+		return false;
+	}
+	void *value = (char *)conf + item->offset;
+	char *str = (char *)item->formatter(value);
+	char *buf = x_strdup("");
+	reformat(&buf, "%s = %s", key, str);
+	printer(buf, conf->item_origins[item->number], context);
+	free(buf);
+	free(str);
+	return true;
+}
+
 bool
 conf_print_items(struct conf *conf,
                  void (*printer)(const char *descr, const char *origin,
                                  void *context),
                  void *context)
 {
-	char *s = x_strdup("");
-
-	reformat(&s, "base_dir = %s", conf->base_dir);
-	printer(s, conf->item_origins[find_conf("base_dir")->number], context);
-
-	reformat(&s, "cache_dir = %s", conf->cache_dir);
-	printer(s, conf->item_origins[find_conf("cache_dir")->number], context);
-
-	reformat(&s, "cache_dir_levels = %u", conf->cache_dir_levels);
-	printer(s, conf->item_origins[find_conf("cache_dir_levels")->number],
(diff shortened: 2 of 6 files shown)


afbjorklund:
:+1: Looks good to me

gasche:
Minor note: it was not obvious to me from the codebase how one is supposed to fetch a configuration value from its key (as a string) and a `conf`. The solution, if an `item` exists for the configuration key, is to use `(char *)conf + item->offset`, as is done in the existing `handle_conf_setting` function.

I think that it could be nice for code readability/discoverability to have this line of code exposed as (static) getter and setter functions in the `conf` module:

```c
static void *get_item_value(struct conf *conf, struct conf_item *item);
void set_item_value(struct conf *conf, struct conf_item *item, void *value);
```

src/ccache.c
```diff
@@ -59,6 +59,7 @@ static const char USAGE_TEXT[] =
   "    -M, --max-size=SIZE   set maximum size of cache to SIZE (use 0 for no\n"
   "                          limit); available suffixes: k, M, G, T (decimal) and\n"
   "                          Ki, Mi, Gi, Ti (binary); default suffix: G\n"
+  "    -k, --get-config=K    get the value of the configuration key K\n"
```
jrosdahl: Please keep the option list in sorted order.

src/ccache.c
```diff
@@ -3461,6 +3462,7 @@ ccache_main_options(int argc, char *argv[])
 		{"help",          no_argument,       0, 'h'},
 		{"max-files",     required_argument, 0, 'F'},
 		{"max-size",      required_argument, 0, 'M'},
+		{"get-config",    required_argument, 0, 'k'},
```
jrosdahl: Please keep the option list in sorted order.

src/ccache.c
```diff
@@ -3470,7 +3472,7 @@ ccache_main_options(int argc, char *argv[])
 	};
 
 	int c;
-	while ((c = getopt_long(argc, argv, "cChF:M:o:psVz", options, NULL)) != -1) {
+	while ((c = getopt_long(argc, argv, "cChF:M:k:o:psVz", options, NULL)) != -1) {
```
jrosdahl: Please keep the option list in sorted order. Also, it would be nice to keep the line [a bit narrower](https://github.com/ccache/ccache/blob/master/CONTRIBUTING.md#code-style).

src/ccache.c
```diff
@@ -3533,6 +3535,16 @@ ccache_main_options(int argc, char *argv[])
 		}
 		break;
 
+		case 'k': // --get-config
```
jrosdahl: Please keep the option list in sorted order.

src/conf.c
```diff
@@ -19,13 +19,15 @@
...
 typedef bool (*conf_item_verifier)(void *value, char **errmsg);
+typedef const char *(*conf_item_printer)(void *value);
 
 struct conf_item {
 	const char *name;
 	size_t number;
 	conf_item_parser parser;
 	size_t offset;
 	conf_item_verifier verifier;
+        conf_item_printer printer;
```
jrosdahl: Please indent with a tab character.

src/conf.c
```diff
@@ -50,6 +52,19 @@ parse_bool(const char *str, void *result, char **errmsg)
...
+bool_to_string(bool value)
+{
+	return value ? "true" : "false";
+}
+
+static const char *
+print_bool(void *value)
+{
+	bool *b = (bool *)value;
+        return x_strdup(bool_to_string(*b));
```
jrosdahl: Indent using tab.

src/conf.c
```diff
@@ -19,13 +19,15 @@
 
 typedef bool (*conf_item_parser)(const char *str, void *result, char **errmsg);
 typedef bool (*conf_item_verifier)(void *value, char **errmsg);
+typedef const char *(*conf_item_printer)(void *value);
```
jrosdahl: The convention in ccache is to name functions that convert types to string like `format_*`, so please rename this to `conf_item_formatter` and the `print_*` functions accordingly.

src/conf.c
```diff
@@ -535,152 +623,79 @@ conf_set_value_in_file(const char *path, const char *key, const char *value,
 }
 
 bool
-conf_print_items(struct conf *conf,
-                 void (*printer)(const char *descr, const char *origin,
-                                 void *context),
-                 void *context)
+conf_print_value(struct conf *conf, const char *key,
+                 void *context, char **errmsg)
```
jrosdahl: It's a bit surprising that `context` here is not the same thing as the `context` parameter to `conf_print_items`. Perhaps change from `void *context` to `FILE *file`? And perhaps rename to `conf_print_value_to_file` or introduce a `printer` function parameter like in `conf_print_items`? Doing the latter would open up for writing unit tests without having to print to an actual file.

unittest/test_conf.c
```diff
@@ -381,6 +381,50 @@ TEST(conf_set_existing_value)
...
 }
 
+TEST(conf_print_existing_value)
+{
+	struct conf *conf = conf_create();
+	conf->max_files = 42;
+	char *errmsg;
+	{
+		FILE *log = fopen("log", "w");
+                CHECK(log);
```
jrosdahl: Indent with tab here and for the three other `CHECK(log)` calls below.

gasche:
I'll get back to this PR sometimes next week, and do the changes suggested, thanks!

jrosdahl:
> I'll get back to this PR sometimes next week, and do the changes suggested, thanks!

@gasche: Any news here?

gasche:
Sorry, the PR slipped out of my worklist... I just rebased against master, and starting fixing your remarks.

src/ccache.c
gasche: To clarify, you expect to sort on the first component (`get-config` here), not the last component (`k`), right? But then those will not be in the same order as the `getopt_long` string argument?

gasche:
@jrosdahl I tried to address all your demands, let me know if something is still off.

gasche:
(I realize I forgo to take two of your comments into account, I'll try to get to that shortly.)

src/conf.c
gasche: I went for `FILE *file`. I tried first to get a generic `printer` function, but it is not the same signature as for `conf_print_items` and I didn't find an approach that I really liked, so I left this further change for someone else / later.

src/ccache.c
gasche: I wasn't completely sure whether to sort on the short-name or the long-name, but I went for the short name.

src/ccache.c
gasche: I sorted on the short component here (the long name, not the short name). Note that the list in master is already not sorted, with `print-config` after `set-config`.

src/ccache.c
gasche: I cut the line before `!= 1`.

src/conf.c
gasche: Done, thanks!

gasche:
I went over all the comments now. Let me know if further changes are necessary.

jrosdahl:
Applied to master, thanks! (There was still some indentation that was off, but no worries.)

afbjorklund:
Getting some const-correctness warnings from this, the `formatter` usage looks odd.

```
src/conf.c:642:22: error: cast from 'const char *' to 'char *' drops const qualifier [-Werror,-Wcast-qual]
        char *str = (char *)item->formatter(value);
                            ^
src/conf.c:659:22: error: cast from 'const char *' to 'char *' drops const qualifier [-Werror,-Wcast-qual]
        char *str = (char *)item->formatter(value);
                            ^
```

Will open a new PR about it.

afbjorklund:
Added #321

Issue #148, opened 2016-11-07
C99 requires -D_XPG6 on Solaris


dago:
I am trying to compile ccache 3.3.3 on Solaris 10 with Oracle Studio 12 and get a compilation error.

The required flags for C99 are detected as follows:
```
checking for /opt/SUNWspro/bin/cc option to accept ISO C89... none needed
checking for /opt/SUNWspro/bin/cc option to accept ISO C99... -D_STDC_C99=
checking how to run the C preprocessor... /opt/SUNWspro/bin/cc -D_STDC_C99= -E
```
which results in a compilation error with inconsistently set preprocessor flags:
```
/opt/SUNWspro/bin/cc -D_STDC_C99= -DHAVE_CONFIG_H -DSYSCONFDIR=/etc/opt/csw -I. -I. -I/opt/csw/include -I./zlib -xO3 -m32 -xarch=sparc -O -c -o args.o args.c
"/usr/include/sys/feature_tests.h", line 337: #error: "Compiler or options invalid for pre-UNIX 03 X/Open applications  and pre-2001 POSIX applications"
```

The addition of `-D_XPG6` to `CPPFLAGS` solves the issue.

afbjorklund:
Seems like `AC_USE_SYSTEM_EXTENSIONS` will get rid of this, but then again it might be overkill...
https://www.gnu.org/software/autoconf/manual/autoconf-2.60/html_node/Posix-Variants.html

If used, we also need `AC_PREREQ(2.60)` (up from current 2.52)

Also hard to fix all these proprietary platforms bugs, without access to a proper compile farm.
Whether it is Windows or Macintosh or Solaris, it might probably need "downstream" porters ?

I'll leave it to @jrosdahl, how to best address this SunOS issue.

afbjorklund:
Possibly related to 9485354a5542f901dbca9f96930d036bdbdaaf8b

afbjorklund:
Maybe moving `m4_include(m4/feature_macros.m4)` above `AC_PROG_CC_C99` is enough ?

dago:
Hi Anders,

Am 13.11.2016 um 17:26 schrieb Anders Björklund <notifications@github.com>:
> Seems like AC_USE_SYSTEM_EXTENSIONS will get rid of this, but then again it might be overkill...
> https://www.gnu.org/software/autoconf/manual/autoconf-2.60/html_node/Posix-Variants.html
> 
[...]

Issue #214, opened 2018-01-25
Configure fails under SmartOS native zone


jcea:
Configure fails. Checking "config.log" I see this:

```
uname -m = i86pc
uname -r = 5.11
uname -s = SunOS
uname -v = joyent_20180104T174227Z
[...]
configure:2995: checking whether we are cross compiling
configure:3003: gcc -o conftest    conftest.c  >&5
In file included from /usr/include/stdio.h:37:0,
                 from conftest.c:17:
/usr/include/sys/feature_tests.h:400:2: error: #error "Compiler or options invalid; UNIX 03 and POSIX.1-2001 applications       require the use of c99"
 #error "Compiler or options invalid; UNIX 03 and POSIX.1-2001 applications \
  ^
configure:3007: $? = 1
configure:3014: ./conftest
./configure[3016]: eval[1]: ./conftest: not found [No such file or directory]
configure:3018: $? = 127
configure:3025: error: in `/tmp/ccache-3.3.5':
configure:3027: error: cannot run C compiled programs.
If you meant to cross compile, use `--host'.
See `config.log' for more details
```

jcea:
It works if I do `CC=gcc -std=c99` first.

afbjorklund:
Possibly related to #148, but `AC_PROG_CC_C99` is _supposed_ to be able to figure out `-std`...

http://git.savannah.gnu.org/gitweb/?p=autoconf.git;a=blob;f=lib/autoconf/c.m4#l490

jrosdahl:
I will not dedicate time to this issue, but I'll gladly apply any fix if it's clean enough.

jrosdahl:
Closing this issue now due to lack of progress. Also, the next ccache version will require a C++11 compiler.

PR #271, opened 2018-06-11
test: Fix various Solaris shell issues

totph:
Change `sh` to `bash`, use `find` with `-exec rm`, and avoid `+=` on strings. These are issues which require source changes, unlike the [missing -D_XPG6 flag](https://github.com/ccache/ccache/issues/148) or the missing `seq` binary.

### DIFF
--- a/test/run
+++ b/test/run
@@ -267,7 +267,7 @@ fi
 HOST_CCACHE_DIRS="/usr/lib/ccache/bin
 /usr/lib/ccache"
 for HOST_CCACHE_DIR in $HOST_CCACHE_DIRS; do
-    PATH=$(echo -n $PATH | awk -v RS=: -v ORS=: '$0 != "'$HOST_CCACHE_DIR'"' | sed 's/:$//')
+    PATH="$(echo "$PATH:" | awk -v RS=: -v ORS=: '$0 != "'$HOST_CCACHE_DIR'"' | sed 's/:*$//')"
 done
 export PATH
 
--- a/test/suites/base.bash
+++ b/test/suites/base.bash
@@ -495,7 +495,8 @@ EOF
 
     cat >compiler.sh <<EOF
 #!/bin/sh
-export CCACHE_DISABLE=1 # If $COMPILER happens to be a ccache symlink...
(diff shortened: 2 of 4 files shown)


afbjorklund:
Think that we have talked about similar/related Solaris issues before, like #148 and #214 ?

But since it is much harder to test, it means that things outside of Linux will see less testing.

I'm surprised that `/bin/bash` even exists.

Think it was `/usr/local/bin/bash` on BSD...

totph:
I fixed the `/bin/sh` issues by not using `export` with assignment. This is a solaris `sh` bug it seems.

Technically, the `find -delete` option is not POSIX, so use `rm {} +`.

The new `PATH=..` line contains a non Solaris bugfix (the missing `:` would make awk not remove the last path entry, should I extract that into a separate commit?) and the Solaris fix that `sed` returns nothing if the input is not newline terminated.

And as long as these fixes do not break something on another OS, shouldn't they be Ok?

afbjorklund:
They are probably OK, just wondered if it was possible to test (and thus support) it ?
I added "shellcheck", but it doesn't know about SunOS bugs and I checked as bash...

jrosdahl:
> Change sh to bash, use find with -exec rm, and avoid += on strings. These are issues which require source changes, unlike the missing -D_XPG6 flag or the missing seq binary.

@totph: These seem fine to me, except that I don't understand why `+=` should be avoided. Doesn't that work in bash on Solaris?

totph:
>  why+= should be avoided.

Solaris 10 (still supported) only ships bash version 3, which gets confused by `+=` statements:

    $ bash --version
    GNU bash, version 3.00.16(1)-release (sparc-sun-solaris2.10)
    Copyright (C) 2004 Free Software Foundation, Inc. 
    $ i="abc"; i+="def"
    bash: i+=def: command not found

afbjorklund:
Hmm, shellcheck doesn't have any checks for bash3 (only "bash")
If we want to support this ancient OS, we need a legacy container

afbjorklund:
IIRC, it's the same situation on OS X and it's because of GPLv3

jrosdahl:
I've applied this on master now, thanks!

PR #272, opened 2018-06-11
Add support for running ShellCheck on the test suites

afbjorklund:
See: https://www.shellcheck.net/

I left some of the issues excluded, mostly involving "legacy" (sh) and whitespace issues...

You can comment the excludes out, to see what it is complaining about. Ditto for test/run.

``` console
$ shellcheck test/run | grep -c ^In
54
```


jrosdahl:
Applied, thanks.

Issue #274, opened 2018-06-17
Add more details about compression


afbjorklund:
Currently compression is only mentioned briefly mentioned in the manual.

Should expand upon this, and give more information and benchmarks.

This could involve new commands (or tools), to measure and test.

Also makes it easier to compare different compression types and levels.

afbjorklund:
As there are no tools anymore and no different compression types, closing this.

Issue #218, opened 2018-01-27
Idea: Introduce aggregated result files


jrosdahl:
Idea: We could introduce "aggregated result files" (other name suggestions welcome) that store all of a cached result in one file instead of in one to seven files as done on master at the time of writing.

This idea has previously been mentioned [on the mailing list](https://www.mail-archive.com/ccache@lists.samba.org/msg01185.html) and in https://github.com/ccache/ccache/pull/58#issuecomment-254638407 and https://github.com/ccache/ccache/issues/213#issuecomment-359385579.

Currently, a cached result is stored in the cache in the form `$md4hash-$length.o`, `$md4hash-$length.stderr`, `$md4hash-$length.d`, `$md4hash-$length.gcno`, `$md4hash-$length.su`, `$md4hash-$length.dia` and `$md4hash-$length.dwo`. This has a couple of advantages:

* It's a straightforward and easy to understand.
* It's possible to use hard links to copy files from the cache.

There are advantages to storing the result in a single file, though:

* Less i-nodes and directory entries are used. This is good for at least two reasons:
  1. Cache hit/miss performance will be higher due to less (potential) disk seeks and less system calls.
  2. Cleanup performance will be higher since smaller directories need to be scanned, less `stat`s need to be made for the cleanup algorithm's LRU behavior, etc.
* The result can less easily become corrupt. For example, if one of the currently stored files becomes zero length in the cache (which could happen on flaky systems or due to power cycling, etc.) the whole result is invalid but ccache can't detect it and will deliver a broken result.

jrosdahl:
Replying to @afbjorklund's comment in https://github.com/ccache/ccache/issues/213#issuecomment-359567866:

> It got tedious, to pile on new file formats all the time...

I don't see why that would be an issue if a more flexible file format is used (similar to what you sketched further down in https://github.com/ccache/ccache/pull/58#issuecomment-255547818)?

> Not sure how it would work in the context of #98, though ? Harder to share object files, done this way.

I don't think that it will be significantly harder. On the other hand, cache deduplication is very low priority for me since it adds complexity.

> Adding support for .pack files would be great, too. Then you wouldn't have all those pesky small files.

I'd be happy to hear more about your ideas here, but spontaneously I'm a bit sceptical. :slightly_smiling_face: Do you mean the actual Git .pack format or just something similar? And should it then be a per-cache file or per-level-1 or per-level 2 file, etc? How should cleanup be done?

afbjorklund:
[...]

PR #275, opened 2018-06-17
Add command to show compression statistics, and a tool to compress

afbjorklund:
By default only the .manifest files are compressed, but add a tool to measure the amount of compression.

Also do a small wrapper to help compress all files with `--fast`, optionally to recompress using `--best`.

``` console
$ ccache -x
Compressed size: 9.8 MB, 22 files
Uncompressed size: 10.0 MB, 68 files
Compressed files: 32.35 %
Compression ratio: 1.34 % (1.0x)
$ ccache-compress 
$ ccache -x
Compressed size: 2.5 MB, 68 files
Uncompressed size: 10.0 MB, 68 files
Compressed files: 100.00 %
Compression ratio: 75.41 % (4.0x)
```

For large cache directories, compressing will take a *long* time... But decompression should be fast.

Once the benchmarking tools have been updated, they should have more compress options as well...

Part of #274

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -29,6 +29,7 @@ non_3pp_sources = \
     src/ccache.c \
     src/cleanup.c \
     src/compopt.c \
+    src/compress.c \
     src/conf.c \
     src/counters.c \
     src/execute.c \
--- a/misc/ccache-compress
+++ b/misc/ccache-compress
@@ -1,0 +1,38 @@
+#!/bin/sh
+
+# This will compress (or recompress) the ccache cache directory,
+# using the best compression method (level 9) available with gzip.
+
+CCACHE=${CCACHE:-ccache}
+CCACHE_DIR=${CCACHE_DIR:-$HOME/.ccache}
+
+# Set RECOMPRESS=false to avoid recompressing less compressed files.
+# Set DECOMPRESS=true to reverse the process and decompress cache.
+
+RECOMPRESS=${RECOMPRESS:-true}
+DECOMPRESS=${DECOMPRESS:-false}
+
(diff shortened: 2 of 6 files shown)


jrosdahl:
I'd rather wait with this until we have decided the future of aggregated files (#218) and if we should switch to LZ4, either as the only compression algorithm or as the default of several.

Does this sound reasonable?

afbjorklund:
I think it is just one more function to the compression "API", in order to dry-run the compressed file.

jrosdahl:
Now that the work related to aggregated files and compression has settled, I think that this feature would be awesome to have in the upcoming ccache version.

@afbjorklund: Would you be interested in reworking this PR to work with the new file formats and to implement recompression in ccache itself instead of a standalone tool? (If not, I can have a look at it myself, no problem.)

I'm imagining one option for showing compression statistics, just like the proposed `-x` flag, and one option for recompressing the cache. Maybe like this?

```
    -x, --show-compression    show compression statistics
                              
    -X, --recompress=LEVEL    recompress the cache using compression level
                              LEVEL (positive or negative integer, or
                              "uncompressed" for no compression)
```

Thoughts?

afbjorklund:
The main problem with the recompression is the threading, but I suppose it can start out single threaded.

jrosdahl:
> The main problem with the recompression is the threading, but I suppose it can start out single threaded.

I agree, no need to parallelize in a first version.

afbjorklund:
Will close this PR, and make a new one for the new format

afbjorklund:
Reimplemented in #440

PR #277, opened 2018-06-18
Add support for clang -MJ compile command / compilation database

totph:
Using `clang`, the `-MJ output.ccmd` flag writes the compiler call to the given file. After concatenating multiple of these into a single `compile_commands.json` file, it can be used by various clang tools, e.g. clang-tidy [1].

Currently `ccache` silently ignores `-MJ`, i.e. a non-cached call of `clang` will produce a compile command file. However, after deleting it and the object and calling `clang` again with `ccache`, the compile command file will not be restored.

This patch adds some support for caching the compile command file itself.  It is stored with a placeholder in the cache and always modified/copied to contain the current directory.

What is **missing** is proper json support, which means escaping non 7-bit path names, not depending on the json fragment always starting with  `{ "directory":` [2], and even tolerating newlines.  Currently `fatal()` is called instead. Also, the `-fcolor-diagnostics` argument added by `ccache` is always present.

For my purposes this current patch works good enough, so this is more a "FYI" if someone has similar issue than a finished pull request. I don't know if `ccache` ever wants to add a dependency to a json library, so this might remain as-is.

1: https://clang.llvm.org/docs/JSONCompilationDatabase.html
2: from the compile command patch https://reviews.llvm.org/D27140 :

    CDB << "{ \"directory\": \"" << escape(Buf) << "\"";
    CDB << ", \"file\": \"" << escape(Input.getFilename()) << "\"";


afbjorklund:
SIlently ignoring sounds bad either way, so it should at least recognize this `-MJ` flag.

afbjorklund:
Your code needs to check the `clang` compiler, whether it has `-MJ` support.

You can see examples of this in the other similar features, under "PROBE"

afbjorklund:
And it seems that there is a new file born every minute now, so need to hurry up with #218.

totph:
The JSON spec seems to allow any unicode character except `\` and `"`, so maybe naïve rewriting (but still expecting `directory` as the first key) without `\x0123` escaping is enough. Or just extracting the string escape logic from a json library.

afbjorklund:
I made something similar for older and other compilers, by just storing the command and the file names:

https://github.com/ccache/ccache/compare/master...afbjorklund:command

Mostly used for debugging, originally written for ccache version 3.1 (though without the in/out filenames)

It didn't make any attempts to be compatible with the `compile_commands.json` syntax, though.

jrosdahl:
> SIlently ignoring sounds bad either way, so it should at least recognize this -MJ flag.

Yes, fixed in db3f3a9db39f4980b228735c3db5b561df3b3a01.

> And it seems that there is a new file born every minute now, so need to hurry up with #218.

Agreed. No more result file types until we have aggregated result files in place, please... :)

jrosdahl:
@totph: Would you be interested in working more on this PR or should it be closed for now? I have opened a [separate issue](#446) to keep the feature request in the loop even if this PR is closed.

totph:
@jrosdahl Sure. The two questions are: How to handle json and what to do with 8bit/unicode directories.

I would skip json entirely and treat it as a text substitution problem as only the value of "directory" has to be replaced. This would also make it easier to guarantee identical output in case some simple non-json tools expect the format to be stable.

The unicode escaping in clang is handled by this [yaml::escape function](http://llvm.org/doxygen/YAMLParser_8cpp_source.html#l00690) which creates several different (\x \u \U \N etc.) escapes and e.g. turns `/path/f ü Δ "'/` into `"/path/f \xFC \u0394 \"'/"`. Either this is ported to C or only 7 bit directories are cached, the rest always fall back to calling the compiler.

(Or... hear me out... a limited subset of C++ is allowed (maybe even pre-11), just like in gcc. The first benefit would be that the escape function could just be copied, licence wise copying from Apache 2 to  GPL 3 should be Ok. All the string operations would benefit from that as well. Maybe bump the ccache version to 4 and backport everything possible to version 3 for a while.)

jrosdahl:
> json [...] text substitution

It sounds totally fine to just do text substitution in this case; it would be nice to avoid a dependency on a JSON parser.

> only the value of "directory" has to be replaced

Doesn't the value of `"output"` need to be replace as well? And maybe `"file"` if "sloppy file macro" has been enabled (if not enabled the source name will be part of the hash)?

> unicode escaping

I think that only supporting 7-bit directories would be fine, at least for now. That said:

> C++

Your timing is good. I'm actually working on making the code build as C++11. (It will mostly look as the current C code in the short term but will be gradually converted to more C++-like C++ in the long term.)

> Maybe bump the ccache version to 4

Yes, that's the plan.

> and backport everything possible to version 3 for a while.

That's however not the plan. :slightly_smiling_face:

totph:
That sounds good! I'll rebase this PR once you feel your work can be published.

jrosdahl:
> I'll rebase this PR once you feel your work can be published.

Done now.

totph:
To make this work with the new `.result` file (where this output will always be stored) `write...` and `read_embedded_file_entry` need to be adapted.

These functions currently pass through all the data, but now will need to change it for this filetype. Can you think of other file types where this might be necessary? If this is the only one I'm thinking of just hardcoding a test for the `.ccmd` suffix and calling into `compile_command.cpp` from `results.cpp`. Otherwise these function pointers will get function pointer parameters, or result_file will get one as a parameter so that `result_files_add()` can super dynamically change the behavior.

jrosdahl:
> These functions currently pass through all the data, but now will need to change it for this filetype. Can you think of other file types where this might be necessary?

As it is now, the stderr data is copied to a file before being read back again and then written to stderr, so that has improvement potential. Other than that there is currently no other file types that need special treatment.

> If this is the only one I'm thinking of just hardcoding a test for the .ccmd suffix and calling into compile_command.cpp from results.cpp.

That sounds OK as an intermediate solution.

> Otherwise these function pointers will get function pointer parameters, or result_file will get one as a parameter so that result_files_add() can super dynamically change the behavior.

I'm currently (slowly) in progress of rewriting internal APIs in preparation for #414 and other improvements. I haven't yet thought about how the result API will be affected, but it will likely see substantial changes. I think that it would be OK if you want to add a hardcoded special case for .ccmd; I can make up something more generic as part of other refactoring. Or you could wait until I'm done, but I don't have any time frame for that. :slightly_smiling_face:

src/CompileCommand.cpp
```diff
@@ -0,0 +1,242 @@
...
+
+#include "Error.hpp"
+#include "ccache.hpp"
+
+#include <third_party/llvm_yaml_escape.hpp>
+#include <third_party/nonstd/string_view.hpp>
+
+#include <map>
+
+// TODO "-fcolor-diagnostics" is added by ccache
```
totph: Need to check if `args_add(common_args, "-fcolor-diagnostics");` was called. If not, remove.

src/CompileCommand.cpp
```diff
@@ -0,0 +1,242 @@
...
+
+        char* cwd = get_cwd();
+        write_escaped(q, cwd);
+        free(cwd);
+        q.write("\"");
+      } else {
+        checked_advance_out(true);
+      }
+    } else if (g_ccmd_incompletely_hashed_args != nullptr && q.pos_startswith("arguments\"")) {
+      assert(g_ccmd_incompletely_hashed_args->size() > 0); // otherwise infinite loop
```
totph: or a throw, but should never happen

src/ccache.cpp
```diff
@@ -2882,6 +2897,50 @@ cc_process_args(struct args* args,
...
+          (*g_ccmd_incompletely_hashed_args)[std::string("arg_namexxx")] = current_argument;
+          return true;
+        }
+        return false;
+      };
+
+      for (int ii = 1; ii < args->argc; ++ii) {
+        if (add_if_arg_present(args->argv[ii], "-fdebug-prefix-map=")) {
+        } else if (add_if_arg_present(args->argv[ii], "-ffile-prefix-map=")) {
+        } else if (add_if_arg_present(args->argv[ii], "-fmacro-prefix-map=")) {
```
totph: so far these two are only supported by gcc, but they are WIP for clang

src/ccache.cpp
```diff
@@ -2882,6 +2897,50 @@ cc_process_args(struct args* args,
...
+          if (g_ccmd_incompletely_hashed_args == nullptr) {
+            g_ccmd_incompletely_hashed_args = new std::map<std::string, const char*>();
+          }
+          (*g_ccmd_incompletely_hashed_args)[std::string("arg_namexxx")] = current_argument;
+          return true;
+        }
+        return false;
+      };
+
+      for (int ii = 1; ii < args->argc; ++ii) {
```
totph: Is there a better way to iterate over all args again?

src/result.cpp
```diff
@@ -175,13 +181,38 @@ read_embedded_file_entry(CacheEntryReader& reader,
...
         size_t n = std::min(remain, sizeof(buf));
         reader.read(buf, n);
-        if (fwrite(buf, n, 1, subfile.get()) != 1) {
+
+        remain -= n;
+
+        // 64k ought to be enough for anybody's ccmd text file
+        if (remain > 0) {
+          throw Error("stored .ccmd file is too big");
+        }
```
totph: Relying on READ_BUFFER_SIZE remaining larger than any feasible json fragment. Maybe add a static assert. Otherwise would need to duplicate the while loop from below.

totph:
RFC

> Doesn't the value of "output" need to be replace as well? And maybe "file" if "sloppy file macro" has been enabled (if not enabled the source name will be part of the hash)?

These two are not handled yet. Output should be easy, it should be possible to extract the name from `result_file_map->find(FileType::object)` and then look for the argument after `"-o"`. For file, I would just bail if this macro is active.

And, somewhat related: what about relaxing the line limit from 80 to 99?

jrosdahl:
> apart from -fdebug-prefix-map= there are probably other flags which ccache skips, but which might show up in the .ccmd file. They need to make it to the adapt_compile_command_json() function as well.
> [...]
> Need to check if args_add(common_args, "-fcolor-diagnostics"); was called. If not, remove.

This is becoming complicated. Reading this makes me wonder if we instead could *generate* the whole file instead of storing it and then trying to tweak it. Would that be possible? You gave a related comment just below:

> The commandline can not just be copied because the line which is recorded is of the backend:
"bin/clang-8", "-xc++", ..., "--target=x86_64-unknown-linux-gnu"

On my system, Clang dereferences the compiler symlink (`/usr/bin/clang` → `/usr/lib/llvm-8/bin/clang`) and adds `-xc++`, `--driver-mode=g++` and `--target=x86_64-pc-linux-gnu` to `arguments` when constructing the compiler command file.

ccache could easily follow the compiler symlink and also deduce the argument to `-x`, but the other two are probably harder. But maybe they actually don't matter, so that ccache could leave them out? https://clang.llvm.org/docs/JSONCompilationDatabase.html says this about `command`/`argument`: 

_"The compile command executed. After JSON unescaping, this must be a valid command to rerun the exact compilation step for the translation unit in the environment the build system uses"_.

Using the original command line given to ccache (`orig_args` sans the `--ccache-*` arguments, see `failed()`) for the `argument` list would fulfill the contract as I see it. This would make us produce  functionally equivalent output but not exactly the same output as the compiler would. If we're scared about that, we could hide the behavior behind a `sloppy_...` opt-in configuration setting.

What do you think about this idea, is it feasible?

> And, somewhat related: what about relaxing the line limit from 80 to 99?

I happen to like the 80 character width for code readability reasons, so I'd like to keep it.

totph:
>  hide the behavior behind a `sloppiness = ...` opt-in configuration setting.

That's what I have been thinking as well. Most tools probably only require the correct `-D`, `-I`, and maybe some `-W` flags - but if someone actually produces binaries this way it would be quite nasty if `ccache` gets something wrong. So definitely opt-in.

(Still, all the completely ignored or only partially hashed arguments should be processed in some central place in the future).

> wonder if we could _generate_ the whole file

Given the above that would probably suffice. Other differences I see are that `-Dno_space` arguments become space separated `"-D", "no_space"`, and that various `-M..` arguments are skipped. Also, `-O` becomes `-O2`? Not without its pitfalls, but doable without copying more logic verbatim from clang.

> > And, somewhat related: what about relaxing the line limit from 80 to 99?
> 
> I happen to like the 80 character width for code readability reasons, so I'd like to keep it.

New offer, 92? :) - I like code readability too, but good descriptive variable names are a part of that. The resulting awkwardly wrapped lines or worse shortened names run contrary to that. Nobody is editing files on a VT100 or other 80x25 terminals any more!

jrosdahl:
Let's go with generating the file if opted in then. Sounds good?

> (Still, all the completely ignored or only partially hashed arguments should be processed in some central place in the future).

I didn't understand that part.

> New offer, 92? :)

I don't want to debate this as part of this pull request. Please send me an email if you want to discuss this further.

jrosdahl:
Sounds good to me.

totph:
Where would an argument be placed which should not be hashed, but still be passed to the compiler on a cache miss? Is there an argument like this already or would this be the first?

__Or__ should `-MJ` always be withheld from the compiler so that the generated files are consistent and not a mix of the "true" json and what ccache writes on a cache hit? I'm thinking the latter. This would make it possible to also generate these files with ccache+gcc (given the ccmd sloppyness is set). Then ccache could work like build ear/Bear.

totph:
__Proposal__:

  - By default `-MJ` is too hard.
  - If the corresponding sloppiness is set via ccache.conf or the env var, `-MJ` is always handled by ccache and hidden from all compilers. This however means that using gcc, adding `-MJ` plus the sloppyness and then disabling ccache would cause compilation to fail. But that's why it is opt-in, and documented of course.
  - Add the more explicit `--ccache-MJ` option. In addition to writing to the provided `out.ccmd` argument, it also takes `-` which does not write to stdout but next to the `out.o` object file. This makes adding it to the generic C/CXXFLAGS possible without adapting every compiler call. (If both `-MJ` and this command are present, both files are written).

Then ccache can create a `compile_commands.json` file (after joining all `out.ccmd` files). That would be especially helpful when building a project which otherwise can not create such a file (or other build information). With e.g. the CompilationDatabaseProjectManager in Qt Creator these projects can then be viewed and edited with an IDE without manually configuring the list of source files, include paths and defines.

jrosdahl:
> **Proposal**:
> 
> * By default `-MJ` is too hard.
> * If the corresponding sloppiness is set via ccache.conf or the env var, `-MJ` is always handled by ccache and hidden from all compilers. This however means that using gcc, adding `-MJ` plus the sloppyness and then disabling ccache would cause compilation to fail. But that's why it is opt-in, and documented of course.

Sounds like a good idea!

> * Add the more explicit `--ccache-MJ` option. In addition to writing to the provided `out.ccmd` argument, it also takes `-` which does not write to stdout but next to the `out.o` object file. This makes adding it to the generic C/CXXFLAGS possible without adapting every compiler call. (If both `-MJ` and this command are present, both files are written).

This also sounds good. Regarding the `-` argument, perhaps it would be cleaner to have another argument-less option for that behavior (`--ccache-auto-MJ`?)?

> Then ccache can create a `compile_commands.json` file (after joining all `out.ccmd` files).

Could you expand on this idea? How would ccache generate `compile_commands.json`?

src/CompileCommand.cpp
```diff
@@ -0,0 +1,335 @@
...
+      if (second == misc_arg[1] && arg.starts_with(misc_arg)) {
+        return keep_and_maybe_split(arg, misc_arg);
+      }
+    }
+  }
+
+  return ArgStatus::keep();
+}
+
+size_t
```
totph: What if the call is "ccache distcc gcc", then not distcc but json should be in the compile commands file.

src/Config.cpp
```diff
@@ -263,6 +263,8 @@ parse_sloppiness(const std::string& value)
       result |= SLOPPY_LOCALE;
     } else if (token == "modules") {
       result |= SLOPPY_MODULES;
+    } else if (token == "compile_command" || token == "compile_commands") {
```
totph: While a ccache call creates a single compile_command file, these are then joined into a "compile_commands.json" file, so I'd leave this maybe undocumented but related name here and maybe save someone a bit of confusion.

src/ccache.cpp
```diff
@@ -2526,8 +2538,8 @@ cc_process_args(struct args* args,
     }
 
     // These are always too hard.
-    if (compopt_too_hard(argv[i]) || str_startswith(argv[i], "-fdump-")
-        || str_startswith(argv[i], "-MJ")) {
+    if (compopt_too_hard(argv[i]) || str_startswith(argv[i], "-fdump-")) {
+      // -MJ + e.g. -MM (=TOO_HARD) - should still write out.ccmd?
```
totph: How "intrusive" (up to: what do to do in case of CCACHE_DISABLE), should this behaviour be? I think always writing the file instead of the compiler is a must to remain consistent. This would mean all the args have to be parsed before early bailout is possible.

src/ccache.cpp
```diff
@@ -2936,6 +2948,73 @@ cc_process_args(struct args* args,
...
+    if (str_eq(argv[i], "--ccache-MJ")) {
+      i++;
+      if (i == argc) {
+        cc_log("Missing argument to --ccache-MJ");
+        result = false;
+        goto out;
+      }
+      output_ccmd = x_strdup(argv[i]);
+      output_ccmd_auto = false;
+      exitfn_add_nullary(write_compile_command_json_if_requested);
```
totph: Instead of adding a lot of write_ccmd() calls in various paths I decided to use the exit function.

src/ccache.cpp
```diff
@@ -3804,6 +3884,7 @@ ccache(int argc, char* argv[])
   }
 
   if (g_config.disable()) {
+    // and -MJ + sloppy? not --cache-MJ
```
totph: See above.

jrosdahl:
@totph: Do you think that you will continue this PR?

totph:
Sure, let me catch up. I see the Args class is done!

src/Args.hpp
```diff
@@ -63,6 +63,10 @@ class Args
   // Remove all arguments with prefix `prefix`.
   void erase_with_prefix(nonstd::string_view prefix);
 
+  // Remove all arguments matching `name` and (if not the last) the one
+  // following it. E.g. given '-A' and [-A, b, c, -A, d, e, -A] -> [c, e].
+  void erase_matching_and_next(nonstd::string_view name);
```
totph, line 68: Require to hide `-MJ foo.ccmd` and `--ccache-MJ` from the compiler.

jrosdahl:
@totph: What do you think, should we close this PR or will you work on it?

totph:
I have not forgotten about this, but I need to look at the compile commands logic in the latest clang again. It was updated a few versions back and I want to make sure that the comments are referring to the correct source files and that the output is identical.

adenisonafifi:
Hello, poking my head into this thread to say I was trying to integrate `gen-cdb-fragment-path` into my build system (https://reviews.llvm.org/D66555) for compilation database generation and noticed that it made builds uncacheable due to the compiler flag being marked TOO_HARD.

Curious if this is still underway or if there's an opportunity to pitch in.

totph:
Thanks for reminding me, I plan to finish this in the next 2 or 3 months.

totph (commented):
This is mostly done, from my side the UI and docs are final, but open to suggestions.

In `compilation_database.cpp` I still have to adapt the argument processing to the current version of clang (`CC1Options.td` is gone) and the tests need to be adapted.

Adding support for `-gen-cdb-fragment-path` will be done in a followup patch.

doc/MANUAL.adoc
```diff
@@ -726,6 +726,14 @@ safe to use, but not all file systems support the feature. For such file
 systems, ccache will fall back to use plain copying (or hard links if
 <<config_hard_link,*hard_link*>> is enabled).
 
+[#config_generate_compilation_database]
+*generate_compilation_database* (*CCACHE_GENERATE_COMPILATION_DATABASE* or *CCACHE_NOGENERATE_COMPILATION_DATABASE*, see _<<Boolean values>>_ above)::
+
+    If true, `ccache` writes a compilation database fragment for any compiler.
+    The output will be written next to the object file with a `.cdb.json`
```
jrosdahl, line 733: Would it make sense not to use a `.json` extension since the file is not a valid JSON file?

src/compilation_database.cpp
```diff
@@ -0,0 +1,292 @@
...
+//
+// See include/clang/Driver/CC1Options.td and include/clang/Driver/Options.td
+// for the group definitions, and see the "JoinedOrSeparate" for the splitting:
+//
+// relevant are: CC1Option Preprocessor_Group clang_i_Group I_Group M_Group
+//
+// cat include/clang/Driver/Options.td |
+//   grep -E 'JoinedOrSeparate.*(CC1Option|(Preprocessor|M|clang_i|I)_Group)'
+//
+// This function duplicates this behavior.
```
jrosdahl, line 99: Do we really need to duplicate Clang's behavior? If ccache is called to wrap a certain compilation command , wouldn't it be OK to just put that compilation command as-is in the compilation database?

I would like to avoid having compiler-specific knowledge in ccache unless it's strictly necessary. More complexity means more bugs and more code to update and maintain. Specifically, it would be good not to have to make updates in this area to match future Clang versions.

unittest/test_llvmEscape.cpp
```diff
@@ -0,0 +1,48 @@
...
+  std::string in = R"(fo"o\bar)";
+  in.push_back('\0');
+  in.push_back('Z');
+  in.push_back('\n');
+  std::string out = llvm::yaml::escape(in);
+
+  CHECK(out == R"(fo\"o\\bar\0Z\n)");
+}
+
+TEST_CASE("2 byte utf8")
```
jrosdahl, line 34: Why is there a need to handle UTF-8 specially at all? JSON parsers are required to support UTF-8 natively.

test/CMakeLists.txt
```diff
@@ -27,6 +27,7 @@ addtest(cleanup)
 addtest(color_diagnostics)
 addtest(config)
 addtest(cpp1)
+addtest(compile_command)
```
jrosdahl, line 30: The compile_commands suite fails in all CI jobs that execute it. Please have a look.

PR #278, opened 2018-06-21
Improvement and upgrade of Docker build

afbjorklund:
Make the docker build target better and faster, by reusing and mounting instead of rebuilding and copying.

Also upgrade the ubuntu version from `xenial` to `bionic`, and make it possible to use other base images.


afbjorklund:
I created an organization on https://hub.docker.com, and will push the prebuilt images there.

jrosdahl:
Nice!

Would it be a good idea to lock versions instead of using latest? E.g., rename `base/centos` to `base/centos-7`, etc? Using "latest" makes me a bit uneasy...

Also, maybe `base` should be `docker`,`docker-base` or `docker-images`?

afbjorklund:
Actually I wanted it to just use `latest`, if you look at the actual images they're just tagged e.g. `centos`.

https://hub.docker.com/r/ccache/build/tags/

```
REPOSITORY                                  TAG                 IMAGE ID            CREATED             SIZE
ccache/build                                fedora              f3daf7674645        6 weeks ago         513MB
ccache/build                                debian              3973ab08258a        6 weeks ago         341MB
ccache/build                                centos              ab89d6316d19        6 weeks ago         355MB
ccache/build                                alpine              199307394dc2        6 weeks ago         197MB
ccache/build                                latest              ca5e84251a92        6 weeks ago         320MB
ccache/build                                ubuntu              ca5e84251a92        6 weeks ago         320MB
```

But I could remove the `:latest` from the Dockerfile (it would still be used/added on by docker, though)

----

I have heard that Docker Inc. is persuing their "docker" trademarks agressively, and don't want to use "images" since it sounds it would contain binaries rather than just descriptions. Will consider alternatives.

Basically we don't need to keep _all_ of these, but I wanted to show the syntax of both `apt` and `yum` ? Historically we just used one that resembled the development environment, which was `ubuntu:xenial`.

So doing something like `make docker`, would make things more reproducible on my end (i.e. laptop).

But the `make travis` is the only one that is versioned, it uses `ubuntu:trusty` in order to match Travis.

jrosdahl:
>  [...] and don't want to use "images" since it sounds it would contain binaries rather than just descriptions.

Note that you wrote "This directory contains base **images**" in base/README. :wink: I don't see a problem with naming the directory docker-images even if it only contains the dockerfiles (we do that at work, so maybe I'm biased), but just "base" feels wrong. How about "build-images"?

jrosdahl:
@afbjorklund: This PR is nice and I would like to merge it. Have you thought anything more about naming?

afbjorklund:
@jrosdahl : nope, but will do!

jrosdahl:
New suggestion for the directory name: "buildenv". @afbjorklund: Are you OK with that? If so I can make the changes before merging.

afbjorklund:
I'm OK with "buildenv", recently used "containers" (but then again it's mostly about "images" anyway)

Issue #226, opened 2018-02-13
Improve debuggability of cache misses


jrosdahl:
### Background

Sometimes it's hard to understand why you don't get cache hits when you think you should. The only documented way to debug is to set `log_file` and inspect the log. However, the log file doesn't really tell much about why a cache miss occurred since that depends on what has been hashed. A related minor nuisance is that logs from parallel ccache invocations will be intermingled in the log.

There is an undocumented way of debugging what has been hashed: build ccache with `-DCCACHE_DEBUG_HASH` set and set `CCACHE_DEBUG_HASH` at runtime. All data that is hashed will then be appended to a file called `ccache-debug-hash.bin` in the current working directory. It works to inspect that file if you execute a single compilation with ccache, move the file away, execute a compilation once more and compare the resulting file with the moved file. But for a real scenario, when you build a full source tree, that method is not practical.

### What to do

Here is an idea on how to make it easier to debug cache misses and also ccache in general: **Store logs and hashed data next to the output object file.** This will make it easy to compare two different builds: just compare two logs and/or hashed data files for the same object file in the two different build trees.

Perhaps something like this:

1. Add a suitable configuration option, maybe just a boolean named `debug` (`CCACHE_DEBUG`).
2. If set, save all logs (from `cc_log`) in a buffer until `cc_process_args` (which sets  `output_obj`) has returned.
3. Write all buffered logs to `output_obj.ccache-log` or similar and keep appending new logs to that file.
4. Write all hashed data to `output_obj.ccache-input` or similar.
5. To make it easier to interpret the `.ccache-input` file, consider adding some headings that delimit data hashed by `calculate_common_hash`, `calculate_object_hash` in preprocessor and direct mode and `hash_source_code_file` (called from `manifest_get`), and maybe more.
6. Remove the `CCACHE_DEBUG_HASH` stuff from `mdfour.c`.
7. Make sure to document the new feature properly.

tdihp:
I'm wondering whether saving each state of the hash update instead of the entire hashed data, along with why it's changed (CWD, compiler binary, extra file, etc), would be a cheaper solution while able to identify the first difference of a cache miss.

jrosdahl:
> I'm wondering whether saving each state of the hash update instead of the entire hashed data, along with why it's changed (CWD, compiler binary, extra file, etc), would be a cheaper solution while able to identify the first difference of a cache miss.

[...]

PR #279, opened 2018-06-24
Add "debug" option, for debugging hash misses

afbjorklund:
Hi @jrosdahl:

Here is the requested feature, for being able to debug the hash functions of the cache.

This version adds both a "debug" text mode, as well as the original "binary" hash input.

Thus you can verify the checksum to see it matches, but also be able to read and diff it.

Note that there are actually _three_ main hashes, as well as some other hash as well.

Implements #226


jrosdahl:
Awesome! I'm reviewing this now.

src/ccache.c
```diff
@@ -894,7 +913,7 @@ process_preprocessed_file(struct mdfour *hash, const char *path, bool pump)
 				free(cwd);
 			}
 			if (should_hash_inc_path) {
-				hash_string(hash, inc_path);
+				hash_buffer(hash, inc_path, strlen(inc_path));
```
jrosdahl: Why this change?

src/ccache.c
```diff
@@ -2051,6 +2070,10 @@ from_cache(enum fromcache_call_mode mode, bool put_object_in_manifest)
 		break;
 	}
 
+	if (conf->debug) {
+		debug_end(true);
```
jrosdahl: Calling `debug_end` explicitly means that we won't get a log e.g. when `failed()` is called. How about conditionally registering `debug_end` as an exit function (see `exitfn_add_nullary`) instead?

src/ccache.h
```diff
@@ -136,6 +137,7 @@ bool hash_file(struct mdfour *md, const char *fname);
 void cc_log(const char *format, ...) ATTR_FORMAT(printf, 1, 2);
 void cc_bulklog(const char *format, ...) ATTR_FORMAT(printf, 1, 2);
 void cc_log_argv(const char *prefix, char **argv);
+bool cc_copylog(const char *path);
```
jrosdahl: Perhaps rename to something like `cc_dump_log_buffer`?

src/ccache.h
```diff
@@ -243,6 +245,7 @@ void wipe_all(struct conf *conf);
 int execute(char **argv, int fd_out, int fd_err, pid_t *pid);
 char *find_executable(const char *name, const char *exclude_name);
 void print_command(FILE *fp, char **argv);
+char *string_command(char **argv);
```
jrosdahl: I suggest renaming to `format_command` analogous to the other similar functions.

src/execute.c
```diff
@@ -348,3 +348,21 @@ print_command(FILE *fp, char **argv)
...
+{
+	size_t len = 0;
+	for (int i = 0; argv[i]; i++) {
+		len += (i == 0) ? 0 : 1;
+		len += strlen(argv[i]);
+	}
+	len += 1;
+	char *buf = x_calloc(1, len + 1);
+	for (int i = 0; argv[i]; i++) {
+		strcat(buf, (i == 0) ? "" : " ");
```
jrosdahl: For very long command lines, this O(n²) algorithm might hurt performance... It would be nice to get away with the `strcat`s and keep a pointer to remember where to copy the next argument instead.

src/hash.c
```diff
@@ -18,6 +18,60 @@
...
 #define HASH_DELIMITER "\000cCaChE"
+#define HASH_DEBUG_DELIMITER "### "
+
+// binary input, for hashing
+char *debug_hash_bin;
+
+// text input, for debugging
+char *debug_hash_txt;
+
+void hash_debug(const char *bin, const char *txt)
```
jrosdahl: Rename to `hash_init_debug`?

src/hash.c
```diff
@@ -18,6 +18,60 @@
...
+		debug_hash_bin[strlen(debug_hash_bin)-1] = md->identifier;
+
+		FILE *f = fopen(debug_hash_bin, "a");
+		fwrite(s, 1, len, f);
+		fclose(f);
+	}
+}
+
+static void
+hash_debug_buffer(struct mdfour *md, const void *s, size_t len)
```
jrosdahl: Let's see, `hash_binary_buffer` hashes a (binary?) buffer, but `hash_debug_buffer` doesn't hash a buffer... Rename to something like `hash_note_input` instead? And I think I would prefer renaming `hash_binary_buffer` back to `hash_buffer`.

src/hash.c
```diff
@@ -18,6 +18,60 @@
...
+{
+	mdfour_update(md, (unsigned char *)s, len);
+	if (!md->identifier || len == 0) {
+		return;
+	}
+	if (debug_hash_bin) {
+		// log to different files, for the different hash types
+		debug_hash_bin[strlen(debug_hash_bin)-1] = md->identifier;
+
+		FILE *f = fopen(debug_hash_bin, "a");
```
jrosdahl: It would be nice to keep the hash debug files open to avoid three system calls for each hash invocation... :thinking:

src/util.c
```diff
@@ -58,12 +64,22 @@ init_log(void)
...
 
+static void
+append_log(const char *s, size_t len)
+{
+	assert(logbuffer);
+	if (logsize + len > logsize) {
+		logbuffer = x_realloc(logbuffer, logsize + len + 1);
+		logsize = logsize + len;
+	}
+	strncat(logbuffer, s, len);
```
jrosdahl: O(n²) here as well – it would be nice to update and use a buffer size variable instead of implicitly `strlen`-ing the potentially large string for each call.

src/ccache.c
```diff
@@ -3360,12 +3383,18 @@ ccache(int argc, char *argv[])
...
+	}
+
 	struct mdfour common_hash;
 	hash_start(&common_hash);
+	mdfour_identify(&common_hash, 'c');
 	calculate_common_hash(preprocessor_args, &common_hash);
 
 	// Try to find the hash using the manifest.
 	struct mdfour direct_hash = common_hash;
+	mdfour_identify(&direct_hash, 'd');
```
jrosdahl: As suggested in bullet 5 in #226, would it be possible to add a debug log file header here saying "here comes the direct mode part", and similar for the common and preprocessor parts?

jrosdahl:
By the way: If you agree to my suggestions but don't have time to address them near term (or don't want to :slightly_smiling_face:) I can work on them if you like.

src/ccache.c
jrosdahl: > for the binary version I wanted to keep the exact same format so that the checksums would match what is being used as the key in the cache.

Certainly, that's good.

> I did that (well, implicitly) for the text version,

I guess it's possible to deduce the three sections if you happen to know how ccache works in detail, but I think that it would be very nice with explicit section markers in the `*.ccache-input` file, perhaps like this:

```
=== COMMON ===
3
### ext
i
[...]

=== DIRECT MODE ===
### manifest version
1
### arg
-g
[...]

=== PREPROCESSOR MODE ===
### arg
-g
[...]
```

src/hash.c
afbjorklund: Actually `hash_buffer` _is_ the same function as before, the other two are just internal (static) helpers. One doing the hashing and logging to the binary files, and one logging to the text file.

src/util.c
afbjorklund: Hmm, I don't think `strcat` calls `strlen` does it ? Was careful to use the `len` parameter, actually.

src/ccache.c
```diff
@@ -3360,12 +3383,18 @@ ccache(int argc, char *argv[])
 
 	cc_log("Object file: %s", output_obj);
```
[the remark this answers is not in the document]
afbjorklund: Done

doc/MANUAL.adoc
```diff
@@ -749,6 +756,26 @@ A source language e.g. specified with *-x* was unsupported by ccache.
...
+
+When investigating why you are not getting hits, and what information
+that ccache is actually hashing, you can enable the "debugging" mode.
+
+In this mode, ccache will add some additional output files next to the
+object files with results from the hashing and a copy of the debug log.
+
+There are several binary hash files created, for different partial input:
+
+* Common (*'c'*)
```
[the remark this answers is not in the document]
afbjorklund: Done

doc/MANUAL.adoc
```diff
@@ -749,6 +756,26 @@ A source language e.g. specified with *-x* was unsupported by ccache.
...
+* Direct mode (*'d'*)
+* Preprocessor mode (*'p'*)
+
+The hash is the checksum of the common + direct or common + preprocessor.
+
+There is also a human readable text file, for easier reading and diffing.
```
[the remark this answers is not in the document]
afbjorklund: Done

src/ccache.c
```diff
@@ -476,6 +476,25 @@ clean_up_internal_tempdir(void)
...
+{
+	char *hash_bin = format("%s%s", path, ".ccache-hashX");
+	char *hash_txt = format("%s%s", path, ".ccache-input");
+	hash_debug(hash_bin, hash_txt);
+	free(hash_bin);
+	free(hash_txt);
+}
+
+static void
+debug_end(bool hit)
```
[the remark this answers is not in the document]
afbjorklund: Done

src/ccache.c
```diff
@@ -476,6 +476,25 @@ clean_up_internal_tempdir(void)
...
+{
+	(void) hit;
+	char *path = format("%s%s", output_obj, ".ccache-log");
```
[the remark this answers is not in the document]
afbjorklund: Think I prefer the suffix to be separate (and same as the others), but thanks!

src/ccache.c
afbjorklund: There is no change for the binary buffer, but for the text buffer it inserts a newline between strings. Which means that there will be a lot of newlines inserted into the input, making it harder to read...

src/ccache.c
afbjorklund: Done

src/ccache.h
afbjorklund: Done

src/ccache.h
afbjorklund: Done.

src/execute.c
afbjorklund: Hmm, strcat should be O(n) and not O(n²) ? It is supposed to keep going until it finds NUL.

afbjorklund: Of course, that would be `strcpy` and not `strcat` - you are right, will change them as suggested

src/util.c
afbjorklund: It doesn't call it on the *source* string, but it does call it on the *destination* string. My bad.

jrosdahl:
Thanks, looks good!

What about my question about registering `debug_end` as an exit function?

jonnyyu:
Thanks for the great work of 'ccache' and keep improving its 'debug' experience. 
This is very important, as fighting with cache miss is a challenging but very necessary work.

I am trying to nail down a cache miss issue in my codebase using this changes. 
I see the .ccache-input helps me to understand how manifest has is generated.
But I guess in my case the cache miss should from some header file changes. Is there any log to show what header file change caused the cache miss?

Thanks!

afbjorklund:
@jonnyyu :  Is this about direct mode (manifest) or preprocessed mode (cpp) ?
You will need to use the `--dump-manifest` tool, or to decipher those `# ` lines.

jrosdahl:
Thanks, merged to master.

I changed the filenames in 7c5d9aec6de2d5c78230682b91d9804fdbf05ffe, hope that's OK with you. I also rewrote the documentation in e3dbefeb2e675f014f29fdc94da34e56454c56c7.

jrosdahl:
Also, I intend to do some refactoring to remove knowledge about the debug mode from the MD4 struct and code.

afbjorklund:
Sounds good, both the changes and the refactoring to remove it from the mdfour code.

PR #282, opened 2018-07-03
Add handling of separate -target argument

afbjorklund:
Otherwise you will get an error in the log:

```
x86_64-pc-linux-gnu is not a regular file, not considering as input file
```

The argument seems to be specific to clang.


jrosdahl:
Applied on 3.4-maint, thanks!

Issue #287, opened 2018-07-24
CUDA build times out in Travis


afbjorklund:
https://travis-ci.org/ccache/ccache/jobs/407497038

```
No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received
```

afbjorklund:
Travis output:
```
--2018-07-24 13:46:45--  http://developer.download.nvidia.com/compute/cuda/repos/ubuntu1404/x86_64/7fa2af80.pub
Resolving developer.download.nvidia.com (developer.download.nvidia.com)... 192.229.211.70, 2606:2800:21f:3aa:dcf:37b:1ed6:1fb
Connecting to developer.download.nvidia.com (developer.download.nvidia.com)|192.229.211.70|:80... connected.
HTTP request sent, awaiting response... Read error (Connection reset by peer) in headers.
```

afbjorklund:
Looks like http is being blocked, works ok with https:

```
--2018-07-24 14:06:54--  https://developer.download.nvidia.com/compute/cuda/repos/ubuntu1404/x86_64/7fa2af80.pub
Resolving developer.download.nvidia.com (developer.download.nvidia.com)... 192.229.211.70, 2606:2800:21f:3aa:dcf:37b:1ed6:1fb
Connecting to developer.download.nvidia.com (developer.download.nvidia.com)|192.229.211.70|:443... connected.
HTTP request sent, awaiting response... 200 OK
Length: 1633 (1.6K) [application/octet-stream]
Saving to: â€˜7fa2af80.pubâ€™


 0% [                                       ] 0           --.-K/s              
100%[======================================>] 1,633       --.-K/s   in 0s      

2018-07-24 14:06:54 (105 MB/s) - â€˜7fa2af80.pubâ€™ saved [1633/1633]

OK
```

afbjorklund:
Got some other flaky build, so the whole cuda package infrastructure is a bit flaky it seems...

https://travis-ci.org/ccache/ccache/jobs/414963561

`W: Failed to fetch http://developer.download.nvidia.com/compute/cuda/repos/ubuntu1404/x86_64/InRelease  Could not connect to developer.download.nvidia.com:80 (192.229.211.70), connection timed out`

jrosdahl:
[...]

PR #289, opened 2018-07-25
process_preprocessed_file: Move gnu_getcwd() out of tight loop

mgulick:
Change process_preprocessed_file from calling getcwd() once per line
in the preprocessed source file to once at the start of the function.
The performance of getcwd() on Mac seems to be terrible compared to
Linux.  On a macOS 10.13 build machine, this change improves
process_preprocessed_file runtime on a 10MB preprocessed file from 75
seconds to .75 seconds.

### DIFF
--- a/src/ccache.c
+++ b/src/ccache.c
@@ -781,6 +781,11 @@ process_preprocessed_file(struct mdfour *hash, const char *path, bool pump)
 		included_files = create_hashtable(1000, hash_from_string, strings_equal);
 	}
 
+	char *cwd = NULL;
+	if (!conf->hash_dir) {
+		cwd = gnu_getcwd();
+	}
+
 	// Bytes between p and q are pending to be hashed.
 	char *p = data;
 	char *q = data;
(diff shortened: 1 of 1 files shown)


afbjorklund:
Looks good to me :+1:

afbjorklund:
This Travis build is even weirder, it "forgot" to install `gperf` for some reason ?
https://travis-ci.org/ccache/ccache/jobs/408099229

The other failure is from the known CUDA compiler installation issue.

afbjorklund:
See #290 and #287 for the Travis errors

mgulick:
Please let me know if there's anything you'd like me to do for this pull request or the travis build.  As you noted though the errors look unrelated to this change.  Thanks!

afbjorklund:
I guess we just have to wait until @jrosdahl is back from vacation

jrosdahl:
> I guess we just have to wait until @jrosdahl is back from vacation

It's more like "back from parenthood". :slightly_smiling_face:

jrosdahl:
Applied, thanks!

mgulick:
I see now that there is indeed a new warning from the clang static analyzer with this change.  Apparently the analyzer doesn't realize that the 'if (!conf->hash_dir)' is the same condition in both locations.  A couple of options come to mind that would probably fix it:
at line 887:
`if (!conf->hash_dir && cwd) { ...` 
or (since conf->hash_dir must be false if cwd is not null)
`if (cwd) { ...`
or
```
if (!conf->hash_dir && cwd) {
   assert(cwd != NULL)
```
The first option seems the most straightforward.  The second seems a bit too clever to me, and possibly even something the compiler would optimize the first option to on its own.

jrosdahl:
Since `process_preprocessed_file` only is called once, I'll just make the `gnu_getcwd()` call unconditional.

PR #291, opened 2018-07-28
Make sure to install gperf for analyze

### LINKED ISSUE #290: Clang analyzer not working in Travis
Seems like the refactored file no longer installs "gperf" ?

The addons: apt: packages: was overridden, without it...

`/bin/sh: 1: gperf: not found`

afbjorklund:
Also move the 32-bit libraries to the 32-bit target (only).

Closes #290

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,6 @@ addons:
             - gperf
             - elfutils
             - zlib1g-dev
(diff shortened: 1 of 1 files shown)


jrosdahl:
Applied, thanks!

PR #293, opened 2018-08-10
Improve printing of “stats zeroed” statistics

jrosdahl:
* Refactored how formatting functions work in “ccache -s” output. They now return a formatted string (or NULL for “don’t print statistics line at all”) instead of printing the value.
* Added a new format_timestamp function.
* Added STATS_ZEROTIMESTAMP to the common stats_info array instead of being a special case.
* Let the STATS_ZEROTIMESTAMP be 0 for new caches so that the “stats zeroed statistics line only shows up when “ccache -z” actually has been called.

@afbjorklund: What do you think?

### DIFF
--- a/doc/MANUAL.adoc
+++ b/doc/MANUAL.adoc
@@ -736,6 +736,12 @@ The compiler was instructed to write its output to standard output using *-o
 | preprocessor error |
 Preprocessing the source code using the compiler's *-E* option failed.
 
+| stats updated |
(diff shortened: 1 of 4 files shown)


afbjorklund:
Sounds good to me. I don't think the feature is terribly important, but this change should make it more obvious what it is printing... I guess the important feature is to get per-build statistics, but that's harder.

Issue #213, opened 2018-01-14
CoW: copy-on-write (btrfs/xfs clone) support


PaulBone:
Hi,

I'm working on some patches for adding direct support to clone a file (like a hardlink) on btrfs.  Btrfs supports copy-on-write, you can copy a file without duplicating its data blocks.

I want to ask, is the best way to expose this as an option to the hard_link configuration option?  For example hard_link = true, hard_link = false and hard_link = reflink (or btrfs_clone or some other value).  Or should this be done if supported and fallback to a regular copy (since they have the same semantics anyway).

Thanks.

PaulBone:
Oh, I don't seem to be able to assign myself this bug. Nevertheless I have started work on it.

afbjorklund:
Hardlink support isn't very popular, because it is hard to use (avoid strip) and messes with compression.

Maybe btrfs clone with copy-on-write is better (more userfriendly), but then again nobody is using btrfs...

afbjorklund:
> Oh, I don't seem to be able to assign myself this bug. 
Only project maintainers can do that (same with labels)

tdihp:
My two cents here: since BTRFS and the like already have compression builtin, a custom command for copying the object file, which would enforce whatever copying behavior of the underlying filesystem as the user want, would be great.

BTW @afbjorklund  while I'm not convinced this hard link trick would benefit build time a lot, but I do think using BTRFS might be useful for snapshot to manage and transfer cache state, when having a lot of machines, or to reproduce issues.

jrosdahl:
I plan to introduce "fat result files" in some future ccache version, where all produced data (object file, dependency file and standard error output, etc.) is stored in a single file. This has performance benefits. If `hard_link = true`, my plan is to store the object file next to the "fat result file", but that won't be the default. This means that a BTRFS_IOC_CLONE ioctl can't be made by default. Therefore I think that we need a new configuration option for activating BTRFS_IOC_CLONE (and in the future to trigger storing the object file separately). We can't reuse the `hard_link` option since it in its environment variable form has [boolean semantics only](https://ccache.samba.org/manual/latest.html#_boolean_values).

afbjorklund:
We did some of this, for the memcached branch. It got tedious, to pile on new file formats all the time...

https://github.com/ccache/ccache/pull/58#issuecomment-255547818

Not sure how it would work in the context of #98, though ? Harder to share object files, done this way.

[...]

PR #296, opened 2018-08-12
Add reflink support, for file cloning

afbjorklund:
Somewhat similar to the existing hardlink, but without the downsides of sharing files...

But supported only on BTRFS and newer kernel versions of XFS -  and Linux only, for now.

See: https://lwn.net/Articles/581558/ and https://lwn.net/Articles/747633/ (and `cp --reflink`)

Your mileage may vary, mostly helps with *large* files. Can actually be slower, for small files!

Implements #213


PaulBone:
I've tested this with reflink=true and reflink=false, I also tested with my ccache on the same or a different BTRFS volume.  I think that on seperate volumes reflink will fail.  I tested this by building spidermonkey with a hot cache, then touching some .h files and timing the builds.

 seperate volume, reflink=true
#
# real    0m30.386s
# real    0m23.367s
# real    0m23.408s
#
# seperate volume, reflink=false
#
# real    0m19.146s
# real    0m19.016s
# real    0m17.887s# same volume reflink=true
#
# real    0m19.982s
# real    0m22.450s
# real    0m19.758s
#
# same volume, reflink=false
#
# real    0m23.004s
# real    0m20.520s
# real    0m18.856s

So yeah, it seems it's slightly slower, which is surprising, but maybe it's just more synchronous than a regular copy.  But it also uses way less disk space so I like it for that alone.

afbjorklund:
Thanks for testing, seems like there truly is no silver bullet (same story as with `splice`*) ?

\* where we [tried to](https://github.com/ccache/ccache/compare/master...afbjorklund:uncompressed-fast) copy files in kernel space instead of user space, and it ended up... slower.

jonnyyu:
@PaulBone did you test with SSD or HDD?

PaulBone:
> @PaulBone did you test with SSD or HDD?

SSD.

PaulBone:
> Thanks for testing, seems like there truly is no silver bullet (same story as with `splice`*) ?

I'd like to stress that even though it's a bit slower it's still very valuable.  For me it'll save disk space and won't have the problems of accidentally modifying a hardlink and therefore updating the file in the cache.

afbjorklund:
I think that it is pretty safe to keep this as an optional feature, for people who want to use it.

ddiss:
Anything holding this one up from being merged? I look forward to being able to use a more space efficient cache.

jrosdahl:
> Anything holding this one up from being merged? I look forward to being able to use a more space efficient cache.

Except for the fact that the feature branch is not in sync with master, I think mainly two things:

1. Real life constraints.
2. I intend to (any year now) introduce [aggregated result files](#218) and after that introduce support for different storage backends. As part of this, I'm toying with the idea of scrapping support for hard links altogether because i) they are error-prone and ii) will introduce complexity in the storage layer if left. If we add reflink support now, it would be silly to remove reflink support again shortly thereafter, so reflink support must in that case be part of the aggregated result files rework. And I don't have a clear picture of how that would fit in the design yet. So in other words, I'm still undecided if I want the feature in its current form.

ddiss:
Thanks for the explanation...

> I intend to (any year now) introduce aggregated result files and after that introduce support for different storage backends. As part of this, I'm toying with the idea of scrapping support for hard links altogether because i) they are error-prone and ii) will introduce complexity in the storage layer if left.

I'm not using the hardlink functionality, so wouldn't miss it if dropped.

> If we add reflink support now, it would be silly to remove reflink support again shortly thereafter, so reflink support must in that case be part of the aggregated result files rework. And I don't have a clear picture of how that would fit in the design yet. So in other words, I'm still undecided if I want the feature in its current form.

Understood.

afbjorklund:
It was only the configuration system being overhauled again, nothing major.

afbjorklund:
We can leave hardlinks and reflinks, but **only** if we support both aggregated and individual cache files... ?

If we make the "zipped" cache files mandatory, then it doesn't work do link (since there are no cache files)

It doesn't work with compression, either.

And only works on those weird filesystems.

jrosdahl:
Closing this PR now as it won't be merged in its current form due to #411.

ivan-volnov:
Thank you @afbjorklund for the idea and the starting implementation.  I really like it and use this code in my fork.

afbjorklund:
@ivan-volnov : thanks, but I don't think this code will end up getting used.

I'm not sure it gave any major time-saving benifits anyway, doing reflinks ?
Only some space-saving benefits, and nobody cares about those anyway.

Compression is now default, and is the primary source of reduced space...

ivan-volnov:
@afbjorklund I have a project that generates about 7Gb of object files in Debug build.
It take a while to write them on disk from cache. Even on my fast SSD, hardlinks/reflinks are faster than a regular copy. Because you have to read ~600Mb compressed files in cache and write 7Gb uncompressed object files every time you change git branch. I believe it's just a SDD killing.
On the other hand clone files is just a magic! You compile a source once and then just do a simple clone wherever you want.
Yes, I don't care about space-savings and can afford me a huge cache directory. But on fact It's the only place where I store my object files.
I added support for APFS cloning and I also plan to add a filesystem compression on cache. With a filesystem compression my 7Gb will be turned into ~600Mb cache with zero space of my object files in the build directory.

afbjorklund:
Interesting use case! And of course I was interested enough to write it, even if ccache didn't want it.

jrosdahl:
@ivan-volnov: Thanks for sharing your use case! Stories like these are very valuable for understanding how ccache actually is used and how to prioritize between features.

Regarding reflinks, please note that the idea of supporting reflinks has not been discarded (#213 is still open) – it's only this PR that was closed since its context now is obsolete and a potential future solution will look very different.

agrygorenko:
@afbjorklund I believe @ivan-volnov described common case of the codebases with extensive use of template meta programming. 
For example In a code base I'm working with, 500.000 lines of code generates over 30GB cache. Having this feature would allow me substantially improve ccache perforce on CI.

afbjorklund:
The main problem for me was that it required a special file system... Ended up with up with a disk image, just to be able to actually test it.

Issue #98, opened 2016-06-11
Cache deduplication


jrosdahl:
From https://bugzilla.samba.org/show_bug.cgi?id=11711:

```
I've found that there are many identical files stored in my ccache.  After a little
investigation, there are many ways this can happen:

-adding/removing a blank line in a source file (including adding a comment, or
other preprocessor directive that is removed by the C preprocessor)
-changing "useless" code (e.g. an empty namespace, an unused class)
-changing compiler options that don't affect the output (e.g. adding -Wall on
clean code)
-I'm sure the list continues...

Of course, it would be unreasonable for ccache to accurately predict every possible
way that an identical object file could be created.  However, I think it is reasonable
for ccache to identify that it has created an identical object so it can avoid storing
it more than once.

Assuming this de-duplication could be performed, better hit rates could be obtained
with a ccache of the same size (alternatively, a smaller cache could maintain the
same hit rate).

(my brainstorming follows, feel free to disregard)

This probably requires installing additional files in the cache for the hash of each
object.  So, the flow for a miss might look like this:

run compiler
generate hash of generated files (*.o, *.d, *.stderr, others?)
lookup hash in cache
if hit
  install manifest-like file to redirect to the existing cache entry?
  copy/link that object to the proper destination
else //miss
  install object(s) in cache
  install manifest (or something)? at the location of the hash of the object
  copy/link the new object to the proper destination

If you're willing to break backwards-compatibility, it might make more sense to install
the compiler products (*.o,*.d, *.stderr) in the location that corresponds with that
file's own hash (instead of the preprocessor result's hash as is currently done).
Then, manifest-like files would have to be generated for both direct hits and
preprocess hits, and both would refer to the product's hashes.
```

itensionanders:
This is another something that needs to be properly benchmarked, since it will introduce double the amount of stats and files (for objects). We are doing something in similar in memcached, when it is splitting up large files into small chunks based on their content. This could be improved further, to split  the files in the right places for deduplication (like bup-split). But those would mean even *more* files.

Sooner or later all those loose files would need to be packed together, perhaps similar to how git does it. But it has a rather different access pattern compared to compiling, so most likely needs adapting...
Already have to adopt the binary/multi format of the o/stderr/dia/dep container, to allow for more items like dwo/gcno/whatnot. They are stored in one file, to avoid having to look for multiple keys:

```
/* blob format for storing:

    char magic[4]; # 'CCH1', might change for other version of ccache
[...]

PR #298, opened 2018-08-14
Add optional deduplication mode

afbjorklund:
This is a simple version of the deduplication mode, without making it mandatory...

It stores a "binary" hash with a link to the original hash, similar to the manifest file.

Then it can store a "reference" (.ref) to this binary, instead of storing another copy.

Another feature that adds runtime overhead (time), in favor of disk usage (space).

Implements #98


afbjorklund:
Currently the .binary file just contains the hash prefix of all the other files, not actually the data (as in #218)

jrosdahl:
Looks like a good implementation at a first glance, thanks! However, I would like to postpone this to after #218. Does that sound reasonable?

afbjorklund:
Perfectly reasonable, as I'm not fully sure trading speed for size is a good deal.

afbjorklund:
With the current implementation of the aggregated files, this idea would still work as-is.

i.e. you would still get the key, just use it to refer to 1 aggregated file instead of 7 single

jrosdahl:
> With the current implementation of the aggregated files, this idea would still work as-is.
> i.e. you would still get the key, just use it to refer to 1 aggregated file instead of 7 single

My suggestion to wait with this was due to implementation dependencies, not that the idea wouldn't fly.

afbjorklund:
Closed, due to result files.

PR #300, opened 2018-08-21
Bail out on compiler option -save-temps=obj

### LINKED ISSUE #299: Bail out on compiler option -save-temps=obj
Currently ccache bails out in the presence of -save-temps but does not bail out in the presence of -save-temps=obj

```
$ ./ccache-3.4.2/ccache /bin/gcc -x c -c -save-temps=obj main.c
$ ls
ccache-3.4.2  ccache-3.4.2.tar.bz2  main.c  main.o
$ ./ccache-3.4.2/ccache /bin/gcc -x c -c -save-temps main.c
$ ls
ccache-3.4.2  ccache-3.4.2.tar.bz2  main.c  main.i  main.o  main.s
$
```

afbjorklund:
Also add a test, and handle default (=cwd)

Closes #299

### DIFF
--- a/src/compopt.c
+++ b/src/compopt.c
@@ -35,6 +35,8 @@ static const struct compopt compopts[] = {
 	{"--output-directory", AFFECTS_CPP | TAKES_ARG}, // nvcc
 	{"--param",         TAKES_ARG},
 	{"--save-temps",    TOO_HARD},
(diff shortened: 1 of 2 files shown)


jrosdahl:
Applied, thanks!

PR #301, opened 2018-08-25
Add depend mode (avoid the preprocessor)

### LINKED ISSUE #234: Idea: Bypass preprocessor when creating cache entries
### Background
Direct mode gives us much faster cache hits because the preprocessor
is bypassed.  In this issue I'd like to share an idea on how to also
*create* cache entries without using a separate preprocessor step.  The
idea is to reduce unnecessary overhead when creating the cache entry.

Rationale:
 - when using `run_second_cpp` (which is the default now), ccache
   executes the preprocessor just to determine the object hash, and
   will do so for every cache miss.  While compiling, the same work is
   done again.
 - when executing a massively parallel build using ccache and distcc,
   the system controlling the build can get a fairly high load because of
   all these ccache-only preprocessor executions.
 - having only hashed compiler flags and per file checksums as in direct mode might 
   be just enough for practical caching purposes

### Method
CCache currently extracts the actual used files from the preprocessor
output.  When we pass the `-MD`/`-MMD` and `-MF` flags to the
compiler, it is also possible to get this list by parsing the
dependency output.

In short, the procedure will look like this:
 - on a direct cache miss, immediately run the actual compiler (with `-MD`)
 - get the include files from the .d file and calculate their hashes
 - use these hashes for both the manifest file and to create/extend
   the object hash
   
So this really is some hybrid of "direct mode" and `run_second_cpp`,
though the name of the latter doesn't really apply anymore in this case.

Here are some caveats when using this approach:
 - it is not possible to automatically distinguish system headers from
   other headers anymore, which e.g. makes it hard to have all
   contributing files in the manifest, but leave out the system headers
   from the .d files
 - changes such as `-D` flags that do not affect the build (because the
   preprocessor output is still the same) will be treated as cache misses
 - it does not work with the unify feature

I'm currently working on a prototype that expects the `-MD` or `-MMD`
flags to be already present.  In theory it's also possible to let CCache add
these flags when needed.

### Request for feedback
What do you think?  Is this something that might be incorporated into
CCache?  Do you have an idea on how to name this feature?  Anything else?

afbjorklund:
When using *run_second_cpp* (which is the default now), ccache executes
the preprocessor just to determine the object hash, and will do so for
every cache miss. While compiling, the same work is done again.

When executing a massively parallel build using ccache and distcc,
the system controlling the build can get a fairly high load because of
all these ccache-only preprocessor executions.

This is based on the work of @gjkloosterman, with remaining items addressed.

Closes #234


gjkloosterman:
@afbjorklund Thanks for picking this up and taking care of the loose ends!

jrosdahl:
@afbjorklund & @gjkloosterman: Merged to master now, thanks!

I made some improvements on top of your changes (https://github.com/ccache/ccache/compare/c424870...665320c); don't hesitate to review them if wanted.

afbjorklund:
Thanks

PR #310, opened 2018-09-12
Fix configure arg --without-bundled-zlib

rkitover:
If the result of the AC_ARG_WITH check is 'no', assume system zlib.


afbjorklund:
LGTM

jrosdahl:
Thanks, applied on master.

PR #311, opened 2018-09-12
Minor fixes for cygwin/msys2

rkitover:
Add -Wno-error=implicit-fallthrough to dev.mk so that zlib will compile
with -Werror.

Disable -Wdeprecated-declarations in the mkstemp() shim in util.c so
that it compiles with -Werror.

Fix a couple of erroneous checks in zlib that assume cygwin is the same
as win32.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -17,6 +17,7 @@ case $host in
 esac
 
 AC_SUBST(extra_libs)
+AC_SUBST(extra_cflags)
 AC_SUBST(include_dev_mk)
 AC_SUBST(test_suites)
 AC_SUBST(disable_man)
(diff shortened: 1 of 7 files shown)


rkitover:
Added autoconf check for `-Wimplicit-fallthrough`, build should pass now.

afbjorklund:
Have you reported these issues in zlib already ? Maybe that mkstemp is used in error, see 6354a40908f6975c1d7549506fe4ac3dce953ba4

It would be great if you were able to test the other MSYS fixes, that are proposed in: #246

rkitover:
Hi, yes if the mkstemp check for cygwin is a historical artifact, I would suggest removing it, there are very very few people using old versions of cygwin. If there are macros for cygwin/msys2 versions, I could use those as well, I will look. One case of old cygwin usage, is the original msys (not msys2) there are still a very few people sometimes using the original msys releases.

I'll figure out some way to detect all of this, if not by macros then with autoconf.

I will try that PR in my environment.

rkitover:
Forgot to say, I will check on the zlib upstream situation as well. The WIDECHAR thing has been around for a while.

rkitover:
@afbjorklund I tried to research the `mkstemp()` issue some more, and was not able to see anything relevant, other than people having issues because `mkstemp()` **forces** binary mode.

And this doc:

http://www.cygwin.com/cygwin-ug-net/using-textbinary.html

```
The mkstemp() and mkstemps() calls force binary mode.
Use mkostemp() or mkostemps() with the same flags as
open() for more control on temporary files.
```

I think we can safely assume that `mkstemp()` will force binary under any cygwin?

afbjorklund:
> I think we can safely assume that mkstemp() will force binary under any cygwin?

It certainly was broken at the time, and most people don't care about text/binary.

Maybe it is this commit ? The one with `mkostemps`... Doesn't really matter.
https://cygwin.com/git/gitweb.cgi?p=newlib-cygwin.git;a=commitdiff;h=8092f467704a09cbfffb553e38f6ca1408c0d9ed

rkitover:
@afbjorklund in regards to the `mkstemp()` issue, the comment says:

```
mkstemp() on some versions of cygwin don't handle binary files, so override.
```

I can find no documentation for this claim, the cygwin documentation says binary files are always created.

If it ever did not do this, there is no way to check with a configure test because the behavior would depend on mount point options.

For the time being I just removed the check completely in this PR.

As for zlib, I'm afraid I hit a dead end. The mailing list is dead and requires confirmation by someone who is not responding. There are public repos or issue tracking. I don't think having these minor changes in the zlib directory would be a big deal for the time being.

jrosdahl:
@afbjorklund: LGTY?

jrosdahl:
Thanks, applied on master.

Note that the zlib fixes will most likely be lost when we upgrade zlib the next time...

PR #315, opened 2018-09-16
Add option for enabling more compiler warnings

afbjorklund:
`./configure --enable-more-warnings`

Address most warnings, except those about conversion between float/double and int32_t/int64_t.

We also get some warnings from included system headers, that we need to suppress (for `-Werror`)

Now passes `gcc -Wextra` (including pedantic) and `clang -Weverything` (with some exceptions)


jrosdahl:
Thanks, merged to master (with some tweaks).

PR #316, opened 2018-09-16
Add target for running clang-tidy on ccache

afbjorklund:
http://clang.llvm.org/extra/clang-tidy/

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -1,0 +1,6 @@
+---
+Checks:          'clang-diagnostic-*,clang-analyzer-*,-clang-analyzer-alpha*,-clang-analyzer-optin.performance.Padding'
+HeaderFilterRegex: 'src/.*|unittest/.*'
+CheckOptions:    
+...
+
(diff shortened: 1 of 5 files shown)


jrosdahl:
Thanks for trying it out!

I tried running clang-tidy 6.0 (Ubuntu package version 1:6.0-41~exp5~ubuntu1) and was not very impressed. I only saw bogus or uninteresting warnings, including the strcpy one you addressed (which I don't want to apply). Do you think it's worth applying this anyway?

afbjorklund:
It's like uncrustify and cppcheck and the other tools, I suppose doesn't hurt to run once in a while ?

Both clang tidy and clang static analyzer seem to be clean now, not sure it is worth to run automatically...

Maybe keep the target around ? I thought `compiledb` was neat...

Although I need to fix it, since the silent make rules default broke it.

afbjorklund:
It was not the silent mode, it was versioned compilers...
The current regular expression doesn't handle `clang-5.0`

https://github.com/nickdiego/compiledb/blob/master/compiledb/parser.py#L28:_L29

``` py
cc_compile_regex = re.compile("^.*-?g?cc$|^.*-?clang$")
cpp_compile_regex = re.compile("^.*-?[gc]\+\+$|^.*-?clang\+\+$")
```

Thus the compile_commands.json ended up empty (`[]`). Will open a bug report.

afbjorklund:
See https://github.com/nickdiego/compiledb/issues/54

jrosdahl:
I applied the target parts in ea9abd8bb3d70a761a362756aaea17108faa7a4a. Thanks!

PR #319, opened 2018-10-03
Add configure option for silent make rules

afbjorklund:
Adopted from AM_SILENT_RULES, default to no

``` console
$ make V=0
    [CC] src/main.o
    [CC] src/args.o
    [CC] src/ccache.o
    [CC] src/cleanup.o
    [CC] src/compopt.o
    [CC] src/conf.o
    [CC] src/counters.o
    [CC] src/execute.o
    [CC] src/exitfn.o
    [CC] src/hash.o
    [CC] src/hashutil.o
    [CC] src/language.o
    [CC] src/lockfile.o
    [CC] src/manifest.o
    [CC] src/mdfour.o
    [CC] src/stats.o
    [CC] src/unify.o
    [CC] src/util.o
    [CC] src/version.o
    [CC] src/hashtable.o
    [CC] src/hashtable_itr.o
    [CC] src/murmurhashneutral2.o
    [CC] src/snprintf.o
    [LD] ccache
```

https://www.gnu.org/software/automake/manual/html_node/Automake-Silent-Rules.html

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -1,3 +1,5 @@
+verbose = @verbose@
+
 srcdir = @srcdir@
 builddir = @builddir@
 VPATH = @srcdir@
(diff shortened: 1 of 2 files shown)


jrosdahl:
I suggest scrapping the configure option and make this the only option. Or do you see any reason to make it opt-in?

jrosdahl:
> I suggest scrapping the configure option and make this the only option.

I made it so.

Applied to master, thanks!

afbjorklund:
I just didn't want to make any surprising changes, and the automake macro allowed either variant...

PR #321, opened 2018-10-20
Fix const-correctness of the config formatter

afbjorklund:
Both strdup and format are returning "char *"

```
src/conf.c:642:22: error: cast from 'const char *' to 'char *' drops const qualifier [-Werror,-Wcast-qual]
        char *str = (char *)item->formatter(value);
                            ^
src/conf.c:659:22: error: cast from 'const char *' to 'char *' drops const qualifier [-Werror,-Wcast-qual]
        char *str = (char *)item->formatter(value);
                            ^
```

### DIFF
--- a/src/conf.c
+++ b/src/conf.c
@@ -19,7 +19,7 @@
 
 typedef bool (*conf_item_parser)(const char *str, void *result, char **errmsg);
 typedef bool (*conf_item_verifier)(void *value, char **errmsg);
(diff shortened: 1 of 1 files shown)


gasche:
Thanks! Sorry for the mishap.

afbjorklund:
No worries, everyone hates const :-)

afbjorklund:
Note that you need to build with `clang` and `--enable-more-warnings`, to even _see_ this issue...

PR #322, opened 2018-10-20
Fix missing header include / prototype

afbjorklund:
`no previous prototype for function 'unify_hash'`

When running with `--enable-more-warnings` that is


jrosdahl:
Thanks! I'll enable the extra warnings in dev mode by default so that I won't make mistakes like this in the future...

PR #326, opened 2018-11-07
Support gcc-8 -ffile-prefix-map

### LINKED ISSUE #325: gcc-8 -ffile-prefix-map support
Since gcc-8 there is -fmacro-prefix-map which translates __FILE__ strings in addition to mapping debug-info paths. See https://gcc.gnu.org/bugzilla/show_bug.cgi?id=70268

-ffile-prefix-map combines -fmacro-prefix-map and -fdebug-prefix-map into one option.

-fmacro-prefix-map will affect preprocessed output, so can be ignored in command-line hash. -ffile-prefix-map is then equivalent to -fdebug-prefix-map.

osctobe:
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=70268

Resolve #325.

### DIFF
--- a/src/ccache.c
+++ b/src/ccache.c
@@ -1786,6 +1786,12 @@ calculate_object_hash(struct args *args, struct hash *hash, int direct_mode)
 		if (str_startswith(args->argv[i], "-fdebug-prefix-map=")) {
 			continue;
 		}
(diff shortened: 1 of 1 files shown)


jrosdahl:
Applied on master, thanks.

PR #328, opened 2018-11-16
merge master to dev/memcached

jonnyyu:


jrosdahl:
The merge looks correct, but it took some time to verify it, and I don't want to spend time on doing that in the future. The dev/memcached won't be kept in sync with master going forward, as expained in https://github.com/ccache/ccache/pull/58#issuecomment-440415224. Sorry.

PR #329, opened 2018-11-19
Fix the build on IBM i / PASE

friedkiwi:
ccache building was broken on IBM i/PASE.

PASE is the Unix subsystem within IBM i (which is a development of the legacy OS/400 product line). PASE is binary compatible with AIX, however, subtle differences exist due to the operating system not being Unix-like _at all_. Applications which compile and run also run on IBM i, and most source trees which build on AIX will also build on IBM i.

The build was broken on IBM i for two reasons:

_ar needs -X64 on AIX and i_
`ar` on AIX and IBM i needs to explicitly define if the archive being created is for 32 or for 64 bit code. When this parameter is not specified, `ar` on AIX and IBM i fails hard without any output since GCC on AIX and IBM i does not rely on binutils for `ld` and `ar`.  I have updated the `configure.ac` script to check if it's being run on IBM i (which identifies as `OS400` through `uname`) or AIX and then use the right parameter. I've defaulted on `-X64` because 32 bit code is only rarely being used anymore and should not be used anymore for new applications. 

_IBM i does not properly define usleep_
The `usleep()` routine is available in PASE/IBM i since it's a direct port of AIX, however, it is not directly accessible due to the PASE headers not including the right definition. This PR adds an ifdef to add this extern so the linker fixes this up automatically.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -16,6 +16,7 @@ case $host in
         ;;
     *os400* | *aix*)
         AR="ar -X64"
+        ;;
 esac
 
 AC_SUBST(disable_man)
--- a/configure.ac
+++ b/configure.ac
@@ -14,6 +14,8 @@ case $host in
         windows_os=yes
         AC_DEFINE(_WIN32_WINNT,0x0600, Windows Vista or newer is required)
         ;;
+    *os400* | *aix*)
+        AR="ar -X64"
 esac
 
 AC_SUBST(disable_man)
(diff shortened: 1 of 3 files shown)


jrosdahl:
You can disregard my "Please also retarget the pull request to the 3.5-maint branch." comment – I'll apply this to master instead when ready.

By the way, @friedkiwi, do you need any more feedback?

jrosdahl:
Ping @friedkiwi.

jrosdahl:
I fixed the review comments myself and pushed to your branch.

PR #330, opened 2018-11-22
Support multiple -fsanitize-blacklist arguments

### LINKED ISSUE #318: Add support for multiple -fsanitize-blacklist arguments
Clang supports multiple `-fsanitize-blacklist` files which had been implemented [here](https://reviews.llvm.org/D7368). This should be supported by ccache as well. This is a follow up on #174.

jgalenson:
This modifies the code to support multiple -fsanitize-blacklist
arguments, which prevents ccache from incorrectly using a cached
result when one of the blacklist files has changed.

This fixes #318.

### DIFF
--- a/src/ccache.c
+++ b/src/ccache.c
@@ -237,7 +237,10 @@ static bool profile_use = false;
 static bool profile_generate = false;
 
 // Sanitize blacklist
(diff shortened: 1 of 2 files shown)


jrosdahl:
Thanks!

PR #333, opened 2018-11-26
Skip '-index-store-path' when building with Xcode

jonnyyu:
This option is used to enable 'Indexing while Building' in Xcode9,
which does not affect the build output.

The reason we have to skip it is it introduces a path contains
uuid, for example:

-index-store-path /Users/lanya/Library/Developer/Xcode/DerivedData/NEUer-bjvoyplxzoxgkpgkiodfvurkgzwn/Index/DataStore

This causes build command is different between builds. So skip it in manifest hashing.

### DIFF
--- a/src/confitems.c
+++ b/src/confitems.c
@@ -136,8 +136,8 @@ confitem_parse_sloppiness(const char *str, void *result, char **errmsg)
 			*value |= SLOPPY_PCH_DEFINES;
 		} else if (str_eq(word, "time_macros")) {
 			*value |= SLOPPY_TIME_MACROS;
-		} else if (str_eq(word, "no_clang_index_store")) {
-			*value |= SLOPPY_NO_CLANG_INDEX_STORE;
+		} else if (str_eq(word, "clang_index_store")) {
+			*value |= SLOPPY_CLANG_INDEX_STORE;
 		} else {
 			*errmsg = format("unknown sloppiness: \"%s\"", word);
 			free(p);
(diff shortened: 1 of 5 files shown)


jrosdahl:
Thanks for the patch.

If we let ccache skip `-index-store-path` and its argument, won't we risk bad behavior on cache hits? E.g. like this:

1. `ccache clang -index-store-path foo ...` → cache miss; index is stored in foo
2. Clean build tree, or whereever foo is located.
3. `ccache clang -index-store-path foo ...` → cache hit; **index is not stored in foo**

jonnyyu:
Yes, the index will not be restored when cache hit. 

I did more search for how the unique path is calculated and found the blog below.
https://pewpewthespells.com/blog/xcode_deriveddata_hashes.html 

So the essential problem is the unique identifier is calculated from the full path to .xcodeproj file, so it's breaks ccache relative path function when this -index-store-path option exists.

jrosdahl:
> So the essential problem is the unique identifier is calculated from the full path to .xcodeproj file, so it's breaks ccache relative path function when this -index-store-path option exists.

I don't understand this part. In what way does the relative path function get broken? I thought that the issue simply was that the hash part of the index store path changes between builds so that ccache's hash always will be different. But when skimming the blog post you mentioned, it looks like the hash is only based on the project configuration file location, which should be stable between builds?

But regardless of whether `-index-store-path` and its argument are included in the hash or not, ccache will not correctly replicate the compiler's behavior since the index store won't be updated on cache hits either way.

My proposal is to add a new sloppiness option for this, maybe called "clang_index_store". If set, skip the option just like in your proposed patch. Does this sound reasonable?

src/confitems.c
```diff
@@ -136,6 +136,8 @@ confitem_parse_sloppiness(const char *str, void *result, char **errmsg)
 			*value |= SLOPPY_PCH_DEFINES;
 		} else if (str_eq(word, "time_macros")) {
 			*value |= SLOPPY_TIME_MACROS;
+		} else if (str_eq(word, "no_clang_index_store")) {
```
jrosdahl: Except for `no_system_headers`, which I regard as a mistake (see #337), previous sloppiness names can be substituted for X in "ccache should be sloppy about X", so I think that it should be called `clang_index_store` without the `no_` prefix.

Also, the new value needs to be documented in `doc/MANUAL.adoc`. Maybe something like this?

```
*clang_index_store*::
    Ignore the Clang compiler option *-index-store-path* and its argument when
    computing the manifest hash. This is useful if you use Xcode, which uses an
    index store path derived from the local project path. Note that the index
    store won't be updated correctly on cache hits if you enable this option.
```

jrosdahl:
Thanks!

PR #334, opened 2018-11-26
Handle several levels of nonexistent directories in make_relative_path

jonnyyu:
Currently, ccache supports calculating relative path
for 1 level non-exist path. That is to say,
if the given path does not exist, however if its
parent directory exists, then ccache can calculate
the relative path correctly.

Unfortunately this doesn't fit the needs.
Xcode build system always adds these paths into header search path:
xxxxx/DerivedResources/x86-64
xxxxx/DerivedResources

these paths are build outputs for build rules.
For projects which doesn't use build rule to generate files
these directories do not exist.

So this change refine the logic of make_relative_path 
to recursively go up find the nearest existing directory
and use the remaining path as path_suffix.

### DIFF
--- a/src/ccache.c
+++ b/src/ccache.c
@@ -762,7 +762,7 @@ make_relative_path(char *path)
 		// path doesn't exist.
 		char *dir = dirname(path);
 		// find the nearest existing directory in path
-		while(stat(dir, &st) != 0) {
+		while (stat(dir, &st) != 0) {
 			char *parent_dir = dirname(dir);
 			free(dir);
 			dir = parent_dir;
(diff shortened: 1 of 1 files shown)


jrosdahl:
Thanks, looks good.

It would be nice to add a unit test case that proves that the patch fixes a problem. If you would like to write one, see `test/test_argument_processing.c` for related test cases. If not, I can do it.

jrosdahl:
Also, I suggest modifying the commit message to something more descriptive. Perhaps "Handle several levels of nonexistent directories in make_relative_path"?

jrosdahl:
Thanks.

PR #335, opened 2018-11-27
Handle bug where clang doesn't unlink output file

### LINKED ISSUE #331: Clang integrated assembler breaks hard linked cache entries.
ccache doesn't work well with clang integrated assembler. Here is a reproduction below. `expected.o` and `actual.o` at the last line should match. Both ccache and clang here are ToT as of today

```bash
#!/bin/bash
export CCACHE_DIR="${PWD}/ccache"
export CCACHE_HARDLINK=1
ccache -C

echo '.octa 0x0' > failcase.s
ccache clang -c failcase.s -o failcase.o
cp failcase.o expected.o

echo '.octa 0x1' > failcase.s
ccache clang -c failcase.s -o failcase.o

echo '.octa 0x0' > failcase.s
ccache clang -c failcase.s -o failcase.o
cp failcase.o actual.o

cmp expected.o actual.o
```

IIUC, this happens because:
 - At the first compilation, clang creates `failcase.o` and writes the content, and ccache stores it as a cache entry using a hard link.
 - At the second compilation, clang **overwrites** `failcase.o`, and the stored cache entry, as they share the same backing file.
 - At the third compilation, ccache gives the stored cache entry as the result, which is overwritten by the previous compilation.

I'm not sure this should be fixed at clang side or ccache side, but either of them should unlink the output file at the beginning of the compilation.

afbjorklund:
When using assembler, clang doesn't unlink output.

On the other hand, gcc seems to handle this fine ?

Closes #331

### DIFF
--- a/src/ccache.c
+++ b/src/ccache.c
@@ -1242,6 +1242,9 @@ to_cache(struct args *args)
 	args_add(args, "-o");
 	args_add(args, output_obj);
 
+	if (conf->hard_link)
+		x_unlink(output_obj);
+
 	if (generating_diagnostics) {
 		args_add(args, "--serialize-diagnostics");
 		args_add(args, output_dia);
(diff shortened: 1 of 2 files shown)


afbjorklund:
Possibly the unlink should only done for assembler files ?

jrosdahl:
Thanks!

PR #341, opened 2018-12-16
Add missing getopt_long.c source file to dist

### LINKED ISSUE #340: ccache 3.5 release archive is missing getopt_long.c
The ccache 3.5 download link on https://ccache.samba.org/download.html is missing getopt_long.c, but the 3.5 tag on this repo and master show the file present. Was this a minor oversight when creating the source code release packages, or has there been any effort to remove it from the ccache source (it doesn't look that way from what I can tell)?

afbjorklund:
Missed to include this, when making compilation optional

Closes #340

### DIFF
--- a/dev.mk.in
+++ b/dev.mk.in
@@ -89,6 +89,7 @@ source_dist_files = \
     src/confitems_lookup.c \
     src/envtoconfitems.gperf \
     src/envtoconfitems_lookup.c \
+    src/getopt_long.c \
     src/main.c \
     src/zlib/*.c \
     src/zlib/*.h \


jrosdahl:
This makes src/getopt_long.c be included twice in source_dist_files if configure doesn't detect getopt_long in the system libraries, but that seems to be harmless. :slightly_smiling_face: Another solution would be to make getopt_long.c compilable without warnings like before (e.g. [like this](https://stackoverflow.com/a/26541331/331771)) and include getopt_long.c unconditionally in 3pp_sources again...

PR #343, opened 2019-01-03
Fix clang warning with older compiler versions

afbjorklund:
Don't hardcode the clang version, apparently clang-3.8
did have the warning even though clang-3.4 did not...


afbjorklund:
@jrosdahl : maybe this could also be included in 3.5.2 ?

jrosdahl:
> @jrosdahl : maybe this could also be included in 3.5.2 ?

I'm hoping I don't have to release a 3.5.2. :slightly_smiling_face: But if/when I have to, the fix will be included.

PR #349, opened 2019-01-11
Support .d dependency file generation via environment variables

madman2003:
Previously "DEPENDENCIES_OUTPUT="test.d" gcc -c test.c" and
"SUNPRO_DEPENDENCIES="test.d" gcc -c test.c" would not cache
the .d dependency files.


jrosdahl:
Note: This was reverted in fec405539bca90ebd0992ae6606bd10eec3a64da.

PR #362, opened 2019-03-14
Two changes are contained in this commit:

pavsa:
1) Depend mode fix for a case with unchaged common hash (unchanged source file and other parameters
that affect the hash), but with some changes in the header file[s] the source file depends on.
This allows for multiple versions of the object file and dependency file to exist in the cache for a single source file.
Added simple unit test in test/suites/depend_different_headers.bash to verify this behavior.

2) Depend mode also forces the paths of the include files to be relative if the base dir is specified. This
was previously only done if preprocessor was run.

### DIFF
--- a/src/ccache.c
+++ b/src/ccache.c
@@ -1040,7 +1040,7 @@ use_relative_paths_in_depfile(const char *depfile)
 		cc_log("Base dir not set, skip using relative paths");
 		return; // nothing to do
 	}
-	if (!has_absolute_include_headers) {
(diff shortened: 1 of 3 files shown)


afbjorklund:
Why not make two commits ?

Mac (BSD) doesn't support `readlink -f`, so you will have to write that in some portable other fashion...

```
readlink: illegal option -- f
usage: readlink [-n] [file ...]
```

That is why Travis failed.

afbjorklund:
@pavsa : You don't have to close the PR, just reorder the commits a bit and change the readlink ?

PR #363, opened 2019-03-16
Support multiple versions of dependencies per one manifest

pavsa:
This was previously only done if preprocessor was run.

### DIFF
--- a/src/ccache.c
+++ b/src/ccache.c
@@ -1479,10 +1479,6 @@ to_cache(struct args *args, struct hash *depend_mode_hash)
 			failed();
 		}
 		update_cached_result_globals(object_hash);
-
-		// It does not make sense to update an existing manifest file in the depend
-		// mode.
-		x_unlink(manifest_path);
 	}
 
 	if (stat(output_obj, &st) != 0) {
--- a/test/run
+++ b/test/run
@@ -385,6 +385,7 @@ hardlink
 direct
 direct_gcc
 depend
+depend_different_headers
 basedir
 compression
 readonly
--- a/test/suites/depend_different_headers.bash
+++ b/test/suites/depend_different_headers.bash
@@ -1,0 +1,181 @@
+SUITE_depend_different_headers_SETUP() {
+    unset CCACHE_NODIRECT
+
+    BASEDIR=`pwd`
+    BASEDIR1="${BASEDIR}/test/dir1/"
+    BASEDIR2="${BASEDIR}/test/dir2/"
+    BASEDIR3="${BASEDIR}/test/dir3/"
+    BASEDIR4="${BASEDIR}/test/dir4/"
+
+    mkdir -p ${BASEDIR1} ${BASEDIR2} ${BASEDIR3} ${BASEDIR4}
+
+    cat <<EOF >${BASEDIR1}/test.c
+#include "header.h"
+#include <stdio.h>
+
+void test(){
+#ifdef CHANGE_THAT_AFFECTS_OBJECT_FILE
+printf("with change");
+#else
+printf("no change");
+#endif
+}
+EOF
+    cp -f "${BASEDIR1}/test.c" ${BASEDIR2}
+    cp -f "${BASEDIR1}/test.c" ${BASEDIR3}
+    cp -f "${BASEDIR1}/test.c" ${BASEDIR4}
+
+    cat <<EOF >"${BASEDIR1}/header.h"
+void test();
+EOF
+
+    cat <<EOF >"${BASEDIR2}/header.h"
+#define CHANGE_THAT_AFFECTS_OBJECT_FILE
+void test();
+EOF
+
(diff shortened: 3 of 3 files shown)


src/ccache.c
```diff
@@ -1040,7 +1040,7 @@ use_relative_paths_in_depfile(const char *depfile)
 		cc_log("Base dir not set, skip using relative paths");
 		return; // nothing to do
 	}
-	if (!has_absolute_include_headers) {
+	if (!has_absolute_include_headers && !conf->depend_mode) {
```
jrosdahl: Two thoughts:

1. Would you like to write a test case that fails without this fix?
2. I think that a better approach would be to update `has_absolute_include_headers` in `object_hash_from_depfile`, similar to what is done in `process_preprocessed_file`. What do you think?

test/run
```diff
@@ -385,6 +385,7 @@ hardlink
 direct
 direct_gcc
 depend
+depend_different_headers
```
jrosdahl: No need to create a new suite containing only one test. I suggest adding the test case to the depend suite instead.

test/suites/depend_different_headers.bash
```diff
@@ -0,0 +1,178 @@
...
+
+    cat <<EOF >"${BASEDIR4}/header.h"
+#include "header2.h"
+void test();
+EOF
+    cat <<EOF >"${BASEDIR4}/header2.h"
+static void some_function(){};
+EOF
+
+    backdate "${BASEDIR1}/header.h" "${BASEDIR1}/test.c"  "${BASEDIR2}/header.h" "${BASEDIR2}/test.c" "${BASEDIR3}/header.h" "${BASEDIR3}/test.c" "${BASEDIR4}/header.h" "${BASEDIR4}/header2.h" "${BASEDIR4}/test.c"
```
jrosdahl: I suggest splitting this long line into several for readability.

test/suites/depend_different_headers.bash
```diff
@@ -0,0 +1,178 @@
...
+
+    DEPFLAGS="-MD -MF test.d"
+}
+
+function generate_reference_compiler_output() {
```
jrosdahl: [Please remove the function keyword](http://mywiki.wooledge.org/BashPitfalls#function_foo.28.29).

test/suites/depend_different_headers.bash
```diff
@@ -0,0 +1,178 @@
...
+    expect_equal_object_files reference_test.o test.o
+    expect_equal_files reference_test.d test.d
+    expect_stat 'cache hit (direct)' 1
+    expect_stat 'cache hit (preprocessed)' 0
+    expect_stat 'cache miss' 2
+    expect_stat 'files in cache' 5      # 2x .o, 2x .d, 1x manifest
+
+    # Compile dir3
+    # dir3 header change does not change object file compared to dir1, but ccache still adds
+    # additional .o/.d file in the cache due to different contents of the header file
```
jrosdahl: Thank you for good comments!

test/suites/depend_different_headers.bash
```diff
@@ -0,0 +1,178 @@
...
+    generate_reference_compiler_output
+    CCACHE_DEPEND=1 CCACHE_BASEDIR=${BASEDIR4} $CCACHE_COMPILE $DEPFLAGS -c test.c
+    expect_equal_object_files reference_test.o test.o
+    expect_equal_files reference_test.d test.d
+    expect_different_files reference_test.d ${BASEDIR1}/test.d
+    expect_stat 'cache hit (direct)' 5
+    expect_stat 'cache hit (preprocessed)' 0
+    expect_stat 'cache miss' 4
+    expect_stat 'files in cache' 9
+}
```
jrosdahl: [The last newline is missing](https://thoughtbot.com/blog/no-newline-at-end-of-file).

src/ccache.c
pavsa: But the dependency file first has to be relativizated before ```object_hash_from_depfile``` is called since it only calculates the hash. 

Maybe this would be better? 

```
        if (conf->depend_mode)
        {
                has_absolute_include_headers = 1; // force the relativization of dependency paths since we do not yet know if the compiler produced absolute paths or not without parsing the dependency file first
        }

	if (generating_dependencies) {
		use_relative_paths_in_depfile(output_dep);
	}

	if (conf->depend_mode) {
		struct file_hash *object_hash =
			object_hash_from_depfile(output_dep, depend_mode_hash);
```

pavsa: I've separated this issue into #366

test/run
pavsa: I wanted to do that, but the suite setup for that one test and also the test itself is too long and elaborate to fit in there.

test/suites/depend_different_headers.bash
pavsa: Thanks!

jrosdahl:
Thanks!

PR #366, opened 2019-03-22
Enable generating relative paths in dependency file in depend mode.

pavsa:
This was previously only done if preprocessor was run.

Added test case.

### DIFF
--- a/src/ccache.c
+++ b/src/ccache.c
@@ -1040,7 +1040,7 @@ use_relative_paths_in_depfile(const char *depfile)
 		cc_log("Base dir not set, skip using relative paths");
 		return; // nothing to do
 	}
-	if (!has_absolute_include_headers) {
+	if (!has_absolute_include_headers && !conf->depend_mode) {
 		cc_log("No absolute path for included files found, skip using relative"
 		       " paths");
 		return; // nothing to do
--- a/test/suites/depend.bash
+++ b/test/suites/depend.bash
@@ -67,6 +67,14 @@ SUITE_depend() {
     expect_stat 'files in cache' 3
 
     # -------------------------------------------------------------------------
+    TEST "Dependency file paths converted to relative if CCACHE_BASEDIR specified"
+        
+    CCACHE_DEPEND=1 CCACHE_BASEDIR="`pwd`" $CCACHE_COMPILE $DEPSFLAGS_CCACHE -c "`pwd`/test.c"
+    if grep -q "[^.]/test.c" "test.d"; then
+        test_failed "Dependency file does not contain relative path to test.c"
+    fi
+
+    # -------------------------------------------------------------------------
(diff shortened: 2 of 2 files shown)


src/ccache.c
```diff
@@ -1040,7 +1040,7 @@ use_relative_paths_in_depfile(const char *depfile)
 		cc_log("Base dir not set, skip using relative paths");
 		return; // nothing to do
 	}
-	if (!has_absolute_include_headers) {
+	if (!has_absolute_include_headers && !conf->depend_mode) {
```
jrosdahl: [Continuing the discussion from #363]

> But the dependency file first has to be relativizated before object_hash_from_depfile is called since it only calculates the hash.

My thought was that `object_hash_from_depfile` could be moved before `use_relative_paths_in_depfile` – then `has_absolute_include_headers` can be set by the former and read by the latter. I.e. something like this:

```diff
--- a/src/ccache.c
+++ b/src/ccache.c
@@ -1135,7 +1135,11 @@ object_hash_from_depfile(const char *depfile, struct hash *hash)
 			if (str_endswith(token, ":") || str_eq(token, "\\")) {
 				continue;
 			}
-			remember_include_file(x_strdup(token), hash, false, hash);
+			if (!has_absolute_include_headers) {
+				has_absolute_include_headers = is_absolute_path(token);
+			}
+			char *path = make_relative_path(x_strdup(token));
+			remember_include_file(path, hash, false, hash);
 		}
 	}
 
@@ -1468,10 +1472,6 @@ to_cache(struct args *args, struct hash *depend_mode_hash)
 		failed();
 	}
 
-	if (generating_dependencies) {
-		use_relative_paths_in_depfile(output_dep);
-	}
-
 	if (conf->depend_mode) {
 		struct file_hash *object_hash =
 			object_hash_from_depfile(output_dep, depend_mode_hash);
@@ -1485,6 +1485,10 @@ to_cache(struct args *args, struct hash *depend_mode_hash)
 		x_unlink(manifest_path);
 	}
 
+	if (generating_dependencies) {
+		use_relative_paths_in_depfile(output_dep);
+	}
+
 	if (stat(output_obj, &st) != 0) {
 		cc_log("Compiler didn't produce an object file");
 		stats_update(STATS_NOOUTPUT);
```

It's basically just an optimization to avoid having to read/parse/write the dep file file a second time when we know there are no absolute paths in it. (Even better would be to refactor the two reads into one...)

What do you think?

pavsa: That looks good.

jrosdahl: Yes, that would be OK to me.

jrosdahl: I'll merge the PR as is and make the above mentioned change.

pavsa: Thank you, I've been a little short on time this week to finish it.

PR #369, opened 2019-03-29
Memcached as second backend

bestdeveloperever:


afbjorklund:
That code looked familiar...

Issue #368, opened 2019-03-28
-g0 option not handled optimally


pavsa:
Hi,

Setup:
- using CCACHE_BASEDIR
- using CCACHE_DEPEND

I've noticed that having -g0 in gcc params causes pwd being hashed into common hash, despite -g0 meaning no debug symbols (the same as not including -g0). This causes such cached files to be unavailable for other users of the cache.

If -g is detected, it forces ```generating_debuginfo = true```, but there's no checking if it is -g0 or not, which turns off debugging info. Or am I mistaken?

https://github.com/ccache/ccache/blob/128109ebb7ffd2a2168677cae42d0e0200f0dbd7/src/ccache.c#L2623-L2640

Isn't this how it should be handled?
```
+               if (str_eq(argv[i], "-g0")) {
+                   continue;
+               }
                if (str_startswith(argv[i], "-g")) {
                        generating_debuginfo = true;

```

afbjorklund:
I think you might be right, historically we have used either of `-O2 -g` or `-O0`...
i.e. without debugging symbols was the default, so there was no need to disable

Patch above sounds reasonable, combine with if below and make a PR ?
``` C
if (str_startswith(argv[i], "-g") && !str_eq(argv[i], "-g0")) {
```

jrosdahl:
@pavsa: Thanks for noticing this. Would you be interested in fixing it? Otherwise I can have a look at it.

I suggest making sure that this case is handled as well: `gcc -g3 -g0 ...`. I.e., the last `-g` option "wins". For extra niceness, add a test case.

pavsa:
I will look at it tomorrow and I'll add a test case. Thank you.

hwti:
Shouldn't `-g1`/`-g2` disable `generating_debuginfo_level_3` ?

`-gdwarf-3` sets `generating_debuginfo_level_3`, but it has nothing to do with the level.

Also, I'm skeptical about `using_split_dwarf`, in https://github.com/ccache/ccache/issues/149 my test showed that at least for clang, `-gsplit-dwarf -g1` means `-g1` (so ccache would expect a dwo when clang wouldn't generate one).

jrosdahl:
> Shouldn't -g1/-g2 disable generating_debuginfo_level_3 ?

[...]

Issue #149, opened 2016-11-07
Issues with split debug info on clang


hwti:
With clang, unless -fno-split-dwarf-inlining is used (on trunk, it isn't supported on current 3.9.0), the order between -g* and -gsplit-dwarf matters, see https://github.com/llvm-mirror/clang/blob/master/lib/Driver/Tools.cpp#L4776.

But ccache doesn't preserve the order :
> Command line: clang++ -g1 -gsplit-dwarf -c -o test.o test.cpp
...
Executing /usr/bin/clang++ -gsplit-dwarf -g1 -fcolor-diagnostics -E test.cpp
...
Running real compiler
Executing /usr/bin/clang++ -gsplit-dwarf -g1 -fcolor-diagnostics -c -o test.o test.cpp
Compiler didn't produce a split dwarf file
Failed; falling back to running the real compiler
Executing /usr/bin/clang++ -g1 -gsplit-dwarf -c -o test.o test.cpp

In this case, "-g1 -gsplit-dwarf" (same as "-gsplit-dwarf") becomes "-gsplit-dwarf -g1" (same as "-g1"), which doesn't write a dwo.
So when starting from a clean build tree with a clean cache, ccache runs the compiler again with the original arguments like in the traces above.
But later, if the dwo already exists :
 - the compiler only writes the .o (not referencing the dwo)
 - ccache caches the new .o with the old .dwo

If the user looks at the file dates at this point, he can see the issue (the .dwo is older than the .o).
But after cleaning the build tree, it becomes really confusing, since on the next build ccache restores the cached .o and .dwo, so :
 - both files have the same date
 - the .o has only limited debug information (-g1)
 - the .dwo is perfectly valid, but it isn't loaded by gdb since the .o doesn't reference it

Could ccache do something here ?
If the order was preserved, this specific case would work, but "-gsplit-dwarf -g1" would always fail, re-run the compiler, and nothing would be cached (since clang wouldn't create the dwo).
Perhaps the only solution would be to launch clang with "-###", then parse the cc1 arguments, since the driver handles the conflicting or overlapping options before calling the frontend. But it would be a massive change, and would require clang-specific code, since the cc1 options are different (and could change between versions...).

PR #371, opened 2019-04-02
Handle -g0 properly

pavsa:
Solves #368 

I have reused testing in depend mode as it has ideal setup this would fail.

Also removed the depend_different_headers.bash due to not being needed anymore.


jrosdahl:
Oops, I was too fast merging this. And GitHub annoyingly doesn't allow me to reopen the pull request after reverting. :slightly_frowning_face:

src/ccache.c
```diff
@@ -2634,6 +2634,11 @@ cc_process_args(struct args *args, struct args **preprocessor_args,
 
 		// Debugging is handled specially, so that we know if we can strip line
 		// number info.
+		if (str_eq(argv[i], "-g0")) {
```
jrosdahl: This is not quite enough since `-g3 -g0` still will enter the if statement below and for instance set `conf->run_second_cpp = true`.

And more importantly: since `-g0` isn't added to `stripped_args` it won't be sent to the compiler so the compiler will then use `-g3`.

test/suites/depend.bash
```diff
@@ -78,7 +78,7 @@ EOF
     backdate "$BASEDIR3/header.h" "$BASEDIR3/test.c"
     backdate "$BASEDIR4/header.h" "$BASEDIR4/test.c" "$BASEDIR4/header2.h"
 
-    DEPFLAGS="-MD -MF test.d"
+    DEPFLAGS="-g1 -g0 -MD -MF test.d"
```
jrosdahl: I would prefer a separate test case for this (in the base suite) to make it possible to understand to purpose of the flags.

jrosdahl:
@pavsa: I don't know if you got mail about my review comments since the PR has been closed already, so here's an extra ping that I left some feedback.

src/ccache.c
pavsa: Yes, there's also another problem with:
1. conf->unify = true
2. -g1 was parsed and conf->unify was set to false
3. -g0 was parsed, but we can't restore conf->unify back to true, as there is no information of state of conf->unify before the -g1 caused it to be set to false

How do we solve this?

jrosdahl: Right. Thinking more about this, I now remember that we used to have smarter processing of `-g*` options (#92) but it was reverted (b8aeb9826fcbf4627df8c0d9a9b110c4523ea0f1) due to being buggy (#149).

The best solution I see is to remember the last `-g*` option and only act (maybe disable unify, maybe set , etc.) after we have looped over all arguments. But it is quite complex to figure out how all options interact, so it's not clear to me how that should be implemented. For instance, `-ggdb3 -g0` seems to be equivalent to `-g3 -ggdb0`, but `-ggdb3 -g2` is not equivalent to `-g3 -ggdb2`...

So maybe we should keep things safe and not try to handle `-g3 -g0` like I suggested. We could still handle the simple case in #368 like you did, but the argument then needs to be added to `stripped_args`.

Hmm... It does seem like `-g*0` disables all debug info enabled by previous options, though, so maybe we could make use of that fact. I made an attempt in 749fa11f3d59 on a separate branch. What do you think about that approach?

pavsa: Looks good to me. Will you push it?

jrosdahl: Thanks for having a look. I'll push it after writing some tests.

PR #372, opened 2019-04-03
Implement memcached as backend

bestdeveloperever:


jrosdahl:
Thanks for your interest in ccache!

Some generic and specific comments:

1. From https://github.com/ccache/ccache/blob/master/CONTRIBUTING.md#contributing-code: _"If you plan to implement major changes it is wise to open an issue on GitHub (or send a mail to the mailing list) asking for comments on your plans before doing the bulk of the work. That way you can avoid potentially wasting time on doing something that might not end up being accepted."_
2. Please don't just dump a pile of code in a pull request without any description of what it is about, what your intentions and expectations are, the status of the code in question, etc etc. We can of course guess some of this from just reading the code diff, but that will be guesswork and it will take time to try to understand something that is better described by you as a pull request author.
3. Much of this code is obviously taken from @afbjorklund's work in #58. I guess you then already have read [my comment](https://github.com/ccache/ccache/pull/58#issuecomment-435699712) regarding the way forward, specifically the five bullets at the end? From glancing over the diff, it looks like it's kind of the same design as #58 with the addition of an API struct that is tailored to fit the current code. There are several problems here: a) It's a partial implementation of step 5 without doing steps 1-4 before. b) It's easy to see that the implementation is not complete since it doesn't handle the `.cov`, `.su` and `.dwo` files. One of the reasons for doing #218 and the other steps I mentioned is to make backends not have to know about which individual output files there are.

If you want to drive the memcached backend feature forward, the proper way at this stage is to discuss and design, not to implement.

afbjorklund:
I already commented in ~~#369~~: I do **not** approve of losing copyright and history, neither does the other developers. Please see LICENSE, and like @jrosdahl mentioned we already talked a lot about it...

jrosdahl:
No description and no reply → closing.

PR #379, opened 2019-04-15
Copy rather than link depend from the cache

### LINKED ISSUE #378: Hardlinked dep files don't work well with automake
On an example below, the second `mv` failed with `mv: 'foo.d.tmp' and 'foo.d' are the same file` message. The ccache version here was [latest](https://github.com/ccache/ccache/commit/6d71b6179a6e3bf26e2c17162254cdddd66c297b) as of today.

```bash
#!/bin/bash
cd "$(dirname "$0")"

export CCACHE_DIR="${PWD}/ccache"
export CCACHE_HARDLINK=1

rm -rf "${CCACHE_DIR}" foo.{c,h,d,d.tmp}

touch foo.h foo.c
sleep 1s

# ccache stores foo.d.tmp by a hardlink.
ccache gcc -c -MMD -MF foo.d.tmp -include foo.h foo.c
mv foo.d.tmp foo.d

# ccache pulls foo.d.tmp by a hardlink.
# The resulting foo.d.tmp points the same inode to foo.d.
ccache gcc -c -MMD -MF foo.d.tmp -include foo.h foo.c

# mv rejects to override a file if the dest and src file points the same inode.
mv foo.d.tmp foo.d
```

We can see this pattern of build in [automake](https://github.com/autotools-mirror/automake/blob/596e9e130f38ae068ac6150b2d62eafd544900f6/lib/am/depend2.am#L97) and automake-generated build systems.

I'm not sure if we should fix this in ccache side ([again](https://github.com/ccache/ccache/issues/331)), but if we do, just avoiding the hardlink on deps file at [get_file_from_cache](https://github.com/ccache/ccache/blob/6d71b6179a6e3bf26e2c17162254cdddd66c297b/src/ccache.c#L1253) and [its caller](https://github.com/ccache/ccache/blob/6d71b6179a6e3bf26e2c17162254cdddd66c297b/src/ccache.c#L2245) should work.

afbjorklund:
The depends are small enough to avoid the copy, since it leads
to issues with mv when using automake to generate the Makefile.

Closes #378

### DIFF
--- a/src/ccache.c
+++ b/src/ccache.c
@@ -1250,12 +1250,12 @@ move_file_to_cache_same_fs(const char *source, const char *dest)
 	do_copy_or_move_file_to_cache(source, dest, false);
 }
 
-// Copy or link a file from the cache.
+// Helper method for get_file_from_cache and copy_file_from_cache.
 static void
-get_file_from_cache(const char *source, const char *dest)
+do_copy_or_link_file_from_cache(const char *source, const char *dest, bool copy)
 {
 	int ret;
-	bool do_link = conf->hard_link && !file_is_compressed(source);
+	bool do_link = !copy && conf->hard_link && !file_is_compressed(source);
 	if (do_link) {
 		x_unlink(dest);
 		ret = link(source, dest);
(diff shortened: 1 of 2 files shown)


jrosdahl:
Thanks!

PR #380, opened 2019-04-16
Add support for two level cache

pavsa:
Would this be any use? I can polish it more if there's an interest to merge it.

cache_dir_l0 is added to configuration.
If set, ccache performs its first lookup from this directory.
If it fails, it does a second lookup in *cache_dir* and continues as usual.


jrosdahl:
It's a fine idea, and also an idea has that has come up before. I would like to wait with adding support for multilevel caches like this until I/we have found time to work on generalizing how backends work and how to configure them.

One question though: Why only enable it for the depend mode?

pavsa:
For simple reason: I have not tested it in other modes.

afbjorklund:
Probably should call something else than "multi level cache" too, to avoid confusion with nlevels ?

afbjorklund:
@pavsa : how do you transport the files, from "some central compile server" to the L0 cache ?

pavsa:
By rsync

pavsa:
> 
> 
> Probably should call it something else than "multi level cache" too, to avoid confusion with NLEVELS ?

That is a better name :) I admit it is confusing with the NLEVELS. Everything can be renamed in the end.

> 
> 
> @pavsa : how do you transport the files, from "some central compile server" to the L0 cache ?
By rsync.

afbjorklund:
I don't think this feature has _anything_ to do with depend mode, so that part can probably be dropped...

We used a similar approach for cache located on NFS, when people were scared of using memcached.
Basically the only difference was that files were copied from the central cache, over to the local server.

The main reason for this is that you wanted to get local hits next time, if you built the same thing again.

klimkin:
I would love to have this feature! This way we can set level-0 to our CI cache to speedup builds after new code pulls.

megabyde:
Any updates on this feature? This would greatly improve caching experience in shared environment with nightly CI builds.

pavsa:
> 
> 
> Any updates on this feature? This would greatly improve caching experience in shared environment with nightly CI builds.

Well it works as is, so you can use it as is with older branch as I still do. 
Until the API discussion is resolved, this is as good as it gets.

afbjorklund:
There is a simple FileBackend for the StorageBackend, that will accomplish this:

https://github.com/ccache/ccache/issues/414#issuecomment-848196357

```C++
bool
FileBackend::put(const std::string& url, const std::string& path)
{
  try {
    Util::copy_file(path, url, true);
  } catch (const std::exception& e) {
    LOG("Failed to put {} to file cache: exception: {}", path, e.what());
    return false;
  }
  LOG("Succeeded to put {} to file cache", path);
  return true;
}

bool
FileBackend::get(const std::string& url, const std::string& path)
{
  try {
    Util::copy_file(url, path, false);
  } catch (const std::exception& e) {
    LOG(
      "Failed to get {} from file cache: exception: {}", path, e.what());
    return false;
  }
  LOG("Succeeded to get {} from file cache", path);
  return true;
}

```


It will look in the local directory first and in the remote directory second, though.

jrosdahl:
I consider this done as of #857 so I'll close this one.

It's currently not possible to configure ccache to query a secondary storage _before_ the primary storage, but that can be done in the future.

PR #381, opened 2019-04-16
Improve CMake+CUDA compatibility: add "cu" language, recognize -dc as compilation only argument

akors:
Hi, I was trying to use ccache from a CMake project to compile CUDA files and ran into two issues:

First, CMake always passes `-x cu` to NVCC. This is recognized by ccache as an explicit language setting, but the language "cu" is not understood by ccache. The first commit add the "cu" language as an alias to "cuda".

Second, we use [separable compilation](https://docs.nvidia.com/cuda/cuda-compiler-driver-nvcc/index.html#using-separate-compilation-in-cuda) which requires CMake to pass the `-dc` or `--device-c` option. This option as stated [here](https://docs.nvidia.com/cuda/cuda-compiler-driver-nvcc/index.html#options-for-specifying-compilation-phase-device-c) implies the `-c` option already.
Previously, the `-c` option was missing and ccache didn't cache the compiled files. The second commit makes ccache accept `-dc`  instead of `-c` if the compiler is recognized as NVCC.

I don't know if this is The Right Way™ to repair this, but it definitely fixes my use case.

### DIFF
--- a/src/ccache.c
+++ b/src/ccache.c
@@ -2300,6 +2300,7 @@ cc_process_args(struct args *args, struct args **preprocessor_args,
                 struct args **compiler_args)
 {
 	bool found_c_opt = false;
+	bool found_dc_opt = false;
 	bool found_S_opt = false;
 	bool found_pch = false;
 	bool found_fpch_preprocess = false;
@@ -2505,6 +2506,12 @@ cc_process_args(struct args *args, struct args **preprocessor_args,
 		// We must have -c.
 		if (str_eq(argv[i], "-c")) {
 			found_c_opt = true;
+			continue;
+		}
+
+		// when using nvcc with separable compilation, -dc implies -c
+		if ((str_eq(argv[i], "-dc") || str_eq(argv[i], "--device-c")) && guessed_compiler == GUESSED_NVCC) {
+			found_dc_opt = true;
 			continue;
 		}
 
@@ -3013,7 +3020,7 @@ cc_process_args(struct args *args, struct args **preprocessor_args,
 			if (language_for_file(argv[i])) {
 				cc_log("Multiple input files: %s and %s", input_file, argv[i]);
 				stats_update(STATS_MULTIPLE);
-			} else if (!found_c_opt) {
+			} else if (!found_c_opt && !found_dc_opt) {
 				cc_log("Called for link with %s", argv[i]);
 				if (strstr(argv[i], "conftest.")) {
 					stats_update(STATS_CONFTEST);
@@ -3152,7 +3159,7 @@ cc_process_args(struct args *args, struct args **preprocessor_args,
 		goto out;
 	}
 
-	if (!found_c_opt && !found_S_opt) {
+	if (!found_c_opt && !found_dc_opt && !found_S_opt) {
 		if (output_is_precompiled_header) {
 			args_add(common_args, "-c");
 		} else {
@@ -3353,6 +3360,10 @@ cc_process_args(struct args *args, struct args **preprocessor_args,
 
 	if (found_c_opt) {
 		args_add(*compiler_args, "-c");
+	}
+
+	if (found_dc_opt) {
+		args_add(*compiler_args, "-dc");
 	}
 
 	for (size_t i = 0; i < arch_args_size; ++i) {
--- a/src/ccache.c
+++ b/src/ccache.c
@@ -3183,7 +3183,7 @@ cc_process_args(struct args *args, struct args **preprocessor_args,
 		goto out;
 	}
 
-	if (!conf->run_second_cpp && str_eq(actual_language, "cuda")) {
+	if (!conf->run_second_cpp && str_eq(actual_language, "cu")) {
 		cc_log("Using CUDA compiler; not compiling preprocessed code");
 		conf->run_second_cpp = true;
 	}
--- a/src/language.c
+++ b/src/language.c
@@ -88,6 +88,7 @@ static const struct {
 	{"objc++-cpp-output",        "objective-c++-cpp-output"},
 	{"objective-c++-header",     "objective-c++-cpp-output"},
 	{"objective-c++-cpp-output", "objective-c++-cpp-output"},
+	{"cu"  ,                     "cuda-output"},
 	{"cuda",                     "cuda-output"},
 	{"assembler-with-cpp",       "assembler"},
 	{"assembler",                "assembler"},
--- a/src/language.c
+++ b/src/language.c
@@ -1,4 +1,4 @@
-// Copyright (C) 2010-2018 Joel Rosdahl and other contributors
+// Copyright (C) 2010-2019 Joel Rosdahl and other contributors
 //
 // See doc/AUTHORS.adoc for a complete list of contributors.
 //
@@ -64,8 +64,7 @@ static const struct {
 	{".HXX", "c++-header"},
 	{".tcc", "c++-header"},
 	{".TCC", "c++-header"},
-	{".cu",  "cuda"},
-	{".ic",  "cuda-output"},
+	{".cu",  "cu"},
 	{NULL,  NULL}
 };
 
@@ -80,6 +79,7 @@ static const struct {
 	{"c++",                      "c++-cpp-output"},
 	{"c++-cpp-output",           "c++-cpp-output"},
 	{"c++-header",               "c++-cpp-output"},
+	{"cu",                       "cpp-output"},
 	{"objective-c",              "objective-c-cpp-output"},
 	{"objective-c-header",       "objective-c-cpp-output"},
 	{"objc-cpp-output",          "objective-c-cpp-output"},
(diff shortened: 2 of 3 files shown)


src/ccache.c
```diff
@@ -3255,7 +3262,7 @@ cc_process_args(struct args *args, struct args **preprocessor_args,
 		goto out;
 	}
 
-	if (!found_c_opt && !found_S_opt) {
+	if (!found_c_opt && !found_dc_opt && !found_S_opt) {
 		if (output_is_precompiled_header) {
```
akors: I'm a bit unsure about this. I don't think NVCC does precompiled headers at all, so I don't know what the conditions are for `output_is_precompiled_header` to become true.

jrosdahl:
Looks fine to me. @seeraven: Would you like to review this PR?

Would it be possible to add something to the test suite for the new functionality?

akors:
> Would it be possible to add something to the test suite for the new functionality?

Looking at the test suite, I'm not sure I am capable of creating a useful test case. Also, the test suite fails on my machine (if it is started by `make test` ?) even for the unmodified master branch and for the v3.6 tag.

Is it possible that one of you guys add that? My suggestion is two changes:

Add `-x cu` to any current nvcc invocations in the test suite, cache must be hit.
Add `-dc` and test basically the same way as `-c` is tested.

seeraven:
The PR looks fine for me, but I also would like to have tests for the new behaviour. Like  @akors , I have also a failed test in the test suite cleanup:
```
Test suite:     cleanup
Test case:      Forced cache cleanup, size limit
Failure reason: Found 0 (expected 3) *.o files in /home/clemens/ccache/testdir.9111/.ccache
```
If you edit `test/run` and remove the line with `cleanup` all tests succeed. Would that help you @akors to write the test (you can simply extend the `test/suited/nvcc*.bash` files). Unfortunately I won't have time to write the tests.

akors:
@seeraven 

> If you edit test/run and remove the line with cleanup all tests succeed.

Not for me:

```
make test
  CC       src/version.o
  LD       ccache
  LD       unittest/run
  TEST     unittest/run
PASSED: 518 assertions, 103 tests, 10 suites
  TEST     ./test/run
Compiler version: gcc (GCC) 8.3.1 20190223 (Red Hat 8.3.1-2)

Running test suite base........................
FAILED

Test suite:     base
Test case:      CCACHE_NOHASHDIR
Failure reason: Expected "cache hit (preprocessed)" to be 2, actual 1
```
This is using the v3.6 tag.

> Unfortunately I won't have time to write the tests.

Me neither, unfortunately - especially since I would have to get familiarized with your test suite and debug the failures of the unchanged code first.

But off the bat and untested, what about this? I think this shouldn't change much about the tests, except that they would fail without the changes.

```diff
diff --git a/test/suites/nvcc.bash b/test/suites/nvcc.bash
index 3b713c0..bf577ca 100644
--- a/test/suites/nvcc.bash
+++ b/test/suites/nvcc.bash
@@ -53,7 +53,7 @@ nvcc_tests() {
     # instead of comparing the binary object files, we compare the dumps of
     # "cuobjdump -all -elf -symbols -ptx -sass test1.o".
     nvcc_opts_cpp="-Wno-deprecated-gpu-targets -c --x c++"
-    nvcc_opts_cuda="-Wno-deprecated-gpu-targets -c"
+    nvcc_opts_cuda="-Wno-deprecated-gpu-targets -dc -x cu"
     nvcc_opts_gpu1="--generate-code arch=compute_50,code=compute_50"
     nvcc_opts_gpu2="--generate-code arch=compute_52,code=sm_52"
     ccache_nvcc_cpp="$CCACHE $REAL_NVCC $nvcc_opts_cpp"
```

akors:
The tests work on Ubuntu, and I have added the suggested patch.

Unfortunately, this revealed an issue with the patch, namely that

Another issue turned up, namely that ccache tries to pass "cuda-output" to NVCC as the explicit language. I'll have to investigate this, so please don't merge yet.

akors:
> Another issue turned up, namely that ccache tries to pass "cuda-output" to NVCC as the explicit language. I'll have to investigate this, so please don't merge yet.

I believe this is fixed with the latest commit that I added.

I also added a test to the suite based on the other tests in the file. This test suite fails at the nvcc_nocpp2 step, which is with the preprocessing handling when `CCACHE_NOCPP2` is set.

In that case, I get redefinition errors because with device linking `-dc`, NVCC is stricter about that. Unfortunately, I don't know what this `CCACHE_NOCPP2` mode is and whether it actually makes sense for NVCC.

jrosdahl:
As noted above, the CI tests fail. Any news on this?

akors:
> As noted above, the CI tests fail. Any news on this?

CUDA device linking and the `CCACHE_NOCPP2` do not play well with each other. I don't understand enough about the latter to remedy this, sorry.

I left this PR open for any other contributors willing to look into that.

jrosdahl:
@akors: I've had a quick look at this and see if I understand enough to fix and merge the PR.

I think that the `-dc` + `CCACHE_NOCPP2` case is easy to fix: just force `run_second_cpp` if the guessed compiler is NVCC since it apparently doesn't play well with `!run_second_cpp`.

> Another issue turned up, namely that ccache tries to pass "cuda-output" to NVCC as the explicit language.

I don't understand why this is a problem and why 7fd95154cdde49f6ab05cb4a79c535da6b527ded would solve it. Could you perhaps expand on that?

Note that the code in question only runs in the `!run_second_cpp` case. The logic is like this: If `-x L` is specified explicitly then pass `-x P`, where P is the preprocessed language for L, to the final compilation when compiling preprocessed source code from the previous step (this is what `!run_second_cpp` means). But 7fd95154cdde49f6ab05cb4a79c535da6b527ded in effect just undos the logic; if `-x L` is specified in the original command line (and L is not preprocessed) then the code now adds `-x L` instead of `-x P` when sending input of type P to the compiler, which is incorrect. And if `-x P` (where P is preprocessed), it adds `-x P` (since P is the preprocessed language of P).

jrosdahl:
Digging deeper.

b2d6d53d7cbb303573ccce79ddfe74acad08c03f added mappings for extensions `.cu` and `.ic` to languages `cuda` and `cuda-output`, respectively, and then specifies that `cuda-output` is the preprocessed language for `cuda`. NVCC from Cuda 10.1 indeed understands the `.cu` extension:

```
% touch empty.cu
% nvcc -c empty.cu && echo success
success
```

However, it does not understand the `.ic` extension and also neither `cuda` nor `cuda-output` languages:

```
% touch empty.ic
% nvcc -c empty.ic
nvcc fatal   : Don't know what to do with 'empty.ic'
% nvcc -x cuda empty.cu
nvcc fatal   : Value 'cuda' is not defined for option 'x'
% nvcc -x cuda-output empty.cu
nvcc fatal   : Value 'cuda-output' is not defined for option 'x'
```

However, language `cu` is known, so 8d89d8cbfc9a46cccafe9e7a31007102b0411397 looks correct, (and it's not an alias but the correct language):

```
% nvcc -x cu -c empty.cu && echo success
success
```

@seeraven: Any thoughts here since you authored b2d6d53d7cbb303573ccce79ddfe74acad08c03f? Should we just remove the `.ic` extension and let language be `cu`, not mapped to a preprocessed language?

jrosdahl:
I previously wrote:

> I think that the -dc + CCACHE_NOCPP2 case is easy to fix: just force run_second_cpp if the guessed compiler is NVCC since it apparently doesn't play well with !run_second_cpp.

Now I get it – `run_second_cpp` is already forced to true if CUDA is used, but it checks the "cuda" language instead of the new "cu" language that @akors added: https://github.com/ccache/ccache/blob/6d76231f10b002f04625d619f38377d0b73c4e38/src/ccache.c#L3179-L3181

jrosdahl:
I took the liberty of removing the .ic extension and cuda/cuda-output languages in b488b8b and fixed the issue that made the nvcc_nocpp2 test suite crash. I also removed 7fd95154cdde49f6ab05cb4a79c535da6b527ded since I think that it is incorrect as explained in https://github.com/ccache/ccache/pull/381#issuecomment-516585756.

@akors & @seeraven: It would be great if you could verify that I haven't broken anything and that the `-dc` option now works as expected.

jrosdahl:
I now noticed that clang recognizes `-x cuda` so I'll add that language back. I haven't found any compiler that recognizes the `cuda-output` language or the `.ic` extension, though.

jrosdahl:
Or maybe not since it would make the mapping from the `.cu` extension ambiguous: If it's NVCC it should be mapped to "cu" but if it's Clang it should be mapped to "cuda". Confusing.

I guess the common use case is NVCC, so I'll ignore "clang -x cuda" for now.

akors:
Hey there, thanks for the fixes and the merge! This is very appreciated.
I tested the latest master and I can confirm that the caching works.

seeraven:
Finally I've got time to test it. I've built a version based on 3.7.4 and cherry-picked the relevant commits. This works like a charm when using the new cmake. Would be happy to see it in one of the next official releases. :-)

PR #384, opened 2019-04-21
Add possibility to export manifest as json

afbjorklund:
The manifest format is binary and optimized for performance and storage...
When discussing the manifest, it can be useful to export it to simple JSON.

Note that this format potentially duplicates both file strings and file entries.
If you are concerned about that (even for debug output), you can try gzip.

----
Includes sample python program, and a [JSON Schema](http://json-schema.org/) for the format used.

### DIFF
--- a/misc/check-manifest.py
+++ b/misc/check-manifest.py
@@ -1,0 +1,52 @@
+#!/usr/bin/env python
+
+import sys
+import os
+import json
+import hashlib
+
+try:
+    md4 = hashlib.md4
+except AttributeError:
+    from Crypto.Hash.MD4 import MD4Hash as md4
+except ImportError:
+    raise Error("please install pycrypto")
+
+
+def hash(file):
+    return md4(file.read()).hexdigest()
+
+mf = json.load(open(sys.argv[1]))
+for object in mf['objects']:
+    print "%s-%d" % (object['hash'], object['size'])
+    ok = True
+    sloppy = False
+    for file in object['files']:
+        path = file['path']
+        if os.path.exists(path):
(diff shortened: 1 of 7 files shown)


jrosdahl:
Who/what do you see will be the consumer of the JSON manifest data? Do you envision other tools than the check-manifest.py script?

The reason I'm asking is that I like the basic idea of the check-manifest.py script, but I'm less sure about whether the intermediary JSON format is of enough value. Not that the code would be much baggage, I'm just trying to understand the potential use cases here.

I originally implemented the --dump-manifest functionality as a separate script (removed in 8c60a3874565e24d31d168df18b764e352c371ff), but I chose to make it a proper ccache command to avoid knowledge about the manifest format outside of the ccache binary. Even though the check-manifest.py does not depend on the exact manifest binary format, it still depends on the general manifest structure. So I guess what I'm angling for is if you think that it would be OK or even better to just implement functionality similar to check-manifest.py in ccache itself as a debug option?

afbjorklund:
I have only used it for debugging purposes, not sure if it is useful for a user.
One of the programs seemed generic enough to make for an example...

afbjorklund:
Note that the check-manifest.py performs the same function as a normal ccache cache check.

Just that it might be easier to follow along in the python code with json, instead of the real thing ?

jrosdahl:
Closing this as I would prefer not duplicating logic outside the ccache code.

afbjorklund:
Plan was to use the same export format for the result file, using base64 encoding for the file contents...

I guess the `--dump-result` will just create files with the contents, rather than writing them on console ?

PR #385, opened 2019-04-21
Use the localtime_r function, where available

afbjorklund:
This will clear up the security alerts, from lgtm:

https://lgtm.com/projects/g/ccache/ccache/alerts


jrosdahl:
Thanks! However, this is a midair collision with 753ca2b9341b, so closing.

PR #389, opened 2019-04-22
The clang --analyze option is too hard (for now)

### LINKED ISSUE #388: Always invoke compile when static analysis option (`--analyze`) present
The clang toolset offers integrated static analysis via the `--analyze` command line option.  When this option is present, `ccache` should either ALWAYS invoke the compiler or should cache the static analysis results and report those back (IMHO it would be simpler and more correct to always invoke the compiler).

afbjorklund:
See https://clang-analyzer.llvm.org/

Closes #388

### DIFF
--- a/src/compopt.c
+++ b/src/compopt.c
@@ -30,6 +30,7 @@ struct compopt {
 };
 
 static const struct compopt compopts[] = {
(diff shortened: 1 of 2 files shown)


jrosdahl:
Thanks!

PR #391, opened 2019-04-23
Correctly handle @file on Windows

### LINKED ISSUE #95: Command too long when @file is used on Windows
From https://bugzilla.samba.org/show_bug.cgi?id=9875:

> Windows has a limitation of 8K characters for commands.
> 
> Since compilation flags for large projects might exceed this limit (e.g. it has many include paths), @file.opt is commonly used.
> 
> The problem is that ccache expands @file and executes the preprocessor with raw arguments. This fails for very long commands...

zmike:
the @file syntax means that the process reads command arguments from the
specified file. this is commonly used in order to shorten commands which
would otherwise be longer than the maximum length limit: many build systems
do this in all cases to avoid hitting this limit.

when @file is used on Windows, the original arguments must be preserved exactly
without any modification; this means that expansion can occur, but the original
args must be passed through in the command in order to continue using the
@file syntax

fixes #95

### DIFF
--- a/src/ccache.c
+++ b/src/ccache.c
@@ -104,6 +104,9 @@ char *current_working_dir = NULL;
 // The original argument list.
 static struct args *orig_args;
 
+// Whether to use @file on Windows
+static bool use_atfile;
+
 // The source file.
 static char *input_file;
 
@@ -1411,7 +1414,7 @@ to_cache(struct args *args, struct hash *depend_mode_hash)
 		tmp_stderr = format("%s.tmp.stderr", cached_obj);
 		tmp_stderr_fd = create_tmp_fd(&tmp_stderr);
 
-		status = execute(args->argv, tmp_stdout_fd, tmp_stderr_fd, &compiler_pid);
+		status = execute(args->argv, tmp_stdout_fd, tmp_stderr_fd, &compiler_pid, use_atfile);
 		args_pop(args, 3);
 	} else {
 		// The cached object path is not known yet, use temporary files.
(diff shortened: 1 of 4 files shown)


jrosdahl:
Thanks for working on this!

However, I'm afraid the solution is not that simple. From the discussion in https://bugzilla.samba.org/show_bug.cgi?id=9875: _"ccache needs to expand the file because it potentially needs to remove or rewrite arguments in the file, but ccache could very well construct a new temporary argument file on Windows when the command line size is larger than some threshold."_

zmike:
Hm so instead of using `orig_args` when `use_orig` is set, it would instead need to write `args` to a tmpfile and then do `@tmpfile` as args? This seems maybe doable, I'll look into it.

zmike:
I'll make another PR when I have this working.

jrosdahl:
> Hm so instead of using orig_args when use_orig is set, it would instead need to write args to a tmpfile and then do @tmpfile as args?

I imagine that a good solution would look something like this:

* Leave `cc_process_args` as is.
* Let the Windows version of `execute` check the command line size. If it is larger than some threshold (8000 bytes or whatever the limit is on Windows): create a temporary file, write all arguments to it and pass `@tmpfile` to the compiler instead.
* No need for a `use_atfile` boolean or for changing `execute`'s signature.

zmike:
I've been experimenting a bit now that I have `@file` syntax working more generally using tmpfiles. The issue with using "The Limit" is that this seems a bit irregular; the MSDN docs claim that it's 2^15, but I encountered the issue with a command that was only 24k characters long. I think using 2^13 (8192) might be a good breakpoint?

jrosdahl:
> I think using 2^13 (8192) might be a good breakpoint?

Since Orgad Shaneh mentioned "8K characters" in https://bugzilla.samba.org/show_bug.cgi?id=9875, 8192 bytes sounds good to me.

zmike:
Alrighty, this is good timing since I just finished some (excruciatingly long) test builds successfully.

PR #392, opened 2019-04-24
 correctly handle @file syntax on Windows

### LINKED ISSUE #95: Command too long when @file is used on Windows

zmike:
the @file syntax means that the process reads command arguments from the
specified file. this is commonly used in order to shorten commands which
would otherwise be longer than the maximum length limit: many build systems
do this in all cases to avoid hitting this limit.

when a command exceeds 8192 characters on on Windows, ccache now writes
the parsed/modified arguments to a tmpfile and then runs the command using
that tmpfile with @tmpfile in order to preserve this mechanism and avoid hitting
the length limit

fixes #95

### DIFF
--- a/src/ccache.h
+++ b/src/ccache.h
@@ -289,7 +289,7 @@ typedef int (*COMPAR_FN_T)(const void *, const void *);
 #endif
 
 #ifdef _WIN32
-char *win32argvtos(char *prefix, char **argv);
+char *win32argvtos(char *prefix, char **argv, int *length);
 char *win32getshell(char *path);
 int win32execute(char *path, char **argv, int doreturn,
                  int fd_stdout, int fd_stderr);
--- a/src/execute.c
+++ b/src/execute.c
@@ -169,9 +169,24 @@ win32execute(char *path, char **argv, int doreturn,
 	const char *ext = strrchr(path, '.');
 	char full_path_win_ext[MAX_PATH] = {0};
 	add_exe_ext_if_no_to_fullpath(full_path_win_ext, MAX_PATH, ext, path);
-	BOOL ret =
-		CreateProcess(full_path_win_ext, args, NULL, NULL, 1, 0, NULL, NULL,
-		              &si, &pi);
+	BOOL ret = FALSE;
+        if (length > 8192) {
+                char *tmp_file = format("%s.tmp", path);
+                FILE *fp = create_tmp_file(&tmp_file, "w");
+                char atfile[MAX_PATH + 3];
+                fwrite(args, 1, length - 1, fp);
+                fclose(fp);
+                if (ferror(fp)) {
+                        cc_log("Error writing @file; this command will probably fail:\n%s", args);
+                }
+                snprintf(atfile, sizeof(atfile), "\"@%s\"", tmp_file);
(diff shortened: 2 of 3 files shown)


jrosdahl:
Thanks!

jrosdahl:
Hm, when re-reading the code I see that I was a bit too quick to merge this. But what's done is done. :-)

I made some follow-up fixes in 93cefa21a5ab and cb492b98bea4. @zmike: Do they look OK to you?

Also, I don't quite understand the `length` parameter. The length is set to `k` in `win32argvtos`, but it looks like `k` represents how long the argument string at most can be (to know how much to overallocate), not the actual argument string size?

Do you think that the patch below would make sense?

```diff
diff --git a/src/execute.c b/src/execute.c
index af95a343cee6..3ee095346c46 100644
--- a/src/execute.c
+++ b/src/execute.c
@@ -30,7 +30,6 @@ win32argvtos(char *prefix, char **argv, int *length)
 {
 	int i = 0;
 	int k = 0;
-	*length = 0;
 	char *arg = prefix ? prefix : argv[i++];
 	do {
 		int bs = 0;
@@ -53,9 +52,9 @@ win32argvtos(char *prefix, char **argv, int *length)
 	char *ptr = malloc(k + 1);
 	char *str = ptr;
 	if (!str) {
+		*length = 0;
 		return NULL;
 	}
-	*length = k;
 
 	i = 0;
 	arg = prefix ? prefix : argv[i++];
@@ -85,6 +84,7 @@ win32argvtos(char *prefix, char **argv, int *length)
 	} while ((arg = argv[i++]));
 	ptr[-1] = '\0';
 
+	*length = ptr - str - 1;
 	return str;
 }
 
@@ -174,7 +174,7 @@ win32execute(char *path, char **argv, int doreturn,
 		char *tmp_file = format("%s.tmp", path);
 		FILE *fp = create_tmp_file(&tmp_file, "w");
 		char atfile[MAX_PATH + 3];
-		fwrite(args, 1, length - 1, fp);
+		fwrite(args, 1, length, fp);
 		fclose(fp);
 		if (ferror(fp)) {
 			cc_log("Error writing @file; this command will probably fail: %s", args);
```

zmike:
Whoops, I actually had the cleanup patch and the `length - 1` hunk done locally on my Windows test machine but that didn't make it back to my work machine with the git repo. The fwrite size change in particular is necessary since otherwise it writes a garbage character at the end of the file.

As for the arg length change, I was being a bit lazy with this; as I mentioned, the enforced length limit is a bit flaky across different systems regardless of what the docs say, so I figured that if the string could have been up to 8k characters long then probably it was fine to be safe and use `@file`.

I think these are good changes overall.

jrosdahl:
OK, I pushed it as e96d60be2a1c. I would appreciate any testing of it since I don't run the Windows parts myself.

zmike:
Sure, I'll kick off a build later today after I finish my current builds.

zmike:
No issues, seems good!

jrosdahl:
Thanks!

PR #395, opened 2019-04-25
Fix NULL argument in profile directory trace

hwti:
With -fprofile-*=path, we currently print :
"Setting profile directory to (null)".

This fixes the build with gcc 9, which failed with :
error: ‘%s’ directive argument is null [-Werror=format-overflow=]


jrosdahl:
Thanks!

Issue #397, opened 2019-04-28
'-MF /dev/null' fails due to file rename: ccache: error: Failed to create temporary file for /dev/null.tmp: Permission denied


trofi:
Minimal reproducer:
```
$ ccache /usr/bin/x86_64-pc-linux-gnu-gcc -O2 -pipe -c -MF /dev/null -MQ /dev/null -MMD -MP -x c /dev/null -o /dev/null
ccache: error: Failed to create temporary file for /dev/null.tmp: Permission denied

$ /usr/bin/x86_64-pc-linux-gnu-gcc -O2 -pipe -c -MF /dev/null -MQ /dev/null -MMD -MP -x c /dev/null -o /dev/null
<ok>
```

Originally observed when built `git-2.21.0` using `ccache-3.7`. `git` seems to probe `-MF` feature support this way:
https://github.com/git/git/blob/14c0f8d3ab6c36672189cd2dd217f4617d12ccba/Makefile#L1220-L1224

afbjorklund:
We have some checks for -o /dev/null, probably need to extend those for -MF /dev/null too

afbjorklund:
Even better would be to have some release tests for high-profile projects, such as kernel or git.

jrosdahl:
Thanks for the report.

Just a note on the reproducer: `ccache /usr/bin/gcc -O2 -pipe -c -MF /dev/null -MQ /dev/null -MMD -MP -x c /dev/null -o /dev/null` is not enough to trigger the problem with an empty ccache, but `ccache /usr/bin/gcc -O2 -pipe -c -MF /dev/null -MQ /dev/null -MMD -MP -x c /dev/null` (without `-o /dev/null`) is.

afbjorklund:
There might be a slight problem with the implementation, I think we are skipping the name of the output file when we hash the arguments ? That might mean that we _could_ get a cache hit, that is missing the dep ?

afbjorklund:
If that is indeed the case, then we should make sure to hash "/dev/null" (as a special case). Will add a test for it to verify, just something that I missed when I looked it through - similar to that logging and relpath...

PR #400, opened 2019-04-29
Make sure to cache the missing dep file

afbjorklund:
If we first run with -MF /dev/null, then we will get a
"cache file missing" error on the next run - avoid that.

Make sure to cache the special /dev/null value, so that
it differs from the normal (ignored) dep file name...

From #397

### DIFF
--- a/src/ccache.c
+++ b/src/ccache.c
@@ -2043,10 +2043,15 @@ calculate_object_hash(struct args *args, struct hash *hash, int direct_mode)
 				hash_delimiter(hash, "arg");
 				hash_string_buffer(hash, args->argv[i], 3);
 
-				bool separate_argument = (strlen(args->argv[i]) == 3);
-				if (separate_argument) {
-					// Next argument is dependency name, so skip it.
-					i++;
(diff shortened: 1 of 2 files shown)


jrosdahl:
Good catch! I suggest adding a `hash_delimiter` before hashing `output_dep`.

I've been thinking some about this, and an alternative solution to handling `-MF /dev/null` could be to just bail out with `STATS_UNSUPPORTED_OPTION` if `output_dep` is `/dev/null`. Here I'm assuming that there's no real point in caching stuff from a probe call like that. What do you think?

jrosdahl:
> an alternative solution to handling -MF /dev/null could be to just bail out with STATS_UNSUPPORTED_OPTION if output_dep is /dev/null

I went with ahead with this in 253301e7ac56.

jrosdahl:
I changed my mind: it's conceivable that somebody would like to add `-MF /dev/null` as part of a probe that is run more frequently than just once per build. I'll apply this instead. Thanks!

afbjorklund:
Sure, that is easer.

PR #403, opened 2019-05-01
Misc. source comment and documentation typos

luzpaz:
Found via `codespell -q 3 -L uint,thru -S ./src/zlib`


jrosdahl:
Thanks. I don't want the changes to `m4/feature_macros.m4` since it's just bundled third party code, but the rest seem OK.

PR #405, opened 2019-05-02
Update the correct stats file, for manifest

### LINKED ISSUE #402: Number of files according to stats goes out of sync after auto-cleanups
Hello. I am having long-term issues with ccache stats going out of sync compared to actual filesystem contents. I created a shell script that can be used to at least partially reproduce this issue.

Steps to reproduce the issue:

1. Create an empty directory
2. Copy http://atom-symbol.net/f/2019-05-01/ccache-test.sh to the directory
3. Run `ccache-test.sh`
4. Compare printed "files in cache" counts

Example `ccache -s` output:

    files in cache 87

Expected output:

    files in cache 103

There is also minor discrepancy in the "cleanups performed" field. With a single `ccache -s; ccache -c; ccache -s` it can change from 34 to 37.

ccache version 3.7
Filesystem: ext4

afbjorklund:
There are global variables used for updating stats,
and those were wrong for the manifest file update.

So make sure to reset those variables, before adding
the manifest (this means flushing the pending updates)

Closes #402

### DIFF
--- a/src/ccache.c
+++ b/src/ccache.c
@@ -236,6 +236,9 @@ static char *cpp_stderr;
 // belongs (<cache_dir>/<x>/stats).
 char *stats_file = NULL;
 
+// The stats_file to use for manifest
+static char *manifest_stats;
+
 // Whether the output is a precompiled header.
 bool output_is_precompiled_header = false;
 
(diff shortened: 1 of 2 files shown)


jrosdahl:
It would be nicer to introduce parameters to the `stats_*` functions and stop using global variables, but that refactoring can wait to another day.

jrosdahl:
Thanks!

PR #406, opened 2019-05-02
Allow using BLAKE2b checksum instead of bundled MD4

afbjorklund:
This is mostly for doing performance comparisons, before switching the default checksum.

As discussed, we might want to use more than just 128 bits (out of the available 512 bits).


afbjorklund:
Currently this is a compile-time flag. It could be made into a runtime option, if wanted (probably not)

afbjorklund:
As per discussion in #412 this should be made into runtime configuration, so that we can compare.

afbjorklund:
Made into runtime configuration, but still needs `--enable-blake2` to compile.

afbjorklund:
Now it is back to a compiletime configuration again, removed the `checksum` setting.

afbjorklund:
Removed the `total` counter, and increased the digest from 16 bytes to 20 bytes (out of the 64)

afbjorklund:
Closing this branch where it is optional, will do another PR where it is simply changed/replaced...

PR #407, opened 2019-05-02
Fix object size verification

### LINKED ISSUE #382: Object verification assumes that header sizes are stored in the manifest
Setup:
depend mode, no sloppiness enabled.

I stumbled on a cache miss and I was investigating why. Turns out, I have this header that has `__DATE__` in it. It is then included in some source file.  

Generated manifest for this source file, which includes this header, has incorrect field size in "File info entry": 25 bytes more than the real file. I assume that is macro expansion to a real date here:

https://github.com/ccache/ccache/blob/5a17c1a62dea3dc40d585430a17889a4afe3e217/src/hashutil.c#L106-L116

But the stored size in manifest comes not from the actual header file size, but from the number of bytes hashed, which is now different with the date expansion.

This logic now causes cache miss when checking the header for stat change, as number of bytes hashed is compared to raw file size:

https://github.com/ccache/ccache/blob/5a17c1a62dea3dc40d585430a17889a4afe3e217/src/manifest.c#L380-L382

Solution might be to store real file size, then do the macro expansion during hashing and then check the hashes, which should be equal provided the date is still the same.

pavsa:
Changed manifest format to save the actual file size along with hashed content size.
File size field in manifest updated to 64bits.
Manifest version set to 2.

Fixes #382.

### DIFF
--- a/src/manifest.c
+++ b/src/manifest.c
@@ -24,22 +24,23 @@
 
 // Sketchy specification of the manifest disk format:
 //
-// <magic>         magic number                        (4 bytes)
-// <version>       file format version                 (1 byte unsigned int)
-// <hash_size>     size of the hash fields (in bytes)  (1 byte unsigned int)
-// <reserved>      reserved for future use             (2 bytes)
+// <magic>                     magic number                        (4 bytes)
+// <version>                   file format version                 (1 byte unsigned int)
+// <hash_size>                 size of the hash fields (in bytes)  (1 byte unsigned int)
+// <reserved>                  reserved for future use             (2 bytes)
 // ----------------------------------------------------------------------------
-// <n>             number of include file paths        (4 bytes unsigned int)
-// <path_0>        path to include file                (NUL-terminated string,
+// <n>                         number of include file paths        (4 bytes unsigned int)
+// <path_0>                    path to include file                (NUL-terminated string,
 // ...                                                  at most 1024 bytes)
 // <path_n-1>
 // ----------------------------------------------------------------------------
-// <n>             number of include file hash entries (4 bytes unsigned int)
-// <index[0]>      index of include file path          (4 bytes unsigned int)
-// <hash[0]>       hash of include file                (<hash_size> bytes)
-// <size[0]>       size of include file                (4 bytes unsigned int)
-// <mtime[0]>      mtime of include file               (8 bytes signed int)
-// <ctime[0]>      ctime of include file               (8 bytes signed int)
+// <n>                         number of include file hash entries (4 bytes unsigned int)
+// <index[0]>                  index of include file path          (4 bytes unsigned int)
+// <hash[0]>                   hash of include file                (<hash_size> bytes)
+// <size[0]>                   size of include file                (8 bytes unsigned int)
+// <hashed_content_size[0]>    bytes passed through hash function  (4 bytes unsigned int)
+// <mtime[0]>                  mtime of include file               (8 bytes signed int)
+// <ctime[0]>                  ctime of include file               (8 bytes signed int)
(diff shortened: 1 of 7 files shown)


test/suites/depend.bash
```diff
@@ -294,6 +294,29 @@ EOF
     expect_stat 'files in cache' 9
 
     # -------------------------------------------------------------------------
+    TEST "__DATE__ in header file results in direct cache hit as the date remains the same"
```
jrosdahl: This functionality is not specific to the depend mode, so please move it to direct.bash instead.

src/manifest.c
```diff
@@ -767,7 +780,8 @@ manifest_dump(const char *manifest_path, FILE *stream)
 		hash = format_hash_as_string(mf->file_infos[i].hash, -1);
 		fprintf(stream, "    Hash: %s\n", hash);
 		free(hash);
-		fprintf(stream, "    Size: %u\n", mf->file_infos[i].size);
+		fprintf(stream, "    File size: %lu\n", mf->file_infos[i].size);
+        fprintf(stream, "    Hashed bytes: %u\n", mf->file_infos[i].hashed_content_size);
```
jrosdahl: Indentation is off here.

src/manifest.c
```diff
@@ -335,7 +339,8 @@ write_manifest(gzFile f, const struct manifest *mf)
 	for (uint32_t i = 0; i < mf->n_file_infos; i++) {
 		WRITE_INT(4, mf->file_infos[i].index);
 		WRITE_BYTES(mf->hash_size, mf->file_infos[i].hash);
-		WRITE_INT(4, mf->file_infos[i].size);
+		WRITE_INT(8, mf->file_infos[i].size);
+		WRITE_INT(4, mf->file_infos[i].hashed_content_size);
```
jrosdahl: The description of the disk format at the top of manifest.c needs to be updated with the new fields.

pavsa:
Looks 

> 
> 
> Please have a look at fixing the 32-bit build errors reported by Travis.

Thanks. Fixed now.

jrosdahl:
Thanks.

pavsa:
Btw, I did not switch to using stat64 for getting 64bit file sizes. 
We are only storing 32bit sizes in 64bit integer so far.

jrosdahl:
> Btw, I did not switch to using stat64() for getting 64bit file sizes. We are only storing 32bit sizes in 64bit integer so far.

On a 64-bit Linux system, the stat call works with 64-bit file sizes already (and `struct stat`'s `st_size` field is 64 bits, etc), so no need to change anything there. For 32-bit systems we could build with `-D_FILE_OFFSET_BITS=64` to enable 64-bit file sizes (but I'm not sure if we should bother). Don't know if something is needed for portability for other 64-bit systems, though. Do you?

jrosdahl:
Just a quick heads up: I decided to switch the order of file size and hashed size in the manifest in c23187eace89, so you might want to clear your local cache to avoid having to debug cache misses due to differences in the manifest content...

pavsa:
> 
> 
> > Btw, I did not switch to using stat64() for getting 64bit file sizes. We are only storing 32bit sizes in 64bit integer so far.
> 
> On a 64-bit Linux system, the stat call works with 64-bit file sizes already (and `struct stat`'s `st_size` field is 64 bits, etc), so no need to change anything there. For 32-bit systems we could build with `-D_FILE_OFFSET_BITS=64` to enable 64-bit file sizes (but I'm not sure if we should bother). Don't know if something is needed for portability for other 64-bit systems, though. Do you?

Oh yes, you're right. I think we should leave it as is then.

Issue #350, opened 2019-01-24
Large command lines in the debug log get interrupted in the middle


DrASK:
ccache version: 3.4.1; compiler is gcc; os is lunux/ubuntu

I'm writing a Python script to run through a 3.9GB logfile and it's encountering the following two lines (yes, they are long):

Line the first:
[...]

PR #409, opened 2019-05-05
Implement optional logging to syslog

afbjorklund:
The file logger has some issues with large log lines and many jobs,
so provide the "syslog" alternative when doing the ccache logging.

For #350

### DIFF
--- a/doc/MANUAL.adoc
+++ b/doc/MANUAL.adoc
@@ -465,6 +465,14 @@ might be incorrect.
     the specified file. This is useful for tracking down problems.
 +
 If set to "syslog", then ccache will log using syslog() instead of a file.
+Then you can use syslogd configuration to filter into a file, example:
++
+-------------------------------------------------------------------------------
+# log ccache to file
+:programname, isequal, "ccache"         /var/log/ccache
+# remove from syslog
+& ~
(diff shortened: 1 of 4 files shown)


doc/MANUAL.adoc
```diff
@@ -463,6 +463,16 @@ might be incorrect.
 
     If set to a file path, ccache will write information on what it is doing to
     the specified file. This is useful for tracking down problems.
++
+If set to "syslog", then ccache will log using syslog() instead of a file.
+Then you can use syslogd configuration to filter into a file, example:
```
jrosdahl: I think that it would be a good idea to mention /etc/rsyslog.conf and /etc/rsyslog.d here. Perhaps like this?

```
If set to *syslog', ccache will log using `syslog()` instead of to a file. If
you use rsyslogd, you can add something like this to `/etc/rsyslog.conf` or a
file in `/etc/rsyslog.d`:
```

afbjorklund: Not sure it was rsyslogd specific, but I suppose that the file was indeed intended for /etc/rsyslog.d

jrosdahl:
Thanks.

Issue #262, opened 2018-05-10
Environment variable or configuration option to store statistics in a different location


BenLubar:
Example use case: I have a build server that might be building multiple versions of a program simultaneously, or building a pull request in a sandbox with read-only access to ccache. It would be nice to be able to display statistics about cache hits and misses for each individual build.

afbjorklund:
I think this would be a very good feature, currently the log is the only way to get per-build information.
But the default log is way too verbose (more of a debugging feature), I even started on a "stats log"*...

* one that would _only_ output the "Result:".

afbjorklund:
Current workaround:
``` console
$ export CCACHE_LOGFILE=$PWD/.ccache.log
```

Display results:
``` console
$ egrep -o "Result: .*" .ccache.log  | sort | uniq -c
```

As mentioned the downside is that it generates a lot of output, which slows the build and fills the disk...
But at least you would be able to show statistics for each individual build, even accumulated if wanted.

afbjorklund:
One idea that I had previously, was to add a "stats-only" log file. This way it would not be as bloated as the verbose log, but not have the locking issues of the stats. Think I even had some old code doing this...

afbjorklund:
Here it was: https://github.com/ccache/ccache/compare/master...afbjorklund:statslog-3.1

afbjorklund:
`export CCACHE_STATSLOG=$PWD/.stats.log`

`ccache --log-stats`

PR #410, opened 2019-05-05
Add a statsd client, for sending stats over UDP

afbjorklund:
This will send stats to a local "statsd" server

```
{ counters: 
   { 'statsd.bad_lines_seen': 0,
     'statsd.packets_received': 1,
     'statsd.metrics_received': 1,
     'ccache.stats.called_for_preprocessing': 0,
     'ccache.stats.unsupported_compiler_option': 0,
     'ccache.stats.compiler_check_failed': 0,
     'ccache.stats.cache_miss': 1 },
  timers: {},
  gauges: { 'statsd.timestamp_lag': 0 },
  timer_data: {},
  counter_rates: 
   { 'statsd.bad_lines_seen': 0,
     'statsd.packets_received': 0.1,
     'statsd.metrics_received': 0.1,
     'ccache.stats.called_for_preprocessing': 0,
     'ccache.stats.unsupported_compiler_option': 0,
     'ccache.stats.compiler_check_failed': 0,
     'ccache.stats.cache_miss': 0.1 },
  sets: {},
  pctThreshold: [ 90 ] }
```

The local statsd will flush stats to e.g. [graphite](http://graphite.readthedocs.org/)

For details see https://github.com/statsd/statsd

For #262

### DIFF
--- a/LICENSE.adoc
+++ b/LICENSE.adoc
@@ -450,6 +450,37 @@ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 SOFTWARE.
 -------------------------------------------------------------------------------
 
+src/statsd-client.[hc]
+~~~~~~~~~~~~~~~~~~
+
+Trivial statsd client in C, implements the statsd protocol
+Downloaded from <https://github.com/romanbsd/statsd-c-client>.
+
+-------------------------------------------------------------------------------
+The MIT License (MIT)
+
+Copyright (c) 2012 Roman Shterenzon
+
+Permission is hereby granted, free of charge, to any person obtaining
+a copy of this software and associated documentation files (the
+"Software"), to deal in the Software without restriction, including
+without limitation the rights to use, copy, modify, merge, publish,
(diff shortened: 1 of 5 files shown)


jrosdahl:
Spontaneously, this does not seem very interesting to me. As I see it, it's not really a solution to #262, if that's what you mean with "for #262". I'd rather see a solution with a "stats log file" similar to what you mentioned in https://github.com/ccache/ccache/issues/262#issuecomment-412266841 but only containing the counter updates in a terse format. And then a ccache option for parsing and summarizing the file. If we want to be able to send stats to statsd or graphite or whatever, then there could be one or several separate programs that do so with the stats log file as input (or via a named pipe for "continuous updates").

Does this sound reasonable?

afbjorklund:
Sure, we can add the stats log, it's a separate (though stale) implementation.
https://github.com/ccache/ccache/compare/3.1-maint...afbjorklund:statslog-3.1

See also this article, for why UDP and some other background on metrics:
https://codeascraft.com/2011/02/15/measure-anything-measure-everything/

There was something weird with the stats file names, but I forgot what it was...
Can't really remember the reason _why_ I wrote the source file name to the log ?

PR #420, opened 2019-05-09
Fix make check with "funny" locales

llunak:
### Description ###

'ch' in cs_CZ locale is technically just one character, so directory
name 'ccache' fails the test

### DIFF
--- a/test/run
+++ b/test/run
@@ -273,6 +273,8 @@ EOF
 # =============================================================================
 # main program
 
+export LC_ALL=C
+
 if pwd | grep '[^A-Za-z0-9/.,=_%+-]' >/dev/null 2>&1; then
     cat <<EOF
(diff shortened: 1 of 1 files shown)


jrosdahl:
The change looks OK, but I don't understand why it's needed, i.e. what causes the test failure. So for my own understanding, could you explain what you mean with "'ch' in cs_CZ locale"? And which test fails and why?

llunak:
$ echo ccache | grep '[^A-Za-z0-9/.,=_%+-]'
cca**ch**e

In czech, "ch" is a digraph and is considered to be one letter. Which is most probably why the above case fails (although it happens only on openSUSE 15.0, on the previous 42.3 it's ok, so it may be also a bug, since technically 'ch' goes after 'h' and before 'i' in the alphabet, but I don't know).

Issue #413, opened 2019-05-05
 Switch from zlib to zstd


jrosdahl:
### Background

zstd is better than zlib in all aspects, and with negative compression levels it's as good as lz4 even for compression performance. The low overhead means that we can enable compression by default without negative consequences for the absolute majority of users.

### What to do

- Add zstd support.
  - ~Bundle libzstd for optional usage, just like we do with zlib.~ Let's try not bundling it.
- Remove zlib support.
  - **Rationale**:  Not keeping support for zlib means less dependencies and less complex code.

_**Edit:** Changed from lz4 to zstd according to discussion below._

jrosdahl:
Related old pull request with broader scope: #118.

afbjorklund:
This could probably also do with some benchmarks.

afbjorklund:
Not sure if a rebase is needed, seems that the old code still works "good enough" without it...
There's some strange things going on with latest LZ4 versions, so I've stayed on the previous:

LZ4 v1.9.1
LZ4 v1.9.0
> Warning : this version has a known bug in the decompression function which makes it read a few bytes beyond input limit. Upgrade to v1.9.1 is recommended.

LZ4 v1.8.3
LZ4 v1.8.2
> This is maintenance release, mainly triggered by a data corruption that can only occur in v1.8.2, at level 9 (only), for some "large enough" data blocks (> 64 KB)

**LZ4 v1.7.5**

The main reason for staying with gzip (-1) as default was that the difference was not "big enough",
adding a new compression meant invalidating the cache - and everyone has gzip, noone has lz4.

afbjorklund:
Rebased the old code: #421 (could be cleaned up further, but #118 had too much cruft in it now)

afbjorklund:
The performance seems to again vary a lot, so thankful for someone else testing it out as well...

afbjorklund:
Hmm, the above was **only** headers - and the stderr and manifest are still using gzip compression...

Seems like there might be some merit to the [later versions of LZ4](https://github.com/lz4/lz4/releases/tag/v1.8.2) and the special path for short files:

![plotcycles](https://user-images.githubusercontent.com/10364051/57486741-062d8300-72af-11e9-941f-af28335886f8.png)

afbjorklund:
If we want to do this for the manifest and result files, we need some kind of [gzFile](https://www.zlib.net/manual.html#Gzip) replacement...
Either work with a local memory buffer, or something implementing a similar stdio buffer interface.

[...]

PR #421, opened 2019-05-09
Allow using LZ4f compression instead of the default Gzip

afbjorklund:
This is mostly for doing performance comparisons, before switching the default compression.

It is a rebased version of #118, which had gotten stale (config and tests and python3 happened)


afbjorklund:
This should probably be squashed further, before merging. Then again, it might not be merged anyway...

afbjorklund:
Squashed the bug fixes into the original commit, makes the changes easier to read.

afbjorklund:
See #413

PR #424, opened 2019-05-11
clang's -ftime-trace is too hard

llunak:
It produces extra .json file with timing information about the compilation.

(https://aras-p.info/blog/2019/01/16/time-trace-timeline-flame-chart-profiler-for-Clang/)

### DIFF
--- a/src/compopt.c
+++ b/src/compopt.c
@@ -69,6 +69,7 @@ static const struct compopt compopts[] = {
 	{"-fno-working-directory", AFFECTS_CPP},
 	{"-fplugin=libcc1plugin", TOO_HARD}, // interaction with GDB
 	{"-frepo",          TOO_HARD},
+	{"-ftime-trace",    TOO_HARD}, // clang
 	{"-fworking-directory", AFFECTS_CPP},
 	{"-gtoggle",        TOO_HARD},
 	{"-idirafter",      AFFECTS_CPP | TAKES_ARG | TAKES_CONCAT_ARG | TAKES_PATH},


Trass3r:
Somehow this doesn't work for me in 3.7.7 even though it's clearly [there](https://github.com/ccache/ccache/blob/v3.7.7/src/compopt.c#L94), any ideas?
```
=== CCACHE 3.7.7 STARTED =========================================
Config: (default) cache_dir_levels = 2
Config: (default) compiler = 
Config: (default) compiler_check = mtime
Config: (default) compression = false
Config: (default) compression_level = 6
Config: (default) cpp_extension = 
Config: (environment) debug = true
Config: (default) depend_mode = false
Config: (default) direct_mode = true
Config: (default) disable = false
Config: (default) extra_files_to_hash = 
Config: (default) hard_link = false
Config: (default) hash_dir = true
Config: (default) ignore_headers_in_manifest = 
Config: (default) keep_comments_cpp = false
Config: (default) limit_multiple = 0.8
Config: (environment) log_file = /tmp/ccache.log
Config: (default) max_files = 0
Config: (/home/runner/work/llvm-project/llvm-pro
Config: (default) path = 
Config: (default) pch_external_checksum = false
Config: (default) prefix_command = 
Config: (default) prefix_command_cpp = 
Config: (default) read_only = false
Config: (default) read_only_direct = false
Config: (default) recache = false
Config: (default) run_second_cpp = true
Config: (default) sloppiness = 
Config: (default) stats = true
Config: (default) temporary_dir = 
Config: (default) umask = 
...
Compiler option -ftime-trace is unsupported
Failed; falling back to running the real compiler
```

Trass3r:
I guess there's no way to make it work together with ccache?
Even if it properly cached the json files ClangBuildAnalyzer might produce bogus numbers because of cold OS file cache effects?! And if it was just ignored you'd get non-meaningful ones.

llunak:
What would be the point? If -ftime-trace is used, it's wanted that the compiler is always run.

Trass3r:
For big projects it's just infeasible to build without something like ccache. Still getting these analyses would be nice.

PR #427, opened 2019-05-19
Fix PCH detection in depend mode (+test improvements)

llunak:
(I can split this if you dislike so many commits, but it mostly belongs together. I was looking for the PCH bug and had to extend the tests quite a bit before I finally found it.)

This fixes ccache failing to hash pch file in depend mode if that pch file is introduced only using -include. The rest is adding more PCH tests and removing the needless duplication between GCC and Clang.

### DIFF
--- a/doc/MANUAL.adoc
+++ b/doc/MANUAL.adoc
@@ -1058,10 +1058,12 @@ things to make it work properly:
 +
 --
 ** use the *-include* compiler option to include the precompiled header
-   (i.e., don't use *#include* in the source code to include the header); or
+   (i.e., don't use *#include* in the source code to include the header),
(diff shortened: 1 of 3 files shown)


jrosdahl:
Excellent, thanks!

llunak:
Thank you for merging. I'd like to add a configure check for the bug, meaning the ccache version that doesn't have it. Is that going to be 3.8?

jrosdahl:
> Thank you for merging. I'd like to add a configure check for the bug, meaning the ccache version that doesn't have it. Is that going to be 3.8?

Yes, unless some more severe bug pops up which motivates a 3.7.2 release – in that case your bug fix will be included there as well.

PR #408, opened 2019-05-02
Add aggregated file storing all the results

afbjorklund:
From #218

PR #428, opened 2019-05-26
Refactor: save tmp stderr until done storing

afbjorklund:
We might need it for the alternate codepaths

If not using a standard file store, it might
not be possible to move stderr instead of copy.

For those cases, use the temporary file instead.

---
_Broken out from #408, if not ready for all of it_
_This just restructures the code, for extending..._

_By doing it separately like this, ahead of time,_
_the other changes need to modify less things_

### DIFF
--- a/src/ccache.c
+++ b/src/ccache.c
@@ -1330,9 +1330,9 @@ copy_file_from_cache(const char *source, const char *dest)
 
 // Send cached stderr, if any, to stderr.
 static void
-send_cached_stderr(void)
+send_cached_stderr(const char *path_stderr)
 {
-	int fd_stderr = open(cached_stderr, O_RDONLY | O_BINARY);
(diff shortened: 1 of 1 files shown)


jrosdahl:
I took all of #408, so closing this one.

afbjorklund:
> I took all of #408, so closing this one.

Yeah, that is what I was afraid of (it ending up on `dev/aggregation`) and why I created this for `master`...

It would be easier if we could deliver refactorings separately from features, less "noise" in the diffs etc. ?

jrosdahl:
> Yeah, that is what I was afraid of (it ending up on dev/aggregation) and why I created this for master...

I don't understand why I deserved such a snarky comment. Let me try to reformulate it:

> "OK. However, I would still prefer it if this change could be delivered separately from the rest since it's purely refactoring."

Sure, let's do it that way then.

PR #436, opened 2019-06-16
Use Zstd compression (zlibWrapper) in addition to Zlib

afbjorklund:
This switches from zlib to zstd compression, using the zlibWrapper of libzstd.

The zlib option is left as a legacy compile option, otherwise zstd is **always** used.

### DIFF
--- a/src/result.c
+++ b/src/result.c
@@ -18,6 +18,10 @@
 #include "compression.h"
 #include "result.h"
 
+#ifdef HAVE_LIBZSTD
+#include "zstd_zlibwrapper.h"
+#endif
+
 // Result data format:
 //
 // <result>      ::= <header> <body> ; <body> is potentially compressed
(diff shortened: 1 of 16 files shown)


afbjorklund:
This could be used instead of LZ4 (#413)

See https://facebook.github.io/zstd/

shawnl:
Looks good.

 Is zlib linked as NEEDED when using the wrapper, or does zstd dlopen it?

src/result.c
```diff
@@ -407,21 +423,47 @@ bool result_put(const char *path, struct filelist *list, int compression_level)
...
-	} else {
+		break;
+
+	case COMPR_TYPE_ZLIB:
 		compressor = &compr_zlib;
+		break;
+
+#ifdef HAVE_LIBZSTD
+	case COMPR_TYPE_ZSTD:
+		ZWRAP_useZSTDcompression(1);
```
shawnl: You need to support variable compression levels here.

afbjorklund:
> Is zlib linked as NEEDED when using the wrapper, or does zstd dlopen it?

We could drop it later on, but for now it is always included:

`src/zstd/libzstd_zlibwrapper.a -lzstd -lm  -lz`

src/result.c
afbjorklund: The levels are still handled by the "zlib" compressor.

It is unfortunate that they have different ranges, but.

jrosdahl:
> This switches from zlib to zstd compression, using the zlibWrapper of libzstd.

Please consider adding a zstd implementation of the compression API instead of using the zlib wrapper. I want zlib to be removed.

Also, I want to be able to use negative zstd levels.

afbjorklund:
> Please consider adding a zstd implementation of the compression API instead of using the zlib wrapper. I want zlib to be removed.

That is probably a bigger project, will not handle that in this PR.

afbjorklund:
> Also, I want to be able to use negative zstd levels, so that's one thing to keep in mind.

Pretty sure that I had added support for negative levels (for another reason, but anyway)

PR #437, opened 2019-06-16
Allow using zstd compresssion instead of zlib

afbjorklund:
This is the version without the zlibWrapper, using native zstd as a compression implementation.

See #413 (and #436)

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -37,11 +37,13 @@ non_3pp_sources = \
     src/compr_none.c \
     src/compr_zlib.c \
     src/compression.c \
+    src/compr_zstd.c \
     src/conf.c \
     src/confitems.c \
     src/counters.c \
     src/decompr_none.c \
     src/decompr_zlib.c \
+    src/decompr_zstd.c \
     src/execute.c \
     src/exitfn.c \
     src/hash.c \
--- a/configure.ac
+++ b/configure.ac
@@ -163,6 +163,15 @@ else
     LIBS="$LIBS -lz"
 fi
 
+AC_ARG_ENABLE(zstd,
+  [AS_HELP_STRING([--enable-zstd],
+    [use zstd compression instead of zlib])])
+if test x${enable_zstd} = xyes; then
+    CPPFLAGS="$CPPFLAGS -DUSE_ZSTD"
+    AC_CHECK_HEADERS(zstd.h)
+    AC_CHECK_LIB(zstd, ZSTD_versionString)
+fi
+
 AC_ARG_ENABLE(man,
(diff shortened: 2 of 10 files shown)


afbjorklund:
Can't win, the clang analyzer complains about unreachable code but gcc complains about end reached.

shawnl:
@afbjorklund you have to add a default case.

afbjorklund:
There was no default case before ? Anyway, there is no default (left, all enums are used).

 But I suppose I could try adding one, just to keep the analyzer happy. "Painting it green"

afbjorklund:
Nope, clang didn't like: `error: default label in switch which covers all`

jrosdahl:
Nice! The build error will disappear if you rebase your branch to latest master.

One thing I noticed when skimming the libzstd API is that the checksum does not seem to be enabled by default. Indeed, after some tries I managed to flip a bit in the result file without libzstd noticing. And if I understand correctly, the function used for enable checksumming is one of the "advanced and experimental functions", unlocked with `-DZSTD_STATIC_LINKING_ONLY`, which as the name implies is not supposed to be used with a dynamic library. So I guess we should to link statically until the API becomes non-experimental? I found some related discussion here: https://github.com/facebook/zstd/issues/1111.

@shawnl: Since you seem to have libzstd experience, any insights here?

shawnl:
@jrosdahl If we ask they will probably declare what we want stable, so we don't need to use that macro, AND can this change would be retroactive (with the macro when we need it, as long as the symbol is there, I know Debian/Ubuntu builds will all the symbols visible), and then we add a configure test.

afbjorklund:
> One thing I noticed when skimming the libzstd API is that the checksum does not seem to be enabled by default. Indeed, after some tries I managed to flip a bit in the result file without libzstd noticing.

This (and the file magic bytes) was the reason that I used the "frame" version of some of the other compressions, such as LZ4 (i.e. lz4f vs lz4), rather than a more simple "block" or "stream" format.

Since zstd is so new, it will probably take them a while to figure these practical things out ? After all, it took quite some time (years) before LZMA/xz and LZ4/lz4f got into their stable state of today.

afbjorklund:
Note when testing this, that you need `--enable-zstd` in order to use zstd (otherwise it uses zlib)

jrosdahl:
Thanks!

Regarding the checksum, I would like to avoid static linking, and I also would like to avoid depending on experimental APIs even if they might be non-experimental in the future. Instead, my plan is to simply perform checksumming myself (using XXH64, just like zstd does). This has the added benefit of checksumming non-compressed data as well.

PR #440, opened 2019-07-04
Add command to show compression statistics

afbjorklund:
This will only show information about the files that is knows about
(right magic bytes). So the file count might differ from what is shown
with the regular statistics (which shows all files, including old ones).

The terminology used here is a bit confused, the compression ratio is
supposed to grow upwards. Sometimes known as "space savings" instead,
so list both values (ratio and savings) to make the output more obvious.

### DIFF
--- a/src/ccache.c
+++ b/src/ccache.c
@@ -75,6 +75,7 @@ static const char USAGE_TEXT[] =
 	"                              human-readable format\n"
 	"    -s, --show-stats          show summary of configuration and statistics\n"
 	"                              counters in human-readable format\n"
+	"    -x, --show-compression    show compression statistics\n"
 	"    -z, --zero-stats          zero statistics counters\n"
 	"\n"
 	"    -h, --help                print this help text\n"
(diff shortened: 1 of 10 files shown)


src/ccache.c
```diff
@@ -75,6 +75,7 @@ static const char USAGE_TEXT[] =
 	"                              human-readable format\n"
 	"    -s, --show-stats          show summary of configuration and statistics\n"
 	"                              counters in human-readable format\n"
+	"    -x, --comp-stats          show compression statistics\n"
```
jrosdahl: Most long options begin with a verb, so maybe rename to `--show-statistics`?

src/compress.c
```diff
@@ -0,0 +1,107 @@
...
+	}
+
+	if (strstr(p, "CACHEDIR.TAG")) {
+		goto out;
+	}
+
+	size_t uncompressed_size;
+	bool is_compressed;
+	if (str_endswith(p, ".manifest")) {
+		uncompressed_size = manifest_size(fname, &is_compressed);
```
jrosdahl: I now realize that I should have mentioned that there is a better way of doing compression statistics and recompression now. Sorry.

The file formats and "common header" functions have already been prepared to make this use case (compression statistics and recompression) efficient, see 5905237ebb917bcb1bb773431f8e8703899194cb and 318f980eb9e7424ccd166ee3de795beb0d355984 (also pushed 7839217088c7 to master just now). Since there is a content size field in the header, there's no need to dry-run decompression to calculate the content size.

So for compression inspection something like this can be done:

```c
FILE *f = fopen(path, "r");
struct common_header header;
common_header_initialize_for_reading(
	&header,
	f,
	magic, // MANIFEST_MAGIC or RESULT_MAGIC,
	version, // MANIFEST_VERSION or RESULT_VERSION,
	NULL,
	NULL,
	NULL);
// Use header.content_size for uncompressed size and stat(path, ...) for actual size.
fclose(f);
```

For recompression:

1. Use `common_header_initialize_for_reading` to read the header but with non-NULL `decompressor` and `decompressor_state` arguments.
2. If `header.compression_type` and `header.compression_level` are what we want, we're done. Otherwise:
3. Open a new file and use `common_header_initialize_for_writing`.
4. Feed data from `decompressor.read` to `compressor.write`. No need to know anything about the actual body format.

src/ccache.c
jrosdahl: Fixed typo in the above comment, I meant `--show-compression` like suggested in https://github.com/ccache/ccache/pull/275#issuecomment-508140412.

afbjorklund: Done

src/compress.c
afbjorklund: Done, recompression is left to do (not in this PR)

jrosdahl:
Thanks.

PR #443, opened 2019-07-21
Get rid of gperf postprocessing

ivan-volnov:
A trying to make things easier. Gperf postprocessing isn't necessary.


afbjorklund:
> Gperf postprocessing isn't necessary.

Did you try with gperf < 3.1 (as it says) ?

ivan-volnov:
> Did you try with gperf < 3.1 (as it says) ?

Yes, I tried it with 3.0.3 and 3.0.4

jrosdahl:
Thanks for the suggestion! I appreciate making things easier.

In this case, however, removing the mechanism for adding item numbers makes it harder for the developer, not easier. It was deliberately added recently: 3ff1d5a38412c76d08be5f1d1d1982be2ff34760

Instead I think that we should just get rid of the gperf altogether. It's a cool concept, but it's apparent that it's too much of a hassle for little gain.

I'm currently working on C++-ifying the source code (not rewriting, only compiling the current code as C++ and making it possible to use C++ for new code and to slowly convert old code). I plan to address the lookup tables as part of that. Stay tuned.

@ivan-volnov: By the way, maybe you would like to join the [ccache gitter room](https://gitter.im/ccache/ccache) to discuss ccache development informally?

ivan-volnov:
> In this case, however, removing the mechanism for adding item numbers makes it harder for the developer, not easier. It was deliberately added recently:

Oh, I see now. But I suppose it really adds more harm than benefit. Consider this fix to be applied before removing gperf completely.

>Instead I think that we should just get rid of the gperf altogether. It's a cool concept, but it's apparent that it's too much of a hassle for little gain.

I couldn't agree more.

> I'm currently working on C++-ifying the source code...

That would be great!

> By the way, maybe you would like to..

Yes, sure.

afbjorklund:
> In this case, however, removing the mechanism for adding item numbers makes it harder for the developer, not easier

Right, I do remember those merges from hell. Working on several features, you have to renumber...

afbjorklund:
> I'm currently working on C++-ifying the source code (not rewriting, only compiling the current code as C++ and making it possible to use C++ for new code and to slowly convert old code).

Sounds like the worst of both worlds, but maybe there is some hidden benefit to compiling C as C++

jrosdahl:
@ivan-volnov wrote:
> But I suppose it really adds more harm than benefit. Consider this fix to be applied before removing gperf completely.

Not sure I understand. What's the harm more specifically?

@afbjorklund wrote:
> Right, I do remember those merges from hell. Working on several features, you have to renumber...

Exactly, 3ff1d5a was made for you. :slightly_smiling_face:

@afbjorklund wrote:
> Sounds like the worst of both worlds, but maybe there is some hidden benefit to compiling C as C++

Here's why: C++-ifying in the sense converting *all* C code to C++ idioms in one go is just not feasible. To be able to convert little by little, I want to be able to call new C++ code from old C code, and that won't work if the old code is compiled as C (unless exposing extern "C" interfaces for all new C++ code, but that way lies madness).

jrosdahl:
This is obsolete as of 728acfac984f so closing.

Issue #442, opened 2019-07-19
CCACHE_MAXSIZE does not handle filesystems with transparent compression


trofi:
It's an unstream version of two Gentoo downstream bugs/patches:
- https://bugs.gentoo.org/456178, patch: https://gitweb.gentoo.org/repo/gentoo.git/tree/dev-util/ccache/files/ccache-3.4-size-on-disk.patch?id=5e1fcb74fa533737d02b61738bcdeed84be5c308
- https://bugs.gentoo.org/649440, patch: https://gitweb.gentoo.org/repo/gentoo.git/tree/dev-util/ccache/files/ccache-3.6-disable-sized-cleanup.patch?id=5e1fcb74fa533737d02b61738bcdeed84be5c308

The problem:

Some filesystems (btrfs and zfs) support transparent compression. Users expect `CCACHE_MAXSIZE` to control roughly on-disk filesystem usage (compressed file blocks), not uncompressed size. But ccache effectively measures uncompressed size with some old code: https://github.com/ccache/ccache/blob/master/src/util.c#L968-L981

```c
// Return size on disk of a file.
size_t
file_size(struct stat *st)
{
	size_t size = st->st_blocks * 512;
	if ((size_t)st->st_size > size) {
		// Probably a broken stat() call...
		size = (st->st_size + 1023) & ~1023;
	}
	return size;
}
```

Perhaps we can just use `st->st_blocks * 512` nowadays. That is what first patch does.

Unfortunately that introduces instability into ccache's test suite if ran on such a filesystem: cleanup happens (or does not happen) depending on how well compressed content is. https://github.com/ccache/ccache/blob/master/test/suites/cleanup.bash#L67-L91

We can carry test workaround downstream as-is but would be nice to have an upstream fix for that as well (or a knob to disable such tests).

Thank you!

jrosdahl:
> Perhaps we can just use st->st_blocks * 512 nowadays. That is what first patch does.

Right. The check was added 17 years ago in 2cf82676eb72e, presumably to work around some broken legacy system. I agree that we can just remove it now.

> We can carry test workaround downstream as-is but would be nice to have an upstream fix for that as well (or a knob to disable such tests).

I would prefer not to remove it but to add some kind of knob.

jrosdahl:
Fixed by #444.

PR #444, opened 2019-07-21
Fix CCACHE_MAXSIZE with filesystem compression

ivan-volnov:
Just created a PR for https://github.com/ccache/ccache/issues/442

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,7 @@ matrix:
   # Job 1: Linux native GCC
   - os: linux
     compiler: gcc
-    env: V=1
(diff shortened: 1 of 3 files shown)


jrosdahl:
Thanks.

As noted in https://github.com/ccache/ccache/issues/442#issuecomment-513557115, I think that the first part is OK but removing the test case is not.

Here's my suggestion: Only run the test case in question if some environment variable is set, say `ENABLE_CACHE_CLEANUP_TESTS`. Then set that variable in the Travis-CI environment (see `.travis.yml`). What do you think?

jrosdahl:
Thanks.

Issue #412, opened 2019-05-05
Switch from MD4 to BLAKE2b and bump manifest on-disk format


jrosdahl:
### Background

MD4 is old technology that works very well for ccache's purposes as far as we know. However, the much more modern BLAKE2b algorithm has ~more or less exactly the same performance as~ ≈25% worse performance than MD4 (at least for large files on 64-bit systems) and has much better hash characteristics.

### What to do

- Switch from MD4 to BLAKE2b.
  - Determine a suitable digest size. Maybe 160 or 192 bits would be a good sweet spot? (Using a very large digest size leads to unnecessarily long key sizes in file names and internal data structures.)
   - ~Bundle libb2 for optional usage, just like we do with zlib.~ Do like what's done for libzstd instead.
- Don't include the size of hashed data in keys anymore.
- ~Bump manifest version to 3.~ No longer needed since we no longer strive to have a `--dump-manifest` that can understand all different versions. Keep version 2 as is on master; the latest released version of ccache (3.7.2 at the time of writing) has manifest version 1.

jrosdahl:
@pavsa wrote in https://github.com/ccache/ccache/issues/404#issuecomment-489449909:

> But is faster though?

The reason for switching was not to make it faster, only safer. I thought that the performance was very similar. But...

> I've done a simple benchmark of 1GB of data: [...]

When I benchmarked MD4 vs BLAKE2b previously, I got very similar performance results for the two algorithms, but I can't reproduce it now so I must have made some mistake. And I think I know what it was: I used an md4sum program I had laying around based on ccache's code. But it was compiled a long time ago, and compilers have probably advanced a bit since then...

When comparing `ccache --hash-file` from master with the same command with ccache compiled from #406, I get a slowdown as well: libb2's BLAKE2b is around 50% slower than ccache's MD4 implementation on my system. That's a bit disappointing. :slightly_frowning_face:

jrosdahl:
@afbjorklund has made a prototype of this in #406.

pavsa:
libb2 (if compiled with --enable-fat) runtime selection of optimized code. 
I have retested it (on a different machine now, that is has older cpu) with that enabled and also tested md5+sha1 from openssl:

blake2b: time took 1.4s   758.5 MB/s
MD4 time took 0.9s   1150.6 MB/s
MD5 time took 1.5s   673.7 MB/s
SHA1 time took 1.4s   758.5 MB/s

So I think we will have a hard time finding anything faster than md4.

[...]

PR #445, opened 2019-07-21
Use BLAKE2b checksum instead of bundled MD4

afbjorklund:
This is one third of #412

* Needs libb2 to be installed

* Removing mdfour is TODO

Based on previous ~#406~

### DIFF
--- a/.travis.sh
+++ b/.travis.sh
@@ -3,8 +3,8 @@
 make clean
 make travis CC=gcc
 make travis CC=clang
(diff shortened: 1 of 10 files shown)


afbjorklund:
Apparently it is **not** possible to compile libb2, with the clang flags that ccache has.

And there are some issues with cross-compiling, as the CPU detection doesn't work...

afbjorklund:
Think I would much rather have libb2 as a requirement, then mess around with this...

jrosdahl:
Thanks.

PR #448, opened 2019-07-28
Fix gperf generation when running on MSYS2

afbjorklund:
Apparently gperf doesn't like DOS linefeeds

As observed when testing PR #170


jrosdahl:
Thanks.

I'm working on removing gperf completely, but I have less spare time than expected right now so this fix makes sense untill I have landed my changes.

PR #457, opened 2019-08-30
Handle -MF=file.d as understood by the Green Hills compilers

Shionigami:
### Description ###

Remove '=' from destination path.
In case compiler option contains an '=' char, remove it from the dest path.

Fix: Removes "=" char from destination path if compiler option contains "="
Example: cxrh850.exe -MD -MF=<path_to_file>
Before fix: dest path: **=**<path_to_file>
After fix: dest path: <path_to_file>

### DIFF
--- a/src/args.c
+++ b/src/args.c
@@ -24,11 +24,9 @@ args_init(int init_argc, char **init_args)
 	args->argc = 0;
 	args->argv = (char **)x_malloc(sizeof(char *));
 	args->argv[0] = NULL;
-
-    for (int i = 0; i < init_argc; i++) {
-        args_add(args, init_args[i]);
-    }
-
+	for (int i = 0; i < init_argc; i++) {
+		args_add(args, init_args[i]);
+	}
 	return args;
 }
 
--- a/src/args.c
+++ b/src/args.c
@@ -24,9 +24,20 @@ args_init(int init_argc, char **init_args)
 	args->argc = 0;
 	args->argv = (char **)x_malloc(sizeof(char *));
 	args->argv[0] = NULL;
-	for (int i = 0; i < init_argc; i++) {
-		args_add(args, init_args[i]);
-	}
+
+    for (int i = 0; i < init_argc; i++) {
+        if(strstr(init_args[i], "-MF") == NULL) {
+            args_add(args, init_args[i]);
+        } else {
+            char *token;
+            token = strtok(init_args[i], "=");
+            while( token != NULL ) {
+                args_add(args, token);
+                token = strtok(NULL, "=");
+            }
+        }
+    }
+
 	return args;
 }
 
--- a/src/args.c
+++ b/src/args.c
@@ -26,16 +26,7 @@ args_init(int init_argc, char **init_args)
 	args->argv[0] = NULL;
 
     for (int i = 0; i < init_argc; i++) {
-        if(strstr(init_args[i], "-MF") == NULL) {
-            args_add(args, init_args[i]);
-        } else {
-            char *token;
-            token = strtok(init_args[i], "=");
-            while( token != NULL ) {
-                args_add(args, token);
-                token = strtok(NULL, "=");
-            }
-        }
+        args_add(args, init_args[i]);
     }
 
 	return args;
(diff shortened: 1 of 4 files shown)


jrosdahl:
I don't understand what problem your suggested change aims to solve. Please describe in more detail (give an example) what you think is the problem that should be fixed. Thanks!

Shionigami:
In some cases, like with GreenHills compiler, there are some options like -MF=<path>. Then, it occurs that the destination path still contains the "=" character.
Example:
  cxrh850.exe -MD -MF=<path to file>
  Destination folder, before fix: **=**<path to file>
  After fix: <path to file>

jrosdahl:
> In some cases, like with GreenHills compiler, there are some options like -MF=. Then, it occurs that the destination path still contains the "=" character.

OK, I see – you had that information in the original description but is was inside comments (`<!-- ... -->`) so it was hidden. I've edited the PR description and improved the PR title now.

Making `-MF=file.d` mean "the dependency file is named `file.d`" to fix the Green Hills case would break the GCC/Clang case if somebody for some reason uses the `-MF=file.d` form (the dependency file will then be named `=file.d` which is perfectly OK). But maybe that behavior be OK in practice since it's unlikely that GCC/Clang users use that form.

src/args.c
```diff
@@ -24,9 +24,20 @@ args_init(int init_argc, char **init_args)
 	args->argc = 0;
 	args->argv = (char **)x_malloc(sizeof(char *));
 	args->argv[0] = NULL;
-	for (int i = 0; i < init_argc; i++) {
-		args_add(args, init_args[i]);
-	}
+
+    for (int i = 0; i < init_argc; i++) {
+        if(strstr(init_args[i], "-MF") == NULL) {
```
jrosdahl: This is not the right place for parsing specific options – the `args_*` functions are generic and should stay that way. Please move the logic to the place in `cc_process_args` where `-MF` is handled.

src/ccache.c
```diff
@@ -785,13 +785,25 @@ print_included_files(FILE *fp)
 static char *
 make_relative_path(char *path)
 {
+    // if(path[0] == '=') { // Remove first char if string starts with "="
```
jrosdahl: Please remove leftover commented code like this.

src/ccache.c
```diff
@@ -785,13 +785,25 @@ print_included_files(FILE *fp)
...
+    // }
+
 	if (str_eq(conf->base_dir, "") || !str_startswith(path, conf->base_dir)) {
 		return path;
 	}
 
 #ifdef _WIN32
 	if (path[0] == '/') {
-		path++;  // Skip leading slash.
+		char *p = NULL;
```
jrosdahl: It looks like you have managed to include the content of 93322f0acf49 in the PR. Please rebase against latest 3.7-maint.

src/util.c
```diff
@@ -1159,7 +1160,15 @@ x_realpath(const char *path)
 	p = realpath(path, ret);
 #elif defined(_WIN32)
 	if (path[0] == '/') {
-		path++;  // Skip leading slash.
+		char *p = NULL;
```
jrosdahl: This looks like a fix that is unrelated to the issue of `-MF`? Please create a separate pull request for this instead.

src/util.c
```diff
@@ -1159,7 +1160,15 @@ x_realpath(const char *path)
...
+		if (isalpha(path[1]) && path[2] == '/') {
+			// Transform /c/path... to c:/path...
+			p = format("%c:/%s", path[1], &path[3]);
+		} else {
+			p = x_strdup(path+1); // Skip leading slash.
+		}
+		free(path);
```
jrosdahl: As you maybe have seen from the failing CI test, you can't free path here since it's not owned by the function.

src/util.c
```diff
@@ -464,6 +463,7 @@ copy_file(const char *src, const char *dest, int compress_level)
 int
 move_file(const char *src, const char *dest, int compress_level)
 {
+    cc_log("move_file() -> copy_file()");
```
jrosdahl: Please remove debug traces like this.

Issue #453, opened 2019-08-19
Report effectiveness percentage by accounting time taken


socketpair:
### Description ###
I mean calculation of percentage not by number of hit/miss but by user+system time.
In my scenario, it seems, small .c files are cached, and big ones are not. So, I see good percentage in `ccache -s`, but actually my build seems not speed up significantly.

afbjorklund:
This is an interesting idea! We have something similar in the `misc/performance` script, where it will run a build several times (with different settings) and report the speedup (or slowdown):

```
Without ccache:                              20.94 s (100.00 %) ( 1.00 x)
With ccache, preprocessor mode, cache miss:  21.85 s (104.34 %) ( 0.96 x)
With ccache, preprocessor mode, cache hit:    0.95 s (  4.53 %) (22.06 x)
With ccache, direct mode, cache miss:        21.90 s (104.60 %) ( 0.96 x)
With ccache, direct mode, cache hit:          0.15 s (  0.72 %) (139.34 x)
With ccache, depend mode, cache miss:        22.42 s (107.07 %) ( 0.93 x)
With ccache, depend mode, cache hit:          0.12 s (  0.57 %) (174.49 x)
```

This python script only shows clock time though, and not user/system time from time/times. For a concurrent build, it would be more interesting to show kernel times - like when using `time(2)`

```
0.70user 0.01system 0:00.72elapsed 99%CPU (0avgtext+0avgdata 50680maxresident)k
0inputs+2296outputs (0major+12359minor)pagefaults 0swaps
```

----

For it to work though, there are some things that needed to be added to the cache:
1) For each compiled file (hash result), there needs to be a file with the times (or rusage)
2) For the entire cache, there needs to be some aggregated counters about the total

But it is definitely doable, one would use the time without children for the cache lookup/hit.
And then use the time **with** children for the cache miss/compile, right after the `wait`.


However, you _should_ look into why compiling your big files is failing to be cached by ccache.

afbjorklund:
Did some proof-of-concept code in 2352161ca149e22694286fdd42166b78da9a0182

Using the POSIX format (`time -p`), which should be "portable" enough for storing as text:

```
real 3.14
user 3.01
sys 0.13
```

The structures are in microseconds, but on Linux the resolution seems to be clocks (1/100s)

```C
           struct timeval {
               time_t      tv_sec;     /* seconds */
               suseconds_t tv_usec;    /* microseconds */
           };
```

[...]

PR #461, opened 2019-09-02
Timing

afbjorklund:
### Description ###

Implementation of a compile time cache (with counters) for ccache, as suggested in #453
It will store the compile time(1) taken for each miss, in order to calculate the time saved later.

```
real 3.054660
user 2.948000
sys 0.096000
```

This enables ccache to show the cache efficiency and the performance speedup/slowdown.
It also counts the total amount of time (ms) used by the program, divided into: real, user, sys

```
cache hit (direct)                     1
cache miss                             1
cache hit rate                     50.00 %
time (real)                         3.34 s
time (user)                         3.10 s
time (sys)                          0.02 s
cpu utilization                    93.41 %
time (cache)                        0.17 s
time (compile)                      2.95 s
time (saved)                        3.04 s
cache efficiency                   49.42 %
performance                         1.98 x
```

The format is similar to the output with `time -p`, only with increased resolution (microseconds).

Time counters are stored using milliseconds, since ccache uses integers rather than floating point.

----

Conversion to C++ left as an exercise for the reader.

Thus the comparison is for `3.7-maint`, not `master`.


jrosdahl:
@afbjorklund: What would you like me to do with this PR?

afbjorklund:
My plan was to leave it up for review (it was originally just a PoC), and if it looks good we can convert it to c++ ? (I’m guessing 3.7.x doesn’t take any new features at this point)

jrosdahl:
> My plan was to leave it up for review (it was originally just a PoC), and if it looks good we can convert it to c++ ? (I’m guessing 3.7.x doesn’t take any new features at this point)

OK, thanks the clarification. I was a bit unsure since, as you say, 3.7-maint is in bug-fix-only mode so I was a bit surprised to see it as the target branch. Why not make the PoC based on master instead? Porting it to master won't primarily be about converting to C++ but to think about how to add the counters to the result files.

I can do a high-level review but I don't think that it makes sense to review the actual code changes since they will look quite different based on master. Sounds reasonable?

jrosdahl:
Some thoughts after a brief look at the code:

1. Storing and showing the real time doesn't sound like a good idea to me since it necessarily depends on other activity on the machine at the time of the cache miss, and that likely won't be representative of the situation on a cache hit.

2. Would it make sense to combine user and sys time to one stored measurement, or is there much value in keeping them separate?

3. There's an elephant in the room: the cost of automatic cleanup won't (and, as I see it, can't in a good way) be included in the measurements. Is it OK enough for the statistics to "lie" about it by just ignoring it?

4. Adding nine new rows to the `ccache -s` output feels a bit excessive – it's already quite long when a number of error counters start to show up. Would it be possible to condense the information to one or two lines in some way? Or, maybe hide it behind another new status option (`ccache --show-extended-stats`, `ccache --show-stats --verbose`)?

5. Would it be possible to find some more descriptive configuration name than `time`? Or perhaps just make it always true since the overhead of the sys calls is very small? When porting it to result files there won't be any I/O cost to speak of.

afbjorklund:
Storing and showing the real time is not necessary (for showing the efficiency), it is used to calculate a _different_ metric: how much of the time was actually spent on CPU (and indirectly how much was wait)

As per time(1):
```
              U      Total number of CPU-seconds that the process used directly (in user mode), in seconds.
              S      Total number of CPU-seconds used by the system on behalf of the process (in kernel mode), in seconds.
              E      Elapsed real (wall clock) time used by the process, in [hours:]minutes:seconds.
              P      Percentage of the CPU that this job got.  This is just user + system times divided by the total running time.  It also prints a percentage sign.
```

Also, keeping those 3 counters for the _total_ time spent in ccache is also not required for the initial feature. So maybe those additions should have been kept separate, but all the data was already being collected.

Leaving that feature out, and hiding the counters, the total output would be only two lines:

```
cache hit (direct)                   431
cache hit (preprocessed)              68
cache miss                           502
cache hit rate                     49.85 %
cache efficiency                   55.18 %
performance                         2.37 x
```

But then it would be slightly more "magic" how those numbers were calculated, unlike hit rate ?

----

> Would it make sense to combine user and sys time to one stored measurement, or is there much value in keeping them separate?

They are always used together, so it is mostly useful for _other_ purposes - such as seeing how much "overhead" is generated ("sys" time) and how much is actually spent on compilation ("user" time)

As described in the issue the idea was to keep the same format as the `time -p` output, although then I modified that to include more accuracy on the timing - i.e. microseconds instead of clocks (1/100)

> There's an elephant in the room: the cost of automatic cleanup won't (and, as I see it, can't in a good way) be included in the measurements. Is it OK enough for the statistics to "lie" about it by just ignoring it?

The automatic cleanup is more like a "garbage collect", it probably won't affect individual compiles ? But you are right that it would be "fair" to include it in the time spent by ccache, i.e. the global counters.

I haven't actually measure the time needed for the `gettimeofday` and `getrusage`, but I doubt the cost is very high - should be in the milliseconds. Even when calling it up to four times, as per the code...

jrosdahl:
> Storing and showing the real time is not necessary (for showing the efficiency), it is used to calculate a different metric: how much of the time was actually spent on CPU (and indirectly how much was wait)

My point is that it seems like a useless metric to store and display.

For example, say that you compile a file and that it takes 5 s wall clock time on your machine with high load. It will then appear as if `real_compile` is 5 s for all future compilations taken from that cache result. On the other hand, it looks like the per-result `real_compile` value is only used in a `cc_log` at the moment.

Similarly, I don't see how real time would be of interest for global counters. Compare these two cases:

1. You have a 1 CPU system with 10 files which you compile serially, each taking 1 s. The total build time will be 10 s and the global `time (real)` counter will be **10** s.
2. On the same 1 CPU system you compile the 10 files in parallel, each taking 10 s (because there are 9 other competing for CPU resources). The total build time will be 10 s but the global `time (real)` counter will be **100** s.

Have I misunderstood something?

> They are always used together, so it is mostly useful for other purposes - such as seeing how much "overhead" is generated ("sys" time) and how much is actually spent on compilation ("user" time)

Mmm, yes, I know what user and sys time is, I'm just not seeing a benefit of displaying them separately (and therefore neither of storing them separately) for the purpose of displaying metrics about how much CPU time is saved (or lost) by using ccache.

> The automatic cleanup is more like a "garbage collect", it probably won't affect individual compiles ?

If automatic cleanup is triggered, it will directly affect an individual compilation since it will be delayed until the cleanup is ready, which could be a very long time.

> But you are right that it would be "fair" to include it in the time spent by ccache, i.e. the global counters.

Right. The problem is that cleanup typically is extremeIy I/O-bound, so measuring CPU won't tell the whole story. The same goes for normal cache misses/hits but to a lesser extent, by the way: extra I/O overhead that is caused by ccache won't be directly accounted for in user/sys time counters.

afbjorklund:
I removed the real time, and the utilization counters.

jrosdahl:
Thanks for working on this.

Have you had time to consider my ["thoughts"](https://github.com/ccache/ccache/pull/461#issuecomment-529210432) number 4 and 5 yet?

Also, could you maybe reason a bit about why you chose to present both an efficiency and a performance number? I don't find any of them intuitive, which could just mean that they should be [documented properly](https://ccache.dev/manual/3.7.4.html#_cache_statistics), but it could also mean that they could be improved. I would spontaneously consider a single value called "saved CPU time" (in percent) more intuitive:

* A high value means that the cache is working well. For instance, if it's 90% it means that to produce object files whose cost would have been X CPU seconds if ccache hadn't been used actually only cost 0.1 X CPU seconds when using ccache.
* A low value means that the cache isn't working well. If it's 0%, you have neither gained nor lost CPU time by using ccache. If it's negative you are losing CPU time by using ccache.

afbjorklund:
The new output is shorter, and there are less counters. Was there something else ?

I believe both the percentage and the speedup was taken directly from `perf.py`
Looking at it again, I think the percentage might actually have be presented "backwards".

Currently a "100% (1.0x)" means that you haven't gained or lost anything wrt time.
Less than 1.0x means lose (cache overhead), more than 1.0x means win (cache hit).

jrosdahl:
> The new output is shorter, and there are less counters. Was there something else ?

Yes, "thought 4" (maybe condense information more or hide it behind another status option?) and "thought 5" (find a more descriptive configuration name than "time" or make it always true).

> I believe both the percentage and the speedup was taken directly from perf.py
> Looking at it again, I think the percentage might actually have be presented "backwards".

The performance script presents speedup/slowdown in percent and as a factor where the relationship between the two is `percent = 100 / factor`. But the "cache efficiency" and "performance" statistics seem to be something else and have a different relationship.

afbjorklund:
This needs to have a better name and status, and rewritten in c++ for master

Closing PR, it's still there if you want to reuse any of it (like using `getrusage`)

jrosdahl:
Thanks @afbjorklund! I'll certainly reuse most parts of it if implemented in the future.

PR #465, opened 2019-09-11
Add standalone string_view.hpp from boost

totph:
- I specifically need this for the `-MJ` compile command switch, see #277.
- It would be helpful for C to C++ interop in general, e.g. unlike `std::string` it avoids copies when wrapping `char*` strings.
- Switching `ccache` to native C++17 which contains `string_view` probably won't happen for a while.

For the patches to `string_view.hpp` see https://github.com/totph/boost_string_view, maybe clone this repo into the ccache project itself for reference and boost updates. I did this manually because the `bcp` tool reportedly creates quite a large dependency tree and it can't know that C++11 support is a given.

License wise the Boost Software License is one-way compatible to the GPL 3.

Or instead of this making the entirety of boost a dependency would work as well :)

### DIFF
--- a/src/third_party/string_view.hpp
+++ b/src/third_party/string_view.hpp
@@ -1,0 +1,635 @@
+/*
+   Copyright (c) Marshall Clow 2012-2015.
+   Copyright (c) Beman Dawes 2015
+   Copyright (c) Glen Joseph Fernandes 2019 (glenjofe@gmail.com)
+
+   Distributed under the Boost Software License, Version 1.0. (See accompanying
+   file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+    For more information, see http://www.boost.org
+
+    Based on the StringRef implementation in LLVM (http://llvm.org) and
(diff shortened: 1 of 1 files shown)


jrosdahl:
I'm considering starting to use https://github.com/martinmoene/optional-lite, and then I noticed that the same author has a string_view implementation as well: https://github.com/martinmoene/string-view-lite. How does your implementation compare to that one?

totph:
I used the boost version to be on the safe side and because it mentioned that LLVM's `StringRef` (which I am trying to replace, see below) inspired the boost version.

This one has the advantage of already being stand-alone and having a few convenience methods added, looks fine to me.

It also works with the stand-alone UTF8-`escape()` function that compile commands support will need: [escape() function from LLVM's YAMLParser.cpp](https://github.com/totph/llvm/blob/utf8_escape_standalone/lib/Support/YAMLParser.cpp)

jrosdahl:
> I'm considering starting to use https://github.com/martinmoene/optional-lite

Done now, and I also added `nonstd::string_view` in 6587b941651c so closing this one.

PR #466, opened 2019-09-12
Fix test suite SDKROOT detection in OSX

00003481:
### Description ###
(APPLE OSX) fixes the SDKROOT detection, so that it works also for the command line tools

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -66,12 +66,12 @@ if test "$ac_compiler_clang" = yes; then
     more_warnings="$more_warnings -Wno-padded"
     more_warnings="$more_warnings -Wno-shorten-64-to-32"
     more_warnings="$more_warnings -Wno-sign-conversion"
-    more_warnings="$more_warnings -Wno-undefined-func-template"
+#   more_warnings="$more_warnings -Wno-undefined-func-template"
     more_warnings="$more_warnings -Wno-weak-vtables"
 
 # TODO: Enable these in the future:
     more_warnings="$more_warnings -Wno-old-style-cast"
-    more_warnings="$more_warnings -Wno-zero-as-null-pointer-constant"
+#   more_warnings="$more_warnings -Wno-zero-as-null-pointer-constant"
 fi
 
 AC_ARG_ENABLE(more_warnings,
--- a/test/run
+++ b/test/run
@@ -378,25 +378,15 @@ else
 fi
 
 if $HOST_OS_APPLE; then
-    # Grab the developer directory from the environment or try xcode-select
-    if [ "$XCODE_DEVELOPER_DIR" = "" ]; then
-      XCODE_DEVELOPER_DIR=`xcode-select --print-path`
-      if [ "$XCODE_DEVELOPER_DIR" = "" ]; then
-        echo "Error: XCODE_DEVELOPER_DIR environment variable not set and xcode-select path not set"
+#   if somebody has the skill to deal with miltiple xcode versions/releases
+#   he/she should be capable of specifying a full sdk-path
+    if [ "$SDKROOT" = "" ]; then
+      SDKROOT=`xcrun --sdk macosx --show-sdk-path 2>/dev/null`
+      if [ "$SDKROOT" = "" ]; then
+        echo "Error: xcrun --show-sdk-path failure"
         exit 1
       fi
     fi
-
-    # Choose the latest SDK if an SDK root is not set
-    MAC_PLATFORM_DIR=$XCODE_DEVELOPER_DIR/Platforms/MacOSX.platform
-    if [ "$SDKROOT" = "" ]; then
-        SDKROOT="`eval ls -f -1 -d \"$MAC_PLATFORM_DIR/Developer/SDKs/\"*.sdk | tail -1`"
-        if [ "$SDKROOT" = "" ]; then
-            echo "Error: Cannot find a valid SDK root directory"
-            exit 1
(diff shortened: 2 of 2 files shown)


00003481:
The pull request provides the support for the “command line tools” environment

The “command line tools” is the component that provides a development ( compilers and related utilities ) environment
similar to the Linux environment without installing the bulkier XCODE 
( 200MB download instead of 6.1GB, 1GB installed instead of  12GB)

The current code parses the SDK path according to the XCODE installation ONLY
And ignores a command line tools environment 

The command 
xcode-select -p

Returns 
/Library/Developer/CommandLineTools
For the command line tools environment

and
/Applications/Xcode.app/Contents/Developer
For the XCODE environment 


While 
xcrun --sdk macosx --show-sdk-path 2>/dev/null 
Returns the fully resolved SDK PATH

/Library/Developer/CommandLineTools/SDKs/MacOSX10.14.sdk
For the command line tools environment

/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.14.sdk
For the XCODE environment 

it fixes also the bug of the missing 2>/dev/null 

As far as backwards compatibility
when testing on the older El_Capitan OSX Version 10.11
The command  xcrun --sdk macosx --show-sdk-path 2>/dev/null     works

But the support is already flaky

[enrico@enrico-core2 ccache_my.git]$make
  CXX      src/main.o
  CXX      src/AtomicFile.o
error: unknown warning option '-Wno-undefined-func-template'; did you mean '-Wno-undefined-internal'? [-Werror,-Wunknown-warning-option]
error: unknown warning option '-Wno-zero-as-null-pointer-constant'; did you mean '-Wno-int-to-void-pointer-cast'?
      [-Werror,-Wunknown-warning-option]
make: *** [src/AtomicFile.o] Error 1

[enrico@enrico-core2 ccache_my.git]$clang --version
Apple LLVM version 8.0.0 (clang-800.0.42.1)
Target: x86_64-apple-darwin15.6.0
Thread model: posix
InstalledDir: /Library/Developer/CommandLineTools/usr/bin 

Note: just commenting the setting of the two flags provided a good installation 

Right now I am testing on OS[X] 10.13 ( High Sierra ) and 10.14 ( Mojave ) , with 10.11 ( El Capitan ) for backwards compatibility


IMO for full portability across different compiler kinds and versions
The compiler flags should be checked , I remember using a long time ago
Something like AC_CHECK_COMPILER_FLAG

jrosdahl:
> The “command line tools” is the component that provides a development ( compilers and related utilities ) environment similar to the Linux environment without installing the bulkier XCODE [...]

Is it the case that the Xcode environment depends on the command line tools component so that the latter is always available?

> But the support is already flaky
> [...]
> error: unknown warning option '-Wno-undefined-func-template'; did you mean '-Wno-undefined-internal'? [-Werror,-Wunknown-warning-option]
> [...]
> IMO for full portability across different compiler kinds and versions The compiler flags should be checked [...]

Now that's a different issue: it means that it's not possible to use "Apple LLVM version 8.0.0 (clang-800.0.42.1)" for **developing** ccache. But it's still possible to **compile and use** ccache with that compiler when compiled from a [release archive](https://github.com/ccache/ccache/blob/master/doc/INSTALL-from-release-archive.md). Compiling ccache [directly from the source repository](https://github.com/ccache/ccache/blob/master/doc/INSTALL.md) intentionally requires a more specific environment.

That said, I think that it in this case makes sense to make it easier to build ccache in dev mode from the source repository with other compilers than the ones used by the CI builds, so adding configure checks for the flags sounds good. Done in 4642591a0ccb. Thanks for the suggestion!

test/run
```diff
@@ -378,25 +378,15 @@ else
 fi
 
 if $HOST_OS_APPLE; then
-    # Grab the developer directory from the environment or try xcode-select
-    if [ "$XCODE_DEVELOPER_DIR" = "" ]; then
-      XCODE_DEVELOPER_DIR=`xcode-select --print-path`
-      if [ "$XCODE_DEVELOPER_DIR" = "" ]; then
-        echo "Error: XCODE_DEVELOPER_DIR environment variable not set and xcode-select path not set"
+#   if somebody has the skill to deal with miltiple xcode versions/releases
```
jrosdahl: I don't quite understand this comment. Do you mean there's a risk that `$SDKROOT` isn't set to an absolute path?

configure.ac
```diff
@@ -66,12 +66,12 @@ if test "$ac_compiler_clang" = yes; then
     more_warnings="$more_warnings -Wno-padded"
     more_warnings="$more_warnings -Wno-shorten-64-to-32"
     more_warnings="$more_warnings -Wno-sign-conversion"
-    more_warnings="$more_warnings -Wno-undefined-func-template"
+#   more_warnings="$more_warnings -Wno-undefined-func-template"
```
jrosdahl: These are no longer needed, please rebase.

00003481:
I did not want to hack too much the script, but the use of an exported variable is useless IMO

Xcode-select takes care of setting the proper environment and directories 

And 
xcrun --sdk macosx --show-sdk-path  prints the full path


The current XCODE_DEVELOPER_DIR IS the <root> 
and after that You search by hand for the proper SDK

My approach gets in one shot the full SDK path

The comment was a warning not to make things too complicated

Enrico

00003481:
will create a new pull request

PR #467, opened 2019-09-14
Cleanup in the cc_reset function

brenoguim:
# Description ###

I was reading the codebase the other day and noticed that it's moving from C to C++. Call me crazy, but modernizing code (small step by small step) is a very relaxing hobby :)

I guess the reason for the first change is obvious, as there is no need to reset the field twice.
The second one is just to improve readability a bit and avoid writing each field name twice.

Let me know if you are interested in such changes. I'd be willing to submit these kinds of PRs every now and then.

### DIFF
--- a/doc/NEWS.adoc
+++ b/doc/NEWS.adoc
@@ -41,11 +41,11 @@ Release date: 2019-09-12
 Improvements
 ~~~~~~~~~~~~
 
-- Added support for GCC 9’s `-gz[=type]` option (previously ccache would think
-  that “-gz” alone would enable debug information, thus potentially including
-  the current directory in the hash).
+- Added support for the `-gz[=type]` compiler option (previously ccache would
+  think that “-gz” alone would enable debug information, thus potentially
+  including the current directory in the hash).
 
-- Added support for converting paths like “/c/users” into relative paths on
+- Added support for converting paths like “/c/users/...” into relative paths on
   Windows.
 
 


brenoguim:
Ooops, I didn't realize GitHub would update the pull request with all extra commits I had fork.

I'll see how to undo that, but in the meantime, is this kind of change something you are interested in? I mean, the general lines of reducing manual memory management.

jrosdahl:
Thanks for the PR.

> I was reading the codebase the other day and noticed that it's moving from C to C++. Call me crazy, but modernizing code (small step by small step) is a very relaxing hobby :)

I have similar hobbies. For ccache, I'm trying to restrain myself from doing too much superficial improvements at the moment, though. :slightly_smiling_face: 

> Let me know if you are interested in such changes. I'd be willing to submit these kinds of PRs every now and then. [...] is this kind of change something you are interested in? I mean, the general lines of reducing manual memory management.

I'm much in favor of graceful code cleanups in general. However, since I'm currently working on relatively large structural code changes, some of the changes you propose will clash with that effort. I think that I would like to pass on some of the changes for the time being.

Here are some quick comments:

* ff090f8efcce38d887d7fa006a0d776c68fcba3d and 6d5c383068b69e9e85e9428e098af614728ed4cc look fine.
* The "Use std::vector instead of manually managed array" commits look fine in principle, haven't reviewed details.
* 1d3ca5b2e39e16c30a38bfe862880f5ca19c0add looks fine (that must be some leftover from when the Config class was implemented in another way).
* "Use std::string instead of manually managed char*": yes.
* "Use std::vector to avoid manual memory management": why vector instead of string?
* b3c3cce44f40f3eb9ac884adf2392e5454f1e39f and the following commits: It's a nice idea, but spontaneously I think I would rather see converting code to use fmt::format and std::string directly instead of using such an intermediate solution. I'm currently reworking code related to compression, manifests and results, and it's likely that doing C++-ification or other cleanups in those areas will lead to both syntactical and semantical conflicts with my local work.

Let me try to come up with some parts of the code that can be improved without conflicts... A couple that come to mind immediately are the args, counters and lockfile routines – they can be converted to proper classes without clashing (much) with what I'm doing. :slightly_smiling_face: Also, if you want to sync efforts, you can join https://gitter.im/ccache/ccache or mail me directly.

brenoguim:
> ff090f8 and 6d5c383 look fine.

Cool, I'll cleanup this PR to contain only those two instead, as I originally wanted.

> 1d3ca5b looks fine (that must be some leftover from when the Config class was implemented in another way).

I'll add that one too.

> "Use std::vector to avoid manual memory management": why vector instead of string?

The call to `strtok_r` takes a  `char*` as it needs to change the array to do the tokenization. `std::string::c_str()` returns a `const char*`.

> b3c3cce and the following commits: It's a nice idea, but spontaneously I think I would rather see converting code to use fmt::format and std::string directly instead of using such an intermediate solution. 

Yes, I realized that a bit too late.  I'll just drop that altogether.

> Let me try to come up with some parts of the code that can be improved without conflicts... A couple that come to mind immediately are the args, counters and lockfile routines – they can be converted to proper classes without clashing (much) with what I'm doing. 🙂 Also, if you want to sync efforts, you can join https://gitter.im/ccache/ccache or mail me directly.

I didn't know about the gitter community, that will be much better to sync and understand what is going on.

I actually got to ccache because I wanted implement a cache for a few scripts and I wanted to see if it was possible to implement another "compiler" command line parser easily.
But seemed harder than I thought, so I wrote some lame bash script for the time being.

Thank you for taking the time to review and comment the PR.

jrosdahl:
> Cool, I'll cleanup this PR to contain only those two instead, as I originally wanted.

Great, thanks.

> The call to strtok_r takes a char* as it needs to change the array to do the tokenization. std::string::c_str() returns a const char*.

Right, but one can use `&s[0]` to get a non-const `char*` and, AFAIK, it's legal to modify the string using that pointer. (BTW, I just learned that `.data()` returns `char*` in C++17.) But sure, `std::vector<char>` works as well.

PR #472, opened 2019-09-21
Small cleanup cc reset

brenoguim:
As discussed before, these are cleanups to the cc_reset function.


afbjorklund:
Could have been done with a macro even before, if it had been worth it.

Issue #414, opened 2019-05-05
Design and implement a backend API and plugin framework


jrosdahl:
### Background

We want to refactor the code in preparation for introducing support for alternative/additional backends. See also #58.

### What to do

1. The API primitives should work for both the file system backend and other alternative backends.
2. The API should preferably include a way for the backend to read configuration. Should we have namespaces in the configuration to separate different backends?
3. Decide if the API should be open (usable by plugins not bundled with ccache, thus aptly documented and versioned) or closed (plugins are only used as an internal mechanism for improved code structure).
4. Design the plugin framework.
5. Make a plan on how to configure the interaction of backends in a generic way.
    - For instance, one might want to configure in which order backends should be queried and which backends should be read-only.
    - Or if we want to simplify things, we could decide that we only support **one** optional backend in addition to the standard one.

pavsa:
I think this issue should receive the highest priority now that various backend patches are starting to appear so that people can begin to implement various backends cleanly that would be easily merged to the master, instead of patches that work for one backend only.

Also, if we are on the edge of switching to C++, this might be a very sensible time to do it. If we are to support querying multiple backends, the current code state is very inflexible for such scenario. 

We should definitely support querying multiple backends, even of the same type.
For example, I would want to query two filesystem backends in two different directories, but one is read only, other one is read/write. 

There should be also an operation to verify the integrity of the cache contents.

jrosdahl:
> I think this issue should receive the highest priority

Right, it has quite high priority in the sense that I'm keeping it mind when doing various refactoring. (As mentioned in another issue, I've begun sketching an API.) It's not the highest priority for me at the moment, though – I think that polishing things related to #411 and #413 is more important right now.

That said, if you or anybody else also is interested in designing an API, please go ahead and post ideas and opinions. One thing in particular to think about is bullet 3 in "What to do" in the issue description. I would personally like it if we could manage to make an open API so that backend implementations can have their own repos, ownership and life cycles.

> now that various backend patches are starting to appear

Which patches do you refer to? I'm not aware of any lately appearing backend-related patches.

> Also, if we are on the edge of switching to C++, this might be a very sensible time to do it.

Related to this: If go for an open API, it will be in C (since there's no stable C++ ABI)... But yes, internal interfaces would be less clunky in C++ form.

> For example, I would want to query two filesystem backends in two different directories, but one is read only, other one is read/write.

Agreed, that's one of the scenarios I had in mind for bullet 5.

[...]

PR #473, opened 2019-10-09
Limit suffix size in ResultFileMap to 15

totph:
Implicit but unenforced limit was 256 before.
Reduces stack usage and ensures SSO for the
std::string (which is used as the map key) is
always possible.

Also fix an out-of-bounds read when logging.

-------
Additional motivation: Make sure this limit is set now before plugins use overly long keys.

And now that I think about it: The key uniqueness needs to be ensured as well.

### DIFF
--- a/src/result.cpp
+++ b/src/result.cpp
@@ -97,6 +97,14 @@ const uint8_t k_embedded_file_marker = 0;
 // File stored as-is in the file system.
 const uint8_t k_raw_file_marker = 1;
 
+void ResultFileMap::register_suffix(std::string suffix, const char* path)
+{
+  if (suffix.size() >= MAX_SUFFIX_SIZE) {
+    throw Error("internal error: suffix too large");
+  }
+  this->emplace(suffix, path);
+}
+
 using ReadEntryFunction = void (*)(CacheEntryReader& reader,
                                    const std::string& result_path_in_cache,
                                    uint32_t entry_number,
@@ -119,8 +127,9 @@ read_embedded_file_entry(CacheEntryReader& reader,
   uint8_t suffix_len;
   reader.read(suffix_len);
 
-  char suffix[256];
+  char suffix[MAX_SUFFIX_SIZE + 1];
   reader.read(suffix, suffix_len);
+  suffix[suffix_len] = '\0';
 
   uint64_t file_len;
   reader.read(file_len);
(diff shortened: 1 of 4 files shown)


src/result.cpp
```diff
@@ -119,8 +127,9 @@ read_embedded_file_entry(CacheEntryReader& reader,
   uint8_t suffix_len;
   reader.read(suffix_len);
 
-  char suffix[256];
```
totph: When first called with the suffix `.ccmd` (from local WIP), `.ccmd` is correctly printed in `cc_log()` below because luckily `suffix` is still zero'd. The next suffix `.o` however is printed as `.ocmd` here.

jrosdahl:
Thanks for noticing the logging bug! It's a regression in 1e5a684058e99. It should definitely be fixed.

That said, I think that the proposed solution complicates the code more than it improves it. I don't quite understand the motivation to enforce an arbitrary limit of the keys in ResultFileMap. I'd say that there's no need to micro-optimize run-once code paths like the one we're talking about. I think that clearer/less code overtrumps micro-optimizations that in practice don't affect performance.

You're correct that the suffix length is not enforced when adding entries to the map; that could very well be done with an assertion. However, I don't think that it would gain much to do so since a ccache developer adding support for a new file type will notice right away that it doesn't work with a larger key than 256 bytes (or if the key is duplicated).

Regarding plugins: what kind of plugins do you have in mind that could affect the key lengths? If you're thinking about storage backend plugins (#414), they won't affect the result data content. The logic for encoding and decoding result data will stay in core ccache code.

If you're OK with my reasoning above, would you like to work on fixing the logging (for instance, just create the `std::string suffix_str` variable earlier and use it in the log calls) or do you want me to do it?

totph:
Ok, fix is now in #475 

And if the cached data types stay in core (though this might seem like a micro optimization again): Why not skip the suffix string entirely and just store an 8 or 16 bit enum directly? This explicit typing would solve the length and uniqueness issues plus add a microscopic speedup. Then the internal names can be properly descriptive.

jrosdahl:
> And if the cached data types stay in core (though this might seem like a micro optimization again): Why not skip the suffix string entirely and just store an 8 or 16 bit enum directly? This explicit typing would solve the length and uniqueness issues plus add a microscopic speedup. Then the internal names can be properly descriptive.

Yes, using an enum sounds like a good idea. I guess the strings solution just felt like the easiest solution at the time when the initial implementation of the result file code was written (0399be2d625a7bd61df545b78f13bca2fa61ceb6).

jrosdahl:
Obsoleted by #478.

PR #475, opened 2019-10-14
Fix an out-of-bounds read when logging

totph:
The stack variable `suffix` is not guaranteed to be zero-terminated.


jrosdahl:
Thanks.

PR #477, opened 2019-10-17
Test output: visibility on light backgrounds / ansi codes

totph:
test output: improve readability on light backgrounds
   
 `bold()` now only writes bold text, not bold and white text

_and_

 test output: only use ansi codes when run in a terminal


jrosdahl:
Thanks.

PR #478, opened 2019-10-17
Convert result type string to an enum

totph:
As discussed in #473, feel free to adapt the enum name or the enum entries to your liking. E.g. to me it was very important that stderr_output had the id 2!

### DIFF
--- a/src/result.cpp
+++ b/src/result.cpp
@@ -48,8 +48,7 @@
 // <embedded_file_entry>  ::= <embedded_file_marker> <suffix_len> <suffix>
 //                            <data_len> <data>
 // <embedded_file_marker> ::= 0 (uint8_t)
-// <suffix_len>           ::= uint8_t
-// <suffix>               ::= suffix_len bytes
+// <embedded_file_type>   ::= uint8_t
 // <data_len>             ::= uint64_t
 // <data>                 ::= data_len bytes
 // <raw_file_entry>       ::= <raw_file_marker> <suffix_len> <suffix> <file_len>
@@ -68,8 +67,7 @@
 // --- [potentially compressed from here] -------------------------------------
 // <n_entries>            1 byte
 // <embedded_file_marker> 1 byte
-// <suffix_len>           1 byte
-// <suffix>               suffix_len bytes
+// <embedded_file_type>   1 byte
 // <data_len>             8 bytes
 // <data>                 data_len bytes
 // ...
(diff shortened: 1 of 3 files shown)


src/result.cpp
```diff
@@ -116,30 +113,26 @@ read_embedded_file_entry(CacheEntryReader& reader,
                          const ResultFileMap* result_file_map,
                          FILE* dump_stream)
 {
-  uint8_t suffix_len;
-  reader.read(suffix_len);
-
-  char suffix[256];
-  reader.read(suffix, suffix_len);
+  FileType type;
+  reader.read(*reinterpret_cast<UnderlyingFileTypeInt*>(&type));
```
jrosdahl: I find these casts a bit ugly. How about working with an `uint8_t` instead and casting when searching instead, i.e. something like

```c++
uint8_t type;
reader.read(type);
// ...
const auto it = result_file_map->find(FileType(type));
```

(or use `UnderlyingFileTypeInt`)? And similar in other places. What do you think?

src/result.hpp
```diff
@@ -25,9 +25,24 @@
...
 
-typedef std::map<std::string /*suffix*/, std::string /*path*/> ResultFileMap;
+using UnderlyingFileTypeInt = uint8_t;
+enum class FileType : UnderlyingFileTypeInt {
+  // These values are written into the cache result file.
+  // This means they must never be changed or removed unless the
+  // result file version is incremented. Adding new values is Ok.
+  object = 0,
+  dependency = 1,
+  stderr_output = 2,
```
jrosdahl: :+1: for value 2 here. :slightly_smiling_face:

src/result.cpp
```diff
@@ -116,30 +113,26 @@ read_embedded_file_entry(CacheEntryReader& reader,
...
 
   bool content_read = false;
-  std::string suffix_str(suffix, suffix_len);
   if (dump_stream) {
     fmt::print(dump_stream,
-               "Embedded file #{}: {} ({} bytes)\n",
+               "Embedded file #{}: type {} ({} bytes)\n",
                entry_number,
-               suffix_str,
+               +static_cast<UnderlyingFileTypeInt>(type),
```
jrosdahl: It would be nice to print a human-readable string instead of (or in addition to) an integer here and in similar places.

jrosdahl:
Thanks. I'll tweak the implementation as commented earlier.

PR #481, opened 2019-10-30
Fix and test interaction between fmt and nonstd::string_view

totph:
Empty string_views could not be handled and formatting
did not stop until the first \0 was encountered.

### DIFF
--- a/src/FormatNonstdStringView.hpp
+++ b/src/FormatNonstdStringView.hpp
@@ -18,6 +18,9 @@
 
 #pragma once
 
+#include "third_party/nonstd/string_view.hpp"
+
+
 // Specialization of fmt::formatter for nonstd::string_view.
 namespace fmt {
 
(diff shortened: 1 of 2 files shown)


jrosdahl:
Thanks!

So there are two bugs:

1. I wrote `{:{}}` when I meant `{:.{}}` – the latter works for non-null `string_view`s.
2. I didn't think about the default-constructed `string_view` case.

Using `fmt::string_view` seems like relying on an implementation detail in `fmt` since (IIUC) its `string_view` is just an internal compatibility type alias, potentially using a custom implementation. Would it be better to not rely on the internal type alias? If so an alternative fix could look like this:

```diff
--- a/src/FormatNonstdStringView.hpp
+++ b/src/FormatNonstdStringView.hpp
@@ -38,7 +37,8 @@ template<> struct formatter<nonstd::string_view>
   format(const nonstd::string_view& sv, FormatContext& ctx)
     -> decltype(ctx.out())
   {
-    return format_to(ctx.out(), "{:{}}", sv.data(), sv.length());
+    auto data = sv.data() ? sv.data() : "";
+    return format_to(ctx.out(), "{:.{}}", data, sv.length());
   }
 };
 
```

Which would you prefer?

totph:
This actually is part of the API, and fmt already handles the nullptr case:
https://fmt.dev/6.0.0/api.html#_CPPv2N3fmt17basic_string_viewE

jrosdahl:
OK, that seems good enough, thanks.

PR #482, opened 2019-11-02
Allow -fmodules in direct depend mode

llunak:
Clang creates the internal on-disk representation of modules
in the background on demand as necessary. This means that ccache does
not need to cache that, all that it needs to cache is the actual
compilation result and the dependencies. Which in the case of modules
includes also the module.modulemap files, and Clang outputs such
dependencies correctly only in the depend mode, in the preprocessed
output there is no way to find out the module dependencies.

Therefore support -fmodules in direct depend mode by more or less
ignoring it. Tested with source build of LLVM/Clang configured
with -DLLVM_ENABLE_MODULES=On (and the build actually occassionally
fails because of module problems, but that happens regardless of ccache).

See also pull request #130 (and this reuses the tests from there).

PS: To be honest, I can't say for 100% sure that this is technically correct, but it appears to work fine (besides the above-mentioned problems, which however do not seem to be ccache's fault).

### DIFF
--- a/src/ccache.cpp
+++ b/src/ccache.cpp
@@ -2570,6 +2570,31 @@ cc_process_args(struct args* args,
       found_fpch_preprocess = true;
     }
 
+    // Modules are handled on demand as necessary in the background,
+    // so there is no need to cache them, they can be in practice ignored.
+    // All that is needed is to correctly depend also on module.modulemap files,
+    // and those are included only in depend mode (preprocessed output does not
+    // list them). Still, not including the modules themselves in the hash
+    // could possibly result in an object file that would be different
+    // from the actual compilation (even though it should be compatible),
+    // so require a sloppiness flag.
+    if (str_eq(argv[i], "-fmodules")) {
+      if (!g_config.depend_mode() || !g_config.direct_mode()) {
+        cc_log("Compiler option %s is unsupported without direct depend mode",
+               argv[i]);
+        stats_update(STATS_CANTUSEMODULES);
+        result = false;
+        goto out;
+      } else if (!(g_config.sloppiness() & SLOPPY_MODULES)) {
+        cc_log(
+          "You have to specify \"modules\" sloppiness when using"
+          " -fmodules to get hits");
+        stats_update(STATS_CANTUSEMODULES);
+        result = false;
+        goto out;
+      }
+    }
+
     // We must have -c.
     if (str_eq(argv[i], "-c")) {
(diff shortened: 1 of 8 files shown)


test/suites/modules.bash
```diff
@@ -0,0 +1,84 @@
...
+EOF
+    backdate module.modulemap
+
+   cat <<EOF >test1.cpp
+#import "test1.h"
+int main() { return 0; }
+EOF
+}
+
+expect_file_contains() {
```
jrosdahl: Please move this function to `test/run` near `expect_file_content`.

src/ccache.cpp
```diff
@@ -2570,6 +2570,19 @@ cc_process_args(struct args* args,
...
     }
 
+    // Modules are handles on demand as necessary in the background,
+    // so there is no need to cache them, they can be in practice ignored.
+    // All that is needed is to correctly depend also on module.modulemap files,
+    // and those are included only in depend mode (preprocessed output does not
+    // list them).
+    if ((!g_config.depend_mode() || !g_config.direct_mode())
+        && str_eq(argv[i], "-fmodules")) {
+      cc_log("Compiler option %s is unsupported without direct depend mode", argv[i]);
```
jrosdahl: I you have the possibility, consider installing clang-format and running `make format`.

src/ccache.cpp
```diff
@@ -2570,6 +2570,19 @@ cc_process_args(struct args* args,
       found_fpch_preprocess = true;
     }
 
+    // Modules are handles on demand as necessary in the background,
```
jrosdahl: Minor: handles -> handled

llunak:
> Right. I agree that the approach seems reasonable, but I get a little uncomfortable about the thought of not reproducing the compiler behavior.

Note that I do the same in Icecream (both PCH and modules options are simply stripped before sending the compile job to a remote node) and it seems to work fine. I still can't quite vouch for modules because of little usage so far and the modules mechanism being more complicated, but I've already used PCHs this way a lot and it's been without problems.

Patch updated.

jrosdahl:
Excellent, thanks!

mbektchiev:
🎉  I can't wait to see this released! What are the plans for the official 4.0 version?

jrosdahl:
> What are the plans for the official 4.0 version?

In terms of dates: no plan other than "when it feels ready enough". In terms of scope: the major thing I would like to find time to work on is #414.

Issue #447, opened 2019-07-28
The future of ccache's support for running on Windows


jrosdahl:
This issue is for discussing the future of ccache's support for running on Windows for caching GCC/Clang output. It's not about building/using ccache with Visual Studio.

### The present

* ccache has had some support for running on Windows for many years. (Here "running on Windows" only means executing on Windows, not support for caching output of Windows-only compilers like Visual Studio.)
* The support has however never been official. The major reason is simply that the ccache development team (@jrosdahl) has little to no personal interest in the Windows port.
* We have CI tests that verify that ccache is possible to build with a MinGW-w64 cross-compiler.
* However, there are currently no CI tests that run on Windows. This means that the state of the Windows support is largely unknown for the ccache developers. See also #247.
* Every once in a while, contributors make various efforts to fix or improve the Windows support, which is appreciated.
* The main problem still remains, though: there are too few people who both care about the Windows port and are expected to contribute to ccache regularly in the future.
* As long as I (@jrosdahl) am the main ccache maintainer and contributor without more people helping out, I'm forced by time and interest constraints to be restrictive in accepting changes that make it harder for me to maintain ccache. It's easy to accept new features that work _now_, but it's a different story to _keep_ them working. (I've written more about this before [here](https://github.com/ccache/ccache/pull/162#issuecomment-289215585), although then the context was also about using ccache for the Visual Studio compiler.)
* The problem I have with maintaining Windows-specific code is that I don't run Windows and don't know much about Windows development environments, so if some Windows-related problem is flagged by CI tests it's hard and timeconsuming for me to understand and fix them. If such problems block a release, I have to disable the tests and release with known-broken Windows support. And I want to dedicate my spare time on things I find fun to do, like implementing new features, not fixing portability issues.
* If there were more people who are interested in (and committed to) the Windows port in the developer team (i.e. more than zero :slightly_smiling_face:), things would be different.

We need to figure out a way forward to avoid wasting time and to reduce friction between people with different goals and perceptions on what ccache should support.

### The future

I see the following possibilities:

1. **Keep the status quo.** I.e. the Windows support continues to be unofficial and releases continue to be untested. Problems will maybe be fixed by drive-by contributions, or maybe not.
    - _Pros_: Nothing would need to be done at the moment.
[...]

PR #483, opened 2019-11-04
Misc C++-ification

totph:
Misc string view stuff I came across, split into separate commits.

Also, what is your opinion on adding `using nonstd::string_view;` to the top of `.cpp` files?

### DIFF
--- a/src/Util.cpp
+++ b/src/Util.cpp
@@ -24,6 +24,8 @@
 #include <algorithm>
 #include <fstream>
 
+using nonstd::string_view;
+
 namespace {
 
 void
@@ -40,7 +42,7 @@ get_cache_files_internal(const std::string& dir,
   std::vector<std::string> directories;
   dirent* de;
   while ((de = readdir(d))) {
-    nonstd::string_view name(de->d_name);
+    string_view name(de->d_name);
     if (name == "" || name == "." || name == ".." || name == "CACHEDIR.TAG"
         || name == "stats" || name.starts_with(".nfs")) {
       continue;
@@ -137,6 +139,26 @@ dir_name(nonstd::string_view path)
     return ".";
   }
   return n == 0 ? "/" : path.substr(0, n);
+}
+
+nonstd::string_view
+get_extension(nonstd::string_view path)
+{
+#ifndef _WIN32
+  const char stop_at_chars[] = "./";
+#else
+  const char stop_at_chars[] = "./\\";
+#endif
+  size_t pos = path.find_last_of(stop_at_chars);
+  if (pos == string_view::npos || path.at(pos) == '/') {
+    return string_view();
+#ifdef _WIN32
+  } else if (path.at(pos) == '\\') {
+    return string_view();
+#endif
+  } else {
+    return path.substr(pos);
+  }
 }
 
 nonstd::string_view
--- a/src/Util.cpp
+++ b/src/Util.cpp
@@ -139,6 +139,16 @@ dir_name(nonstd::string_view path)
   return n == 0 ? "/" : path.substr(0, n);
 }
 
+nonstd::string_view
+get_truncated_base_name(nonstd::string_view path, size_t max_length)
+{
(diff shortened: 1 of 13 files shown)


src/ccache.cpp
```diff
@@ -1559,15 +1559,15 @@ get_result_name_from_cpp(struct args* args, struct hash* hash)
...
-
+    path_stdout = []() -> char* {
+      string_view input_base = Util::base_name(input_file);
+      size_t dot_pos = input_base.find('.');
+      size_t truncate_pos = std::min(size_t(10),
+                                     std::min(input_base.size(), dot_pos));
+      input_base.substr(0, truncate_pos);
+      return x_strdup(
+          fmt::format("{}/{}.stdout", temp_dir(), input_base).c_str());
+    }();
```
totph: Ok, the "improvement" looks terrible with this diff. Mostly to visually separate the logic and the variable noise it adds. But completely optional.

src/Util.cpp
```diff
@@ -139,6 +141,18 @@ dir_name(nonstd::string_view path)
   return n == 0 ? "/" : path.substr(0, n);
 }
 
+nonstd::string_view
+get_extension(nonstd::string_view path)
+{
+  const char stop_at_chars[] = "./";
```
totph: This and the legacy version both stop at `/` - are non-cygwin Windows paths converted from `\` to `/` elsewhere or is only cygwin supported? I only see the `DIR_DELIM_CH` define used one.

src/ccache.cpp
jrosdahl: How about extracting the truncation logic to a named function (`get_truncated_base_name`?)?Then we could unit test it as well.

src/Util.cpp
jrosdahl: Neither non-cygwin Windows nor cygwin Windows is supported officially (yet); see #447. That said, I guess adding backslash to `stop_at_chars` for the `_WIN32` case wouldn't hurt.

src/ccache.cpp
```diff
@@ -32,6 +32,7 @@
 #include "result.hpp"
 #include "unify.hpp"
 
+#include "src/FormatNonstdStringView.hpp"
```
jrosdahl: The `src/` part can be removed here.

src/ccache.cpp
```diff
@@ -3434,12 +3426,10 @@ cc_process_args(struct args* args,
   // Add flags for dependency generation only to the preprocessor command line.
   if (generating_dependencies) {
     if (!dependency_filename_specified) {
-      char* base_name = remove_extension(output_obj);
-      char* default_depfile_name = format("%s.d", base_name);
-      free(base_name);
+      std::string default_depfile_name = Util::change_extension(output_obj, ".d");
```
jrosdahl: I you have the possibility, consider installing clang-format and running `make format`.

src/ccache.cpp
```diff
@@ -4202,8 +4188,8 @@ ccache_main(int argc, char* argv[])
 {
   try {
     // Check if we are being invoked as "ccache".
-    char* program_name = x_basename(argv[0]);
-    if (same_executable_name(program_name, MYNAME)) {
+    std::string program_name = std::string(Util::base_name(argv[0]));
```
jrosdahl:
```suggestion
    std::string program_name(Util::base_name(argv[0]));
```

totph:
Done, now with one more commit and the ConfigTester.

src/Util.cpp
```diff
@@ -169,6 +215,29 @@ get_level_1_files(const std::string& dir,
   get_cache_files_internal(dir, 1, progress_receiver, files);
 }
 
+std::string
+get_path_in_cache(nonstd::string_view name, nonstd::string_view suffix)
```
jrosdahl: When refactoring and modernizing the code I'm generally trying to remove dependencies to global state, e.g. passing the config explicitly to utility functions. Do you think that this would make sense here as well?

PR #484, opened 2019-11-05
Handle properly color diagnostics also in depend mode

llunak:
Compiler invocation in the depend mode uses the original arguments,
which means the added -fcolor-diagnostics (or -fdiagnostics-color)
was dropped.


llunak:
BTW, a version that applies cleanly to 3.7.x is at https://pastebin.com/zMd6Scya , in case that'd be useful.

jrosdahl:
Thanks.

PR #486, opened 2019-11-07
Correct "-MJ" bail out, identify both "-MJ file" and "-MJfile"

totph:
Added test to detect the invalid case TOO_HARD | TAKES_CONCAT_ARG


jrosdahl:
Thanks.

PR #488, opened 2019-11-10
Add handling of __TIMESTAMP__ macro

erijo:
### Description ###

`__TIMESTAMP__` is [expanded by the pre-processor](https://github.com/gcc-mirror/gcc/blob/master/libcpp/macro.c#L397) to `asctime()` of the source file's modification time. Handle `__TIMESTAMP__` similar to `__DATE__` and include asctime(modification time) in the hash when `__TIMESTAMP__` is detected in the source code.


jrosdahl:
Great, thanks.

PR #491, opened 2019-11-17
do not resolve symlinks for CCACHE_BASEDIR

llunak:
When building with builddir != srcdir, it is practically necessary
to have either builddir contain srcdir or the other way around,
otherwise relative paths from one to another will not be the same
for different builds, which will lead to misses with CCACHE_BASEDIR.
In case one wants builddir and srcdir to be elsewhere (for a number
of reasons, different filesystems, build files inside srcdir being
a mess, ...), this setup can be achieved using a symlink. But then
symlink resolving changes the relative path to exactly the broken
setup mentioned above.

The resolving was introduced in 2df269a3121889ebcdfa5d98dfb4d675f690e039
with questionable reasoning (how can CWD possibly contain "/./"?).
The only test that breaks by removing this change is an artifical
test designed to test exactly this (mis)feature, all the rest work fine.

### DIFF
--- a/src/ccache.cpp
+++ b/src/ccache.cpp
@@ -833,7 +833,7 @@ make_relative_path(char* path)
     free(p);
   }
 
-  char* canon_path = x_realpath(path);
+  char* canon_path = x_strdup(path);
   if (canon_path) {
     free(path);
     char* relpath = get_relative_path(get_current_working_dir(), canon_path);
--- a/test/suites/basedir.bash
+++ b/test/suites/basedir.bash
@@ -45,28 +45,6 @@ SUITE_basedir() {
     expect_stat 'cache miss' 2
 
     # -------------------------------------------------------------------------
- if ! $HOST_OS_WINDOWS && ! $HOST_OS_CYGWIN; then
-    TEST "Path normalization"
-
-    cd dir1
-    CCACHE_BASEDIR="`pwd`" $CCACHE_COMPILE -I`pwd`/include -c src/test.c
-    expect_stat 'cache hit (direct)' 0
-    expect_stat 'cache hit (preprocessed)' 0
-    expect_stat 'cache miss' 1
-
-    mkdir subdir
-    ln -s `pwd`/include subdir/symlink
-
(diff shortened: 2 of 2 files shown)


llunak:
Note: In case there actually is a good reason for the feature, I'm fine with adding a config option for this.

jrosdahl:
> But then symlink resolving changes the relative path to exactly the broken setup mentioned above. [...] The resolving was introduced in 2df269a with questionable reasoning (how can CWD possibly contain "/./"?).

The background can be found in this thread: https://www.mail-archive.com/ccache@lists.samba.org/msg00802.html. As can be seen, the original poster did have "/./" in CWD (although I don't understand how). Also, one can of course pass constructed absolute paths that include "/./" to the compiler like the ones in the test case you mentioned. On the other hand, the normalization of CWD, which is what the commit message of 2df269a3121889ebcdfa5d98dfb4d675f690e039 talks about, is done via another realpath call (nowadays present in get_current_working_dir) so that's not affected by your proposed change.

I agree (see my two last mails in said mail thread) that it would be preferable not to use abspath to normalize paths. I remember working on a better way of doing it back then but I gave up, presumably due to lack of time, and I can't find any code left from the attempt.

That said, just removing the call to realpath in make_relative_path makes the path no longer fulfill the preconditions of get_relative_path. The key thing is correctness in the presence of symlinks in the path. For example, `get_relative_path("/a/b/c", "/a/b/d")` currently always returns `../c`, which is incorrect if `/a/b/d` is a symlink pointing to, say, `/a/e`.

> broken setup [...] questionable reasoning [...] artifical test [...] (mis)feature

I would appreciate it if you could tone down the attitude a notch.

jrosdahl:
This should be fixed by e8444a3893cb. Please give it a try.

llunak:
I have meanwhile given up on CCACHE_BASEDIR as I found the CWD hassle not worth it, so I don't have an easy way to test, but if the tests in the commit work, then I assume the setup I used would have worked as well.

PR #492, opened 2019-11-20
Windows define and include restructuring

totph:
Set the _WIN32_WINNT macro only once in the generated config.h.

Include <windows.h> only once in system.hpp, unless these includes
are in third party libs - however third party includes themselves
should always be behind internal includes.


jrosdahl:
Thanks.

PR #494, opened 2019-11-27
Add debug logs in function verify_object

MizuhaHimuraki:
### Add more debug logs in function verify_object
Those logs are controlled by env named CCACHE_DEBUG_VERIFY_OBJECT like CCACHE_DEBUG_INCLUDED or CCACHE_DEBUG_UNIFY

They may be helpful when someone wants to debug cache missing in direct mode

### DIFF
--- a/src/manifest.c
+++ b/src/manifest.c
@@ -368,13 +368,23 @@ static int
 verify_object(struct conf *conf, struct manifest *mf, struct object *obj,
               struct hashtable *stated_files, struct hashtable *hashed_files)
 {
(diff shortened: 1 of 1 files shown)


jrosdahl:
Thanks for the PR.

Logs like this are indeed very helpful when debugging a specific problem, but I'm hesitant to add them to the main ccache code since the logs are, well, very specific. If a user wants to debug such a specific ccache problem they can add them theirselves temporarily, just like you did. Sounds reasonable?

afbjorklund:
Some of the other low-level debugging were protected by shields, to leave them out of release builds...

Issue #493, opened 2019-11-26
Entries of manifest out of range causes direct mode miss


MizuhaHimuraki:
First, I execute a build with a.cpp which include a header named pub.h, ccache created cache and manifest correctly.

Then, I define a macro in the public header  pub.h
> #define A_CPP_NOT_USE "1"

When I build again, in the compiling of a.cpp, the manifest will not hit cache in direct mode since the change of size, but ccache will hit the cache in preprocessor mode, and overwrite the old manifest.

Later, I rollback pub.h, and built one more time, it missed cache in direct mode again, and rollbacked manifest.

Should we put the multiple possible information in manifest instead of simply overwriting?

We use ccache with distributed nfs system in a giant C++ project development, developers may run pretest base on different commit or branch, manifest repeatedly overwrited again and again.

I built the binary with same commit, and I got the following hit rate, I think those all could be hitted in direct mode

> cache hit (direct)                  2150
cache hit (preprocessed)              81
cache miss                             1
cache hit rate                     99.96 %
cleanups performed                     0
files in cache                         1
cache size                         380.9 kB

Maybe we can use an option like "CCACHE_OVERWRITE_MANIFEST" to control this feature?

Looking forward to your reply

jrosdahl:
> First, I execute a build with a.cpp which include a header named pub.h, ccache created cache and manifest correctly.
> 
> Then, I define a macro in the public header pub.h
> 
> > #define A_CPP_NOT_USE "1"
> 
> When I build again, in the compiling of a.cpp, the manifest will not hit cache in direct mode since the change of size, but ccache will hit the cache in preprocessor mode,

Sounds expected so far.

> and overwrite the old manifest.

No, that's not what happens (unless there's some bug, of course). A manifest stores multiple entries and old entries are not overwritten.

It should be noted, though, that manifest files are intentionally not protected via file locks, so what can happen is that some entries may be lost if several compilations want to update same manifest file _at the same time_. The latest update will win and the manifest file won't be corrupted, only potentially missing the entry from the "loser compilation".

> Later, I rollback pub.h, and built one more time, it missed cache in direct mode again

That's unexpected. I cannot reproduce this. Here's a script you can test:

```sh
#!/bin/sh

ccache=ccache
cc=/usr/bin/gcc

echo '#include "pub.h"' >main.c
echo '' >pub.h
sleep 2 # Need to sleep in order to avoid invalidation of a too new header file
[...]

PR #495, opened 2019-11-29
Store mtime/ctime as -1 if sloppy_file_stat is not set

MizuhaHimuraki:
details see: https://github.com/ccache/ccache/issues/493

### DIFF
--- a/src/ccache.c
+++ b/src/ccache.c
@@ -1386,8 +1386,10 @@ update_manifest_file(void)
 		old_size = file_size(&st);
 	}
 
+	bool save_timestamp = (conf->sloppiness & SLOPPY_FILE_STAT_MATCHES);
+
 	MTR_BEGIN("manifest", "manifest_put");
-	if (manifest_put(manifest_path, cached_obj_hash, included_files)) {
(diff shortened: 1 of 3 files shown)


src/ccache.c
```diff
@@ -1371,8 +1371,10 @@ update_manifest_file(void)
 		old_size = file_size(&st);
 	}
 
+	bool save_timestamp = (conf->sloppiness & SLOPPY_FILE_STAT_MATCHES) || output_is_precompiled_header;
```
jrosdahl: I understand that the `|| output_is_precompiled_header` part is because of the CI error you got, but did you investigate why the test failed?

MizuhaHimuraki: The CI error is important and helpful, however I also found the answer in the source code.

https://github.com/ccache/ccache/blob/1a4929bfd73e63ebd1b369ee116f9c57f1931824/src/manifest.c#L391-L399

ccache will check mtime in file_info if it detected the cmd is about pch. So, I keep the timestamp if pch is detected.

jrosdahl: Excellent, thanks.

PR #507, opened 2020-01-14
Some trivial changes via clang-tidy moving towards C++

AlexanderLanin:
### Description ###

Since the project is moving towards C++ a lot of code can be transformed within minutes by clang-tidy. Of course the changes are quite trivial and almost not worth it. Just in their sum they will make the code feel a little bit more like C++.
I've excluded third_party directory.


totph:
Do you have the tooling to run clang tidy on the ccache code base? Then this could become part of the test suite with more and more checks enabled gradually to prevent regressions.

cristianadam:
For the CMake build system one just needs to pass `-DCMAKE_CXX_CLANG_TIDY=clang-tidy;-checks=*` to the configure step, and CMake will call `clang-tidy` on every cpp file with the same parameters it used to compile the file.

AlexanderLanin:
Sure, that's a good way. But keep in mind that most findings have to be resolved by hand. It's only a few that clang-tidy can auto fix. This would force every contributor to have & run clang-tidy.

AlexanderLanin:
Closed for the moment as I'm terrified of the merge conflicts vs #513

PR #508, opened 2020-01-16
Ignore files from VS (Code)

AlexanderLanin:
Visual Studio and Visual Studio Code create temporary/cache files which do not need to be committed


jrosdahl:
This looks fine, except that there are `\r\n` line endings instead of `\n`. Would you like to fix?

PR #509, opened 2020-01-16
Simplify parse_int slightly

AlexanderLanin:
Remove manual int boundary checks as std::stoi already does that (throws out_of_bounds in that case)

### DIFF
--- a/unittest/test_Util.cpp
+++ b/unittest/test_Util.cpp
@@ -329,6 +329,24 @@ TEST_CASE("Util::parse_int")
   CHECK_THROWS_WITH(Util::parse_int("0x"), Equals("invalid integer: \"0x\""));
   CHECK_THROWS_WITH(Util::parse_int("0x4"), Equals("invalid integer: \"0x4\""));
   CHECK_THROWS_WITH(Util::parse_int("0 "), Equals("invalid integer: \"0 \""));
+
+  // check boundary values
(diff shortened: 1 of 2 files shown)


AlexanderLanin:
One of the few places which had unit tests so I feel I can try to change something here without breaking everything :-)

unittest/test_Util.cpp
```diff
@@ -329,6 +329,24 @@ TEST_CASE("Util::parse_int")
   CHECK_THROWS_WITH(Util::parse_int("0x"), Equals("invalid integer: \"0x\""));
   CHECK_THROWS_WITH(Util::parse_int("0x4"), Equals("invalid integer: \"0x4\""));
   CHECK_THROWS_WITH(Util::parse_int("0 "), Equals("invalid integer: \"0 \""));
+
+  // check boundary values
+  if (sizeof(int) == 2) {
```
jrosdahl: Just curious, do you have access to a system that have 16 bit `int`s?

AlexanderLanin: (Un)fortunately no. Especially as a host system that would be kind of strange. The ~2 test lines are quite straight forward and I hope I got them right for 2 byte ints.

However, I did not look at the context of parse_int. I was just simplifying it. Maybe it should be parse_long. I don't know.

PR #511, opened 2020-01-16
Failed tests create a testdir.failed symlink

totph:
When the test suite fails 'testdir.failed' now points to the random
'testdir.9876/'  directory. This make iterative debugging of failing
tests easier by always using the symlink name or re-issuing 'cd $PWD'
when 'cd'-ed into the 'testdir.failed' directory.

All artifacts can still be removed via 'rm -r testdir.*'.

### DIFF
--- a/test/run
+++ b/test/run
@@ -53,7 +53,8 @@ test_failed() {
     echo "ccache -s:"
     $CCACHE -s
     echo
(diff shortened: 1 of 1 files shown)


jrosdahl:
Thanks.

PR #513, opened 2020-01-16
Preparation for simplifying args structure

AlexanderLanin:
I was going to convert the `struct args` to something simple like `std::deque<std::string>` but noticed it was "slightly" more than a few lines that would need changing.
As far as I can tell every commit in this branch is fully working.

Creating draft pull request for initial feedback and since I just saw that @jrosdahl has changed similar files in the last ~2 hours.

### DIFF
--- a/src/args.cpp
+++ b/src/args.cpp
@@ -1,5 +1,5 @@
 // Copyright (C) 2002 Andrew Tridgell
-// Copyright (C) 2009-2020 Joel Rosdahl and other contributors
+// Copyright (C) 2009-2019 Joel Rosdahl and other contributors
 //
 // See doc/AUTHORS.adoc for a complete list of contributors.
 //
@@ -21,10 +21,10 @@
 
 #include "legacy_util.hpp"
 
-struct args*
+Args
 args_init(int init_argc, const char* const* init_args)
 {
-  struct args* args = (struct args*)x_malloc(sizeof(struct args));
+  Args args;
   args->argc = 0;
   args->argv = (char**)x_malloc(sizeof(char*));
   args->argv[0] = NULL;
@@ -34,13 +34,13 @@ args_init(int init_argc, const char* const* init_args)
   return args;
 }
 
-struct args*
+Args
 args_init_from_string(const char* command)
 {
   char* p = x_strdup(command);
   char* q = p;
   char *word, *saveptr = NULL;
-  struct args* args = args_init(0, NULL);
+  Args args;
   while ((word = strtok_r(q, " \t\r\n", &saveptr))) {
     args_add(args, word);
     q = NULL;
@@ -50,15 +50,16 @@ args_init_from_string(const char* command)
   return args;
 }
 
-struct args*
+nonstd::optional<Args>
 args_init_from_gcc_atfile(const char* filename)
 {
   char* argtext;
   if (!(argtext = read_text_file(filename, 0))) {
-    return NULL;
+    return {};
   }
 
-  struct args* args = args_init(0, NULL);
+  Args args;
+
   char* pos = argtext;
   char* argbuf = static_cast<char*>(x_malloc(strlen(argtext) + 1));
   char* argpos = argbuf;
@@ -127,8 +128,8 @@ out:
   return args;
 }
 
-struct args*
-args_copy(struct args* args)
+Args
+args_copy(const Args& args)
 {
   return args_init(args->argc, args->argv);
 }
@@ -140,7 +141,7 @@ args_copy(struct args* args)
 // src is consumed by this operation and should not be freed or used again by
 // the caller.
 void
-args_insert(struct args* dest, int index, struct args* src, bool replace)
+args_insert(Args& dest, const int index, Args& src, const bool replace)
 {
   // Adjustments made if we are replacing or shifting the element currently at
   // dest->argv[index].
@@ -190,22 +191,18 @@ args_insert(struct args* dest, int index, struct args* src, bool replace)
 }
 
 void
-args_free(struct args* args)
+args_free(Args& args)
 {
-  if (!args) {
-    return;
-  }
   for (int i = 0; i < args->argc; ++i) {
     if (args->argv[i]) {
       free(args->argv[i]);
     }
   }
   free(args->argv);
-  free(args);
 }
 
 void
-args_add(struct args* args, const char* s)
+args_add(Args& args, const char* s)
 {
   args->argv = (char**)x_realloc(args->argv, (args->argc + 2) * sizeof(char*));
   args->argv[args->argc] = x_strdup(s);
@@ -215,7 +212,7 @@ args_add(struct args* args, const char* s)
 
 // Add all arguments in to_append to args.
 void
-args_extend(struct args* args, struct args* to_append)
+args_extend(Args& args, const Args& to_append)
 {
   for (int i = 0; i < to_append->argc; i++) {
     args_add(args, to_append->argv[i]);
@@ -224,7 +221,7 @@ args_extend(struct args* args, struct args* to_append)
 
 // Pop the last element off the args list.
 void
-args_pop(struct args* args, int n)
+args_pop(Args& args, int n)
 {
   while (n--) {
     args->argc--;
@@ -235,7 +232,7 @@ args_pop(struct args* args, int n)
 
 // Set argument at given index.
 void
-args_set(struct args* args, int index, const char* value)
+args_set(Args& args, const int index, const char* const value)
 {
   assert(index < args->argc);
   free(args->argv[index]);
@@ -244,7 +241,7 @@ args_set(struct args* args, int index, const char* value)
 
 // Remove the first element of the argument list.
 void
-args_remove_first(struct args* args)
+args_remove_first(Args& args)
 {
   free(args->argv[0]);
   memmove(&args->argv[0], &args->argv[1], args->argc * sizeof(args->argv[0]));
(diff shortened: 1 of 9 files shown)


AlexanderLanin:
Maybe I should squash a little since the commits are rather intensely small

AlexanderLanin:
Sorry I messed up the merge slightly in git so it resulted in two commits. Not sure what happened. But it's all there.

Next up:
* change Args to a std::vector/deque/list, and then from char* to std::string.
* return Args by value instead of out-parameters

But it's already merge-worthy as it is (Especially as it touches so many lines it's a mess when it gets out of sync).
Some of the changes might be non obvious why they are there. It's because I have switched to deque locally and already know what has to be prepared to enable that commit to be simple. Nevertheless also those should already make sense as simple cleanups.

jrosdahl:
Thanks. `struct args` is naturally a good candidate for C++-ification, something that has been on the to do list (although with low priority for me at the moment).

Before reviewing details I would like to understand what the benefit would be to not go to `std::deque` right away. Won't many (or even most) of the lines modified by this PR be modified again in the follow-up commits? I would prefer not reviewing too much that will be thrown away anyway.

Also, I would like to know what your end goal looks like. Will `Args` be an alias of `std::deque<std::string>` with accompanying free functions for things like `args_init_from_string` and `args_strip`, or will it be an `Args` class that wraps `std::deque<std::string>`?

AlexanderLanin:
Honestly, I do not know. As the change is big enough as it is I stopped here for a review.
While you might be right that lines will change again, it would simply be an unmanageable huge change.
Also the change of those lines would be trivial as the big complex thing is changing them to value semantics. Globally removing ".argv" is kind of trivial.

This PR moves Args to value semantics without standard lib.
Next PR should change Args.argv from char** to some_standard_container<char*> and remove argc.
Next PR will change char* to std::string.

Therefore my current plan ends at:
`struct Args { std::deque<std::string> argv; }`
`void keep_free_functions_as_they_are_now(Args&, ...)`

After that is done and we have the code, we can discuss whether Args should be a typedef with free functions or a class. But I suspect that compared to the 3 PRs mentioned above this last optimization is kind of irrelevant.

jrosdahl:
Thanks for the clarification.

Looking at the current large diff and the description of the future PRs, I can't help thinking that there should be a way to make the bulk of the changes in smaller increments (e.g. function by function) and with fewer conversion steps. I have some sketchy ideas on how it could be done and also what I think the end goal should be; I'll try to write them down today or tomorrow (or reject them if they turn out to be infeasible) and ask for your opinions.

AlexanderLanin:
I would say this got out of hand when I started touching execv and execute... which should be removed from this PR. Also it is partially resolved by 42939f091a1ba23f98cd now.
I'm now attempting such a C-to-C++ struct change with more manageable steps in #518.

I have a branch where each of the args free functions was one commit. That's what I had before squashing. It was a useless amount of commits and changes.
What I can (and should) certainly try is to change args but change as little of the other code as possible. Without converting all the functions from pointers to references in the same PR etc.

totph:
Regarding big, possibly conflicting changes: I am currently removing most global state (see #515, more to follow). This will also touch almost all function signatures. One of these project should probably be finished first to avoid needless rebase conflicts. What do you want to merge first @jrosdahl?

AlexanderLanin:
How far is #515? 99%? Then I'd propose #515 to go in first. It's massive. Maybe not regarding lines touched, but regarding how they are touched (Is the change even review-able?)

jrosdahl:
@AlexanderLanin wrote:
> I'm now attempting such a C-to-C++ struct change with more manageable steps in #518.

I don't have bandwidth to look at #518 at the moment, sorry. All my spare time this evening went to this:

I wrote:
> I'll try to write them down today or tomorrow (or reject them if they turn out to be infeasible) and ask for your opinions.

Here's my idea: 6ae3cd615cc3. I also have a mostly working proof-of-concept implementation, including updated test cases, but I didn't manage to finish the last parts today. But I'm quite confident that the approach will work if we choose this path. @AlexanderLanin: What do you think?

@totph wrote:
> One of these project should probably be finished first to avoid needless rebase conflicts. What do you want to merge first @jrosdahl?

I agree with @AlexanderLanin, let's go with your work first.

AlexanderLanin:
Looks like the same idea as in #520 with overloading operators to make it backwards compatible. However you made some better choices along the way where to apply it. I've closed #520.

I guess we have now something like ~4 implementations which have the same direction and fulfill it in different degrees. We could try 5 more for the "best incremental approach", but that's just too much work.
Having said that, I think yours is too big to be sure it doesn't change any functional behavior.
However having yours as a goal I've modified mine to better fit that direction.
Yours can be rebased onto this PR with only 2 conflicts! All the rest stays! So it's 0 wasted effort to merge this one first. I think it's a good review-able intermediate step before applying yours on top.

(Current state: compiles, links, tests fail. I'll not fix it now since there is a huge rework coming anyway after #515)

totph:
Ok, I should be done sometime this weekend.

jrosdahl:
> I guess we have now something like ~4 implementations which have the same direction and fulfill it in different degrees. We could try 5 more for the "best incremental approach", but that's just too much work.

I'm not sure what you're getting at. I haven't proposed trying out more approaches.

> Having said that, I think yours is too big to be sure it doesn't change any functional behavior.

I'm fine if the tests pass and the code looks sane. Do you have any specific concerns? If so please share them.

> However having yours as a goal I've modified mine to better fit that direction.

When I wrote _"I have some sketchy ideas on how it could be done and also what I think the end goal should be; I'll try to write them down today or tomorrow [...] and ask for your opinions"_ I was hoping to convey that it would be a good idea to pause your implementation effort until we agree on the end goal and how to get there. Perhaps I should have been more clear.

> Yours can be rebased onto this PR with only 2 conflicts! All the rest stays!

That may be the case if you by "yours" mean 6ae3cd6, but 6ae3cd6 is only a subset of what I did (_"I also have a mostly working proof-of-concept implementation"_), so no it can't. For the record, I'm talking about 53ea24c6 (now fully working).

> So it's 0 wasted effort to merge this one first. I think it's a good review-able intermediate step before applying yours on top.

I don't understand why it would be a good intermediate step, sorry.

AlexanderLanin:
Absolutely fine for me. I'm just too accustomed to pull requests being rejected for being too big, I'm just damaged in that way ;-) Where I work no one would even consider reviewing a change this big.

On first glance:
* It definitely makes sense to have methods within Args instead of free functions. It's just easier to read as many have an in&out parameter.
* Some of the new functions can probably be named better like add_prefix --> add_to_front / push_front, strip_with_prefix --> remove_all_with_prefix
* Feels like ArgvKeeper is overly complex. Does it really have to be able to outlive Args? Otherwise a `std::vector<char*> to_argv()` within Args would suffice. No copying, to memory management.

jrosdahl:
> Absolutely fine for me. I'm just too accustomed to pull requests being rejected for being too big, I'm just damaged in that way ;-) Where I work no one would even consider reviewing a change this big.

Right, I agree that smaller would be better. (On the other hand, it's of the same magnitude in terms of modified lines as both the first and the current revisions of this PR.)

I guess you have to understand the asymmetry and dynamics of unpaid FOSS maintainer work for this to make sense... Since I maintain ccache in my limited spare time, I (just like many maintainers) have to value my time higher than contributors' time. This may not sound very nice, but it's more or less how it has to be. And not having to review many revisions of something that touches many lines of code saves me time.

> On first glance:

Thanks for your thoughts, much appreciated. Would you be interested in performing a more thorough review once #515 and follow-up PRs have landed?

> [...]
> * Feels like ArgvKeeper is overly complex. Does it really have to be able to outlive Args? Otherwise a `std::vector<char*> to_argv()` within Args would suffice. No copying, to memory management.

No, it doesn't have to outlive `Args` but it has to outlive the call to `Args::to_argv`. With a `std::vector<char*> Args::to_argv()` the caller would have to deallocate the memory manually after use, right? Or do you mean that `Args::to_argv` should not return a new snapshot of the argument array but a pointer an array kept in `Args`, letting `~Args` deallocate it? Sure, that would work in practice and would make the implementation simpler even though it doesn't feel as elegant to me.

AlexanderLanin:
@jrosdahl does it make sense to restart this now? I could help out by rebasing your latest branch?

jrosdahl:
> @jrosdahl does it make sense to restart this now?

Sure!

> I could help out by rebasing your latest branch?

Are you talking about https://github.com/jrosdahl/ccache/commit/fe814883a858f6c303b4a39d28101aaaa861d791? I have a local unpushed rebase of it that I did a while a go. It's not rebased to latest master, though, but I can check if there are many conflicts.

jrosdahl:
> Are you talking about [jrosdahl@fe81488](https://github.com/jrosdahl/ccache/commit/fe814883a858f6c303b4a39d28101aaaa861d791)? I have a local unpushed rebase of it that I did a while a go. It's not rebased to latest master, though, but I can check if there are many conflicts.

It was not much work to rebase it. Done now: [jrosdahl@dcff303e](https://github.com/jrosdahl/ccache/commit/dcff303ebcf747a551a407fe07e84b3f83fee6f2). @AlexanderLanin: Would you like to review it? If so I can make a pull request. Or do you suggest another approach or implementation?

AlexanderLanin:
Branch looks fine, can you create a pull request? I'd be happy to review.
Small conflict to #555 but we'll handle it.

jrosdahl:
> Branch looks fine, can you create a pull request? I'd be happy to review.

OK, thanks. Done in #575.

PR #515, opened 2020-01-19
Make cc_process_args() only use local variables

totph:
plus

Simplify make_relative_path() argument and result ownership

because all the make_relative_path(x_strdup(..)) + free() calls were getting out of hand.

### DIFF
--- a/src/ccache.cpp
+++ b/src/ccache.cpp
@@ -19,6 +19,7 @@
 
 #include "ccache.hpp"
 
+#include "ArgsInfo.hpp"
 #include "Error.hpp"
 #include "FormatNonstdStringView.hpp"
 #include "ProgressBar.hpp"
@@ -33,6 +34,7 @@
 #include "hash.hpp"
 #include "hashutil.hpp"
 #include "language.hpp"
+#include "legacy_globals.hpp"
 #include "manifest.hpp"
 #include "result.hpp"
 #include "stats.hpp"
@@ -53,11 +55,6 @@
 #define TO_STRING(x) STRINGIFY(x)
 
 using nonstd::string_view;
-
-// Global variables used by other compilation units.
-extern char* current_working_dir;
-extern char* stats_file;
-extern unsigned lock_staleness_limit;
 
 static const char VERSION_TEXT[] = MYNAME
   " version %s\n"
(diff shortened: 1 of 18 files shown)


src/ccache.cpp
```diff
@@ -1730,7 +1568,7 @@ hash_nvcc_host_compiler(struct hash* hash,
 
 // Update a hash with information common for the direct and preprocessor modes.
 static void
-hash_common_info(struct args* args, struct hash* hash)
+hash_common_info(struct args* args, struct hash* hash, const ArgsInfo& ccargs)
```
jrosdahl: Perhaps rename ccargs to args_info?

src/ccache.cpp
```diff
@@ -3190,191 +3038,184 @@ cc_process_args(struct args* args,
...
+
+      result.output_obj = Util::change_extension(result.output_obj, extension);
     }
   }
 
-  if (seen_split_dwarf) {
-    char* p = strrchr(output_obj, '.');
-    if (!p || !p[1]) {
+  if (result.seen_split_dwarf) {
+    size_t pos = result.output_obj.find('.');
```
jrosdahl: This doesn't look correct; `strrchr` corresponds to `.rfind`.

src/ccache.cpp
```diff
@@ -3847,10 +3675,40 @@ ccache(int argc, char* argv[])
   // Arguments to send to the real compiler.
   struct args* compiler_args;
   MTR_BEGIN("main", "process_args");
-  if (!cc_process_args(
-        orig_args, &preprocessor_args, &extra_args_to_hash, &compiler_args)) {
+  ArgsInfo ccargs = cc_process_args(g_config,
```
jrosdahl: Rename to `args_info`?

src/legacy_util.cpp
```diff
@@ -780,6 +780,29 @@ format_parsable_size_with_suffix(uint64_t size)
...
 }
 
+// Convert an owned char* string `str` to an std::string and free `str`.
+std::string
+from_owned_cstr(char* str)
+{
+  std::string result;
+  if (str) {
+    result = str;
+    delete (str);
```
jrosdahl:
```suggestion
    delete str;
```

src/legacy_util.cpp
```diff
@@ -780,6 +780,29 @@ format_parsable_size_with_suffix(uint64_t size)
...
+
+// Convert a char* string `str` to a std::string, if `str` is NULL return "".
+std::string
+from_cstr(const char* str)
+{
+  std::string result;
+  if (str) {
+    result = str;
+  }
+  return result;
```
jrosdahl:
```suggestion
  return str ? str : {};
```

totph:
Extracted and tested the `from_cstr` functions and added a ScopeGuard to make returning bool easier. All other issues are addressed.

src/ccache.cpp
```diff
@@ -1168,9 +1016,8 @@ result_name_from_depfile(const char* depfile, struct hash* hash)
       if (!has_absolute_include_headers) {
         has_absolute_include_headers = is_absolute_path(token);
       }
-      char* path = make_relative_path(x_strdup(token));
+      std::string path = make_relative_path(token).c_str();
```
AlexanderLanin:
```suggestion
      std::string path = make_relative_path(token);
```

src/ccache.cpp
```diff
@@ -2766,95 +2621,88 @@ cc_process_args(struct args* args,
...
-      profile_arcs = true;
-      generating_coverage = true;
+      args_info.profile_arcs = true;
+      args_info.generating_coverage = true;
       args_add(common_args, argv[i]);
       continue;
     }
     if (str_startswith(argv[i], "-fprofile-dir=")) {
-      profile_dir = x_strdup(argv[i] + 14);
+      args_info.profile_dir = from_owned_cstr(x_strdup(argv[i] + 14));
```
AlexanderLanin: Isn't this the same?
```suggestion
      args_info.profile_dir = from_cstr(argv[i] + 14);
```

jrosdahl:
Thanks (and thanks @AlexanderLanin for the review)!

PR #517, opened 2020-01-22
Collect forward declarations in one file

totph:
Plus update copyright information of edited files and
finally fix formatting of CacheEntryWriter.cpp.

### DIFF
--- a/src/Config.hpp
+++ b/src/Config.hpp
@@ -20,6 +20,7 @@
 
 #include "system.hpp"
 
+#include "Forward.hpp"
 #include "legacy_util.hpp"
 
 #include "third_party/fmt/core.h"
@@ -30,7 +31,6 @@
 #include <string>
 #include <unordered_map>
 
-class Config;
 extern Config g_config;
 
 class Config
--- a/src/Forward.hpp
+++ b/src/Forward.hpp
@@ -1,4 +1,4 @@
-// Copyright (C) 2019-2020 Joel Rosdahl and other contributors
+// Copyright (C) 2020 Joel Rosdahl and other contributors
 //
 // See doc/AUTHORS.adoc for a complete list of contributors.
 //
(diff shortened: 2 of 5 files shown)


jrosdahl:
> Collect forward declarations in one file

I'm missing why you think that it would be a good idea. To me the current approach is superior; my instinct is to strive to not lump unrelated things together.

> finally fix formatting of CacheEntryWriter.cpp.

That formatting is due to a known bug in clang-format. It's unfortunate but the fix will disappear next time somebody runs `make format`, or (like I do) edit the code with automatic code formatting turned on, so we'll just have to live with it. Or, if you think that it's extremely annoying, add `// clang-format off/on` comments, but I don't think we should bother.

totph:
> > Collect forward declarations in one file
> 
> I'm missing why you think that it would be a good idea. To me the current
> approach is superior; my instinct is to strive to not lump unrelated things together.

These things are indeed unrelated, but then most header files would only contain a single forward include instead of having several copies of the same forward declarations in various files. So, fewer lines and duplications, but one "mudball" file. A matter of taste, certainly.

> bug in clang-format.

Ah, that was it, 7 is the last version with the bug, my mistake. I updated and the issue went away.

AlexanderLanin:
Hi, why are there even forward declarations? Currently I see you have 3, one of which would be removed by followup of #518.

The reason I'm doubtful about forward declaration is that I've heard several people talk about potential overuse: The point being the cpp files that include the header in question need the definition anyway: the implementer needs to work on it and the caller needs to provide it. Exceptions are potentially rare. Forward declarations are especially useful when you headers are so huge, that you don't need most of it in most of the cpp files - however this has been resolved in parallel.

totph:
Forward declarations are required for two mutually dependent files (which was the reason for this commit, it was unintentional and might be better now with the "split ccache.hpp" commit). But they are also a way to cut down on compile times by reducing the number of required includes in header files by only using pointers or references instead of the value type. This project might be small and still compile quickly, but everything counts!

jrosdahl:
@totph: Are you fine with me passing on this one?

totph:
Sure, it's a purely cosmetic change.

PR #518, opened 2020-01-22
Refactoring: make 'struct hash' behave as a value semantic object

AlexanderLanin:
Currently hash is allocated on heap which makes it necessary to handle it as a pointer and to remember to free it. While this commit is obviously not the final achievable result it seems trivial to integrate it while already achieving fewer memory leaks (e.g. cpp_hash in ccache.cpp)


src/hash.cpp
```diff
@@ -63,32 +64,37 @@ do_debug_text(struct hash* hash, const void* s, size_t len)
...
 {
-  free(hash);
+  pimpl->state = other.state;
+  pimpl->debug_binary = NULL;
+  pimpl->debug_text = NULL;
 }
 
+
+// Force destructor to cpp file.
+// Required so unique_ptr can call delete on a known object.
```
jrosdahl:
```suggestion
// The destructor needs to be in the cpp file for std::unique_ptr to see
// the complete class definition.
```

src/hash.hpp
```diff
@@ -42,14 +44,25 @@ bool digests_equal(const struct digest* d1, const struct digest* d2);
 // struct hash represents the hash algorithm's inner state.
 struct hash;
 
-// Create a new hash state.
-struct hash* hash_init(void);
+// value semantic hash to replace 'struct hash' in the future
+class Hash {
```
jrosdahl: How about implementing the class in `Hash.hpp`/`Hash.cpp` instead?

AlexanderLanin: You are technically quite right, but as not everyone has a filesystem which is case sensitive it feels like this should be avoided. I would therefore rename the files as a last step.

jrosdahl: Right, I solved that problem for utils by renaming utils to legacy_utils. But OK.

jrosdahl:
@AlexanderLanin: Would you like to rebase this PR?

AlexanderLanin:
Merging this old PR is more effort then restarting from scratch. Closing PR.

PR #520, opened 2020-01-23
Refactoring: make struct args a value semantic object

AlexanderLanin:
By converting args to a value semantic object, sometimes it becomes Args, sometimes Args& and sometimes Args*.

- Option 1) Change all code where args is on the stack to use &args to call the Args* functions.
This would need to be undone once those functions accept Args&.
Hunderts of lines changed to one direction and then back.
- Option 2) Change all code where args is a pointer to use *args to call the Args& functions.
This would need to be undone once those args become Args&.
Hunderts of lines changed to one direction and then back.
- Option 3) Allow mixed mode for the moment. Implicit cast to pointer is introduced within struct Args.

### DIFF
--- a/src/args.cpp
+++ b/src/args.cpp
@@ -1,5 +1,5 @@
 // Copyright (C) 2002 Andrew Tridgell
-// Copyright (C) 2009-2020 Joel Rosdahl and other contributors
+// Copyright (C) 2009-2019 Joel Rosdahl and other contributors
 //
 // See doc/AUTHORS.adoc for a complete list of contributors.
 //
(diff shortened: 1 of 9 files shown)


AlexanderLanin:
@jrosdahl please give me some feedback how/if we want to proceed here. At the moment I would propose very small very trivial pull requests. They are trivial but affect many lines and constantly run into merge conflicts. So small changes which are easily and quickly review-able seem the way to go. Like this one.
But I'd like to resolve conflicts only on one of the PRs :-)

jrosdahl:
See https://github.com/ccache/ccache/pull/513#issuecomment-577897735.

PR #522, opened 2020-01-23
Make Config noncopyable and add clear_and_reset() method

totph:
Prevents changes from getting lost by writing them to a
temporary copy of the main Config object.

Also restore deleted comments from the move of
primary/secondary_config_path.


totph:
If tests require copies a `clone()`/`clone_into(Config& other)` method could be added.

PR #525, opened 2020-01-27
Added 'make check_format' to makefile rules

totph:
clang-format has no option to change the exit code, so
check manually. Additionally, clang-format < 9 updates
timestamps of all checked files, this works around that.

### DIFF
--- a/dev.mk.in
+++ b/dev.mk.in
@@ -199,7 +199,18 @@ shellcheck: test/suites/*.bash
 
 .PHONY: format
 format:
-	$(CLANG_FORMAT) -i $(non_third_party_headers) $(non_third_party_sources) $(test_sources)
+	@echo $(non_third_party_headers) $(non_third_party_sources) $(test_sources) | xargs -n1 -P8 \
+	    misc/clang-format.sh
+
+# Not using parallel execution because target is most likely being run on non-interactive CI system,
+# so no user is waiting for immediate results, and it avoids possibly interleaved output.
+.PHONY: check_format
+check_format:
+	@[ -t 1 ] && export cf_diff_color="--color=always"; \
+	echo $(non_third_party_headers) $(non_third_party_sources) $(test_sources) | xargs -n1 -P1 \
+	    misc/clang-format.sh --check || \
+		{  echo; echo "Error: Sources are not formatted with clang-format."; \
+		   echo 'Run "make format" or apply the above diff.'; }
 
 # pip install compiledb
 compile_commands.json:
--- a/misc/clang-format.sh
+++ b/misc/clang-format.sh
@@ -1,0 +1,46 @@
+#!/bin/bash
+
+# Reformat the given C++ file if required, or if --check
+# is specified, only print the diff.
+# Exits with 0 if the file was already formatted correctly.
+
+set -eu
+
+clang_format="${CLANG_FORMAT:-clang-format}"
+cf_diff_color="${cf_diff_color:-}"
+
+if [[ "${1:-}" == "--check" ]]; then
+    shift
+    check=true
+else
+    check=false
+fi
+
+if [[ $# -ne 1 ]]; then
+    echo "Usage: $0 [--check] file.cpp"
+    exit 1
+fi
+
+file="$1"
+
+if [[ ! -e "$file" ]]; then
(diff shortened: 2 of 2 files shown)


totph:
Weird, `.ONESHELL:` directive in dev.mk breaks libb2 integration:

    make[1]: *** src/third_party/libb2: No such file or directory.  Stop.
    Makefile:170: recipe for target 'src/third_party/libb2/src/.libs/libb2.a' failed
    make: *** [src/third_party/libb2/src/.libs/libb2.a] Error 2
    The command "make" exited with 2.

totph:
I just noticed that clang-format 9 no longer updates all timestamps. But maybe the exit code is useful for CI.

AlexanderLanin:
FYI: I was playing around with formatting the last change (branch, PR, ...) only and I would conclude it's possible, but complexity and effort are too high. Just format the whole damn thing. Also I was not aware of `make format`.
Partially implemented in https://github.com/AlexanderLanin/ccache/pull/2/files
After yours is merged I'll create a PR for travis integration.

Review:
* It's significantly slower: (real 0.6s, user+sys 0.6s) --> (real 2.3s, user+sys 2.0s). Not a deal breaker, but it went from nothing to noticeable (That's why I measured).
* A different (not necessarily better) approach would be to require all changes to be staged/committed (as in git-clang-format), then format in-place and run git-diff. Maybe that's faster due to less temp files?
* Yet another would be to use https://github.com/Sarcasm/run-clang-format.

jrosdahl:
How about a `make format` that just reformats (and exits with status 0) plus a `make check_format` that dry runs formatting and exits with error if changes are detected? The latter could also print a diff of the expected changes so that it would be possible to apply the diff without a having local installation of clang-format. Both could relay to the same script in the `misc` directory or so.

Regarding `src/CacheEntryWriter.cpp` I didn't realize that the formatting bug only exists in newer clang-format versions (looks like it was introduced in version 8 and I run 9), so that's more annoying than [I thought](https://github.com/ccache/ccache/pull/517#issuecomment-577362562). Worked around in 21e95ec61741.

totph:
> https://github.com/Sarcasm/run-clang-format.

Indeed, why re-invent the wheel.

> make check_format

Added. The error message could probably be improved, suggestions?:
"Error: Code needs to be reformatted, run 'make format' or apply the patch above."

AlexanderLanin:
You should probably include https://github.com/Sarcasm/run-clang-format/blob/master/LICENSE into misc/run-clang-format-license or at least add some link to it.

totph:
@AlexanderLanin good point, done.

Ironically, `run-clang-format.py` is _not_ black formatted (python `clang-format` equivalent), but I left it like that.

jrosdahl:
I agree that not reinventing the wheel is typically a good idea, but in this case the Sarcasm/run-clang-format wheel seems a bit overkill to me. And it doesn't solve the problem of `make format` updating timestamps of all files when using clang-format < 9, right?

I was thinking of a simple shell script like this:

```
#!/bin/bash

set -eu

tmp_file=$(mktemp)
trap "rm $tmp_file" EXIT

if [ "${1:-}" = --check ]; then
    shift
    check=true
else
    check=false
fi

clang_format=${CLANG_FORMAT:-clang-format}

status=0
for file in "$@"; do
    $clang_format "$file" >"$tmp_file"
    if ! cmp -s "$file" "$tmp_file"; then
        if $check; then
            echo "Error: $file not formatted with clang-format"
            echo 'Run "make format" or apply this diff:'
            diff -u "$file" "$tmp_file" |
                sed -r -e "s!^---.*!--- a/$file!" -e "s!^\+\+\+.*!+++ b/$file!"
        else
            echo "Reformatted $file"
            cp "$tmp_file" "$file"
        fi
    fi
done

exit $status
```

totph:
> timestamps / clang-format < 9

Not only that, but running clang-format with `-verbose` just prints all files, even if no reformatting took place. Great program, terrible UI.

> slower

To speed up the sequential formatting, the makefile now uses `xargs` so that typing `make format` in a shell or even having a pre-commit hook for it is less annoying (8 -> 16 only yielded .1 fewer seconds). And the `check_format` diff can be colored now.

PR #526, opened 2020-01-27
Move all global state into a new Context struct 

totph:
Work in Progress


totph:
This would be the next step in removing global state.

The first commit mostly adds "Context& ctx" to nearly all function signatures.

The following commits move a single global variable into Context. Every intermediate step at least compiles.

Not all variables are moved (counter_updates, ignore_headers), however the unit test succeed already.

Is there another way to move these without adding a global <-> Context shim for each variable, then then removing it again?

And if this is the way to go should  all these commits be squashed into a single one after the review?

jrosdahl:
> This would be the next step in removing global state.

Thanks for creating an early WIP PR; that makes it easier to understand where you aim to go.

Next time, if you want, you could create a [draft PR](https://github.blog/2019-02-14-introducing-draft-pull-requests/).

> Is there another way to move these without adding a global <-> Context shim for each variable, then then removing it again?

Not sure what you mean here, please expand.

> And if this is the way to go should all these commits be squashed into a single one after the review?

I'd like to keep the individual commits (as long as they are tidy, related, not fixup commits, etc., and as you say at least compile). But perhaps you could batch them in a (small) number of PRs, like 10 commits each or so? That would make reviewing more manageable.

I haven't looked much at the actual changes, but one thing that I happened to see is moving stats_info to Context which seems strange to me. It's just a local constant that I believe is good to keep near the stats code.

I hope that we can round off the large refactorings reasonably soon since I would like to start thinking about how and when to get 4.0 released.

AlexanderLanin:
> I hope that we can round off the large refactorings reasonably soon since I would like to start thinking about how and when to get 4.0 released.

How about some editable overview / list of possible large scale refactorings with some possibility to extend, edit and discuss it? e.g. Issue, Wiki, etc. As they are kind of blocking everything. Noone wants to merge to such a change :D
The next ones I personally see (in no order) are the Args change, Hash change, converting more function to use std::string and running clang-tidy in auto-fix mode (e.g. add const everywhere).
However none of them add any features. All these change "only" enable easier adding of features afterwards.

src/ccache.cpp
```diff
@@ -4048,7 +4048,7 @@ int
 ccache_main(int argc, char* argv[])
 {
   try {
-    // TODO ptr, and LEAK to be available in atexit()
+    // TODO! ptr, and LEAK to be available in atexit()
     Context* ctx = new Context{};
```
totph: Because `Context` is passed as a `void* context` to `atexit`/`exit_fn`-called-functions it can't be a stack variable. To prevent leaks a final exit function is required to to `delete` it, though maybe only on valgrind runs.

src/ccache.cpp
```diff
@@ -3618,7 +3618,7 @@ ccache(Context& ctx, int argc, char* argv[])
 
   initialize(ctx);
 
-  // TODO, now calling init log HERE, not in vlog cc_log_argv
+  // TODO!, now calling init log HERE, not in vlog cc_log_argv
   // i.e. these functions MUST NOT be called until this line runs.
   init_log(ctx.config);
```
totph: `init_log()` was called automatically in the first `cc_log()` call. Either init now happens before any log calls or
`Config::debug()` / `Config::log_file()` (these are required in `init_log()` become global variables. The latter would be easier because `warn_log_fail` **also** requires  `Config::log_file()`.

This means Config would parse these variables and then write them into global variables.

src/legacy_util.cpp
```diff
@@ -186,7 +192,7 @@ static void warn_log_fail(void) ATTR_NORETURN;
 static void
 warn_log_fail(void)
 {
-  // TODO -- fake, legacy debug globals
+  // TODO! -- fake, legacy debug globals
   Config cfg;
```
totph: See `init_log()` above.

src/legacy_util.cpp
```diff
@@ -62,18 +62,24 @@
 #  include <tchar.h>
 #endif
 
+// TODO: remaining non-const globals
+// move Config::log_file() into a global?
+// would enable init_log() without args again
+
```
totph: All the log / syslog related state should probably stay global.

src/stats.cpp
```diff
@@ -293,7 +293,7 @@ stats_flush(void* context)
   Context& ctx = *static_cast<Context*>(context);
   stats_flush_to_file(ctx, ctx.stats_file, ctx.counter_updates);
 
-  // TODO: is cleanup in Context OK?
+  // TODO!: is cleanup in Context OK?
   //  counters_free(ctx.counter_updates);
   //  ctx.counter_updates = NULL;
```
totph: (Ignore, not relevant for now)

totph:
Now with a fresh look at this refactoring PR I think it is possible to split it up into smaller ones which not only compile but pass the CI tests with every commit.

Currently the first commit in this series adds `Context&` to all function signatures. However some changes do more, such as expect a usable `ctx.config` element which does not happen until later. Some still lacked `Context&` which I then added in the last 10 commits.

I will fix this and start with a commit which **only** changes the function signatures and adds an empty `Context` heap object in ccache_main (plus `-Wno-error=unused-parameter`).

Then all the globals can be moved on commit at a time.

Before I can start I have a few minor questions I just asked for in the review.



>>  Is there another way to move these without adding a global <-> Context shim for each variable, then then removing it again?

> Not sure what you mean here, please expand.

I first feared a clean "just-add-Context&" start was not doable so I thought about writing to the Context::some_var and via a setter to the global some_var at the same time to stay in sync, but that should not be required.

> moving stats_info 

I was probably on autopilot, will undo.

> I haven't looked much at the actual changes.

This is mostly here in one dump to check that everything passes the CI in the end. However the first one (not each line, but the concept), the "TODO/WIP: init_log() moved!" and the "stats flush: void* context ptr" might be worth a look.

jrosdahl:
@AlexanderLanin wrote:
> How about some editable overview / list of possible large scale refactorings with some possibility to extend, edit and discuss it? e.g. Issue, Wiki, etc. As they are kind of blocking everything. Noone wants to merge to such a change :D
> The next ones I personally see (in no order) are the Args change, Hash change, converting more function to use std::string and running clang-tidy in auto-fix mode (e.g. add const everywhere).

I have a handful more refactorings on my personal to do list, but I'm a little afraid to open Pandora's box... :slightly_smiling_face: Yes, improving code quality of course makes sense, but in this case I want to go slow – it's OK if code looks C-ish for some time. I want to focus on other things for a while so I suggest stopping and thinking a bit after @totph's global state hunt and the C++-ification of struct args and struct hash. In particular I don't want a big race for converting C-style strings to C++-style; that has to be done little by little over a period of time.

jrosdahl:
@totph wrote:
> I will fix this and start with a commit which only changes the function signatures and adds an empty Context heap object in ccache_main (plus -Wno-error=unused-parameter).
>
> Then all the globals can be moved on commit at a time.

Sounds OK.

Another approach could be to add context parameters to functions only when needed, i.e. as a side effect of moving a global variable into Context. Is this something you have considered?

src/ccache.cpp
jrosdahl: Sounds reasonable to pass ownership of ctx to the last executed exit function.

src/ccache.cpp
jrosdahl: Not sure if I have the full picture yet, but running `init_log` early instead of at the first `cc_*log*` call makes sense.

src/legacy_util.cpp
jrosdahl: Yes, leave them for now.

totph:
> add context parameters to functions only when needed

Sure, but then the first individual global->local commits become unevenly sized, depending on whether they are the first or second variable to be converted. That might make the individual diffs a bit harder to review.

jrosdahl:
> That might make the individual diffs a bit harder to review.

I don't think that will be the case. Not having a large "dummy commit" that only adds unused parameters feels more natural to me.

jrosdahl:
> I don't think that will be the case. Not having a large "dummy commit" that only adds unused parameters feels more natural to me.

It would also make it possible to do things gradually. Right now the first large commit will collide with almost any other non-trivial change to the code, but if it's done variable by variable (or in smaller batches) the merge conflicts will easier to handle. And there will be less pressure to get it all done before other larger efforts can commence. For instance, I currently feel a bit blocked by this; should I wait with some stuff to reduce the merging difficulties of this one or should I go ahead?

totph:
Ok, the first two vars are done.

I could continue with all the others in no time, but then the situation is more or less identical to this huge PR. If the other 3-commit PR is Ok, I'll instead selectively pick the globals first which change most of the function signatures. Then the largest change is out of the way. This will probably include `g_config` though!

jrosdahl:
@totph: I guess this should be closed now?

PR #533, opened 2020-02-06
Use Config for init_log()

totph:
Note that `initialize()` and therefore `init_log()` is called multiple times in `ccache_main_options()`, so for now `init_log()` still needs to tolerate being called multiple times.

### DIFF
--- a/src/logging.cpp
+++ b/src/logging.cpp
@@ -45,6 +45,9 @@
 // Destination for g_config.log_file.
 static FILE* logfile;
 
+// Path to the logfile.
+static std::string logfile_path;
+
 // Whether to use syslog() instead.
 static bool use_syslog;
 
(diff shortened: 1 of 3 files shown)


src/logging.cpp
```diff
@@ -44,6 +44,9 @@
 // Destination for g_config.log_file.
 static FILE* logfile;
 
+// Path to the logfile.
+static char* logfile_path;
```
jrosdahl: Make this an `std::string`?

src/logging.cpp
```diff
@@ -58,28 +61,32 @@ static size_t debug_log_size;
 
 #define DEBUG_LOG_BUFFER_MARGIN 1024
 
-static bool
-init_log(void)
+// Initialize logging. Call only once.
+bool
+init_log(const Config& conf)
```
jrosdahl: I think I'd prefer `config` over `conf`.

jrosdahl:
Obsoleted by #536.

Issue #530, opened 2020-02-05
Recompression with -X <level> leads to errors on Windows


mumpitzstuff:
### How to reproduce ###

1. ccache -X 6

### Actual behavior ###

In the logfile i can see a lot of such errors:
[.590345 21180] can't rename file C:\workspace\PRP_RH850\.ccache/0/0/bc6fa48a05eef2e131dc61603c637e04686ef5.result.tmp.unknown.21180.FzP5tM to C:\workspace\PRP_RH850\.ccache/0/0/bc6fa48a05eef2e131dc61603c637e04686ef5.result OS returned error: ../src/legacy_util.cpp failed with error 183: Eine Datei kann nicht erstellt werden, wenn sie bereits vorhanden ist.

### Environment ###

master branch

jrosdahl:
I can't reproduce this. Somebody running Windows who can help out?

jrosdahl:
Background: The error message comes from [`x_rename`](https://github.com/ccache/ccache/blob/aa64164acb6caf45e9eb5c3ca563d12ccb67ed3b/src/legacy_util.cpp#L1264) which is in its `_WIN32` version is supposed to remove the destination file before moving the source file to the destination path.

mumpitzstuff:
It seems a backslash is missing here.

PRP_RH850.ccache
must be
PRP_RH850\.ccache

Can it be that \. is interpreted as escaped dot? This would be wrong and could be the reason of the problem.

jrosdahl:
> PRP_RH850.ccache
> must be
> PRP_RH850\.ccache

Not sure I understand. Do you mean that you have set `CCACHE_DIR` to `C:\workspace\PRP_RH850\.ccache`?

What output do you get from `ccache -p`?

mumpitzstuff:
Exactly. `ccache -p` shows me the correct path. I only see the wrong path if I try to run the recompression command (the error is written to the ccache log file). Today I tried to replace all backslashes in all paths of the config with slashes. But recompression seems to be broken somehow anyway. I will try to dive deeper next week.

cristianadam:
I was able to reproduce the issue.

With this patch I was able to get better errors:
```patch
--- src/legacy_util.cpp
+++ src/legacy_util.cpp
@@ -1267,9 +1267,8 @@ x_rename(const char* oldpath, const char* newpath)
   return rename(oldpath, newpath);
 #else
   // Windows' rename() refuses to overwrite an existing file.
[...]

PR #535, opened 2020-02-06
Win32: Close reading file for recompression

cristianadam:
Fixes: https://github.com/ccache/ccache/issues/530


jrosdahl:
Thanks!

PR #542, opened 2020-02-16
Context: g_config, stats_file and orig_args moved

totph:
Used the proof of concept #526 to decide between `Config&` or `Context&` if both were possible.

### DIFF
--- a/src/ccache.cpp
+++ b/src/ccache.cpp
@@ -157,7 +157,7 @@ static pid_t compiler_pid = 0;
 static const char HASH_PREFIX[] = "3";
 
 static void
-add_prefix(struct args* args, const char* prefix_command)
+add_prefix(Context& ctx, struct args* args, const char* prefix_command)
 {
   if (str_eq(prefix_command, "")) {
     return;
@@ -170,7 +170,7 @@ add_prefix(struct args* args, const char* prefix_command)
        tok = strtok_r(NULL, " ", &saveptr)) {
     char* p;
 
-    p = find_executable(tok, MYNAME);
+    p = find_executable(ctx, tok, MYNAME);
     if (!p) {
       fatal("%s: %s", tok, strerror(errno));
     }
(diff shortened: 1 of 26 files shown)


src/ccache.cpp
```diff
@@ -3638,16 +3636,17 @@ cache_compilation(int argc, char* argv[])
...
 
-    args_strip(orig_args, "--ccache-");
-    add_prefix(ctx, orig_args, ctx.config.prefix_command().c_str());
+    args_strip(ctx.orig_args, "--ccache-");
+    add_prefix(ctx, ctx.orig_args, ctx.config.prefix_command().c_str());
 
     cc_log("Failed; falling back to running the real compiler");
-    cc_log_argv("Executing ", orig_args->argv);
+    cc_log_argv("Executing ", ctx.orig_args->argv);
+    ctx.free_orig_args = false;
```
totph: ...instead of copying `orig_args` just before `exec`.

src/ccache.cpp
```diff
@@ -1101,12 +1101,12 @@ update_cached_result_globals(Context& ctx, struct digest* result_name)
...
                .c_str());
-  ctx.stats_file = format(
-    "%s/%c/stats", ctx.config.cache_dir().c_str(), result_name_string[0]);
+  ctx.stats_file =
+    fmt::format("{}/{}/stats", ctx.config.cache_dir(), result_name_string[0]);
 }
 
 static bool
-create_cachedir_tag(const std::string& dir)
+create_cachedir_tag(nonstd::string_view dir)
```
totph: Being optimistic here that `cc_log` will accept `string_view` soon.

src/ccache.cpp
```diff
@@ -3337,47 +3349,47 @@ cc_process_args(ArgsInfo& args_info,
...
+  if (ctx.config.run_second_cpp()) {
     args_extend(*extra_args_to_hash, dep_args);
   }
 
   return true;
 }
 
 static void
-create_initial_config_file(const std::string& path)
+create_initial_config_file(Config& config)
```
totph: argument was always `Config.primary_config_path()`

PR #543, opened 2020-02-17
Final Globals -> Context moves

totph:
This removes legacy_globals.cpp


AlexanderLanin:
Feels somewhat unlucky that Context is no longer const. It was **way** easier to understand the concept with a const as that was just "some Context which is apparently needed". Now that it's no longer const it's "some huge structure which contains everything and anything could change anywhere".
Would it be reasonable to extract the non const parts?

totph:
`Context` was always mutable because - just like the globals it replaces - various functions use it to pass data to other functions. Most functions read and write, only the few which only read from it get passed a `const Context&`.

Splitting it would mean adding a mix various const/mutable argument everywhere. It is more feasible  to split and simplify the functions so they only take consts and return the modified data. E.g. just for counter updates `mutable Context&` is required in a lot of places.

jrosdahl:
Thanks for the well-shaven yak!

> E.g. just for counter updates mutable Context& is required in a lot of places.

I have made some follow-up commits to improve on this.

PR #545, opened 2020-02-24
Low hanging fruits of some auto fixable improvements via clang-tidy

AlexanderLanin:
* Incorporating clang-tidy into make/CI is out of scope here
* Do not bother with merge conflicts. This pull request can easily be regenerated. Apply it when no other major changes are in progress.
* For easy review have a look at the individual commits
* All credits go to the clang-tidy team :-)
* This is not meant to solve everything. It just makes the code look a little more like C++


AlexanderLanin:
Hups, I was assuming Context has been merged :D
I'll update this after #526

src/manifest.cpp
```diff
@@ -392,8 +392,8 @@ verify_result(const Context& ctx,
               std::unordered_map<std::string, FileStats>& stated_files,
               std::unordered_map<std::string, digest>& hashed_files)
 {
-  for (uint32_t i = 0; i < result.file_info_indexes.size(); ++i) {
-    const auto& fi = mf.file_infos[result.file_info_indexes[i]];
+  for (unsigned int file_info_indexe : result.file_info_indexes) {
```
jrosdahl: `indexe` -> `index`
`unsigned int` -> `uint32_t`

src/manifest.cpp
```diff
@@ -606,8 +606,8 @@ manifest_dump(const std::string& path, FILE* stream)
     char name[DIGEST_STRING_BUFFER_SIZE];
     fmt::print(stream, "  {}:\n", i);
     fmt::print(stream, "    File info indexes:");
-    for (unsigned j = 0; j < mf->results[i].file_info_indexes.size(); ++j) {
-      fmt::print(stream, " {}", mf->results[i].file_info_indexes[j]);
+    for (unsigned int& file_info_indexe : mf->results[i].file_info_indexes) {
```
jrosdahl: `indexe` -> `index`
`unsigned int&` -> `uint32_t`

Also, I think that we should avoid shadowing the outer loop variable.

src/stats.cpp
```diff
@@ -362,10 +362,9 @@ stats_flush_to_file(const Config& config,
   }
 
   if (!config.log_file().empty() || config.debug()) {
-    for (int i = 0; i < STATS_END; ++i) {
-      if (updates->data[stats_info[i].stat] != 0
-          && !(stats_info[i].flags & FLAG_NOZERO)) {
-        cc_log("Result: %s", stats_info[i].message);
+    for (auto& i : stats_info) {
```
jrosdahl: `i` -> `si` or `info`

unittest/test_ZstdCompression.cpp
```diff
@@ -90,8 +90,8 @@ TEST_CASE("Large compressible Compression::Type::zstd roundtrip")
 TEST_CASE("Large uncompressible Compression::Type::zstd roundtrip")
 {
   char data[100000];
-  for (size_t i = 0; i < sizeof(data); i++) {
-    data[i] = rand() % 256;
+  for (char& i : data) {
+    i = rand() % 256;
```
jrosdahl: No thanks. :slightly_smiling_face: But if you must, name the variable `c` or `ch`.

src/ccache.cpp
```diff
@@ -1513,16 +1513,16 @@ hash_nvcc_host_compiler(const Context& ctx,
 #else
     const char* compilers[] = {"gcc", "g++"};
 #endif
-    for (size_t i = 0; i < ARRAY_SIZE(compilers); i++) {
+    for (auto& compiler : compilers) {
```
jrosdahl: `auto&` -> `auto`

src/ccache.cpp
```diff
@@ -1094,11 +1094,7 @@ create_cachedir_tag(nonstd::string_view dir)
     return false;
   }
   f << cachedir_tag;
-  if (!f) {
-    return false;
-  }
-
-  return true;
+  return !!f;
```
jrosdahl: I would prefer `return bool(f);`.

src/manifest.cpp
AlexanderLanin: I'll attempt to fix that second issue in clang-tidy or file a report otherwise.

src/manifest.cpp
AlexanderLanin: Not sure about the shadowing comment? Outer loop variable is `i`.

jrosdahl: > Not sure about the shadowing comment? Outer loop variable is `i`.

Right, strike that – I misinterpreted the narrow diff view.

src/ccache.cpp
AlexanderLanin: Since I had to think about the type, I just made it explicit char* instead of auto

PR #547, opened 2020-02-25
Make new installations use XDG base directory

### LINKED ISSUE #191: Should default cache_dir to $XDG_CACHE_HOME/ccache
Problems using cache_dir=$HOME/.ccache: it clutters the home dir when showing hidden files. Also, it slows down utilities scanning the home dir.

Advantages using cache_dir=$XDG_CACHE_HOME/ccache: one less dot-dir in the home directory. Also, makes it easy to delete all cache at once. $XDG_CACHE_HOME/ccache will tipically be $HOME/.cache/ccache

Of course, the configuration file is not cache, so keep reading .ccache/ccache.conf if existing, but favor this as new location: $XDG_CONFIG_HOME/ccache.conf 

https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html

(I know I can configure ccache to use whatever cache_dir I want. The proposal is about setting a better default for everyone.)

linkmauve:
This effectively deprecates `~/.ccache` as the default directory.

It is still possible to revert to the old behaviour by either setting
this directory in the secondary config file, or by setting the
`CCACHE_DIR` environment variable.

The new directories are `~/.config/ccache` for the configuration file, and
`~/.cache/ccache` for all other files.  We may want to introduce a third
directory, `~/.local/share/ccache`, for statistics, but I wanted to
validate the new behaviour first.

Fixes #191.

### DIFF
--- a/src/Config.hpp
+++ b/src/Config.hpp
@@ -121,7 +121,7 @@ private:
   std::string m_secondary_config_path;
 
   std::string m_base_dir = "";
-  std::string m_cache_dir = fmt::format("{}/.ccache", get_home_directory());
+  std::string m_cache_dir = "";
   uint32_t m_cache_dir_levels = 2;
   std::string m_compiler = "";
   std::string m_compiler_check = "mtime";
--- a/src/ccache.cpp
+++ b/src/ccache.cpp
@@ -3345,30 +3345,26 @@ tmpdir()
 static void
 set_up_config(Config& config)
 {
-  char* p = getenv("CCACHE_CONFIGPATH");
-  if (p) {
-    config.set_primary_config_path(p);
-  } else {
-    config.set_secondary_config_path(
-      fmt::format("{}/ccache.conf", TO_STRING(SYSCONFDIR)));
-    MTR_BEGIN("config", "conf_read_secondary");
-    // A missing config file in SYSCONFDIR is OK so don't check return value.
-    config.update_from_file(config.secondary_config_path());
-    MTR_END("config", "conf_read_secondary");
+  char* p;
+  config.set_secondary_config_path(
+    fmt::format("{}/ccache.conf", TO_STRING(SYSCONFDIR)));
+  MTR_BEGIN("config", "conf_read_secondary");
+  // A missing config file in SYSCONFDIR is OK so don't check return value.
+  config.update_from_file(config.secondary_config_path());
+  MTR_END("config", "conf_read_secondary");
 
-    if (config.cache_dir().empty()) {
-      fatal("configuration setting \"cache_dir\" must not be the empty string");
-    }
-    if ((p = getenv("CCACHE_DIR"))) {
-      config.set_cache_dir(p);
-    }
-    if (config.cache_dir().empty()) {
-      fatal("CCACHE_DIR must not be the empty string");
-    }
-
-    config.set_primary_config_path(
-      fmt::format("{}/ccache.conf", config.cache_dir()));
+  if (config.cache_dir().empty()) {
+    fatal("configuration setting \"cache_dir\" must not be the empty string");
   }
+  if ((p = getenv("CCACHE_DIR"))) {
+    config.set_cache_dir(p);
+  }
+  if (config.cache_dir().empty()) {
+    fatal("CCACHE_DIR must not be the empty string");
+  }
+
+  config.set_primary_config_path(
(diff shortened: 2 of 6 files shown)


src/ccache.cpp
```diff
@@ -3359,17 +3359,53 @@ set_up_config(Config& config)
     MTR_END("config", "conf_read_secondary");
 
     if (config.cache_dir().empty()) {
-      fatal("configuration setting \"cache_dir\" must not be the empty string");
+      // If the system cache_dir isn’t set, first check for the legacy path.
+      char* home = getenv("HOME");
+      if (!home) {
+        fatal("HOME must be defined");
+      }
```
jrosdahl: This won't work on Windows – use `get_home_directory()` instead.

Tangentially, I would also like to improve the default location on macOS (`~/Library/Preferences/ccache` and `~/Library/Caches/ccache`) and Windows as well (`%APPDATA%/ccache/` instead of `%APPDATA%/.ccache`), but that doesn't have to be part of this PR.

src/ccache.cpp
```diff
@@ -3359,17 +3359,53 @@ set_up_config(Config& config)
...
     }
+
     if ((p = getenv("CCACHE_DIR"))) {
       config.set_cache_dir(p);
+      if (config.cache_dir().empty()) {
+        fatal("CCACHE_DIR must not be the empty string");
+      }
+      config.set_primary_config_path(
+        fmt::format("{}/ccache.conf", config.cache_dir()));
```
jrosdahl: Would it make sense to set the primary config path to `$XDG_CONFIG_HOME/ccache/ccache.conf` if `CCACHE_DIR` is set to `$XDG_CACHE_HOME/ccache`? I think so since there otherwise will be a difference between `ccache` and `CCACHE_DIR=$HOME/.cache/ccache ccache`:

```
% ccache -s | head -n 2
cache directory                     /home/joel/.cache/ccache
primary config                      /home/joel/.config/ccache/ccache.conf
% CCACHE_DIR=$HOME/.cache/ccache ccache -s | head -n 2
cache directory                     /home/joel/.cache/ccache
primary config                      /home/joel/.cache/ccache/ccache.conf
```

src/Config.hpp
```diff
@@ -121,7 +121,7 @@ class Config : NonCopyable
   std::string m_secondary_config_path;
 
   std::string m_base_dir = "";
-  std::string m_cache_dir = fmt::format("{}/.ccache", get_home_directory());
+  std::string m_cache_dir = "";
```
jrosdahl: If `$CCACHE_CONFIGPATH` is set then `config.set_cache_dir` will never be called so `m_cache_dir` will be left unset. On the other hand it would be OK to remove the interpretation of `CCACHE_CONFIGPATH` since it's no longer used in the test suite.

src/ccache.cpp
```diff
@@ -3359,17 +3359,53 @@ set_up_config(Config& config)
...
     }
-    if (config.cache_dir().empty()) {
-      fatal("CCACHE_DIR must not be the empty string");
-    }
-
-    config.set_primary_config_path(
-      fmt::format("{}/ccache.conf", config.cache_dir()));
```
jrosdahl: If `cache_dir` is set in the secondary config then `config.set_primary_config_path` will no longer be called and the primary config thus not read.

Just curious: Is there a reason why you didn't follow (or didn't mention that you rejected) the implementation sketch in https://github.com/ccache/ccache/issues/191#issuecomment-412979701? It takes this and other aspects into account.

src/Config.hpp
linkmauve: Oh, it was there only for the test suite?  I’ll remove support for it then.

src/ccache.cpp
linkmauve: I quickly skimmed through this issue but didn’t notice this particular comment. :/

I’ll have a look at integrating it, it otherwise seems sound.

src/ccache.cpp
linkmauve: Done for `get_home_directory()`.  I don’t have any way to test these two platforms, aside from Wine, so I’ll indeed leave that out for some other contributor.

Although I remember reading that on macOS you have to put some plist file in that Preferences path, so you may want to use XDG base directories there too.

jrosdahl:
@linkmauve: Any update on the status?

jrosdahl:
Since this PR seems to have stalled I'll take over the implementation.

PR #548, opened 2020-02-25
Convert counters to a class

totph:
Also remove file writing logic from `stats_update_size()` and
delete unused declaration of `stats_timestamp()`.


src/Counters.cpp
```diff
@@ -16,36 +16,40 @@
...
-#include "util.hpp"
+#include "Counters.hpp"
 
-TEST_SUITE(stats)
+#include "stats.hpp"
 
-TEST(forward_compatibility)
+#include <algorithm>
+
+Counters::Counters() : m_counters(0, STATS_END)
```
AlexanderLanin: Sorry I don't understand this. Wouldn't this call create 0 elements with value STATS_END?
Why all this complexity: isn't it a fixed size vector (std::array) of STATS_END elements?

totph: Thanks, wrong argument order, fixed.

And a vector is required because future ccache versions might add more stats fields which must not get lost when rewritten by an older version, see the added test case.

AlexanderLanin: We should convert all remaining enums to enum class in the future to avoid such problems. That's exactly what it has been invented for.

totph: Yes, but exactly *here* the automatic decay into an integer is nice to have because of the forward compatibility. Both with `operator[](int)` and `operator[](StatsCounterEnum)` some public `static_cast`ing would be needed. To hide this the counter and the stats functions should probably be jointed so this can be hidden.

src/stats.cpp
```diff
@@ -373,14 +342,7 @@ stats_flush_to_file(const Config& config,
...
 
-  bool should_flush = false;
-  for (int i = 0; i < STATS_END; ++i) {
-    if (updates->data[i] > 0) {
-      should_flush = true;
-      break;
-    }
-  }
-  if (!should_flush) {
+  if (updates.all_zero()) {
```
jrosdahl: This is redundant since the same check is made first in the function. I'll remove it.

jrosdahl:
Thanks. And thanks @AlexanderLanin for the review!

PR #550, opened 2020-02-26
Avoid duplicate stat calls

erijo:
hash_source_code_string will, in read_file, stat the file to get the file
size. But when called from verify_result, the file size is already known so by
using this size the number of stat calls in a normal run can be cut in half.


jrosdahl:
Thanks.

PR #551, opened 2020-02-26
Read/write optimizations

erijo:
This PR improves `read_file` to avoid doing a realloc + read when the whole file has already been read in the first read call. Similar logic is then applied in other places where `read` is used.

Additionally, I've changed `read_embedded_file_entry` to use `write` instead of `fwrite` to write the result file. This since I noticed that each call to `fwrite` resulted in two calls to `write` (on my system). It seemed like it would be more efficient to write directly to the fd instead of going via stdio.

### DIFF
--- a/src/hash.cpp
+++ b/src/hash.cpp
@@ -175,7 +175,7 @@ hash_int(struct hash* hash, int x)
 }
 
 bool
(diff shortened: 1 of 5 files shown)


jrosdahl:
Thanks.

PR #554, opened 2020-03-03
CacheEntryReader: return type fix

totph:


AlexanderLanin:
There is `level = ctx.config.compression_level();` in ccache.hpp.
But I wonder why there was no warning for implicit conversion from int to int8?

PR #562, opened 2020-03-15
Enable check_format and clang-tidy checks in CI

AlexanderLanin:
Add clang-tidy (and check_format) to CI to prevent regressions.
Over time more checks could be enabled as they are fixed in code.

Some checks might require further configuration, this can be done over time.

Important: some checks are style/opinion based. Enabling all of them is not the point!

P.S. I have no idea what .travis.sh is used for. I've added something there too.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ script:
   - ./autogen.sh
   - test -z $BUILD_DIR || { mkdir -p $BUILD_DIR; cd $BUILD_DIR; }
   - ${SRC_DIR:-.}/configure $CONFIGURE
-  - make
+  - test "$NO_COMPILE" || make
   - make ${TEST:-test}
 
 matrix:
@@ -111,3 +111,50 @@ matrix:
         - libb2-dev
     before_install:
     - source ./.travis/install_cuda.sh
+
+  # Job: Clang tidy
+  - os: linux
+    env: V=1 PATH="/usr/bin:$PATH" NO_COMPILE=1 TEST=tidy
+    addons:
+      apt:
+        packages:
+        - elfutils
+        - libzstd1-dev
+        - libb2-dev
+    before_install:
+    - pip install --user compiledb
+
+  # Job: Check formatting
+  - os: linux
+    env: V=1 PATH="/usr/bin:$PATH" NO_COMPILE=1 TEST=check_format
+    addons:
+      apt:
+        packages:
+        - elfutils
+        - libzstd1-dev
+        - libb2-dev
+
+  # New Jobs go here until they are established for some weeks
+  allow_failures:
+
+  # Job: Clang tidy
+  - os: linux
+    env: V=1 PATH="/usr/bin:$PATH" NO_COMPILE=1 TEST=tidy
+    addons:
+      apt:
+        packages:
+        - elfutils
+        - libzstd1-dev
+        - libb2-dev
+    before_install:
+    - pip install --user compiledb
+
+  # Job: Check formatting
+  - os: linux
+    env: V=1 PATH="/usr/bin:$PATH" NO_COMPILE=1 TEST=check_format
+    addons:
+      apt:
+        packages:
+        - elfutils
+        - libzstd1-dev
+        - libb2-dev
--- a/src/.clang-tidy
+++ b/src/.clang-tidy
@@ -1,0 +1,83 @@
+# This .clang-tidy file is used by CI to ensure commits do not worsen the
+# codebase. If you want to improve the codebase this config is not for you.
+# Some checks are highly style dependant. The goal is NOT to activate all
+# of them.
+
+---
+Checks:          '-*,
+                  readability-*,
+                  -readability-implicit-bool-conversion,
+                  -readability-magic-numbers,
+                  -readability-else-after-return,
+                  -readability-named-parameter,
+                  -readability-qualified-auto,
+                  -readability-magic-numbers,
+                  -readability-isolate-declaration,
+                  -readability-inconsistent-declaration-parameter-name,
+                  -readability-simplify-boolean-expr,
+                  performance-*,
+                  -performance-unnecessary-value-param,
+                  -performance-faster-string-find,
+                  modernize-*,
+                  -modernize-avoid-c-arrays,
+                  -modernize-deprecated-headers,
+                  -modernize-loop-convert,
+                  -modernize-pass-by-value,
+                  -modernize-use-auto,
+                  -modernize-use-bool-literals,
+                  -modernize-use-using,
+                  -modernize-use-trailing-return-type,
+                  cppcoreguidelines-*,
+                  -cppcoreguidelines-pro-bounds-array-to-pointer-decay,
+                  -cppcoreguidelines-pro-type-vararg,
+                  -cppcoreguidelines-owning-memory,
+                  -cppcoreguidelines-avoid-magic-numbers,
+                  -cppcoreguidelines-pro-bounds-pointer-arithmetic,
+                  -cppcoreguidelines-no-malloc,
+                  -cppcoreguidelines-init-variables,
+                  -cppcoreguidelines-avoid-c-arrays,
+                  -cppcoreguidelines-pro-bounds-constant-array-index,
+                  -cppcoreguidelines-pro-type-cstyle-cast,
+                  -cppcoreguidelines-avoid-goto,
+                  -cppcoreguidelines-pro-type-member-init,
+                  -cppcoreguidelines-macro-usage,
+                  -cppcoreguidelines-pro-type-reinterpret-cast,
+                  -cppcoreguidelines-pro-type-union-access,
+                  -cppcoreguidelines-narrowing-conversions,
+                  bugprone-*,
+                  -bugprone-signed-char-misuse,
+                  -bugprone-branch-clone,
+                  -bugprone-narrowing-conversions,
+                  cert-*,
+                  -cert-err34-c,
+                  -cert-dcl50-cpp,
+                  -cert-err58-cpp,
+                  clang-diagnostic-*,
+                  clang-analyzer-*,
+                  -clang-analyzer-alpha*,
+                  -clang-analyzer-valist.Uninitialized,
+                  -clang-analyzer-optin.performance.Padding'
+WarningsAsErrors: '*'
+# Exclude subdirectories (currently there is only thrid_party)
+HeaderFilterRegex: 'src/[^\\/]*^'
+CheckOptions:    
+  - key:             readability-braces-around-statements.ShortStatementLines
+    value:           2
+
+  # extremly high limits, they are just here to ensure the situation doesn't
+  # get worse than it already is.
+  # If you hit the limits, please change the code and not the limits!!
+  - key:             readability-function-size.LineThreshold
+    value:           1100
+  - key:             readability-function-size.StatementThreshold
+    value:           760
+  - key:             readability-function-size.BranchThreshold
+    value:           170
+  - key:             readability-function-size.ParameterThreshold
+    value:           6
+  - key:             readability-function-size.NestingThreshold
+    value:           6
+  - key:             readability-function-size.VariableThreshold
+    value:           80
+...
+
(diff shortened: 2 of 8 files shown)


AlexanderLanin:
Example how errors look like: https://travis-ci.com/github/AlexanderLanin/ccache/jobs/298297589 and https://travis-ci.com/github/AlexanderLanin/ccache/jobs/298298685

src/.clang-tidy
```diff
@@ -0,0 +1,83 @@
...
+                  clang-analyzer-*,
+                  -clang-analyzer-alpha*,
+                  -clang-analyzer-valist.Uninitialized,
+                  -clang-analyzer-optin.performance.Padding'
+WarningsAsErrors: '*'
+# Exclude subdirectories (currently there is only thrid_party)
+HeaderFilterRegex: 'src/[^\\/]*^'
+CheckOptions:    
+  - key:             readability-braces-around-statements.ShortStatementLines
+    value:           2
```
jrosdahl: Just curious, why is this needed? If I interpret the documentation correctly the default value (0) should be enough (and what I would like).

.travis.yml
```diff
@@ -111,3 +111,50 @@ matrix:
...
+
+  # Job: Check formatting
+  - os: linux
+    env: V=1 PATH="/usr/bin:$PATH" NO_COMPILE=1 TEST=check_format
+    addons:
+      apt:
+        packages:
+        - elfutils
+        - libzstd1-dev
+        - libb2-dev
```
jrosdahl: Since the jobs succeed as is, perhaps remove this part?

src/.clang-tidy
```diff
@@ -0,0 +1,83 @@
+# This .clang-tidy file is used by CI to ensure commits do not worsen the
```
jrosdahl:
```suggestion
# This .clang-tidy file is used by CI to ensure that commits do not worsen the
```

src/.clang-tidy
```diff
@@ -0,0 +1,83 @@
...
+# codebase. If you want to improve the codebase this config is not for you.
+# Some checks are highly style dependant. The goal is NOT to activate all
```
jrosdahl:
```suggestion
# Some checks are highly style dependent. The goal is NOT to activate all
```

src/.clang-tidy
```diff
@@ -0,0 +1,83 @@
...
+                  -cert-err34-c,
+                  -cert-dcl50-cpp,
+                  -cert-err58-cpp,
+                  clang-diagnostic-*,
+                  clang-analyzer-*,
+                  -clang-analyzer-alpha*,
+                  -clang-analyzer-valist.Uninitialized,
+                  -clang-analyzer-optin.performance.Padding'
+WarningsAsErrors: '*'
+# Exclude subdirectories (currently there is only thrid_party)
```
jrosdahl:
```suggestion
# Exclude subdirectories (currently there is only third_party)
```

src/.clang-tidy
```diff
@@ -0,0 +1,83 @@
...
+HeaderFilterRegex: 'src/[^\\/]*^'
+CheckOptions:    
+  - key:             readability-braces-around-statements.ShortStatementLines
+    value:           2
+
+  # extremly high limits, they are just here to ensure the situation doesn't
+  # get worse than it already is.
```
jrosdahl:
```suggestion
```

unittest/.clang-tidy
```diff
@@ -0,0 +1,27 @@
...
+  - key:             readability-braces-around-statements.ShortStatementLines
+    value:           2
+
+  # extremly high limits, they are just here to ensure the situation doesn't
+  # get worse than it already is
+  # If you hit the limits, please change the code and not the limits!!
+  # Note: some limits disabled due to TEST_SUITE macro.
+  # It generates hundreds of statements, branches and variables.
+  - key:             readability-function-size.LineThreshold
+    value:           130
```
jrosdahl: FWIW, I don't consider 130 to be "extremely high". Which limit would you like to have?

unittest/.clang-tidy
```diff
@@ -0,0 +1,27 @@
+---
+Checks:          '-*,readability-function-size'
+WarningsAsErrors: '*'
+HeaderFilterRegex: 'unittest/.*'
+CheckOptions:    
+  - key:             readability-braces-around-statements.ShortStatementLines
+    value:           2
+
+  # extremly high limits, they are just here to ensure the situation doesn't
+  # get worse than it already is
```
jrosdahl:
```suggestion
```

src/.clang-tidy
AlexanderLanin: If you prefer 0 than the default is fine :-)
I had done a fix there in parallel and was considering to change the default to 2. So I just made the default explicit here.

.travis.yml
AlexanderLanin: I feel it's risky. Totally unrelated stuff will be unmergeable because of some small bug here. I would rather like to observe some pull requests after this one is merged and remove this in let's say two weeks.

unittest/.clang-tidy
AlexanderLanin: You are correct, only the 99999 a few lines later is "extremely high". I'll make this more clear.

AlexanderLanin: I just searched **a little** what different people advice on function length. Some respected folk say 10 lines. Others say it doesn't matter at all (big monitors etc).
Important part is that a function does only one thing - but that cannot be measured by tools.

A word on limits:
I recently had to deal with such a limit and spend 2 days refactoring a working pull request. It was a 80 char line limit and my function took up 60 of those. Because it did two things and I at least didn't want to hide that from the function name.
It was indeed way better afterwards! But not every contributor has those 2 days. And especially if you don't enjoy clean code as I do this is most definitely extremely demotivating. So I would keep the limit way above 10 here ;-)

src/.clang-tidy
AlexanderLanin: Fixed.

AlexanderLanin:
"An error occurred while generating the build script."

The exact same build script was working a week ago! :-(
Any ideas?
It fails consistently here and also at https://travis-ci.com/github/AlexanderLanin/ccache/jobs/300710189

AlexanderLanin:
Travis support reports bug has been fixed. I'll force push 3 times to verify. Sorry for the spam here!

unittest/.clang-tidy
jrosdahl: > especially if you don't enjoy clean code as I do

You make it sound like there are people who don't enjoy clean code, but I very much doubt that many people dislike clean code. But what's clean or not is subjective, so your clean code is not necessarily my clean code.

.travis.yml
jrosdahl: OK.

src/Counters.cpp
```diff
@@ -26,15 +26,17 @@ Counters::Counters() : m_counters(STATS_END)
 {
 }
 
-unsigned& Counters::operator[](size_t index)
+unsigned&
+Counters::operator[](size_t index)
```
jrosdahl: If you rebase to latest master you'll see that this fails check_format. Or is it only here to trigger the failure?

src/.clang-tidy
```diff
@@ -0,0 +1,82 @@
+# This .clang-tidy file is used by CI to ensure that commits do not worsen the
+# codebase. If you want to improve the codebase this config is not for you.
```
jrosdahl: Isn't this config for me if I want to improve the codebase by enabling some checks or even lower or raise some limits?

src/.clang-tidy
```diff
@@ -0,0 +1,82 @@
...
+                  -clang-analyzer-optin.performance.Padding'
+WarningsAsErrors: '*'
+# Exclude subdirectories (currently there is only third_party)
+HeaderFilterRegex: 'src/[^\\/]*^'
+CheckOptions:    
+  # Always add braces (added here just in case clang-tidy default changes).
+  - key:             readability-braces-around-statements.ShortStatementLines
+    value:           0
+
+  # If you hit the limits, please change the code and not the limits!!
```
jrosdahl:
```suggestion
  # If you hit a limit, please consider changing the code instead of the limit.
```

src/Counters.cpp
AlexanderLanin: This was auto formatted by clang-format-10.
I'll try rebasing to see what happens.

unittest/.clang-tidy
AlexanderLanin: Absolutely true. What I meant is that even then it differs how much people are willing to invest to arrive there.

Back to the topic. I guess 130 is fine for now?! When you hit 130 lines it's time to split your function. It doesn't prevent anyone from doing so before 130 lines :-)
The difficulty is that it's not possible to set exceptions. So the limit should probably remain at 130 and not be lowered.

src/.clang-tidy
AlexanderLanin:
```suggestion
# codebase. The checks and values below are the minimum standard for new code.
(Without claiming that they are 100% correct. They can be modified on demand!)
If you want to improve the codebase try enabling some additional checks or
playing with the configuration values.
```

src/Counters.cpp
AlexanderLanin: Rebased on current master (606bf51) and force-pushed (498a2d5):

    alex@DESKTOP-ALEX:~/ccache$ make check_format
    alex@DESKTOP-ALEX:~/ccache$ make format
    alex@DESKTOP-ALEX:~/ccache$ git status
    On branch clang-tidy
    nothing to commit, working tree clean
    alex@DESKTOP-ALEX:~/ccache$ clang-format --version
    clang-format version 10.0.0

unittest/.clang-tidy
jrosdahl: > I guess 130 is fine for now?!

It's fine for now.

>  When you hit 130 lines it's time to split your function.

Again, that's inherently subjective. I'm personally much more tolerant to long functions than complex functions, even though there of course often is correlation between long and complex.

src/Counters.cpp
jrosdahl: How annoying. Looks like they fixed a bug in clang-format 10, but we can't really require clang-format 10. This makes the CI check kind of useless unless we start requiring clang-format 6-9 or start bundling (or making it easy to download) a clang-format 10 binary.

I guess the same thing goes for clang-tidy in principle.

AlexanderLanin: Changing AlwaysBreakAfterReturnType from AllDefinitions to All just made the declarations **fail**:
[Code](https://github.com/ccache/ccache/compare/41b206f...AlexanderLanin:test), [Travis](https://travis-ci.com/github/AlexanderLanin/ccache/jobs/306889601)

Changing AlwaysBreakAfterReturnType from AllDefinitions to None makes it **pass**:
 [Code](https://github.com/AlexanderLanin/ccache/compare/41b206f...AlexanderLanin:test3), [Travis](https://travis-ci.com/github/AlexanderLanin/ccache/builds/156151361)

Adding old option AlwaysBreakAfterDefinitionReturnType made no difference, **fail**:
[Code](https://github.com/ccache/ccache/compare/41b206f...AlexanderLanin:test2), [Travis](https://travis-ci.com/github/AlexanderLanin/ccache/jobs/306889887)

Adding comments to those two places made it **fail**:
[Code](https://github.com/ccache/ccache/compare/41b206f...8acd1e9), [Travis](https://travis-ci.com/github/AlexanderLanin/ccache/jobs/306891547)

Moved comments by one line made it **pass**:
[Code](https://github.com/ccache/ccache/compare/41b206f...270187f), [Travis](https://travis-ci.com/github/AlexanderLanin/ccache/builds/156152340)

AlexanderLanin: Bundling clang-format for every system is probably a huge job.
Maybe format/check_format scripts should just fail when clang-format version x is not available since ccache (travis) is using version x? Maybe it's not 10 but an older version like 7 or 8 which had the bugfix. Such an old version could be more realistically enforced?

For now one of the two working workarounds on top seem sufficient.

jrosdahl: > Maybe it's not 10 but an older version like 7 or 8 which had the bugfix.

No, versions 6-9 are stable in this regard. I've access to them all (but not 10).

> For now one of the two working workarounds on top seem sufficient.

OK, `// clang-format off/on` comments will have to be good enough. I already had to use them in `CacheEntryWriter.cpp` to work around a clang-format bug.

AlexanderLanin:
Disabled formatting as discussed. I guess everything is fine, but as we already see it's somewhat brittle. So it's good it's marked as may-fail for now.

PR #563, opened 2020-03-16
Backport cuda fixes to v3.7 branch

leezu:
Backport https://github.com/ccache/ccache/pull/381

I ran into various "Illegal instruction" issues on CI servers with libzstd on v4 branch. It would be advisable if ccache could compile libzstd such that it doesn't optimize for the cpu features present on the machine performing the compilation.
In the meantime, it's helpful to backport this fix to the 3.7 branch.

### DIFF
--- a/src/ccache.c
+++ b/src/ccache.c
@@ -2537,6 +2537,7 @@ cc_process_args(struct args *args,
                 struct args **compiler_args)
 {
 	bool found_c_opt = false;
+	bool found_dc_opt = false;
 	bool found_S_opt = false;
 	bool found_pch = false;
 	bool found_fpch_preprocess = false;
@@ -2745,6 +2746,12 @@ cc_process_args(struct args *args,
 		// We must have -c.
 		if (str_eq(argv[i], "-c")) {
 			found_c_opt = true;
+			continue;
+		}
+
+		// when using nvcc with separable compilation, -dc implies -c
+		if ((str_eq(argv[i], "-dc") || str_eq(argv[i], "--device-c")) && guessed_compiler == GUESSED_NVCC) {
+			found_dc_opt = true;
 			continue;
 		}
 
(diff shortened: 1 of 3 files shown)


jrosdahl:
I'm OK with backporting #381, but it looks like this PR only applies one of the four commits in #381. Is it intentional to skip the other three?

Also, I suggeest cherry-picking with `-x` to get references to the original commits.

jrosdahl:
> I ran into various "Illegal instruction" issues on CI servers with libzstd on v4 branch. It would be advisable if ccache could compile libzstd such that it doesn't optimize for the cpu features present on the machine performing the compilation.

Hmm. No `-march`, `-mcpu`, `-mtune` or similar options are used for the zstd object files. zstd does have [runtime CPU detection](https://github.com/facebook/zstd/blob/dev/lib/common/cpu.h) that is [enabled for Clang and GCC ≥4.8](https://github.com/facebook/zstd/blob/66607d0eac54a99946ca14fee6af29fc525275c0/lib/common/compiler.h#L92) so I guess that is failing for you somehow? I don't know what I can do here. Suggestions welcome.

By the way, regarding
https://github.com/apache/incubator-mxnet/pull/17842/commits/c1dc4ab9aebfdc3a9edcb825a560263788eb5dd2: I don't see any use of AVX instructions in zstd. It only seems to use BMI2 if available.

leezu:
> I'm OK with backporting #381, but it looks like this PR only applies one of the four commits in #381. Is it intentional to skip the other three?
> Also, I suggest cherry-picking with `-x` to get references to the original commits.

It's not intentional. The PR is updated now to include all commits and reference the original commits.

> Hmm. No -march, -mcpu, -mtune or similar options are used for the zstd object files. zstd does have runtime CPU detection that is enabled for Clang and GCC ≥4.8 so I guess that is failing for you somehow? I don't know what I can do here.

The runtime CPU detection is only enabled if `__BMI2__` is not defined. I suspect that GCC may define `__BMI2__` if the CPU supports the instruction set, even if `-march, -mcpu, -mtune` is not specified. 

> By the way, regarding
apache/incubator-mxnet@c1dc4ab: I don't see any use of AVX instructions in zstd. It only seems to use BMI2 if available.

That explains why the commit didn't have any effect ;)
I previously observed that GCC generates AVX instructions even though `gcc -Q --help=target` shows that the default target disables AVX and no `-march, -mcpu, -mtune` are specified.

> Suggestions welcome.

If the error is due to gcc defining `__BMI2__`, adding `-mno-bmi2` may help.

jrosdahl:
OK, thanks.

PR #566, opened 2020-03-21
Make cc_log use fmtlib

totph:
This replaces the C `va_list` style `cc_log()` with one which uses the fmtlib internals, so it can finally format `string_view`.

This template expansions leads to quite some binary bloat, from 8.4M to 36M, and every new call adds several kilobytes. Thus I enabled LTO (link time optimization) to reduce the size again (not via autotools yet). However the compile time goes up from 36 to 48 seconds.

I decided to depend on the internals here because the alternative is `cc_log(fmt::format(..))` at every call site. Another possibility would be a macro - which might be justifiable here and could also move the `if (g_logging_enabled)` check before the function call.


jrosdahl:
> This template expansions leads to quite some binary bloat, from 8.4M to 36M, and every new call adds several kilobytes.

Looks like you're measuring unstripped size, but I don't think that's terribly relevant. On my system, the stripped size goes from 0.530 MB (39fd7489defbce33eadd8457ec4a24be7ff854ed) to 0.628 MB bytes (0dc267c2f12643cb437d0b9c15f36d1baf3f8633).

jrosdahl:
Implemented in another way (using perfect forwarding) in 05acfffbe50b.

PR #570, opened 2020-03-25
Ignore fmessage-length=... as a cache parameter

michaeleisel:
I've noticed that my build system (Xcode) will change this, e.g. in one invocation it's `-fmessage-length=20` and in another it's `-fmessage-length=70` for the same file. I'd like to have ccache ignore this flag, since it appears only to affect formatting of diagnostics.

### DIFF
--- a/src/ccache.cpp
+++ b/src/ccache.cpp
@@ -2805,6 +2805,10 @@ process_args(Context& ctx,
       continue;
     }
 
+    if (str_startswith(argv[i], "-fmessage-length=")) {
+        continue;
+    }
+
     // Options taking an argument that we may want to rewrite to relative paths
     // to get better hit rate. A secondary effect is that paths in the standard
     // error output produced by the compiler will be normalized.


AlexanderLanin:
Should this be added to sloppiness instead? I'm not really sure there is such a use case, but in theory someone could set and change this parameter on purpose?

michaeleisel:
sloppiness would be fine by me

jrosdahl:
Sorry, but just ignoring the option will make ccache return the wrong stderr for different values of -fmessage-length. That can't be the default behavior. Like @AlexanderLanin says the behavior has to be opt-in in some way if implemented at all.

I think that a more generic `ignore_options` (`CCACHE_IGNOREOPTIONS`) configuration setting would be better than a specific sloppiness boolean. Then you could set something like `ignore_options = -fmessage-length=*` to get the effect you want.

michaeleisel:
Personally, I'm fine with anything that allows me to ignore this flag

jrosdahl:
> Personally, I'm fine with anything that allows me to ignore this flag

Would you be interested in working on this, or should we create a feature request issue and close this PR?

michaeleisel:
Creating a feature request would be great 👍

jrosdahl:
> Creating a feature request would be great :+1:

Done in #578.

PR #572, opened 2020-03-31
Make ccache build on Solaris 11.4

pstorz:
These changes enable building on Oracle Solaris 11.4 for both i386 and sparc.


jrosdahl:
>  Probably at least compilation could be ensured by cross compiling for Solaris - without making sure that the resulting executable actually works.
Would that make sense? @jrosdahl

I wouldn't be against that.

PR #575, opened 2020-04-05
Add an Args class with backward compatibility for the old args API

jrosdahl:
The idea of this is to make a proper C++ API while at the same time implementing most of the old legacy args API. This makes it possible to refactor call sites gradually instead of in a series of large
refactoring steps. This is an alternative to #513.

Given that `struct args* args` is replaced with `Args args`, the following types of legacy API usage can be kept without source changes:

1. The `args_*` functions can be used as is.
2. `args->argv[i]` still works for read access.

Things that aren’t emulated:

1. `args->argc` (use `args.size()` instead).
2. `args->argv` to access the underlying `char*` array (use `args.to_argv().data()` instead).
3. Mutating the `args->argv[i]` string in place (don’t do that).

In other words, code that uses `args_foo(args, ...)` and `args->argv[i]` can easily be converted to `args.bar(...)` and `args[i]` (with `std::string` arguments instead of `char*`) at will and in arbitrarily small steps.


src/Args.hpp
```diff
@@ -0,0 +1,180 @@
...
+
+  size_t size() const;
+  const std::string& operator[](size_t i) const;
+  std::string& operator[](size_t i);
+
+  // Accessor functions for the legacy API:
+  Args& operator*();
+  const Args* operator->() const;
+
+  class ArgvKeeper : NonCopyable
```
AlexanderLanin, line 55: We started discussing in [#513](https://github.com/ccache/ccache/pull/513#issuecomment-578486206) but haven't quite finished.
I think this can be simplified **a lot** by having to_argv return the c_str() values of the strings instead of this extra struct. Runtime is also improved by avoiding all those string copies and memory allocations.

It's _slightly_ more complex for usage since to_args return value does no longer outlive Args. But I don't think this is an issue. We are accustomed to such stuff with strings, as in c_str().

src/hashutil.cpp
```diff
@@ -313,23 +313,27 @@ hash_command_output(struct hash* hash,
...
+  cc_log_argv("Executing compiler check command ", argv);
 
 #ifdef _WIN32
   PROCESS_INFORMATION pi;
   memset(&pi, 0x00, sizeof(pi));
   STARTUPINFO si;
   memset(&si, 0x00, sizeof(si));
 
-  char* path = find_executable_in_path(args->argv[0], nullptr, getenv("PATH"));
+  char* path = find_executable_in_path(argv[0], nullptr, getenv("PATH"));
```
AlexanderLanin, line 334: Don't rely on to_argv where it's not needed.
```suggestion
  char* path = find_executable_in_path(args[0].c_str(), nullptr, getenv("PATH"));
```

src/hashutil.cpp
```diff
@@ -313,23 +313,27 @@ hash_command_output(struct hash* hash,
...
   if (!path) {
-    path = args->argv[0];
+    path = x_strdup(argv[0]);
```
AlexanderLanin, line 336:
```suggestion
    path = x_strdup(args[0].c_str());
```

unittest/test_Args.cpp
```diff
@@ -0,0 +1,188 @@
...
+  CHECK(Args::from_string("a little string").to_string() == "a little string");
+}
+
+TEST_CASE("Args operations")
+{
+  Args args = Args::from_string("eeny meeny miny moe");
+  Args more_args = Args::from_string("x y");
+  Args no_args;
+
+  SECTION("erase_with_prefix")
```
AlexanderLanin, line 125: use of sections 👍

src/Args.hpp
AlexanderLanin: You could return a vector<char*> to avoid memory management completely.
Call sites would potentially store this locally and use &...[0] to get the char**.

jrosdahl: > I think this can be simplified a lot by having to_argv return the c_str() values of the strings instead of this extra struct.

In the big scheme of things I don't think that such a simple thing can be simplified _a lot_ – then it would have negative complexity. :slightly_smiling_face: But sure, sounds good. Especially after 9131b0595242 since it wouldn't feel quite right not to copy the string data before that commit.

> You could return a vector<char*> to avoid memory management completely.

One reason I saw for not using a vector is that it has to store the terminating null entry, so `args.size()` will then not be equal to `args.to_argv().size()`, which I think feels slightly strange – it would be better if there were no size method at all. But that's a minor thing; I'm OK with a vector.

src/hashutil.cpp
jrosdahl: Done.

src/hashutil.cpp
jrosdahl: Done.

src/Args.hpp
```diff
@@ -0,0 +1,161 @@
...
+  size_t size() const;
+  const std::string& operator[](size_t i) const;
+  std::string& operator[](size_t i);
+
+  // Accessor functions for the legacy API:
+  Args& operator*();
+  const Args* operator->() const;
+
+  // Return the argument list as a vector of raw string pointers. Callers can
+  // use `const_char<char* const*>(args.to_argv().data())` to get an array
```
AlexanderLanin, line 58:
```suggestion
  // use `const_cast<char* const*>(args.to_argv().data())` to get an array
```

jrosdahl: > Did you by any chance test this on Windows?

I don't have (easy) access to Windows so I never test ccache there, except for the unit tests which I run in Wine from time to time, at least before a release. I ran them now and that actually exposed a unit test failure for this PR. Good suggestion!

jrosdahl:
Thanks @AlexanderLanin for the review.

PR #576, opened 2020-04-06
Fix #559 (setting base_dir and cache_dir_levels)

### LINKED ISSUE #559: ccache issue with base_dir option (version 3.7.5, Windows 10, MinGW)
Hello.

I got problems with the base_dir option which does not seems to work well:
- When using the command:
   `ccache -o base_dir=<project_folder>`
- The file ccache.conf contains only the max_size by default (running on Windows Prompt)

- The environment variable CCACHE_BASEDIR seems to be taken in account when I run tests, but, the conversion from absolute to relative path seems to not work. I already tried with both forward and backslashes, but I got different results:
  - Forward slashes: I got errors due to the path length, like when I do not define any base_dir
  - Back slashes: When I check the log, there are only the 4 first chars which are removed. Looks like it is running this part: // Strip \\?\ from the file name.
-> We run it on minGW

For info, we are using ccache 3.7.5 for GHS compatibility.

Thanks.

AlexanderLanin:
Verified manually (on linux only):

    ./ccache --set-config=cache_dir_levels=5
    ./ccache --set-config=base_dir=/tmp
    ./ccache --set-config=max_size=5G
    ./ccache --show-config
    ./ccache --set-config=cache_dir_levels=3
    ./ccache --set-config=base_dir=/home/alex
    ./ccache --set-config=max_size=15G
    ./ccache --show-config

Before this patch it would fail with:
* cache_dir_levels was comparing pointer to <= 8 --> cache directory levels must be between 1 and 8
* base_dir was comparing jibberish with is_absolute_path --> not an absolute path

Those are the only two validations done on config input, therefore I verified with max_size that nothing else got broken here.

//edit: there is no automatic link created from title or commit messages, so here we go: #559 😄

### DIFF
--- a/src/conf.c
+++ b/src/conf.c
@@ -305,9 +305,9 @@ conf_set_value_in_file(const char *path, const char *key, const char *value,
 		return false;
 	}
 
-	char dummy[8] = {0}; // The maximum entry size in struct conf.
-	if (!item->parser(value, (void *)dummy, errmsg)
-	    || (item->verifier && !item->verifier(value, errmsg))) {
+	char parsed[8] = {0}; // The maximum entry size in struct conf.
+	if (!item->parser(value, (void *)parsed, errmsg)
+	    || (item->verifier && !item->verifier(&parsed, errmsg))) {
 		return false;
 	}
 


jrosdahl:
Thanks!

PR #579, opened 2020-04-17
Fix path length issue on Windows 10, fix base_dir: hash issue fixed, …

Shionigami:
Fix for long path on Windows 10:
- Add \\\\?\\ prefix to bypass Windows length limitation (255 characters). Works only with **Windows 10**
-> Then, absolute path becomes: \\\\?\\<absolute_path_to_project>

Fix base_dir issues:
- If path contains double backslashes, replace by single backslashes
Example:
`c:\\project\\folder1/file1`
 -> 
`c:\\\\project\\\\folder1\\file1`

- Do not try to remove \\\\?\\, if it is not part of the path
- In function x_realpath, if path_handle is not INVALID_HANDLE_VALUE, it will remove the first 4 chars, which are supposed to be \\\\?\\, even if the string is not part of the path. 
- Make test first to ensure that the string exists, before removing the 4 first chars.

- Make sure that there path with only back or forward slashes
Example:
`c:**\\**project**\\**folder1**/**file1`
 -> 
`c:\\project\\folder1\\file1`

- Consider backslashes for conditions to remove the last / char from path: 
Consider also the Windows format of the path to remove the last \\

**Note: For Windows Only**

### DIFF
--- a/src/util.c
+++ b/src/util.c
@@ -706,7 +706,7 @@ slash_to_backslash(char* str)
 	}
 	temp[len] = 0;
 
-	return str;
+	return temp;
 }
 #endif
 
@@ -1462,7 +1462,7 @@ common_dir_prefix_length(const char *s1, const char *s2)
 		++p1;
 		++p2;
 	}
-	while ((*p1 && *p1 != '/' && *p1 != '\\') || (*p2 && *p2 != '/' && *p2 != '\\')) {
+	while ((*p1 && (*p1 != '/' || *p1 != '\\')) || (*p2 && (*p2 != '/' || *p2 != '\\'))) {
 		p1--;
 		p2--;
 	}
@@ -1519,7 +1519,7 @@ get_relative_path(const char *from, const char *to)
 	if (strlen(to) > common_prefix_len) {
 		reformat(&result, "%s%s", result, to + common_prefix_len + 1);
 	}
-	for (int i = strlen(result) - 1; i >= 0 && result[i] == '/' && result[i] == '\\'; i--) {
+	for (int i = strlen(result) - 1; i >= 0 && (result[i] == '/' || result[i] == '\\'); i--) {
 		result[i] = '\0';
 	}
 	if (str_eq(result, "")) {
--- a/src/util.c
+++ b/src/util.c
@@ -1452,12 +1452,20 @@ common_dir_prefix_length(const char *s1, const char *s2)
 {
 	const char *p1 = s1;
 	const char *p2 = s2;
+    char separator = NULL;
+
+    if(strchr(p1, '/') != NULL)
+    {
+        separator = '/';
+    } else {
+        separator = '\\';
+    }
 
 	while (*p1 && *p2 && *p1 == *p2) {
 		++p1;
 		++p2;
 	}
-	while ((*p1 && *p1 != '/') || (*p1 && *p1 != '\\') || (*p2 && *p2 != '/')  || (*p2 && *p2 != '\\')) {
+	while ((*p1 && *p1 != separator) || (*p2 && *p2 != separator)) {
 		p1--;
 		p2--;
 	}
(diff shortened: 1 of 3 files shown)


jrosdahl:
Hi,

1. Please have a look at the failing CI tests.
2. Could you please describe in more detail which problems your PR aims to fix?

jrosdahl:
I suggest retargeting this PR to master instead of 3.7-maint – master already has some related improvements. I want to focus on getting master released instead of investing too much effort into 3.7-maint. Selected improvements can be backported to 3.7-maint, though.

Also, please look into adding unit tests that verify the changes (and that implicitly define how the functions are supposed to behave). It's a bit hard to understand what actual problems the changes intend to solve by just reading the modified code and the PR description, and we need to make sure that the fixes don't stop working in the future.

src/util.c
```diff
@@ -690,6 +690,63 @@ create_cachedirtag(const char *dir)
 	return 1;
 }
 
+#ifdef _WIN32
+// Replace forward slashes with backslashes
+// Fixes issues when path got backslashes AND forward slashes
+char* 
+slash_to_backslash(char* str)
```
AlexanderLanin, line 697: Wouldn't this be way simpler with std::string and replace? I know it's char* all around, but it should be std::string anyway.

src/util.c
```diff
@@ -1254,12 +1315,28 @@ strtok_r(char *str, const char *delim, char **saveptr)
...
 int
 create_tmp_fd(char **fname)
 {
-	char *template = format("%s.%s", *fname, tmp_string());
+	char* template = NULL;
+#ifdef _WIN32
+		int maxlen=190;
+		int pathlen = strlen(format("%s.%s", *fname, tmp_string()));
+
+		if(pathlen > maxlen && atof(format("%ld.%ld", (DWORD)(LOBYTE(LOWORD(GetVersion()))), (DWORD)(HIBYTE(LOWORD(GetVersion()))))) > 6.1) {
```
AlexanderLanin, line 1330: Can't this be something like GetVersion() > 0x0601? (I have not looked up the exact semantics, just guessing). You extra LOBYTE and HIBYTE and then concatenate them again wir format and then parse them via atof. Sounds like converting back and forth.

src/util.c
```diff
@@ -1375,12 +1452,20 @@ common_dir_prefix_length(const char *s1, const char *s2)
 {
 	const char *p1 = s1;
 	const char *p2 = s2;
+    char separator = '\0';
```
AlexanderLanin, line 1455: Why bother init to 0? When you omit the default value you will get compiler warnings when the variable is being used without proper initialization. Also it can be made const via
`const char separator = strchr(p1, '/') ? '/' : '\\';`

Shionigami:
> I suggest retargeting this PR to master instead of 3.7-maint – master already has some related improvements. I want to focus on getting master released instead of investing too much effort into 3.7-maint. Selected improvements can be backported to 3.7-maint, though.
> 
> Also, please look into adding unit tests that verify the changes (and that implicitly define how the functions are supposed to behave). It's a bit hard to understand what actual problems the changes intend to solve by just reading the modified code and the PR description, and we need to make sure that the fixes don't stop working in the future.

The problem is that we are still using the version 3.7.6 of ccache in the company, because we are building most of the projects with GHS. That is why I am focusing on the branch 3.7-maint. 
I also wanted to apply the changes to the master branch, but there are a lot of changes, and I did not have time to come back on that topic since today.

> **Fix for long path on Windows 10:**
> - Add \\?\ prefix to bypass Windows length limitation (255 characters). Works only with Windows 10
>  -> Then, absolute path becomes: \\?\<absolute_path_to_project>
Here the goal is to fix the ccache failures, when the path is longer than 255 characters. In that case, ccache crashes with the following error: ccache: error: Failed to create temporary file for [...]

> **Fix base_dir issues:**
> - If path contains double backslashes, replace by single backslashes
> - Example:
>  c:\\project\\folder1/file1
>  ->
>  c:\\\\project\\\\folder1\\file1
- In some cases, there are 2 backslashes in the path, instead of 1 and then, there were some problems to detect the cache here.
 
> - Do not try to remove \\?\, if it is not part of the path
>- In function x_realpath, if path_handle is not INVALID_HANDLE_VALUE, it will remove the first 4 chars, which are supposed to be \\?\, even if the string is not part of the path.
>
>- Make test first to ensure that the string exists, before removing the 4 first chars.
- Here the problem was that the path may be truncated, if "\\?\" is not part of the string.
  Example: c:\program files\...    ->   rogram files\...
- The goal here is to avoid that issue.

>- Make sure that there path with only back or forward slashes
>- Example:
> c:**\\**project**\\**folder1**/**file1
> ->
> c:\\project\\folder1\\file1
> - Consider backslashes for conditions to remove the last / char from path:
> - Consider also the Windows format of the path to remove the last \

- In most of the cases, path with backslashes and forward slashes is not a problem, but in some others, like when you add the string "\\?\", it fails if there are forward slashes (on Windows). 

base_dir functionnality, to enable the cache sharing between different location did not work also, on Windows, because of all these issues. Now we are able to share it, and it may also work on network (not tested yet).

jrosdahl:
Closing this PR since it seems it was superceded by #635.

PR #580, opened 2020-04-20
Diverse CMake preparations

AlexanderLanin:
Diverse preparations for CMake in order to make that pull request slightly smaller.
* Prepare gitignore for builds in subdirectory (common pattern in CMake, but not CMake specific) + Ignore to be generated Version.cpp
* Move travis to buildenv as it's just another image
* Move install_cuda out of .travis as it's not travis specific
* Prepare docker images for C++ and CMake
* Update authors + add script to do so
* Fix required C compiler
* Fix some formatting issues with clang-format 10
* Last but not least: remove old suite_args. No idea how current master is working! It's not even compiling for me on clang 10.


unittest/main.cpp
```diff
@@ -30,7 +29,6 @@ unsigned suite_hashutil(unsigned);
 unsigned suite_legacy_util(unsigned);
 
 const suite_fn k_legacy_suites[] = {
-  &suite_args,
```
jrosdahl, line 33: I don't want to remove the legacy args suite as long as the legacy args API is still in use. (And when we do we should remove `unittest/test_legacy_args.cpp` as well.)

What build errors did you get with Clang 10? I can build with Clang 10 without problems.

doc/AUTHORS.adoc
```diff
@@ -7,32 +7,37 @@ maintained by Joel Rosdahl.
 ccache is a collective work with contributions from many people, including:
 
 * Alexander Korsunsky
+* Alexander Lanin
 * Alexey Tourbin
 * Alfred Landrum
 * Anders F Björklund
+* André Klitzing
 * Andrea Bittau
```
jrosdahl, line 15: Looks like you have sorted using the C locale instead of e.g. en_US? I think you can revert the changes to this file; they are done before a release.

misc/update_authors.sh
```diff
@@ -0,0 +1,13 @@
...
+
+if [ -d .git ]; then
+  echo "Updating authors...\n"
+
+  # Full git history is needed for this to work
+  # (hide stderr and return code since this fails in case full history is already available)
+  git fetch --unshallow 2> /dev/null || true
+
+  # collect "Co-authored-by" and Authors and update doc/AUTHORS.adoc
+  { git log --pretty=format:"%(trailers)" | grep -Po "(?<=Co-authored-by: )(.*)(?= <)"; git log --format="%aN"; } | sed -e 's/^/* /' | sort -udf | perl -00 -p -i -e 's/^\*.*/<STDIN> . "\n"/es' doc/AUTHORS.adoc
```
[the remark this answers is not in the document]
AlexanderLanin: It doesn't do what I thought it does. Hopefully fixed now. I wanted to have reproducible results regardless of used system.

misc/update_authors.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/sh -ex
+set -e
+
+if [ -d .git ]; then
+  echo "Updating authors...\n"
+
+  # Full git history is needed for this to work
+  # (hide stderr and return code since this fails in case full history is already available)
+  git fetch --unshallow 2> /dev/null || true
```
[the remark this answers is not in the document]
AlexanderLanin: Let's call not committing a potential for future improvements, if that's ok :-)

CMake brings the ability to run make package and make source_package. It will allow generating source and binary packages with one single command. Updating authors is kind of a prerequisite for that ability.
I know it's only you and you can do it manually. But I like automating things :-)
In the other PR those packages are both generated by travis, extracted out of source and automatically tested. Fells like a big step towards simplifying releases, delivering weekly binaries, etc.

Even if it's just you, it should be simple :-)

unittest/main.cpp
AlexanderLanin: Hups. I missed adding the new file test_legacy_args.cpp

doc/AUTHORS.adoc
AlexanderLanin: This was mostly to verify update_authors is working. It's still producing slightly different results, e.g. Tomasz before Tomm. I was assuming -d fixes that, but it does not. Any ideas?

misc/update_authors.sh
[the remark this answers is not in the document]
AlexanderLanin: P.S. this is required for travis because it doesn't clone with full history.

doc/AUTHORS.adoc
AlexanderLanin: I've added this file simply to verify update_authors works as expected (no unexpected diff in the file).

misc/update_authors.sh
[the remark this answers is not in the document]
jrosdahl: > Even if it's just you, it should be simple :-)

My reaction is only that adding unneeded complexity to the build and release scripts does not make things simpler _for me_, rather the opposite. If there is a need for it later, sure, it's just that it wasn't obvious from your changes alone.

> But I like automating things :-)

You make it sound like I don't like to automate things, which is incorrect. I already _have_ it simple and automated.

misc/update_authors.sh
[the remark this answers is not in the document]
jrosdahl: Yes, setting locale is correct.

doc/AUTHORS.adoc
```diff
@@ -85,16 +90,15 @@ ccache is a collective work with contributions from many people, including:
...
 * Rafael Kitover
 * Ramiro Polla
 * Robert Yang
 * Robin H. Johnson
 * Rolf Bjarne Kvinge
-* Russell King
 * RW
 * Ryan Brown
 * Sam Gross
-* Steffen Dettmer
```
jrosdahl, line 97: This means that the script doesn't pick up Steffen Dettmer from f2481b1d5988cb4ccc3010c4ad6933e9c93e6d02. I'll have a look.

jrosdahl: Fixed in b20f862db0ffcd16991663d56d69c22db5abfd5a.

AlexanderLanin:
Oh I was assuming that's because of branch change from 3.7 to master.

Thanks for review! I'll rebase CMake PR as soon as I can.

PR #584, opened 2020-04-24
Disable hardlink testing on AFS

megatesla:
AFS _kind of_ supports hardlinks, but not really - you can't make hardlinks across directories, only within them. This basically defeats the point of the ccache hardlink mode - the only way to use it would be if your cache directory is the same as your compilation directory, which is silly. Unfortunately the current hardlink probe doesn't test this case and proceeds with the test suite anyway, which fails.

### DIFF
--- a/test/suites/hardlink.bash
+++ b/test/suites/hardlink.bash
@@ -1,6 +1,7 @@
 SUITE_hardlink_PROBE() {
-    touch file1
-    if ! ln file1 file2 >/dev/null 2>&1; then
+    mkdir dir;
+    touch dir/file1;
+    if ! ln dir/file1 file2 >/dev/null 2>&1; then
         echo "file system doesn't support hardlinks"
     fi
 }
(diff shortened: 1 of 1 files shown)


test/suites/hardlink.bash
```diff
@@ -1,6 +1,7 @@
 SUITE_hardlink_PROBE() {
-    touch file1
-    if ! ln file1 file2 >/dev/null 2>&1; then
+    mkdir dir;
```
AlexanderLanin, line 2:
```suggestion
    # AFS kind of supports hardlinks, but not really - you can't make hardlinks across directories, only within them.
    mkdir dir;
```

test/suites/hardlink.bash
```diff
@@ -1,6 +1,8 @@
 SUITE_hardlink_PROBE() {
-    touch file1
-    if ! ln file1 file2 >/dev/null 2>&1; then
+    # AFS kind of supports hardlinks, but not really - you can't make hardlinks across directories, only within them.
```
jrosdahl, line 2:
```suggestion
    # Probe hard link across directories since AFS doesn't support those.
```

test/suites/hardlink.bash
```diff
@@ -1,6 +1,8 @@
...
+    mkdir dir;
+    touch dir/file1;
```
jrosdahl, line 4:
```suggestion
    mkdir dir
    touch dir/file1
```

PR #586, opened 2020-04-24
Generic function for parsing integers

AlexanderLanin:
There are lots of places where ccache is parsing integers and mostly they are parsed to a slightly wrong datatype and then cast implicitly or explicitly to the right one. This function will enable parsing the exact type needed and providing proper error messages for out of bounds etc.

One example application of parsing compression_level is included in this commit. The rest I would apply after this one has been accepted and merged.

I'm truly amazed at the number of lines required for this function. But there is no way to make it shorter. At least none I can figure out.
I was not able to find such a function off the shelf anywhere.

--
Off topic I've included a fix for casting between signed and unsigned bytes. Conversion relies on undefined behavior. One has to use memcpy for that. C++20 will provide [std::bit_cast](https://en.cppreference.com/w/cpp/numeric/bit_cast).

### DIFF
--- a/src/Util.hpp
+++ b/src/Util.hpp
@@ -27,6 +27,7 @@
 
 #include <functional>
 #include <memory>
+#include <sstream>
 #include <string>
 #include <utility>
 #include <vector>
@@ -238,10 +239,55 @@ bool matches_dir_prefix_or_file(nonstd::string_view dir_prefix_or_file,
 // On Windows: Backslashes are replaced with forward slashes.
 std::string normalize_absolute_path(nonstd::string_view path);
 
-// Parse a string into an integer.
-//
-// Throws Error on error.
-int parse_int(const std::string& value);
+/// Parse a string into an integer.
+/// Integer type must be passed as template parameter.
+///
+/// Pass a human readable error_string for better error messages.
+/// Function will throw Error on error.
+template<class IntType>
+IntType
+stoi(const std::string& value, const char* const error_string = "number")
+{
+  // Don't treat int8_t as ASCII char.
+  if (std::is_same<IntType, int8_t>::value
+      || std::is_same<IntType, uint8_t>::value) {
+    const int temp = Util::stoi<int>(value, error_string);
+    if (temp < static_cast<int>(std::numeric_limits<IntType>::min())
+        || temp > static_cast<int>(std::numeric_limits<IntType>::max())) {
+      throw Error(
+        fmt::format("invalid {} (out of range): \"{}\"", error_string, value));
+    }
+    return temp;
+  }
+
+  // Stream operator doesn't handle negative values for unsigned types.
+  if (std::is_unsigned<IntType>::value && value[0] == '-')
+    throw Error(fmt::format(
+      "invalid {} (negative value passed where positive was expected): \"{}\"",
+      error_string,
+      value));
+
+  std::istringstream ss(value);
+  IntType result;
+  try {
+    ss >> result;
+  } catch (...) {
+    // Internal error in Stream operator (rare).
+    throw Error(
+      fmt::format("invalid {} (exception): \"{}\"", error_string, value));
+  }
+
+  // Invalid format etc.
+  if (!ss)
+    throw Error(fmt::format("invalid {}: \"{}\"", error_string, value));
+
+  // Not entire string has been parsed.
+  if (!ss.eof())
+    throw Error(fmt::format(
+      "invalid {} (no suffix allowed): \"{}\"", error_string, value));
+
+  return result;
+}
 
 // Read file data as a string.
 //
--- a/unittest/.clang-tidy
+++ b/unittest/.clang-tidy
@@ -1,3 +1,10 @@
+# This .clang-tidy file is used by CI to ensure that commits do not worsen the
+# codebase. The checks and values below are the minimum standard for new code.
+# (Without claiming that they are 100% correct. They can be modified on demand!)
+# If you want to improve the codebase try enabling some additional checks or
+# playing with the configuration values.
+# Some checks are highly style dependent. Activating all of them is NOT the goal.
+
 ---
 Checks:          '-*,readability-function-size'
 WarningsAsErrors: '*'
@@ -11,7 +18,7 @@ CheckOptions:
   # Note: some limits "disabled" due to TEST_SUITE macro.
   # The macro generates hundreds of statements, branches and variables.
   - key:             readability-function-size.LineThreshold
-    value:           130
+    value:           200
   - key:             readability-function-size.StatementThreshold
     value:           999999
   - key:             readability-function-size.ParameterThreshold
--- a/unittest/test_Config.cpp
+++ b/unittest/test_Config.cpp
@@ -235,19 +235,37 @@ TEST_CASE("Config::update_from_file, error handling")
   SECTION("invalid unsigned")
   {
     Util::write_file("ccache.conf", "max_files =");
-    REQUIRE_THROWS_WITH(
-      config.update_from_file("ccache.conf"),
-      Equals("ccache.conf:1: invalid unsigned integer: \"\""));
+    REQUIRE_THROWS_WITH(config.update_from_file("ccache.conf"),
+                        Equals("ccache.conf:1: invalid number: \"\""));
 
     Util::write_file("ccache.conf", "max_files = -42");
     REQUIRE_THROWS_WITH(
       config.update_from_file("ccache.conf"),
-      Equals("ccache.conf:1: invalid unsigned integer: \"-42\""));
+      Equals("ccache.conf:1: invalid number (negative value passed "
+             "where positive was expected): \"-42\""));
 
     Util::write_file("ccache.conf", "max_files = foo");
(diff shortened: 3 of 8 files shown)


AlexanderLanin:
Interestingly I cannot reproduce the warning as reported by Travis CI on the first commit. Anyway it's fixed now and I simply hope that enabled/disabled warnings will be more transparent with CMake. At least more understandable to me.

I've also extended the allowed test function size as I was hitting the limit without the test looking like it requires refactoring. Therefore the limit was wrong. I've extended it by 50% to avoid small increments.

unittest/test_Config.cpp
```diff
@@ -228,19 +228,39 @@ TEST_CASE("Config::update_from_file, error handling")
...
       config.update_from_file("ccache.conf"),
-      Equals("ccache.conf:1: invalid unsigned integer: \"\""));
+      Equals("ccache.conf:1: invalid unsigned integer: (empty string)"));
 
     Util::write_file("ccache.conf", "max_files = -42");
     REQUIRE_THROWS_WITH(
       config.update_from_file("ccache.conf"),
-      Equals("ccache.conf:1: invalid unsigned integer: \"-42\""));
+      Equals("ccache.conf:1: invalid unsigned integer (signed value passed "
+             "where unsigned was expected): \"-42\""));
```
jrosdahl, line 237: I'm sorry, but I think that this error message is worse than the previous one.

unittest/test_Config.cpp
```diff
@@ -228,19 +228,39 @@ TEST_CASE("Config::update_from_file, error handling")
     Util::write_file("ccache.conf", "max_files =");
     REQUIRE_THROWS_WITH(
       config.update_from_file("ccache.conf"),
-      Equals("ccache.conf:1: invalid unsigned integer: \"\""));
+      Equals("ccache.conf:1: invalid unsigned integer: (empty string)"));
```
jrosdahl, line 231: I prefer the old error message.

src/Util.hpp
```diff
@@ -230,10 +231,62 @@ bool matches_dir_prefix_or_file(nonstd::string_view dir_prefix_or_file,
...
-//
-// Throws Error on error.
+/// Parse a string into an integer.
+/// Integer type must be passed as template parameter.
+///
+/// @param errorString for better error messages pass a user readable string.
+/// @throw Error on error.
+template<class IntType>
+IntType
+stoi(const std::string& value, const char* const errorString = "number")
```
jrosdahl, line 241:
```suggestion
stoi(const std::string& value, const char* const error_string = "number")
```

AlexanderLanin:
> Honestly this kind of feels like a solution in search of a problem. Is this type of improvements worth the implementation and review effort? There are many other things I would like to spend my time on.
As usual, when I started I did not expect it to get so complex. I was looking through the signed/unsigned warnings and did want to to introduce such casting and bounds checking on every place it's needed. This is not some important feature, just slow fixing of compiler warnings. It will improve user experience overall by improved catching of invalid config options. But it's not a big and shiny feature for 4.0 :-(

jrosdahl:
I understand. Sorry if some frustration about being short of spare time shone through.

> There are lots of places where ccache is parsing integers and mostly they are parsed to a slightly wrong datatype and then cast implicitly or explicitly to the right one.

Perhaps you could start by listing where those lots of places are so I can get some more context?

AlexanderLanin:
As you see in my quick response to your findings I totally get having not enough time.

src/Util.hpp
```diff
@@ -230,10 +231,62 @@ bool matches_dir_prefix_or_file(nonstd::string_view dir_prefix_or_file,
...
 std::string normalize_absolute_path(nonstd::string_view path);
 
-// Parse a string into an integer.
-//
-// Throws Error on error.
+/// Parse a string into an integer.
+/// Integer type must be passed as template parameter.
+///
+/// @param errorString for better error messages pass a user readable string.
+/// @throw Error on error.
```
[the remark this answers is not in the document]
AlexanderLanin: Sure, I will change it. But what exactly is the convention to be used? From what I see it's mostly informal, but I wanted to add a sentence specifically to that parameter.
Rather few places seem to document such stuff.

One example might be:

    // Write data from a buffer to the compressed stream.
    //
    // Parameters:
    // - data: Data to write.
    // - count: Size of data to write.
    //
    // Throws Error on failure.
    virtual void write(const void* data, size_t count) = 0;

Or I just drop the @param and @throw and write it as a sentence.

src/Util.hpp
```diff
@@ -62,7 +63,7 @@ template<>
 inline void
 big_endian_to_int(const uint8_t* buffer, int8_t& value)
 {
-  value = buffer[0];
+  std::memcpy(&value, &buffer[0], 1);
 }
```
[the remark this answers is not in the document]
AlexanderLanin: I think I went overboard here. Not sure what I thought back then. I'll remove this change.

AlexanderLanin:
Ok so clearly "lots of places" was an exaggeration. I don't like the added complexity, but I do like the new more precise error messages.

unittest/test_Config.cpp
AlexanderLanin: improved ?!

src/Util.hpp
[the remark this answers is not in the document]
jrosdahl: Yes, the example you quoted is the style I used when I began C++-ifying code. @totph then [wanted to use a less verbose doc string style](https://github.com/ccache/ccache/pull/483#discussion_r342804473) (with private follow up mail discussion) when he worked on the Context PRs. Perhaps it was a mistake, but I didn't object since I didn't feel very strongly about it, and that's the style I've used since then.

What I meant more specifically by not introducing another doc string convention was to not use Doxygen-style markup (`///`, `@param`, etc) since that would be yet another style.

Here's what a @totph-style doc string could look like in this case:

```c++
// Parse `string` into an integer. The integer type is given by the template
// parameter. `error_string` should be a user-readable string describing the
// expected type.
//
// Throws `Error` on error.
template<class IntType>
IntType
stoi(const std::string& string, const char* error_string = "number")
```

src/Util.hpp
```diff
@@ -238,10 +239,55 @@ bool matches_dir_prefix_or_file(nonstd::string_view dir_prefix_or_file,
...
+    ss >> result;
+  } catch (...) {
+    // Internal error in Stream operator (rare).
+    throw Error(
+      fmt::format("invalid {} (exception): \"{}\"", error_string, value));
+  }
+
+  // Invalid format etc.
+  if (!ss)
+    throw Error(fmt::format("invalid {}: \"{}\"", error_string, value));
```
jrosdahl, line 282:
```suggestion
  if (!ss) {
    throw Error(fmt::format("invalid {}: \"{}\"", error_string, value));
  }
```

src/Util.hpp
```diff
@@ -238,10 +239,55 @@ bool matches_dir_prefix_or_file(nonstd::string_view dir_prefix_or_file,
...
+    }
+    return temp;
+  }
+
+  // Stream operator doesn't handle negative values for unsigned types.
+  if (std::is_unsigned<IntType>::value && value[0] == '-')
+    throw Error(fmt::format(
+      "invalid {} (negative value passed where positive was expected): \"{}\"",
+      error_string,
+      value));
```
jrosdahl, line 268:
```suggestion
  if (std::is_unsigned<IntType>::value && value[0] == '-') {
    throw Error(fmt::format(
      "invalid {} (expected non-negative value): \"{}\"",
      error_string,
      value));
  }
```

src/Util.hpp
```diff
@@ -238,10 +239,55 @@ bool matches_dir_prefix_or_file(nonstd::string_view dir_prefix_or_file,
...
 
-// Parse a string into an integer.
-//
-// Throws Error on error.
-int parse_int(const std::string& value);
+/// Parse a string into an integer.
+/// Integer type must be passed as template parameter.
+///
+/// Pass a human readable error_string for better error messages.
+/// Function will throw Error on error.
```
jrosdahl, line 246: `///` -> `//`

src/Util.hpp
```diff
@@ -238,10 +239,55 @@ bool matches_dir_prefix_or_file(nonstd::string_view dir_prefix_or_file,
...
+template<class IntType>
+IntType
+stoi(const std::string& value, const char* const error_string = "number")
```
jrosdahl, line 249:
```suggestion
stoi(const std::string& value, const char* error_string = "number")
```

unittest/.clang-tidy
```diff
@@ -1,3 +1,10 @@
+# This .clang-tidy file is used by CI to ensure that commits do not worsen the
+# codebase. The checks and values below are the minimum standard for new code.
+# (Without claiming that they are 100% correct. They can be modified on demand!)
+# If you want to improve the codebase try enabling some additional checks or
+# playing with the configuration values.
+# Some checks are highly style dependent. Activating all of them is NOT the goal.
```
jrosdahl, line 6:
```suggestion
#
# Some checks are highly style dependent. Activating all of them is NOT the goal.
```

AlexanderLanin:
I'll try to revisit this PR, but I'm focusing on CI at the moment. It's just uselessly listed in the PR section at the moment.

PR #587, opened 2020-04-29
Add support for icpc

sebrowne:
-xHost should be ignored as a compiler language for icpc.

### DIFF
--- a/src/ccache.cpp
+++ b/src/ccache.cpp
@@ -2345,6 +2345,10 @@ process_args(Context& ctx,
     }
     if (str_startswith(argv[i], "-x")) {
       if (args_info.input_file.empty()) {
+        if (str_eq(&argv[i][2], "Host")) {
+            // Skip -xHost for Intel compiler support
+            continue;
+        }
         explicit_language = &argv[i][2];
       }
       continue;


src/ccache.cpp
```diff
@@ -2345,6 +2345,10 @@ process_args(Context& ctx,
     }
     if (str_startswith(argv[i], "-x")) {
       if (args_info.input_file.empty()) {
+        if (str_eq(&argv[i][2], "Host")) {
+            // Skip -xHost for Intel compiler support
+            continue;
+        }
```
jrosdahl, line 2351: This means that `ccache icpc -xHost -c foo.cpp` will make ccache invoke `icpc -c foo.cpp`, i.e. `-xHost` will be thrown away which I guess isn't what you intended.

The logic should look something like this (outside of the `if (... "-x") {` statement):

```c++
    if (str_eq(argv[i], "-xHost")) {
      // -xHost is an ordinary Intel compiler option, not a language
      // specification.
      args_add(common_args, argv[i]);
      continue;
    }
```

What do you think?

sebrowne: Sounds good to me, my mistake reading through the code.

jrosdahl: > My project may have a few other changes coming through related to the Intel compiler, as we find them.

Sounds good.

PR #588, opened 2020-05-03
Don't depend on how echo interprets backslash escapes

liljenzin:
The escape sequence '\n' should be translated to a new line character
when printed to version.cpp. Default mode might be implementation
dependent, thus add the command line switch to make it portable. At
least in Fedora /usr/bin/echo defaults to -E.

### DIFF
--- a/dev.mk.in
+++ b/dev.mk.in
@@ -109,7 +109,8 @@ dist_files = \
     $(built_dist_files)
 
 ifneq ($(shell sed 's/.*"\(.*\)".*/\1/' src/version.cpp 2>/dev/null),$(version))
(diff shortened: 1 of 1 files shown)


liljenzin:
Apparently older version of echo didn't understand the command line switch. Thus split the command into two instead as it should be fully portable.

jrosdahl:
OK, so that's the [second regression](https://github.com/ccache/ccache/commit/ec04de03ebd59f83f2cd5f90a9071fedb523b3a9) in 827237fd... I should have spotted both before accepting it.

By the way, the portable way to print newlines is to use `printf(1)`.

AlexanderLanin:
Maybe the next big thing could really be to establish testing with docker

PR #589, opened 2020-05-08
Support clang options for compiling with sampled profiles

liljenzin:
Add support for -fprofile-sample-use and -fprofile-sample-accurate.


jrosdahl:
Thanks. I'll backport it to 3.7-maint.

Issue #524, opened 2020-01-26
Share cache entries for different object file names when using -MD/-MMD


jrosdahl:
### Description ###

Ccache includes the object file paths (the argument to `-o`) in the hash, so these two compilations won't share cache entries:

```
ccache cc -MMD -c foo.c -o foo1.o
ccache cc -MMD -c foo.c -o foo2.o
```

The reason for doing this is that the `.d` file will be different in the two cases since it contains the name of the object file.

It should be possible to let Ccache share cache entries and adjust the `.d` file appropriately when copying it from the cache.

Related issue: #359.

PR #590, opened 2020-05-11
Share cache for different object file names with -MD/-MMD

dyadav7:
This is the fixes for #524.
Please check it and pull.

Thanks
Deepak

### DIFF
--- a/src/ArgsInfo.hpp
+++ b/src/ArgsInfo.hpp
@@ -58,6 +58,9 @@ struct ArgsInfo
   // Is the compiler being asked to output dependencies?
   bool generating_dependencies = false;
 
+  // Option MD/MMD
+  bool seen_MD_MMD = false;
+
   // Is the compiler being asked to output coverage?
   bool generating_coverage = false;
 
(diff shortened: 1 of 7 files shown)


jrosdahl:
Thanks for working on this. There is however some road left to travel before reaching the goal.

* I think that it's too much of a hack to make the `result_*` routines rewrite the dependency file content, sorry. The result code should only know how to read and write result files; it needs to stay agnostic to the actual contents of the actual embedded files. The way I would like to see it implemented is to pass an implementation of a  "local file interface" into `result_get` and `result_put` instead of the `result_file_map` parameter. Rewriting the dependency file content can then be done by the caller's implementation of the interface.
* The functionality needs automated testing. See the `test` directory for inspiration.
* Speaking of tests, please look into the failing CI tests.
* Question: Wouldn't it work to just remove the `seen_MD_MMD` flag and rely on the existing `generating_dependencies` (then `-Wp,-MD,...` and `-Wp,-MMD,...` will be handled as well)?

dyadav7:
I will take care of review comments and send a new pull request.
First i would like to like -MD/-MMD. Then i will add support for  -Wp,-MD and -Wp,-MMD

PR #594, opened 2020-05-19
Visual C/C++ compiler support

schiele:
This in fact is only a forward port of https://github.com/ccache/ccache/pull/506 to current master. Feel free to integrate this into the original pull request and keep the discussion concentrated in there.

Note that at this point in time I did not do extensive testing besides building on Ubuntu 16.04.6 LTS (Xenial Xerus) and running the included test suite there.

### DIFF
--- a/src/argprocessing.cpp
+++ b/src/argprocessing.cpp
@@ -221,7 +221,7 @@ process_arg(Context& ctx,
   }
 
   // Special case for -E.
-  if (args[i] == "-E") {
+  if (args[i] == "-E" || args[i] == "/E") {
     return STATS_PREPROCESSING;
   }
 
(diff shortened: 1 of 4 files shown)


jrosdahl:
Thanks for your contribution.

> Feel free to integrate this into the original pull request and keep the discussion concentrated in there.

If you want to collaborate on Visual C++ support I suggest making a pull request against @cristianadam's branch instead and, as you say, keep the discussion in #506.

schiele:
> If you want to collaborate on Visual C++ support I suggest making a pull request against @cristianadam's branch instead and, as you say, keep the discussion in #506.

Good point. Will do it that way.

PR #605, opened 2020-06-15
Max old

sjamgade:
Adding one more way to control the ccache size. The commit message would give more clarity.

I am not sure if the second commit is adding value or complexity. (need feedback)

I am writing documentation right now. Looking for some help on how to approach it.

### DIFF
--- a/doc/MANUAL.adoc
+++ b/doc/MANUAL.adoc
@@ -94,6 +94,32 @@ Common options
     Clear the entire cache, removing all cached files, but keeping the
     configuration file.
 
+*`--dump-manifest`* _PATH_::
+
+    Dump manifest file at _PATH_ in text format. This is only useful when
+    debugging ccache and its behavior.
+
+*`--evict-older-than`* _NUM_::
+
+    This is another cleanup mechanism (independent of max_files and max_size).
+    It instructs ccache to perform cleanup based on age of the stored files.
+    The value _NUM_ is duration and means any cache entry older than it from CURRENT_TIME
+    should be deleted. _NUM_ should be an unsigned number with an optional suffix:
+    d(ays)/s(econds). Default being d(ays).
+    This cleanup can be performed from the cli with this option only. During the
+    cleanup max_files/max_size will not be taken into consideration.
+
+
+*`-k`* _KEY_, *`--get-config`* _KEY_::
+
+    Print the value of configuration option _KEY_. See
+    <<_configuration,CONFIGURATION>> for more information.
+
+*`--hash-file`* _PATH_::
+
+    Print the hash (160 bit BLAKE3) of the file at _PATH_. This is only useful
+    when debugging ccache and its behavior.
+
 *`-h`*, *`--help`*::
 
     Print an options summary page.
--- a/doc/MANUAL.adoc
+++ b/doc/MANUAL.adoc
@@ -94,11 +94,6 @@ Common options
     Clear the entire cache, removing all cached files, but keeping the
     configuration file.
 
-*`--dump-manifest`* _PATH_::
-
-    Dump manifest file at _PATH_ in text format. This is only useful when
-    debugging ccache and its behavior.
-
 *`--evict-older-than`* _NUM_::
 
     This is another cleanup mechanism (independent of max_files and max_size).
@@ -108,17 +103,6 @@ Common options
     d(ays)/s(econds). Default being d(ays).
     This cleanup can be performed from the cli with this option only. During the
     cleanup max_files/max_size will not be taken into consideration.
-
-
-*`-k`* _KEY_, *`--get-config`* _KEY_::
-
-    Print the value of configuration option _KEY_. See
-    <<_configuration,CONFIGURATION>> for more information.
-
-*`--hash-file`* _PATH_::
-
-    Print the hash (160 bit BLAKE3) of the file at _PATH_. This is only useful
-    when debugging ccache and its behavior.
 
 *`-h`*, *`--help`*::
 
--- a/doc/MANUAL.adoc
+++ b/doc/MANUAL.adoc
@@ -99,8 +99,8 @@ Common options
     This is another cleanup mechanism (independent of max_files and max_size).
     It instructs ccache to perform cleanup based on age of the stored files.
     The value _NUM_ is duration and means any cache entry older than it from CURRENT_TIME
-    should be deleted. _NUM_ should be an unsigned number with an optional suffix:
-    d(ays)/s(econds). Default being d(ays).
+    should be deleted. _NUM_ should be an unsigned number with a suffix:
+    d(ays)/s(econds).
     This cleanup can be performed from the cli with this option only. During the
     cleanup max_files/max_size will not be taken into consideration.
 
--- a/src/Util.cpp
+++ b/src/Util.cpp
@@ -976,4 +976,39 @@ write_file(const std::string& path,
   file << data;
 }
 
+unsigned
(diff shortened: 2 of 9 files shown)


jrosdahl:
Hi, thanks for your contribution.

I'm hesitant to add the feature as currently designed. ccache can calculate when the _max cache size_ and _max files_ limits have been reached due to the bookkeeping of _cache size_ and _files in cache_. The limits are checked every time a result is placed in the cache due to a cache miss. The automatic cleanup can therefore be triggered exactly when needed. In contrast, there is no bookkeeping that can be used to know when to trigger cleanup for _max old_ (except for checking mtimes of all cache files on every cache miss, which obviously is a bad idea). In your proposal, the _max old_ limit will not kick in if ccache never puts a new result in the cache (e.g. if ccache isn't used at all) or if the _max cache size_ or _max files_ limits are never reached.

Here's another idea: How about making it possible to run `ccache --evict-older-than AGE`? Then you can run that command periodically from crontab or similar to remove old entries, thus driving time-based cache cleanup based on a clock as it should.

sjamgade:
I find you point on-point :-)
thanks for the idea too.

do you imagine the other option as extension to `ccache --cleanup` (take `max_files, max_size` into consideration) and should be supplied with `--cleanup` only. Or something completely independent like do-what-I-asked and do not pay attention to  other limits (max_*)

sjamgade:
I have change the implementation as you suggested. What do you think ?

sjamgade:
okay the CI should be fixed now.

dyadav7:
A suggestion: I generally would cleanup files older than certain number of days. So, The option` --evict-older-than` should have a way to specify days also.  e.g.  `--evict-older-than 100days`

sjamgade:
Yes I too thought of that extension but then I prefer to add  `$((ndays*24*3600))` to my script. I do not think the added complexity of parsing logic is worth it. But if @jrosdahl also prefers it. I will be happy to add it as a following PR. :-)

src/ccache.cpp
```diff
@@ -86,6 +86,8 @@ Common options:
                               (normally not needed as this is done
                               automatically)
     -C, --clear               clear the cache completely (except configuration)
+    -e, --evict-older-than N  delete files older than N seconds (this will not
```
jrosdahl, line 89: I don't think we should dedicate a short option to this since it's a very narrow use case in my mind, so I suggest going with just `--evict-older-than`.

jrosdahl, line 89: How about using days as the unit instead of seconds? That would address @dyadav7's suggestion. I don't see a realistic need for other units. Do you?

src/ccache.cpp
```diff
@@ -2273,6 +2275,7 @@ handle_main_options(int argc, const char* const* argv)
   static const struct option options[] = {
     {"cleanup", no_argument, nullptr, 'c'},
     {"clear", no_argument, nullptr, 'C'},
+    {"evict-older-than", required_argument, nullptr, 'e'},
     {"dump-manifest", required_argument, nullptr, DUMP_MANIFEST},
     {"dump-result", required_argument, nullptr, DUMP_RESULT},
```
jrosdahl, line 2280: Keep the list sorted:

```suggestion
    {"dump-manifest", required_argument, nullptr, DUMP_MANIFEST},
    {"dump-result", required_argument, nullptr, DUMP_RESULT},
    {"evict-older-than", required_argument, nullptr, 'e'},
```

src/cleanup.cpp
```diff
@@ -54,6 +54,7 @@ void
 clean_up_dir(const std::string& subdir,
              uint64_t max_size,
              uint32_t max_files,
+             time_t max_old,
```
jrosdahl, line 57:
```suggestion
             time_t max_age,
```

And similar elsewhere.

src/ccache.cpp
sjamgade: With smallest unit (seconds) a user could can target any unit as needed. Days feels restrictive. And with smaller unit I can better contain results while repetitively compiling same project using git-bisect.

However I dont mind parsing S(econd)/D(ay) to decide what the arg means.  And default to days.

sjamgade: ok.

jrosdahl: > And with smaller unit I can better contain results while repetitively compiling same project using git-bisect.

I don't understand this use case. Can you tell me a bit more why you feel the need to evict results based on time when compiling the same project repetitively?

Regardless, if you want to introduce units then use proper symbols: "s" for second and "d" for day, i.e. lowercase.

src/cleanup.cpp
sjamgade: done

src/ccache.cpp
sjamgade: The use case is: Keep ccache archive for each commit, and remove the unneeded archives once the target commit is found.

sjamgade:
Sorry for the force-push, My git-fu isnt strong enough. 
I will squash the commits once we are good to go. :-)

src/ccache.cpp
jrosdahl: OK, thanks for the clarification. But why not let the normal LRU eviction handle this?

jrosdahl:
> Sorry for the force-push, My git-fu isnt strong enough.
> I will squash the commits once we are good to go. :-)

Force-pushing is perfectly fine. And you don't have to squash unless you want to; I'll squash the commits when merging the PR in the end anyway.

test/suites/cleanup.bash
```diff
@@ -179,4 +179,22 @@ SUITE_cleanup() {
...
+    prepare_cleanup_test_dir $CCACHE_DIR/a
+    touch $CCACHE_DIR/a/now.result
+    $CCACHE -F 0 -M 0 >/dev/null
+
+    $CCACHE --evict-older-than 1d >/dev/null
+    expect_file_count 1 '*.result' $CCACHE_DIR
+    expect_stat 'files in cache' 1
+
+    $CCACHE --evict-older-than 1  >/dev/null
+    expect_file_count 1 '*.result' $CCACHE_DIR
```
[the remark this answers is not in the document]
sjamgade: without suffix we default to days. 

I guess you mean the test below this one. 
So fixed that to 10s

src/Util.cpp
```diff
@@ -883,4 +883,39 @@ write_file(const std::string& path,
   file << data;
 }
 
+unsigned
+parse_duration_with_suffix_to_seconds(const std::string& value)
```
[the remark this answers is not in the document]
sjamgade: sure. done

doc/MANUAL.adoc
```diff
@@ -94,6 +94,16 @@ Common options
     Clear the entire cache, removing all cached files, but keeping the
     configuration file.
 
+*`--evict-older-than`* _NUM_::
+
+    This is another cleanup mechanism (independent of max_files and max_size).
+    It instructs ccache to perform cleanup based on age of the stored files.
+    The value _NUM_ is duration and means any cache entry older than it from CURRENT_TIME
+    should be deleted. _NUM_ should be an unsigned number with an optional suffix:
+    d(ays)/s(econds). Default being d(ays).
```
jrosdahl, line 103: I would actually prefer to not have a default suffix and require `d` for days. Sounds OK?

sjamgade: Okay for me.
Took a while to get back to this :-)

src/ccache.cpp
jrosdahl: @sjamgade: Have you thought about my above question ("why not let the normal LRU eviction handle this?")?

PR #606, opened 2020-06-16
Docker: add Alpine to new docker setup

AlexanderLanin:
Since Alpine has a ccache package, it seems prudent to ensure it works fine.


jrosdahl:
Thanks.

PR #608, opened 2020-06-20
Reduce number of memory allocations in hashing

liljenzin:
Add hash_buffer_once method for hashing a given text immediately using a
stack allocated hasher. Also skip copying the hasher in hash_result as
blake3 is not modifying the hasher.


jrosdahl:
Thanks.

PR #609, opened 2020-06-21
Do not report skipped tests as failures on CMake <3.9

AlexanderLanin:
While CMake 3.8 and before support skipping of tests, they report them as failures. With this change they are reported as Passed. While this is certainly not ideal, it helps automation a lot to have the correct exit codes. Assumption is that noone will use a system with CMake <= 3.8 for development of ccache.

Assuming some tests are Skipped and the rest is Passed:
| |CMake <= 3.8|CMake >= 3.9|
|---|---|---|
|before this PR|Tests reported as Passed or Skipped.<br>Exit code is non zero.|Tests reported as Passed or Skipped.<br>Exit code is 0.|
|after this PR|Tests reported as Passed.<br>Exit code is 0.|Tests reported as Passed or Skipped.<br>Exit code is 0.|


craigscott-crascit:
I confirm that the behavior was changed by [this merge request](https://gitlab.kitware.com/cmake/cmake/-/merge_requests/741) in response to [this issue](https://gitlab.kitware.com/cmake/cmake/-/issues/16822). That change first appeared in CMake 3.9.0.

test/CMakeLists.txt
```diff
@@ -4,12 +4,23 @@ function(addtest name)
...
     WORKING_DIRECTORY ${CMAKE_BINARY_DIR}
   )
-  set(environment CCACHE=${CMAKE_BINARY_DIR}/ccache EXIT_IF_SKIPPED=true)
-  set_tests_properties(
-    "producttest.${name}"
-    PROPERTIES
-    ENVIRONMENT "${environment}"
-    SKIP_RETURN_CODE 125)
+  set_tests_properties("producttest.${name}" PROPERTIES
+    ENVIRONMENT "CCACHE=${CMAKE_BINARY_DIR}/ccache;EXIT_IF_SKIPPED=true")
```
jrosdahl, line 8: (I extracted the environment list to a separate variable since the string list variable syntax felt cleaner.)

jrosdahl:
Good catch, thanks.

test/CMakeLists.txt
AlexanderLanin: Ah sorry, I was confused by set environment and then set tests properties environment. Once I understood what was going on I thought I would make it simpler :-)

PR #610, opened 2020-06-21
Run compilation and tests with multiple threads

AlexanderLanin:
Tested with CTest parameter `--repeat-until-fail 50` locally, locally in docker and on [travis](https://travis-ci.com/github/AlexanderLanin/ccache/builds/172374665). No random problems encountered (Besides Travis killing the MacOS job after a 50 minute timeout).
While this is "only nice" for daily work, it's critical for testing multiple images with docker as the runtime is just too long otherwise.

### DIFF
--- a/ci/build
+++ b/ci/build
@@ -11,11 +11,15 @@ fi
 if [ -n "${SPECIAL:-}" ]; then
     exec "ci/$SPECIAL"
 else
+    [ -z ${JOBS+x} ] && JOBS=$(getconf _NPROCESSORS_ONLN 2>/dev/null)
+    [ -z ${JOBS+x} ] && JOBS=2
+
     mkdir -p ${BUILDDIR:-build}
     cd ${BUILDDIR:-build}
-    ${CMAKE_PREFIX:-} cmake "${CCACHE_LOC:-..}" ${CMAKE_PARAMS:-}
+    ${CMAKE_PREFIX:-} cmake ${CMAKE_PARAMS:-} ${CCACHE_LOC:-..}
     # 4 jobs seem to be a reasonable default for Travis.
-    ${CMAKE_PREFIX:-} cmake --build . ${BUILDEXTRAFLAGS:-} -- -j4
+    ${CMAKE_PREFIX:-} cmake --build . ${BUILDEXTRAFLAGS:-} -- -j${JOBS}
     # Warning: Rare random failures when running with -j4.
-    [ "${RUN_TESTS:-1}" -ne "1" ] || ctest --output-on-failure -j1
+    [ "${RUN_TESTS:-1}" -ne "1" ] || ctest --output-on-failure -j${JOBS}
 fi
+exit 0


ci/build.sh
```diff
@@ -1,15 +1,20 @@
...
-  ${CMAKE_PREFIX:-} cmake ${CCACHE_LOC:-..} ${CMAKE_PARAMS:-}
-  # 4 threads seems a reasonable default for Travis
-  ${CMAKE_PREFIX:-} cmake --build . ${BUILDEXTRAFLAGS:-} -- -j4
-  # Warning: Rare random failures when running with j4.
-  test "${RUN_TESTS:-1}" -eq "1" && ctest --output-on-failure -j1
+  JOBS=$(getconf _NPROCESSORS_ONLN 2>/dev/null) # Linux
+  [ -z "$JOBS" ] && JOBS=$(getconf NPROCESSORS_ONLN) # FreeBSD
+  [ -z "$JOBS" ] && JOBS=$(ksh93 -c 'getconf NPROCESSORS_ONLN') # Solaris
+  [ -z "$JOBS" ] && JOBS=2
+  test "$JOBS" -gt 7 && JOBS=$((JOBS - 1)) # Leave one free for mails, browsing, etc
```
jrosdahl, line 11: Why 7?

ci/build.sh
```diff
@@ -1,15 +1,20 @@
...
-  mkdir -p ${BUILDDIR:-build}
-  cd ${BUILDDIR:-build}
-  ${CMAKE_PREFIX:-} cmake ${CCACHE_LOC:-..} ${CMAKE_PARAMS:-}
-  # 4 threads seems a reasonable default for Travis
-  ${CMAKE_PREFIX:-} cmake --build . ${BUILDEXTRAFLAGS:-} -- -j4
-  # Warning: Rare random failures when running with j4.
-  test "${RUN_TESTS:-1}" -eq "1" && ctest --output-on-failure -j1
+  JOBS=$(getconf _NPROCESSORS_ONLN 2>/dev/null) # Linux
+  [ -z "$JOBS" ] && JOBS=$(getconf NPROCESSORS_ONLN) # FreeBSD
+  [ -z "$JOBS" ] && JOBS=$(ksh93 -c 'getconf NPROCESSORS_ONLN') # Solaris
```
jrosdahl, line 9: Is the `nproc` command available on FreeBSD and Solaris?

AlexanderLanin: nproc is an optional package. e.g. on Alpine it's not available by default.
FreeBSD does something like `alias nproc='sysctl -n hw.ncpu'`, so no. Not by default.
Not sure about Solaris.

I found this on the internet and cannot verify FreeBSD or solaris myself. But multiple websites show this approach. e.g. https://www.mmbyte.com/article/41367.html, https://unix.stackexchange.com/questions/564494/how-to-posix-ly-find-out-how-many-cpu-threads-do-i-have-within-a-shell-script

ci/build.sh
AlexanderLanin: "random" number:
- When you have only 1 core you don't want to keep 1 free ;-)
- Even when you have 2 cores you don't want to keep 50% free!
- I have 8 cores (4*2) and I do want to keep one free. Theoretically it could cost me 12,5% time, but usually my CPU is busy 95% that way. So that last free core is not really free anyway.
Loosing 5% speed so that all other programs keep running fine seems ok.

Somewhere in there is the sweet spot and I picked my personal setup.

jrosdahl: It looks strange to say "Leave one free for mails, browsing, etc" and then use your personal value. Unless it's your personal script, that is. Maybe we should use 63 since I have access to a 64 CPU machine? :slightly_smiling_face:

How about letting the user decide with e.g. `JOBS=7 the-script` or `the-script -j 7` instead?

ci/build.sh
jrosdahl: I feel like I'm missing a piece of the puzzle. Do you plan to add CI jobs for FreeBSD and Solaris?

ci/build.sh
AlexanderLanin: Ok now you got me searching for the optimal number... and of course there are as many opinions as there are people. The range seems to be cores-1 up to cores*2,5. However the speedup when using more jobs then cores seems quite small.

I think for a "small" project like ccache this is not mission critical / can be improved later. Using jobs = cores is a huge improvement over cores = 1.

jrosdahl: OK, right. But what do you think of my suggestion? Maybe the user _does_ want to use only one CPU sometime because of $reasons?

ci/build.sh
AlexanderLanin: In case we pick Cirrus we might run FreeBSD jobs. But I have not found any CI that supports Solaris.

However that was not the main reason I added those lines. It was simply because it was easy to copy paste them here. They are usually mentioned together with the Linux solution when searching how to acquire the number of cores. So I did not have to research them in addition and it doesn't look like they require extra maintenance effort. So I added them.

I actually started with nproc as you suggested, but that failed on Alpine Linux without installing an additional package. I had considered that also, but it seemed overkill because this "uglier" solution is still quite trivial.

Since this script is within "ci" we can drop the other systems. I actually did not think of that. I thought people would be using this script to build, but that's wrong.

ci/build.sh
AlexanderLanin: As I just discovered in the other finding this script is actually only for CI. Users would normally not use it?
Maybe through docker, but that would mean we need to pass -j through all the other scripts too.

Nevertheless using the environment variable could be as trivial as prefixing the first line with `[ -z "$JOBS" ] && `. I can certainly do that.

AlexanderLanin:
rebased

jrosdahl:
Thanks. I made some further tweaks in 876071f9b3d2.

PR #613, opened 2020-06-22
Remove stale licenses

erijo:
The m4 files are no longer used after the switch to cmake in f2341b2.


jrosdahl:
Now I have one thing less on my list of things to clean up after the CMake switch. :slightly_smiling_face:

Issue #617, opened 2020-07-04
The -fdirectives-only option is not passed to the compiler when run_second_cpp=true


deriamis:
### How to reproduce ###

This is most visible when you are using `-fdirectives-only` where it can cause a failure, such as when using ccache as a frontend to Icecream >= 1.2 and GCC as the remote compiler.

1. Use GCC as your compiler for a project
2. Specify the -fdirectives-only option on the compiler command line
3. Watch the option get passed no matter what

### Actual behavior ###

This is straightforward behavior - because ccache doesn't know `-fdirectives-only` is a preprocessor-specific flag, it gets passed to the compiler even when it should not, such as when `run_second_cpp=1`. If you are using ccache as a frontend to Icecream, which in later versions passes `-fdirectives-only` to the preprocessor when remote preprocessing is enabled, the flag can cause a compilation failure when it's also passed to the compiler (specifically GCC in this case).

### Expected behavior ###

The flag should not have been passed to the compiler (should have only been passed to the preprocessor), and compilation should have succeeded.

### Environment ###

ccache version 3.7.7

Description:    Ubuntu 18.04.4 LTS
Release:        18.04
Codename:       bionic

gcc (GCC) 8.2.0

ICECC 1.3.1

jrosdahl:
Thanks for the bug report.

I don't quite understand the issue, though. The reproduction steps only say that the option is passed to the compiler, not why that would be a problem.

>  `-fdirectives-only` […] gets passed to the compiler even when it should not, such as when `run_second_cpp=1`

Do you mean `run_second_cpp = false`? `1` is not a valid value for `run_second_cpp` and `true` means "don't compile the preprocessed output" and it that case `-fdirectives-only` should definitely be passed to the compiler, right?

> If you are using ccache as a frontend to Icecream, which in later versions passes -fdirectives-only to the preprocessor when remote preprocessing is enabled, the flag can cause a compilation failure when it's also passed to the compiler (specifically GCC in this case).

Why can that flag cause a compilation failure? Please provide details.

> The flag should not have been passed to the compiler (should have only been passed to the preprocessor), and compilation should have succeeded.

Why not? And what compilation do you refer to?

That said, it's seems perfectly OK to avoid passing `-fdirectives-only` to the compiler if `run_second_cpp` is `false`, but I would like to understand why sending `-fdirectives-only` to the compiler is a problem in the first place.

deriamis:
> Thanks for the bug report.
> 
[...]

PR #618, opened 2020-07-04
Make ccache aware of -fdirectives-only as a preprocessor option

deriamis:
If the `-fdirectives-only` option is specified on the compler command line, ccache does not correctly flag it as a preprocessor-only option. This can cause problems if the flag were ever passed to a different tool (such as Icecream) that *also* doesn't know the option is preprocessor-only. In such cases, GCC will try to preprocess the same source twice after making changes to it that don't allow it to be preprocessed a second time, leading to a compiler failure.

I tested this change locally, and it solved my particular problem (see #617).

### DIFF
--- a/doc/MANUAL.adoc
+++ b/doc/MANUAL.adoc
@@ -495,6 +495,15 @@ might be incorrect.
     change. The list separator is semicolon on Windows systems and colon on
     other systems.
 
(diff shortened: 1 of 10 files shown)


deriamis:
Moving commits to their own branches.

PR #620, opened 2020-07-04
Add ability to ignore specified compiler options when hashing

### LINKED ISSUE #578: Make it possible to ignore certain compiler options
### Description

Sometimes you may want to exclude certain compiler options from the hash since you either know they don't affect the compiler output (and ccache doesn't know that) or you don't care that they affect the compiler output. ("output" here means files or warnings produced by the compiler.)

A suggestion from https://github.com/ccache/ccache/pull/570#issuecomment-605462053 is to introduce a generic `ignore_options` (`CCACHE_IGNOREOPTIONS`) configuration setting for this. The value could be a list of glob patterns that make ccache exclude matching options from the hash.

deriamis:
This change simply adds the ability to configure ccache to ignore some compiler options when creating a hash. I implemented this as a simple argument list, with the ability to create a prefix by using `*` as the final character, somewhat similar to a wildcard.

I considered a few different ways of doing this that would not involve looping through all the options in the ignore list, but I figured the list was likely to be very short in almost all cases, and I would not be able to use prefix matching if I didn't loop, so I chose not to worry about it much. I also briefly considered using a more powerful pattern-matching scheme like a regex, but I rejected that because - again - the list is likely to be very short in most cases, and a prefix (the simplest solution) is likely to be a good solution almost every time. If we need something more powerful in the future, it shouldn't be too hard to add it without breaking anything.

Closes #578.


src/argprocessing.cpp
```diff
@@ -222,6 +222,14 @@ process_arg(Context& ctx,
     return nullopt;
   }
 
+  for (const std::string& option : ctx.ignore_options) {
+    if (option == args[i] || (Util::ends_with(option, "*") &&
+          Util::starts_with(args[i], option.substr(0, option.length() - 1)))) {
```
AlexanderLanin, line 227: How about extract these two lines of the if statement in a separate function? Maybe the loop as well. Something like `bool optionToBeIgnoredForCache()`?
It doesn't seem like much, but then the if statement would be self documented:
```
if(optionToBeIgnoredForCache(ctx.ignore_options, args[i])) {
      cc_log("Skipping argument: %s", args[i].c_str());
      return nullopt;
}
```

src/Context.hpp
```diff
@@ -107,6 +107,10 @@ class Context : NonCopyable
   // Headers (or directories with headers) to ignore in manifest mode.
   std::vector<std::string> ignore_header_paths;
 
+  // Options to completely ignore on the command line. They will not be passed
+  // to either the preprocessor or the compiler.
+  std::vector<std::string> ignore_options;
```
AlexanderLanin, line 112: ' They will not be passed to either the preprocessor or the compiler' - really? Ccache shall ignore them, but still pass them on unmodified?! Or did I get the idea here wrong?

src/ccache.cpp
```diff
@@ -1881,6 +1881,8 @@ set_up_context(Context& ctx, int argc, const char* const* argv)
   ctx.orig_args = Args::from_argv(argc, argv);
   ctx.ignore_header_paths = Util::split_into_strings(
     ctx.config.ignore_headers_in_manifest(), PATH_DELIM);
+  ctx.ignore_options = Util::split_into_strings(
+      ctx.config.ignore_options(), " ");
```
AlexanderLanin, line 1885: Check for max one \'\*\' char? More precisely check no char but the last one is a '*'?

unittest/test_argprocessing.cpp
```diff
@@ -700,3 +702,27 @@ TEST_CASE("cuda_option_file")
...
+  Args act_cpp;
+  Args act_extra;
+  Args act_cc;
+
+  Util::write_file("foo.c", "");
+  CHECK(!process_args(ctx, act_cpp, act_extra, act_cc));
+  CHECK(exp_cpp == act_cpp);
+  CHECK(exp_extra == act_extra);
+  CHECK(exp_cc == act_cc);
+}
```
AlexanderLanin, line 728: Could you add a matching test where the ignore option does not work?
e.g. -fmessage-length=2* but passing 40.

I know this is paranoid, so just in case you add another commit anyway. I think we've all seen lots of previously working code break unintentionally by seemingly unrelated changes.

AlexanderLanin:
Note: I've reviewed readability only. Not whether that PR makes sense. Although it sounds good :-)

src/Context.hpp
deriamis: That comment was from before I fully understood what I was doing, and I missed changing it accordingly once I did. I’ll fix it.

src/argprocessing.cpp
deriamis: Good idea. I’ll do that.

src/ccache.cpp
deriamis: I was on the fence about using matches at this point, and I was shopping for ideas here. If we like the trailing-star match thing, I can make the check for one and only star at the end part of a setter or something.

unittest/test_argprocessing.cpp
deriamis: No, that’s fair enough. I was mostly demonstrating that it works as intended here as a proof-of-concept. If we like the functionality, I’ll also add a negative test as well.

deriamis:
I took the suggestions and fleshed out the implementation a bit more. Specifically, I moved ignore_options into the context implementation with a getter and setter, and the setter now checks each option for validity before adding it. It looks like I don't have a good way of bailing on a bad ignore_option item (nor am I sure I would want to), so I opted to ignore the ones that are malformed with a log message. Please let me know if I am doing that wrong. I also extracted the logic for checking arguments against the ignore list into a function for better readability.

In the tests, I added a check for invalid items in ignore_options as well as negative tests in the matches. I believe it covers all the bases there now, but let me know if you can imagine any big holes. I am sort of happy-ish with the implementation, but it is kinda weak to only allow a single wildcard at the end IMO.

I'm on the fence about prefix matches, but regexes are also a bit overkill and about an order of magnitude (at least) slower. They should be easy enough to implement at this point if you think it's worth it, though. I suggest we make that decision early, since doing it later is a syntax change to an existing option, and that seems uncool.

src/Context.cpp
```diff
@@ -118,3 +122,17 @@ Context::unlink_pending_tmp_files()
   }
   m_pending_tmp_files.clear();
 }
+
+void
+Context::set_ignore_options(const std::vector<std::string> options)
+{
+  for (const std::string& option : options) {
+    std::size_t n_wildcards = std::count(option.cbegin(), option.cend(), '*');
+    if (n_wildcards == 0 || (n_wildcards == 1 && option.back() == '*')) {
```
jrosdahl, line 131: I'm fine with this limitation.

src/argprocessing.cpp
```diff
@@ -222,6 +236,11 @@ process_arg(Context& ctx,
     return nullopt;
   }
 
+  if (shouldBeIgnoredForCache(args[i], ctx.ignore_options())) {
+    cc_log("Skipping ignored argument: %s", args[i].c_str());
+    return nullopt;
+  }
```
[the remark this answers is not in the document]
deriamis: Ah, I was kinda confused as to how that worked. I also wanted them to be passed but wasn't sure how that actually got accomplished.

deriamis:
I've changed things around as requested (and as they should be). I finally got VIM playing nice with the source, so I was able to identify a few opportunities for improvement, which I also added.

I am not completely sold on the test, though, because while I am able to verify that the hashed arguments match between both runs, the hashes still differ for some reason. I would like to ensure the stronger guarantee that the hashes would match if I can. I get the feeling I am missing something that would make the hashes match. Do I need to write out the manifest file as part of the test to get that to work properly?

src/ccache.cpp
```diff
@@ -1633,6 +1659,7 @@ calculate_result_name(Context& ctx,
     }
 
     ctx.set_manifest_name(hash_result(hash));
+    cc_log("Hashed command line: %s", ctx.hashed_args.to_string().c_str());
```
jrosdahl, line 1662: This is incorrect since there are special cases in lines ≈1451-1561 (e.g. `-fdebug-prefix-map=`, `-MF` and `-fplugin=`) which aren't part of `ctx.hashed_args`. I would prefer removing `ctx.hashed_args` and perform tests using the regular test suite instead.

src/Context.hpp
```diff
@@ -146,6 +153,9 @@ class Context : NonCopyable
   std::string m_result_path;
   mutable std::string m_result_stats_file;
 
+  // Options to ignore for the hash
```
jrosdahl, line 156:
```suggestion
  // Options to ignore for the hash.
```

src/ccache.cpp
deriamis: This was me trying to convince myself that I wasn't going crazy. I forgot to remove it, sorry.

deriamis:
> > I finally got VIM playing nice with the source, so I was able to identify a few opportunities for improvement, which I also added.
> 
> The editor settings are actually not important since the formatting is defined by clang-format, either by you or by me before accepting the PR. See _Code style_ on https://github.com/ccache/ccache/blob/master/CONTRIBUTING.md#code-style.
> 

One reason it was important to me is that I have a VIM plugin that runs clang-format. ;) Kinda immaterial to the improvements I saw, though. I'm still fairly new to modern C++, so I don't catch a lot of the things I should yet. Thank you for kindly helping me improve.

> > I am not completely sold on the test, though, because while I am able to verify that the hashed arguments match between both runs, the hashes still differ for some reason.
> 
> How did you conclude that the hashes differ?
> 

I checked the hashes (the out arguments) for equality after running the function. I suppose it probably won't matter if I get the correct results in the test you suggest, though.

> > I would like to ensure the stronger guarantee that the hashes would match if I can. I get the feeling I am missing something that would make the hashes match. Do I need to write out the manifest file as part of the test to get that to work properly?
> 
> My suggestion is to not test this in the unit test suite but in the regular test suite. That's where all ccache behavior testing is performed. Perhaps like this?
> 
> ```diff
> --- a/test/suites/direct.bash
> +++ b/test/suites/direct.bash
> @@ -1042,4 +1042,27 @@ EOF
>      if [ -n "$data" ]; then
>          test_failed "$manifest contained ignored header: $data"
>      fi
> +
> +    # -------------------------------------------------------------------------
> +    TEST "CCACHE_IGNOREOPTIONS"
> +
> +    CCACHE_IGNOREOPTIONS="-DTEST=1" $CCACHE_COMPILE -DTEST=1 -c test.c
> +    expect_stat 'cache hit (direct)' 0
> +    expect_stat 'cache hit (preprocessed)' 0
> +    expect_stat 'cache miss' 1
> +
> +    CCACHE_IGNOREOPTIONS="-DTEST=1*" $CCACHE_COMPILE -DTEST=1 -c test.c
> +    expect_stat 'cache hit (direct)' 1
> +    expect_stat 'cache hit (preprocessed)' 0
> +    expect_stat 'cache miss' 1
> +
> +    CCACHE_IGNOREOPTIONS="-DTEST=1*" $CCACHE_COMPILE -DTEST=12 -c test.c
> +    expect_stat 'cache hit (direct)' 2
> +    expect_stat 'cache hit (preprocessed)' 0
> +    expect_stat 'cache miss' 1
> +
> +    CCACHE_IGNOREOPTIONS="-DTEST=2*" $CCACHE_COMPILE -DTEST=12 -c test.c
> +    expect_stat 'cache hit (direct)' 2
> +    expect_stat 'cache hit (preprocessed)' 1
> +    expect_stat 'cache miss' 1
>  }
> ```

Awesome! Ok, this makes me feel a bit better about where I'm going. I'll give this a try later this evening and push it back up for you.

src/ccache.cpp
```diff
@@ -1406,6 +1420,16 @@ calculate_result_name(Context& ctx,
 
   // First the arguments.
   for (size_t i = 1; i < args.size(); i++) {
+    // Trust the user if they've said we should not hash a given option.
+    if (option_should_be_ignored(args[i], ctx.ignore_options())) {
+      cc_log("Skipping ignored argument: %s", args[i].c_str());
```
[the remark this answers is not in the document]
deriamis: You're right. Not sure how I missed that.

deriamis:
Alright, latest diff is up. If you'd rather do regexes instead of prefix matching, let me know and I can work it in. Otherwise, I think what I have now matches the requested functionality. Again, thank you for your help getting it ready!

jrosdahl:
Thanks!

PR #621, opened 2020-07-05
GitHub Actions

AlexanderLanin:
This is probably not reasonable to be merged as is. It's mostly to show what an exact replica of travis.yml would look like.


jrosdahl:
Thanks. Having a look at GitHub Actions has been on my to do list for a while now so it's great that you're doing that.

The _Run Clang-Format_ job currently fails like this:

```
[...]
+ exec ci/misc/format-files --all --check
ci/build: 12: exec: ci/misc/format-files --all --check: not found
##[error]Process completed with exit code 127.
```

But that's not indicated in any way; you need to expand the _Build and Test_ step to see that. Do you have any insight here (about the missing indication, not about not finding `ci/misc/format-files`)? ... OK, I guess this is due to `continue-on-error` and https://github.community/t/continue-on-error-allow-failure-ui-indication/16773. So currently the two jobs marked EXPERIMENTAL are kind of useless. But that's not a blocker for me.

What other pros and cons have you found so far compared to Travis?

AlexanderLanin:
I'm certainly no expert on either. As you already discovered there is no "yellow" status in GitHub Actions. No other con so far. Biggest pro is certainly having artifacts. Let's say your commit breaks behavior on MacOS. You get a testdir.zip for MacOS! Other pro: It's faster and it supports Windows.

AlexanderLanin:
I've split the jobs into two sections: "standard tests" and "special tests". Not sure this makes a lot of sense as the possibilities of what to run the tests on are kind of endless. But I wanted to include Ubuntu 16 and 20 with gcc and clang somehow. And that would have been 4 new test cases already.
Maybe it would make more sense to list them kind of as in misc/test-all-systems, here is an approach for that: https://github.com/onqtam/doctest/blob/master/.github/workflows/main.yml

Other:
* fun list of artifacts: https://github.com/AlexanderLanin/ccache/actions/runs/162487451
* doctest does not compile on macos when building from source.zip. No idea what that is about. Maybe it has to do with the next point.
* the build-and-verify scripts do not properly propagate ENV. They are not using ci/build. More cleanup required.

AlexanderLanin:
@jrosdahl shall we pursue this further? It's not a relevant feature for 4.0 yet again, but could have prevented e.g. #630

jrosdahl:
> @jrosdahl shall we pursue this further? It's not a relevant feature for 4.0 yet again, but could have prevented e.g. #630

Yes, that would be appreciated.

AlexanderLanin:
It seems GitHub Actions won't start for this pull request for some reason. Maybe because the file has to be on master first. I don't know. You can see what the execution looks like over here: https://github.com/AlexanderLanin/ccache/actions/runs/199662276

In current state this mostly mimics travis. I have not included anything special here like windows or releases.

Note: The warning "Unexpected input(s) 'CC', 'CXX', 'VERBOSE', 'ENABLE_CACHE_CLEANUP_TESTS', 'CTEST_OUTPUT_ON_FAILURE', valid inputs are ['overwrite']" is a well known bug. Tracked at https://github.com/actions/runner/issues/514. I cannot get rid of it.

.github/workflows/cmake_build.yml
```diff
@@ -0,0 +1,259 @@
...
+        run: sudo apt-get install elfutils libzstd-dev
+  
+      - name: "Setup common ENV"
+        uses: allenevans/set-env@v1.1.0
+        with:
+          CC: ${{ matrix.compiler.CC }}
+          CXX: ${{ matrix.compiler.CXX }}
+          VERBOSE: 1
+          ENABLE_CACHE_CLEANUP_TESTS: true
+          CTEST_OUTPUT_ON_FAILURE: ON
```
jrosdahl, line 39: Would it be possible to set the environment variables on the job level using the [common GitHub actions way](https://docs.github.com/en/actions/configuring-and-managing-workflows/using-environment-variables#about-environment-variables), thereby avoiding using allenevans/set-env and the resulting "Unexpected input(s)" warnings?

.github/workflows/cmake_build.yml
```diff
@@ -0,0 +1,259 @@
...
+        run: zip -r testdirs.zip ${{ matrix.config.BUILDDIR }}/testdir.* -x testdir.failed
+
+      - name: Upload testdirs from failed tests
+        if: failure()
+        uses: actions/upload-artifact@v2
+        with:
+          name: ${{ matrix.config.name }} - testdirs.zip
+          path: testdirs.zip
+
+      # Does it make sense to do this on every matrix element?
```
jrosdahl, line 56: I don't think so.

.github/workflows/cmake_build.yml
```diff
@@ -0,0 +1,259 @@
...
+      - name: "Build binary and verify it"
+        run: ci/build-and-verify-package
+        # ToDo: upload testdirs in case of errors
+
+      - name: Upload binary
+        uses: actions/upload-artifact@v2
+        with:
+          # Does it make sense to upload all of them?
+          # What is even the use case? People using specific versions without compiling on their own.
```
jrosdahl, line 65: I don't think that it makes sense to upload any Linux binaries, actually, unless they are statically linked (except possibly to libc). And if we do they should be a release build.

.github/workflows/cmake_build.yml
```diff
@@ -0,0 +1,259 @@
...
+          name: ${{ matrix.os }}-${{ matrix.compiler.CXX }}-ccache.tar.gz
+          path: build_package_dir_test/ccache-binary.tar.gz
+
+      # Does it make sense to do this on every matrix element?
```
jrosdahl, line 69: No.

.github/workflows/cmake_build.yml
```diff
@@ -0,0 +1,259 @@
...
+      - name: "Create source package and verify it"
+        run: ci/build-and-verify-package-source
+        # ToDo: upload testdirs in case of errors
+
+      - name: Upload source zip
+        # Uploading once is enough
+        if: matrix.os == 'ubuntu-20.04' && matrix.CC == 'gcc'
```
jrosdahl, line 76: This doesn't seem to trigger since there is no such artifact. I guess you meant `matrix.compiler.CC == 'gcc'`? On the other hand I don't see why we would want to publish this artifact.

.github/workflows/cmake_build.yml
```diff
@@ -0,0 +1,259 @@
...
+        uses: actions/upload-artifact@v2
+        with:
+          name: ${{ matrix.os }}-${{ matrix.compiler.CXX }}-ccache-src.tar.gz
+          path: build_package_source_dir_test/ccache-src.tar.gz
+
+  # These mimic the old Travis tests as far as not replaced by standard_tests:
```
jrosdahl, line 82: The "Build with ENABLE_TRACING" job is missing.

.github/workflows/cmake_build.yml
```diff
@@ -0,0 +1,259 @@
...
+  specific_tests:
+    name: ${{ matrix.config.name }}
+    runs-on: ${{ matrix.config.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        config:
+          - {
+            # Job 1: Build (in source directory) on Linux with native GCC in release build mode
```
jrosdahl, line 91: I added job numbers in the .travis.yml comments to make it easier to correlate the jobs to the Travis-CI jobs, but since GitHub Actions jobs are unnumbered we should just remove them since they don't serve a purpose.

.github/workflows/cmake_build.yml
```diff
@@ -0,0 +1,259 @@
...
+
+  # These mimic the old Travis tests as far as not replaced by standard_tests:
+  specific_tests:
+    name: ${{ matrix.config.name }}
+    runs-on: ${{ matrix.config.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        config:
+          - {
```
jrosdahl, line 90: Are the curly braces necessary or could they be removed?

.github/workflows/cmake_build.yml
AlexanderLanin: Yes, but the trouble is one would need to set them at every single step. Maybe that's not an issue if we end up with 1-2 steps. But at the moment I have those build + build binary + build source steps which all need the same env variables. I'll revisit after fixing the rest.

.github/workflows/cmake_build.yml
AlexanderLanin: They are statically linked, but not release. I'll remove it for now.

.github/workflows/cmake_build.yml
AlexanderLanin: The idea was to have a source zip where the Version.cpp is already created. But again I'm not sure of the use case. I'll remove it for now.

.github/workflows/cmake_build.yml
AlexanderLanin: I guess it's new. I've added it.

AlexanderLanin: Yet another idea: how about merging that with release build (Job 1) and uploading the binary. Sounds like then one could say "try that binary and tell me what the traces look like".

.github/workflows/cmake_build.yml
AlexanderLanin: The purpose is rather temporary indeed. I wanted to have them easily comparable to travis.
(Review, keeping in sync until full switch, etc)

.github/workflows/cmake_build.yml
AlexanderLanin: Necessary. Without them:

```
Invalid workflow file
You have an error in your yaml syntax

AlexanderLanin:
Findings addressed, here is what it looks like now: https://github.com/AlexanderLanin/ccache/actions/runs/204643036

.github/workflows/cmake_build.yml
jrosdahl: Hm, not sure I understand. It's possible to set environment variables on the **job** level so they don't have to be set for every single **step**. Like this: https://github.com/jrosdahl/ccache/commit/fc43c6fef32d6a47ed27920765983193f73a164b. CI run: https://github.com/jrosdahl/ccache/runs/964268833. Am I missing something?

.github/workflows/cmake_build.yml
jrosdahl: The optional tracing is something a ccache *developer* would use when investigating performance, so I don't think there is a real use case of refering a user to a prebuilt binary for tracing purposes. (I'm actually not sure if anybody uses the optional tracing (I'm not), so maybe we should just remove it.)

That said, combining builds is fine.

.github/workflows/cmake_build.yml
jrosdahl: This seems to work fine: https://github.com/jrosdahl/ccache/commit/b6f7b62c79ce34a5210a43beb6f12c1235b31b81.

AlexanderLanin: Ok, I guess it has to be the same for all jobs. I tried modifying one job only.

.github/workflows/cmake_build.yml
AlexanderLanin: For some reason I was under the impression it's not possible. In current setup setting those on Step level seems fine. Maybe in future cleanups / improvements it will make sense to move them.

Issue #539, opened 2020-02-12
PCH not detected using Clang and CMake 3.16


cristianadam:
### How to reproduce ###

CMake 3.16 has gotten support for PCHs. I've build a "Hello World" application and build it with MinGW-GCC 8.1 (gcc) and LLVM-MinGW 9.0 (clang).

### CMakeLists.txt
```cmake
cmake_minimum_required(VERSION 3.16)

project(main)

add_executable(main main.cpp)
target_precompile_headers(main PRIVATE <iostream>)

install(TARGETS main)

enable_testing()
add_test(NAME main COMMAND main)
```

### main.cpp
```cpp
#include <iostream>

int main()
{
  std::cout << "Hello world\n";
}
```

### test.cmake 
```cmake
set(ENV{CCACHE_DEBUG} true)
set(ENV{CCACHE_LOGFILE} ${CMAKE_CURRENT_LIST_DIR}/ccache.log)
set(ENV{CCACHE_DIR} ${CMAKE_CURRENT_LIST_DIR}/.ccache)
set(ENV{CCACHE_SLOPPINESS} pch_defines,time_macros)

foreach(step 1 2)
    execute_process(COMMAND cmake -E remove_directory build)

    execute_process(COMMAND cmake
    -G Ninja
    -D CMAKE_BUILD_TYPE=Release
    -D CMAKE_CXX_COMPILER_LAUNCHER=ccache
    -S .
    -B build)

    execute_process(COMMAND ccache -z)

    execute_process(COMMAND cmake --build build)
endforeach()

execute_process(COMMAND ccache -s) 
```

I've build the application running `cmake -P test.cmake`.

### Actual behavior ###

For gcc at the end I've got in the statistics:
```
cache hit (direct)                     0
cache hit (preprocessed)               1
cache miss                             1
cache hit rate                     50.00 %
cleanups performed                     0
files in cache                         2
```

In the [ccache.log](https://gist.github.com/cristianadam/6f7e8bb239c24e306181b8f87ff1a867) file I could see entries like:
```
Detected use of precompiled header:
```

For clang I've got:
```
cache hit (direct)                     2
cache hit (preprocessed)               0
cache miss                             0
cache hit rate                    100.00 %
cleanups performed                     0
files in cache                         4
```
In the [ccache.log](https://gist.github.com/cristianadam/91d38973963ec5fe01bb64bfef62deaf) I couldn't see any pch log entries. 

### Expected behavior ###

[...]

PR #624, opened 2020-07-10
Handle clang PCHs (#539)

ahasselbring:
Closes: #539

### DIFF
--- a/src/ArgsInfo.hpp
+++ b/src/ArgsInfo.hpp
@@ -58,11 +58,8 @@ struct ArgsInfo
   // Is the compiler being asked to output dependencies?
   bool generating_dependencies = false;
 
-  // Option MD/MMD
+  // Seen -MD or -MMD?
   bool seen_MD_MMD = false;
-
-  // Change target of dependency file
-  bool change_dep_file = false;
 
   // Is the dependency makefile target name specified with -MT or -MQ?
   bool dependency_target_specified = false;
--- a/src/ArgsInfo.hpp
+++ b/src/ArgsInfo.hpp
@@ -100,6 +100,10 @@ struct ArgsInfo
   // Clang's -include-pch or -include-pth).
   bool using_precompiled_header = false;
 
+  // Whether Clang is instructed not to include timestamps in the precompiled
+  // header it generates.
+  bool fno_pch_timestamp = false;
+
   // Files referenced by -fsanitize-blacklist options.
   std::vector<std::string> sanitize_blacklists;
 
--- a/src/ResultRetriever.cpp
+++ b/src/ResultRetriever.cpp
@@ -26,7 +26,8 @@
 using Result::FileType;
 using string_view = nonstd::string_view;
 
-ResultRetriever::ResultRetriever(Context& ctx) : m_ctx(ctx)
+ResultRetriever::ResultRetriever(Context& ctx, bool rewrite_dependency_target)
+  : m_ctx(ctx), m_rewrite_dependency_target(rewrite_dependency_target)
 {
 }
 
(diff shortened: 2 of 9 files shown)


ahasselbring:
Thanks for your comments. While writing tests, I found out that things are a bit more complicated:

1. If the call (similar to the one [CMake uses for PCH generation](https://gitlab.kitware.com/cmake/cmake/-/blob/master/Modules/Compiler/Clang.cmake#L105))
```
ccache clang -Xclang -emit-pch -Xclang -include -Xclang pch.h -o pch.h.pch -c empty.c
```
is executed twice, the second call of ccache will find (through [lines 109-115 of `argprocessing.cpp`](https://github.com/ccache/ccache/blob/master/src/argprocessing.cpp#L109) that `pch.h.pch` exists, and therefore assume that it is included as PCH.
This in turn means that ccache will always refuse to use a cached version of the PCH (besides possible other reasons) because the spuriously included file `pch.h.pch` is newer than the cached one (in fact, they should have the same timestamp as `pch.h.pch` was written and then cached immediately after). clang, however, doesn't actually use `pch.h.pch` so it would be safe for ccache to ignore it. I am not yet sure how clang knows that it can/should ignore the precompiled version of `pch.h`, so I am not sure how ccache can find this out.

2. The preprocessor command that ccache makes out of the above command line looks like
```
ccache clang -Xclang -emit-pch -Xclang -include -Xclang pch.h (-fpch-preprocess if pch.h.pch already exists) -E empty.c
```
which still outputs the binary PCH (obviously due to the `-emit-pch` option) and not just the preprocessed source. As far as I understand, ccache derives the files that a call depends on (and therefore when a cached result is out of date) from the output of the preprocessor. Does this mean that ccache will not be able to find out which headers are included (and thus will not know when it needs to call the compiler)?

I derive the following tasks:
1. Find out how ccache can determine that `clang -Xclang -emit-pch -Xclang -include -Xclang pch.h -o pch.h.pch -c empty.c` does not include `pch.h.pch` if that already exists (which means finding out what clang does in this case - maybe it's just the `-emit-pch` option that tells clang not to look for included PCHs, but since clang has chained PCHs, that seems not correct).
2. Eliminate `-Xclang -emit-pch` from the preprocessor call (by making `-emit-pch` a compiler only arg? what about `-emit-pth`? Do `-emit-pch` and `-emit-pth` even exist as anything else than clang-cc1 arguments?), thereby making the preprocessor output look normal and enabling ccache to derive included files properly.
Do you agree with this?

jrosdahl:
> 1. The preprocessor command that ccache makes out of the above command line looks like
> 
> ```
> ccache clang -Xclang -emit-pch -Xclang -include -Xclang pch.h (-fpch-preprocess if pch.h.pch already exists) -E empty.c
> ```
> 
> which still outputs the binary PCH (obviously due to the `-emit-pch` option) and not just the preprocessed source. As far as I understand, ccache derives the files that a call depends on (and therefore when a cached result is out of date) from the output of the preprocessor. Does this mean that ccache will not be able to find out which headers are included (and thus will not know when it needs to call the compiler)?

ccache uses the preprocessed output for two things:

1. Computing (along with other information) a hash that uniquely identifies the result.
2. Remembering which header files were read when compiling. This information is used for future potential ["direct mode"](https://ccache.dev/manual/3.7.10.html#_the_direct_mode) hits. 

For this phase ccache shouldn't have to know the output file (the emitted pch file), so excluding `-emit-pch` from the preprocessor command sounds good. (Not sure if I understood your question so maybe this wasn't the answer you sought?)

> I derive the following tasks:
> 
> 1. Find out how ccache can determine that `clang -Xclang -emit-pch -Xclang -include -Xclang pch.h -o pch.h.pch -c empty.c` does not include `pch.h.pch` if that already exists (which means finding out what clang does in this case - ~maybe it's just the `-emit-pch` option that tells clang not to look for included PCHs, but since clang has chained PCHs, that seems not correct~ apparently, the cc1 option `-include` (in contrast to the frontend option `-include`) never looks for PCHs, so that's an easy fix).

OK.

> 2. Eliminate `-Xclang -emit-pch` from the preprocessor call (by making `-emit-pch` a compiler only arg? what about `-emit-pth`? Do `-emit-pch` and `-emit-pth` even exist as anything else than clang-cc1 arguments?),

No idea. I always have to look up stuff related to PCHs since I don't use them myselves.

> thereby making the preprocessor output look normal and enabling ccache to derive included files properly.
>    Do you agree with this?

Yes.

ahasselbring:
The last commits already integrated the proposed changes, and as far as I can tell everything works as expected now.

jrosdahl:
Thanks.

PR #632, opened 2020-08-09
Support building on MSVC.

rkitover:
With these changes, the project builds with Visual Studio 2019, unit
tests pass and it works correctly with mingw gcc.

NOTE: The very latest version of Visual Studio 2019 is required, because
there was just a necessary fix for template arguments.

Tested building and running unit tests on Windows+MSVC, Windows+MinGW,
Linux and macOS.

- Enable `ZSTD_FROM_INTERNET` by default for MSVC when not using vcpkg
  or conan.

- Add include tests for some standard UNIX headers not available on
  MSVC.

- Add necessary MSVC compiler flags.

- In `Args::from_gcc_atfile()` iterate over the string via `c_str()`
  instead of `cbegin()`, the MSVC string character iterator does not
  include the ending null byte.

- Misc. minor cmake fix-ups.

- Add some headers that are not implicitly included from other headers
  like `<algorithm>`, `<ios>`, `<cstdint>` and `<cstdarg>` in some
  places, gcc does this but MSVC does not.

- Add `std::filesystem` version of `Util::traverse()` when dirent.h is
  not available, which is preferred for performance reasons.

- Add implementations of the following functions that are not available
  in MSVC in Win32Util.cpp: `gettimeofday()`, `localtime_r()`,
  `asprintf()`.

- Add Windows implementation of `getopt_long()` from
  https://www.codeproject.com/Articles/157001/Full-getopt-Port-for-Unicode-and-Multibyte-Microso
  to third_party/win32.

- Add some compatibility typedefs, constants and macros to the `_WIN32`
  section of system.hpp, as well as the prototypes for the functions
  added to Win32Util.cpp.

- Fix up unit tests expecting '/' separated paths to expect paths
  delimited by `DIR_DELIM_CH`.

- Invoke test/run with bash from cmake, necessary on msys2+mingw64, many
  fail, there is more work to do here.

- Set the warning level to `/W4` and silence all the uninteresting
  warning types. Compiles with no warnings now.

- Switch to using standard C++ attributes `[[nodiscard]]` and
  `[[maybe_unused]]` and define macros for gcc for their equivalents.

- `#define DOCTEST_CONFIG_USE_STD_HEADERS` for MSVC only, because it
  requires explicitly including `<ostream>`.

- Add vim files to .gitignore.

### DIFF
--- a/cmake/StandardWarnings.cmake
+++ b/cmake/StandardWarnings.cmake
@@ -137,4 +137,24 @@ elseif(CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
     add_target_compile_flag_if_supported(
       standard_warnings "-Wno-unused-variable")
   endif()
+elseif(MSVC)
+  # Remove any warning level flags added by cmake.
+  string(REGEX REPLACE "/W[0-4]" "" CMAKE_C_FLAGS   "${CMAKE_C_FLAGS}")
+  string(REGEX REPLACE "/W[0-4]" "" CMAKE_ASM_FLAGS "${CMAKE_ASM_FLAGS}")
+  string(REGEX REPLACE "/W[0-4]" "" CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS}")
+
+  target_compile_options(
+    standard_warnings
+    INTERFACE
+    /W4
+    # Ignore bad macro in winbase.h triggered by /Zc:preprocessor
+    /wd5105
+    # Conversion warnings.
+    /wd4244
+    /wd4267
+    # Assignment in conditional.
+    /wd4706
+    # Non-underscore-prefixed POSIX functions.
+    /wd4996
+  )
 endif()
--- a/src/Util.cpp
+++ b/src/Util.cpp
@@ -28,6 +28,10 @@
 #include <algorithm>
 #include <fstream>
 
+#ifndef HAVE_DIRENT_H
+#  include <filesystem>
+#endif
+
 #ifdef HAVE_PWD_H
 #  include <pwd.h>
 #endif
@@ -477,7 +481,7 @@ fallocate(int fd, long new_size)
   int err = 0;
   try {
     write_fd(fd, buf, bytes_to_write);
-  } catch (Error& e) {
+  } catch ([[maybe_unused]] Error& e) {
     err = errno;
   }
   lseek(fd, saved_pos, SEEK_SET);
@@ -780,7 +784,7 @@ is_nfs_fd(int fd, bool* is_nfs)
 }
 #else
 int
-is_nfs_fd([[gnu::unused]] int fd, [[gnu::unused]] bool* is_nfs)
+is_nfs_fd([[maybe_unused]] int fd, [[maybe_unused]] bool* is_nfs)
 {
   return -1;
 }
@@ -957,7 +961,7 @@ int
 parse_int(const std::string& value)
 {
   size_t end;
-  long result;
+  long result = 0;
   bool failed = false;
   try {
     result = std::stoi(value, &end, 10);
@@ -1015,7 +1019,7 @@ uint32_t
 parse_uint32(const std::string& value)
 {
   size_t end;
-  long long result;
+  long long result = 0;
   bool failed = false;
   try {
     result = std::stoll(value, &end, 10);
@@ -1248,7 +1252,7 @@ setenv(const std::string& name, const std::string& value)
 #else
   char* string;
   asprintf(&string, "%s=%s", name.c_str(), value.c_str());
-  putenv(string);               // Leak to environment.
+  putenv(string); // Leak to environment.
 #endif
 }
 
@@ -1304,6 +1308,8 @@ to_lowercase(string_view string)
   return result;
 }
 
+#ifdef HAVE_DIRENT_H
+
 void
 traverse(const std::string& path, const TraverseVisitor& visitor)
 {
@@ -1318,11 +1324,11 @@ traverse(const std::string& path, const TraverseVisitor& visitor)
 
       std::string entry_path = path + "/" + entry->d_name;
       bool is_dir;
-#ifdef _DIRENT_HAVE_D_TYPE
+#  ifdef _DIRENT_HAVE_D_TYPE
       if (entry->d_type != DT_UNKNOWN) {
         is_dir = entry->d_type == DT_DIR;
       } else
-#endif
+#  endif
       {
         auto stat = Stat::lstat(entry_path);
         if (!stat) {
@@ -1349,6 +1355,31 @@ traverse(const std::string& path, const TraverseVisitor& visitor)
     throw Error("failed to open directory {}: {}", path, strerror(errno));
   }
 }
+
+#else // If not available, use the C++17 std::filesystem implementation.
+
+void
+traverse(const std::string& path, const TraverseVisitor& visitor)
+{
+  if (std::filesystem::is_directory(path)) {
+    for (auto&& p : std::filesystem::directory_iterator(path)) {
+      std::string entry = p.path().string();
+
+      if (p.is_directory()) {
+        traverse(entry, visitor);
+      } else {
+        visitor(entry, false);
+      }
+    }
+    visitor(path, true);
+  } else if (std::filesystem::exists(path)) {
+    visitor(path, false);
+  } else {
+    throw Error("failed to open directory {}: {}", path, strerror(errno));
+  }
+}
+
+#endif
 
 bool
 unlink_safe(const std::string& path, UnlinkLog unlink_log)
@@ -1463,6 +1494,10 @@ write_file(const std::string& path,
            const std::string& data,
            std::ios_base::openmode open_mode)
 {
+  if (path.empty()) {
+    throw Error("No such file or directory");
+  }
+
   open_mode |= std::ios::out;
   std::ofstream file(path, open_mode);
   if (!file) {
--- a/src/Util.hpp
+++ b/src/Util.hpp
@@ -26,6 +26,7 @@
 #include "third_party/nonstd/string_view.hpp"
 
 #include <functional>
+#include <ios>
 #include <memory>
 #include <string>
 #include <utility>
@@ -125,8 +126,7 @@ ends_with(nonstd::string_view string, nonstd::string_view suffix)
 
 // Expand all instances of $VAR or ${VAR}, where VAR is an environment variable,
 // in `str`. Throws `Error` if one of the environment variables.
-[[gnu::warn_unused_result]] std::string
-expand_environment_variables(const std::string& str);
+[[nodiscard]] std::string expand_environment_variables(const std::string& str);
 
 // Extends file size to at least new_size by calling posix_fallocate() if
 // supported, otherwise by writing zeros last to the file.
@@ -406,16 +406,13 @@ starts_with(nonstd::string_view string, nonstd::string_view prefix)
 }
 
 // Returns a copy of string with the specified ANSI CSI sequences removed.
-[[gnu::warn_unused_result]] std::string
-strip_ansi_csi_seqs(nonstd::string_view string);
+[[nodiscard]] std::string strip_ansi_csi_seqs(nonstd::string_view string);
 
 // Strip whitespace from left and right side of a string.
-[[gnu::warn_unused_result]] std::string
-strip_whitespace(const std::string& string);
+[[nodiscard]] std::string strip_whitespace(const std::string& string);
 
 // Convert a string to lowercase.
-[[gnu::warn_unused_result]] std::string
-to_lowercase(nonstd::string_view string);
+[[nodiscard]] std::string to_lowercase(nonstd::string_view string);
 
 // Traverse `path` recursively (postorder, i.e. files are visited before their
 // parent directory).
--- a/src/Win32Util.cpp
+++ b/src/Win32Util.cpp
@@ -20,6 +20,9 @@
 
 #include "Util.hpp"
 
+#include <chrono>
+#include <thread>
+
 namespace Win32Util {
 
 std::string
@@ -91,3 +94,75 @@ argv_to_string(const char* const* argv, const std::string& prefix)
 }
 
 } // namespace Win32Util
+
+// From: https://stackoverflow.com/a/58162122/262458
+#ifdef _MSC_VER
+int
+gettimeofday(struct timeval* tp, [[maybe_unused]] struct timezone* tzp)
+{
+  namespace sc = std::chrono;
+  sc::system_clock::duration d = sc::system_clock::now().time_since_epoch();
+  sc::seconds s = sc::duration_cast<sc::seconds>(d);
+  tp->tv_sec = static_cast<long>(s.count());
+  tp->tv_usec =
+    static_cast<long>(sc::duration_cast<sc::microseconds>(d - s).count());
+
+  return 0;
+}
+#endif
+
+void
+usleep(int64_t usec)
+{
+  std::this_thread::sleep_for(std::chrono::microseconds(usec));
+}
+
+struct tm*
+localtime_r(time_t* _clock, struct tm* _result)
+{
+  struct tm* p = localtime(_clock);
+
+  if (p)
+    *(_result) = *p;
+
+  return p;
+}
+
+// From: https://stackoverflow.com/a/40160038/262458
+#ifdef _MSC_VER
+int
+vasprintf(char** strp, const char* fmt, va_list ap)
+{
+  // _vscprintf tells you how big the buffer needs to be
+  int len = _vscprintf(fmt, ap);
+  if (len == -1) {
+    return -1;
+  }
+  size_t size = (size_t)len + 1;
+  char* str = static_cast<char*>(malloc(size));
+  if (!str) {
+    return -1;
+  }
+  // vsprintf_s is the "secure" version of vsprintf
+  int r = vsprintf_s(str, len + 1, fmt, ap);
+  if (r == -1) {
+    free(str);
+    return -1;
+  }
+  *strp = str;
+  return r;
+}
+#endif
+
+// Also from: https://stackoverflow.com/a/40160038/262458
+#ifdef _MSC_VER
+int
+asprintf(char** strp, const char* fmt, ...)
+{
+  va_list ap;
+  va_start(ap, fmt);
+  int r = vasprintf(strp, fmt, ap);
+  va_end(ap);
+  return r;
+}
+#endif
--- a/src/ccache.cpp
+++ b/src/ccache.cpp
@@ -56,6 +56,8 @@
 
 #ifdef HAVE_GETOPT_LONG
 #  include <getopt.h>
+#elif defined(_WIN32)
+#  include "third_party/win32/getopt.h"
 #else
 #  include "third_party/getopt_long.h"
 #endif
--- a/src/system.hpp
+++ b/src/system.hpp
@@ -25,7 +25,10 @@
 
 #include "config.h"
 
-#include <sys/file.h>
+#ifdef HAVE_SYS_FILE_H
+#  include <sys/file.h>
+#endif
+
 #ifdef HAVE_SYS_MMAN_H
 #  include <sys/mman.h>
 #endif
@@ -43,15 +46,35 @@
 #include <csignal>
 #include <cstdarg>
 #include <cstddef>
+#include <cstdint>
 #include <cstdio>
 #include <cstdlib>
 #include <cstring>
 #include <ctime>
-#include <dirent.h>
+
+#ifdef HAVE_DIRENT_H
+#  include <dirent.h>
+#endif
+
 #include <fcntl.h>
-#include <strings.h>
-#include <unistd.h>
-#include <utime.h>
+
+#ifdef HAVE_STRINGS_H
+#  include <strings.h>
+#endif
+
+#ifdef HAVE_UNISTD_H
+#  include <unistd.h>
+#endif
+
+#ifdef HAVE_UTIME_H
+#  include <utime.h>
+#elif defined(HAVE_SYS_UTIME_H)
+#  include <sys/utime.h>
+#endif
+
+#ifdef HAVE_VARARGS_H
+#  include <varargs.h>
+#endif
 
 // AIX/PASE does not properly define usleep within its headers. However, the
 // function is available in libc.a. This extern define ensures that it is
@@ -59,8 +82,6 @@
 #ifdef _AIX
 extern int usleep(useconds_t);
 #endif
-
-extern char** environ;
 
 #define ARRAY_SIZE(array) (sizeof(array) / sizeof((array)[0]))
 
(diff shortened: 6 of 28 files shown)


src/Util.cpp
```diff
@@ -1303,6 +1304,54 @@ traverse(const std::string& path, const TraverseVisitor& visitor)
   } else {
     throw Error("failed to open directory {}: {}", path, strerror(errno));
   }
+#else // _WIN32
```
AlexanderLanin, line 1307: Doesn't MSVC support stuff like std::filesystem::recursive_directory_iterator?

src/Util.hpp
```diff
@@ -117,10 +118,17 @@ bool create_dir(nonstd::string_view dir);
 nonstd::string_view dir_name(nonstd::string_view path);
 
 // Return true if `suffix` is a suffix of `string`.
+// From: https://stackoverflow.com/a/874160/262458
 inline bool
 ends_with(nonstd::string_view string, nonstd::string_view suffix)
 {
-  return string.ends_with(suffix);
```
AlexanderLanin, line 123: Latest MSVC does not support https://en.cppreference.com/w/cpp/string/basic_string/ends_with in its c++=latest setting?

src/Win32Util.cpp
```diff
@@ -91,3 +91,86 @@ argv_to_string(const char* const* argv, const std::string& prefix)
 }
 
 } // namespace Win32Util
+
+// From: https://stackoverflow.com/a/26085827/262458
```
AlexanderLanin, line 95: Why not the shorter more modern looking version a few posts down?
https://stackoverflow.com/a/58162122/12271052

src/Win32Util.cpp
```diff
@@ -91,3 +91,86 @@ argv_to_string(const char* const* argv, const std::string& prefix)
...
+  time = ((uint64_t)file_time.dwLowDateTime);
+  time += ((uint64_t)file_time.dwHighDateTime) << 32;
+
+  tp->tv_sec = (long)((time - EPOCH) / 10000000L);
+  tp->tv_usec = (long)(system_time.wMilliseconds * 1000);
+  return 0;
+}
+#endif
+
+// From: https://stackoverflow.com/a/17283549/262458
```
AlexanderLanin, line 120: I don't know how you handle such sources here, but I think the line below doesn't require a source. It's a rather obvious single line of code.

src/Win32Util.cpp
```diff
@@ -91,3 +91,86 @@ argv_to_string(const char* const* argv, const std::string& prefix)
...
+void
+usleep(int64_t usec)
+{
+  std::this_thread::sleep_for(std::chrono::microseconds(usec));
+}
+
+// From: https://bugs.webkit.org/show_bug.cgi?id=5311
```
AlexanderLanin, line 127: Same here in my personal opinion. There is simply no effort behind it that is worth referring to.

src/system.hpp
```diff
@@ -102,3 +181,11 @@ const size_t READ_BUFFER_SIZE = 65536;
 #else
 #  define IS_TRIVIALLY_COPYABLE(T) std::is_trivially_copyable<T>::value
 #endif
+
+#ifdef max
```
AlexanderLanin, line 185: Maybe this is easier? (before the includes, specifically Windows.h / Windef.h)
#define NOMINMAX 1

unittest/test_Util.cpp
```diff
@@ -658,8 +668,12 @@ TEST_CASE("Util::read_file and Util::write_file")
   CHECK_THROWS_WITH(Util::read_file("does/not/exist"),
                     "No such file or directory");
 
+#ifdef _WIN32
+  CHECK_THROWS_WITH(Util::write_file("", "does/not/exist"), "Invalid argument");
```
AlexanderLanin, line 672: Can the error message be fixed? It's rather unmatching to the error case.

rkitover:
@AlexanderLanin thank you for taking a look, I will make these fixes shortly.

unittest/test_Util.cpp
rkitover: Apparently the test was writing to a file named empty string, when writing to an actual non-existent path, the error is what you would expect. I can't even blame Windows for this, and I blame Windows for a lot of things.

I added a check for empty path, and added a test for a non-empty non-existent path.

rkitover:
@AlexanderLanin everything is taken care of, and I made a couple of other minor improvements like getting the `__cplusplus` macro to work.

rkitover:
@jrosdahl @AlexanderLanin 

So the next step I am planning here, for a followup PR, is fixing the bash tests for msys2+mingw64 and adding CI jobs for msys2 and visual studio, travis supports this.

What do you guys think?

cristianadam:
Wouldn't it be better to wait until GitHub Actions MR is merged?

rkitover:
@cristianadam no problem, I will do that then.

src/Util.cpp
```diff
@@ -1263,6 +1267,21 @@ to_lowercase(string_view string)
 void
 traverse(const std::string& path, const TraverseVisitor& visitor)
 {
+#if __cplusplus >= 201703L
+  if (std::filesystem::is_directory(path)) {
+    for (auto&& p : std::filesystem::directory_iterator(path)) {
```
jrosdahl, line 1272: Using `std::filesystem::directory_iterator` is unfortunately not as efficient as `opendir/readdir` (with `_DIRENT_HAVE_D_TYPE` support) since the `std::filesystem` solution needs to perform a `stat` syscall for each entry to determine its type. Therefore the `std::filesystem` solution needs to be used only if `opendir/readdir` aren't available.

src/Win32Util.hpp
```diff
@@ -20,7 +20,9 @@
 
 #include "system.hpp"
 
+#include <chrono>
 #include <string>
+#include <thread>
```
jrosdahl, line 25: Suggestion: Move `#include <chrono>` and `#include <thread>` to the `.cpp` file instead since they are not required by anything in the header file.

src/Result.hpp
```diff
@@ -22,6 +22,7 @@
 
 #include "third_party/nonstd/optional.hpp"
 
+#include <algorithm>
```
jrosdahl, line 25: Suggestion: Move this to the `.cpp` file instead since it's not required by anything in the header file.

src/ProgressBar.hpp
```diff
@@ -20,6 +20,7 @@
 
 #include "system.hpp"
 
+#include <algorithm>
```
jrosdahl, line 23: Suggestion: Move this to the `.cpp` file instead since it's not required by anything in the header file.

src/ZstdCompressor.hpp
```diff
@@ -23,6 +23,7 @@
 #include "Compressor.hpp"
 #include "NonCopyable.hpp"
 
+#include <algorithm>
```
jrosdahl, line 26: Suggestion: Move this to the `.cpp` file instead since it's not required by anything in the header file.

src/ccache.cpp
```diff
@@ -17,6 +17,16 @@
...
+#include "config.h"
+
+#ifdef HAVE_GETOPT_LONG
+#  include <getopt.h>
+#elif _MSC_VER
+#  include "third_party/win32/getopt.h"
+#else
+#  include "third_party/getopt_long.h"
+#endif
+
```
jrosdahl, line 29: Is there a reason why this needs to be done first in the file instead of at the original location (where `config.h` has already been included)?

src/third_party/win32/getopt.h
```diff
@@ -0,0 +1,136 @@
+/* Getopt for Microsoft C
```
jrosdahl, line 1: Please add a `src/third_party/win32/getopt.*` entry under _Files derived from other sources_  in `LICENSE.adoc`.

unittest/TestUtil.hpp
```diff
@@ -20,6 +20,8 @@
 
 #include "system.hpp"
 
+#include <algorithm>
+#include <ostream>
```
jrosdahl, line 24: Suggestion: Move these to the .cpp file instead since they are not required by anything in the header file.

src/Lockfile.hpp
```diff
@@ -20,6 +20,7 @@
 
 #include "system.hpp"
 
+#include <algorithm>
```
jrosdahl, line 23: Suggestion: Move this to the `.cpp` file instead since it's not required by anything in the header file.

src/ccache.cpp
rkitover: I was getting clashing definitions otherwise, I will see if there's a nicer way to do this.

AlexanderLanin:
I did a quick comparison of opendir vs filesystem as I was surprised to learn that it's slower.
Executed on WSL2 with Ubuntu 18 - no idea what kind of effect that has.
Listing all files in llvm:
* opendir ~105ms
* filesystem ~350ms
* filesystem_recursive ~350ms

Although the recusrive option is like 1000 times more readable I can see that it is in fact slower. Will read up on that some more, but that's off topic.
```
    for (auto&& p : std::filesystem::recursive_directory_iterator(path)) {
      std::string entry = p.path().string();
      visitor(entry, false);
    }
```

src/Util.cpp
AlexanderLanin: I did a quick comparison of opendir vs filesystem as I was surprised to learn that it's slower.
Executed on WSL2 with Ubuntu 18 - no idea what kind of effect that has.
Listing all files in llvm:
* opendir ~105ms
* filesystem ~350ms
* filesystem_recursive ~350ms

Although the recusrive option is like 1000 times more readable I can see that it is in fact slower. Will read up on that some more, but that's off topic.
```
    for (auto&& p : std::filesystem::recursive_directory_iterator(path)) {
      std::string entry = p.path().string();
      visitor(entry, false);
    }
```

src/ccache.cpp
rkitover: Incidentally fixing the other include scopes fixed this problem.

rkitover:
@jrosdahl @AlexanderLanin 

Thank you for supporting this, I realize it adds to the maintenance burden, and I should have discussed this first, but sometimes I just can't stop myself from fixing build issues.

In my experience with adding a Windows CI to our project, most issues are minor and can be fixed without even using the environment, once things are up and running

And if there are other issues I'm sure you have many folks who will be happy to help, including me if I'm available.

I fixed the issues you listed, and also fixed the remaining clang-format issues and the warning spew from the visual studio compiler. There is now only a small page of warnings, here is a screenshot of my build:

build screenshot

cmake/StandardWarnings.cmake
```diff
@@ -137,4 +137,13 @@ elseif(CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
     add_target_compile_flag_if_supported(
       standard_warnings "-Wno-unused-variable")
   endif()
+elseif(MSVC)
```
AlexanderLanin, line 140: for followup commits consider e.g. W4

src/Util.cpp
```diff
@@ -1346,10 +1352,29 @@ traverse(const std::string& path, const TraverseVisitor& visitor)
...
   } else if (errno == ENOTDIR) {
     visitor(path, false);
+#else // If not available, use C++17 std::filesystem implementation.
+  if (std::filesystem::is_directory(path)) {
+    for (auto&& p : std::filesystem::directory_iterator(path)) {
+      std::string entry = p.path().string();
+
+      if (std::filesystem::is_directory(p)) {
+        traverse(entry, visitor);
+      } else if (std::filesystem::is_regular_file(p)) {
```
AlexanderLanin, line 1362: Why is this not simply else? Is it the same reason you don't use `std::filesystem::recusive_directory_iterator`?

src/ccache.cpp
```diff
@@ -17,7 +17,9 @@
 // this program; if not, write to the Free Software Foundation, Inc., 51
 // Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
 
+// clang-format off
```
AlexanderLanin, line 20: how can a simple include be formatted incorrectly?

rkitover: It gets reordered into the block below, another option is to put a single comment in-between.

src/Util.cpp
rkitover: What if it's not a directory and not a type of file that is a regular file? How should that be handled?

I tried the recursive one, but unit tests were failing because the order was not what it was expecting.

src/Util.cpp
AlexanderLanin: @jrosdahl do you have any background why filesystem is not using `_DIRENT_HAVE_D_TYPE` / `DTTOIF`? Is it just poor implementation?

src/ccache.cpp
rkitover: That issue only seems to affect some versions of `clang-format`, but I changed it to a less ugly fix in any case.

cmake/StandardWarnings.cmake
rkitover: Will do!

src/Util.cpp
rkitover: I've cleaned up the code a bit, do you like this version better?

cmake/StandardWarnings.cmake
rkitover: I added `/W4`, now there are much more warnings, but still less than one screenful.

rkitover: And now I silenced all the warnings, none of them were worth keeping.

rkitover:
@jrosdahl @AlexanderLanin 

I did the last change I wanted to do for this change set, which was adding the standard C++ attributes `[[nodiscard]]` and `[[maybe_unused]]` in place of their gcc equivalents, with compatibility macros for gcc. See:

https://docs.microsoft.com/en-us/cpp/cpp/attributes?view=vs-2019

Also added them in a couple of places to reduce the number of warning classes silenced.

Compiles cleanly with no warnings now.

I will wait for your feedback before I go further with all of this.

src/Util.cpp
jrosdahl: It seems like my initial comment was incorrect; I probably misread the code and didn't take my time to look at the actual filesystem API. On my system (GCC 9.3, ext4), `std::filesystem::(recursive_)directory_iterator` uses the expected and efficient system calls (i.e., not one `stat` per file entry), so now I'm surprised that you see a difference. :slightly_smiling_face:

Anyway, I think the current code is OK.

src/Util.cpp
```diff
@@ -1308,6 +1312,8 @@ to_lowercase(string_view string)
 void
 traverse(const std::string& path, const TraverseVisitor& visitor)
 {
+  // Prefer POSIX opendir()/readdir() implementation for performance reasons.
```
jrosdahl, line 1315: Since the implementations share very little I think that it would make sense to write them as separate implementations, i.e.:

```
#ifdef HAVE_DIRENT_H

void
traverse(const std::string& path, const TraverseVisitor& visitor)
{
  // opendir...
}

#else // If not available, use C++17 std::filesystem implementation.

void
traverse(const std::string& path, const TraverseVisitor& visitor)
{
  // std::filesystem...
}

#endif
```

Then we can also avoid the clang-format workaround.

src/system.hpp
```diff
@@ -102,3 +208,9 @@ const size_t READ_BUFFER_SIZE = 65536;
 #else
 #  define IS_TRIVIALLY_COPYABLE(T) std::is_trivially_copyable<T>::value
 #endif
+
+// GCC version of a couple of standard C++ attributes
```
jrosdahl, line 212: :+1:

unittest/main.cpp
```diff
@@ -23,6 +23,8 @@
 
 #define DOCTEST_THREAD_LOCAL // Avoid MinGW thread_local bug
 #define DOCTEST_CONFIG_IMPLEMENT
+#include <ostream>
```
jrosdahl, line 26: Since the defines above configure the behavior of `third_party/doctest.h` below, I suggest moving the `ostream` inclusion above the defines.

unittest/test_Util.cpp
```diff
@@ -21,8 +21,12 @@
 #include "../src/Util.hpp"
 #include "TestUtil.hpp"
 
+#include <ostream>
+//
```
jrosdahl, line 25: Why does `ostream` need to be included before `third_party/doctest.h`?

jrosdahl (commented):
I previously wrote "This doesn't look too intrusive", but now I'm not so sure anymore... :slightly_smiling_face:

Could you perhaps give some background of why we are doing this? What actual benefits would we get by making (and keeping) ccache buildable with MSVC? Is the end result better than something built with MinGW-w64? Or is that you want to be able to use Visual Studio for developing ccache itself?

src/Util.cpp
AlexanderLanin: I've just tried with gcc 9.3 on ext4 and I see lots of stat calls :-(
I'll try opening tickets on clang/gcc. Maybe it's WSL related somehow.

rkitover:
> Could you perhaps give some background of why we are doing this? What actual benefits would we get by making (and keeping) ccache buildable with MSVC? Is the end result better than something built with MinGW-w64? Or is that you want to be able to use Visual Studio for developing ccache itself?

Personally, when I'm in Windows, I do C++ development with the chocolatey + Windows terminal + powershell core + Visual Studio + cmake + ninja stack. I can use terminal vim and various Linux utilities, for other things there are windows utilities or powershell idioms. I actually only use the Visual Studio IDE for the debugger generally, I do my editing in the terminal vim.

In some ways this setup is still missing some conveniences you get in Linux, like terminal mouse support, but overall it is already pretty good.

It took me a few hours to be able to use this setup to do work, and a few weeks to become much more proficient and start doing things like writing my own powershell tools.

I think this is the future for Windows dev, as opposed to using something like Msys2, which is still useful programmatically for some things, for example vcpkg uses it to run some configure scripts.

I was planning to make some improvements on the Msys2 side as well, like getting the bash tests to work.

I started working on this PR after I installed ccache from chocolatey and noticed that it was not actually caching the results of the MSVC build of my project, and had another issue with calling windres correctly. Problems I will look at if I continue with all of this.

This would make it easier for other Windows devs contribute to ccache of course. And there may be various other benefits of closer integration with the Windows ecosystem, I could for example see vcpkg running port builds through ccache as an option.

I will answer your review comments now.

rkitover:
> I don't understand this either. Which clang-format versions? At least versions 6-10 handle it correctly. And what's the less ugly fix?

For some reason the Msys2 version of `clang-format` was ignoring the blank line between includes and reordering them anyway, this is not happening anymore, it must have gotten updated or something, I will remove these things.

unittest/CMakeLists.txt
```diff
@@ -34,4 +34,8 @@ target_link_libraries(
 
 target_include_directories(unittest PRIVATE ${CMAKE_BINARY_DIR} . ../src)
 
+if(MSVC)
+  target_compile_options(unittest PRIVATE /wd4127 /wd4146)
```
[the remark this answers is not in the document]
rkitover: I removed this.

The two warnings are:

```
..\unittest\test_Util.cpp(578): warning C4127: conditional expression is constant
..\unittest\test_Util.cpp(578): note: consider using 'if constexpr' statement instead
..\unittest\test_Util.cpp(584): warning C4127: conditional expression is constant
..\unittest\test_Util.cpp(584): note: consider using 'if constexpr' statement instead
```

This is for the lines such as `if (sizeof(int) == 4) {`.

I fixed this properly by setting a `SIZEOF_INT` `config.h` macro from cmake.

```
..\unittest\test_Util.cpp(585): warning C4146: unary minus operator applied to unsigned type, result still unsigned
```

This is apparently an MSVC bug, it's caused by the constant `-2147483648`. I changed the constant to `-2147483648LL` which silences the warning. I will make a Visual Studio problem report for this.

src/Util.cpp
AlexanderLanin: Found the problem!
1) there is a "bug" in this submission. Will provide suggestion in a second.
2) I measured the speed of recursive_directory_iterator incorrectly. I had a typo and was calling the directory_iterator version while trying to measure recursive_directory_iterator.

Now listing all llvm files (those times do not jitter at all):
opendir ~130ms
directory_iterator as submitted here ~670ms
directory_iterator fixed ~160ms
recursive_directory_iterator ~140ms

P.S. it's not comparable to the numbers above since I checked out a different commit on llvm, have different build directories etc.

unittest/test_Util.cpp
rkitover: I made a cleaner fix for this by defining `DOCTEST_CONFIG_USE_STD_HEADERS` in `TestUtil.hpp`.

src/Util.cpp
AlexanderLanin: Nevermind it's already fixed :) :(

The trick was to not use `std::filesystem::is_directory(path)` and `std::filesystem::is_regular_file(path)` as that info is already available within `p`. `p.is_directory()` and `p.is_regular_file()` return without a stat call.

unittest/CMakeLists.txt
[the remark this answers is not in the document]
AlexanderLanin: It's a known "bug", see https://docs.microsoft.com/de-de/cpp/error-messages/compiler-warnings/compiler-warning-level-2-c4146?view=vs-2019

unittest/test_Util.cpp
AlexanderLanin: Why? I know it's cleaner and more standard compliant to use DOCTEST_CONFIG_USE_STD_HEADERS . But doctest has it off by default intentionally since it's faster to not include but forward declare stuff. And doctest runs tests on all MSVC versions.

And even when I add that define I do not need ostream (on clang and gcc). Why does MSVC need ostream?

unittest/CMakeLists.txt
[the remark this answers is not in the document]
rkitover: Ah thank you, I could change the test to use `INT_MIN` if you don't like the 64 bit comparison.

unittest/test_Util.cpp
rkitover: I timed 3 runs of building the tests on linux with gcc 9.3 without and with the option, on average it increased the real time by `0.38` seconds and the user time by `0.79` seconds, so there is a difference.

I suspect the reason MSVC needs ostream is the same as why I had to add some other standard headers like algorithm, gcc implicitly includes or defines some things from other headers when you include one and MSVC does not.

Without the option I get a big page full of errors, the first of which is:

```none
C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\VC\Tools\MSVC\14.27.29110\include\xstring(426): error C2027: use of undefined type 'std::basic_ostream<char,std::char_traits<char>>'
```

Trying the latest version of doctest did not fix the problem either.

I changed this to set the option **only** for MSVC.

cristianadam: Shouldn't [<iofwd>](https://en.cppreference.com/w/cpp/header/iosfwd) be used instead of `ofstream`?  

Forward declarations shouldn't penalize that much as definitions.

rkitover: @cristianadam I tried this, but it seems MSVC requires `<ostream>` specifically.

cristianadam: Sorry, now I see it was `test_Util.cpp` and not a header file. 

Being in a test file it shouldn't matter that much, right? One can have a precompiled header and it shouldn't be a problem.

rkitover: The issue with headers and doctest is increasing the compile time of test sources, which can be bothersome when you're doing repeated edit/compile/run cycles to run unit tests. I measured the difference for this particular configuration option, and it is fairly minimal, but regardless I changed the option to only be enabled for MSVC.

src/Util.cpp
rkitover: This is done, and it did in fact fix the clang-format issue. This was the last issue in your list.

jrosdahl:
> [...] This would make it easier for other Windows devs contribute to ccache of course.

Thanks for describing your motives. I'm willing to try out supporting MSVC even though it will make it less convenient for non-Windows developers due to increased complexity and test space. It should work if I get help. (And if it becomes too inconvenient the support will risk being dropped.)

> I was planning to make some improvements on the Msys2 side as well, like getting the bash tests to work.

That sounds OK if it won't be too intrusive and hard to maintain.

> I installed ccache from chocolatey and noticed that it was not actually caching the results of the MSVC build of my project

Do I interpret you correctly that you expected ccache to cache MSVC compilations? If so maybe the description of the Chocolatey package needs to be improved? Or why did you expect MSVC caching to work?

src/system.hpp
```diff
@@ -76,17 +97,102 @@ const size_t READ_BUFFER_SIZE = 65536;
...
+#  define mkdir(a, b) _mkdir(a)
 #  define link(src, dst) (CreateHardLink(dst, src, nullptr) ? 0 : -1)
 #  define execv(a, b) win32execute(a, b, 0, -1, -1)
+#  define strncasecmp _strnicmp
+#  define strcasecmp _stricmp
+
+// Visual Studio has deprecated the non-underscore-prefixed names of some POSIX
+// functions.
+#  ifndef __GNUC__
+#    define open _open
```
jrosdahl, line 150: This is actually what triggered my previous comment about not being so sure. Is there no better way?

Redefining common symbols like `open`, `close`, `read` and `write` seems very ugly and error prone. For instance, say that we start using a third party library with a `C::open` method. Then it won't work to call `C::open` since the header will declare an `_open` method (which won't be defined in the library).

rkitover: I noticed that while this is true for GCC, the MSVC preprocessor is smart enough to not do that.

Regardless you're right that this is not particularly great, and I will remove this and just suppress this warning instead.

Since the non-underscore versions of these functions are deprecated, there may be a chance they will be dropped entirely by Microsoft at some point, however I don't think that will happen any time soon, like in the next decade or two.

rkitover:
> Do I interpret you correctly that you expected ccache to cache MSVC compilations? If so maybe the description of the Chocolatey package needs to be improved? Or why did you expect MSVC caching to work?

I vaguely remembered reading some blog post about it working with MSVC, or maybe that was a fork, I don't know.

I was mystified when ccache ran the compiler commands through correctly (except for windres) but my cache remained completely empty.

This is the info of the chocolatey package, it says nothing about not supporting MSVC:

```none
✔ C:\Users\rkitover
rkitover@epycwin > choco info ccache
Chocolatey v0.10.15
ccache 3.7.11 [Approved] Downloads cached for licensed users
 Title: ccache - a fast compiler cache | Published: 7/22/2020
 Package approved as a trusted package on Jul 26 2020 13:46:50.
 Package testing status: Passing on Jul 22 2020 00:37:28.
 Number of Downloads: 952 | Downloads for this version: 420
 Package url
 Chocolatey Package Source: https://github.com/Xav83/chocolatey-packages/tree/develop/automatic/ccache
 Package Checksum: '77WjaEbUpA30hFjeBEYdE/9wiq4k0b0nB59EwdBH+t3PWMFUN7KOVheCObH2nPBN8wmjpybnS3kvtSB3AqX3xg==' (SHA512)
 Tags: ccache C++ cache compiler
 Software Site: https://ccache.dev/
 Software License: https://ccache.dev/license.html
 Software Source: https://github.com/ccache/ccache
 Documentation: https://ccache.dev/manual/latest.html
 Issues: https://github.com/ccache/ccache/issues
 Summary: a fast compiler cache.
 Description: Ccache (or "ccache") is a compiler cache. It speeds up recompilation by caching previous compilations and detecting when the same compilation is being done again
 Release Notes: [Package Release Notes](https://ccache.dev/releasenotes.html)

1 packages found.
```

src/system.hpp
rkitover: This is fixed.

jrosdahl:
Thanks.

AlexanderLanin:
@rkitover @jrosdahl FYI: yet another use case. I'm having trouble setting up the debugger (unknown symbols, variables with no values, etc).
Yes, the build type is DEBUG :p

So now I can try running the ccache unittests with Visual Studio instead.

rkitover:
@AlexanderLanin which mingw build are you using? I recommend using the chocolatey mingw packages. I will try this and see if anything needs to be tweaked, and if so will open a PR.

AlexanderLanin:
Sorry I was unclear probably. I have trouble on Linux.
So instead I switched to Visual Studio for running unittests and immediately found my bug :-)

My point was: Thanks for this change. It helps!

PR #635, opened 2020-08-12
Update for Windows compatibility, fix issues with base_dir -> create relative path during hash calculation

Shionigami:
Environment:
  - Windows 10
  - MinGW64

- Fix for Windows
  - fix the issues detected during build
  - Add windows.h when using MinGW
  - Convert the "\" to "/", then we avoid future issues
- Fix BaseDir function:
  - When base_dir option is set, get the relative path to calculate the file hash in cache. 
  
Only tested on Windows.

### DIFF
--- a/src/Config.hpp
+++ b/src/Config.hpp
@@ -398,12 +398,18 @@ inline void
 Config::set_base_dir(const std::string& value)
 {
   m_base_dir = value;
+#ifdef _WIN32
+  std::replace(m_base_dir.begin(), m_base_dir.end(), '\\', '/');
+#endif
 }
 
 inline void
 Config::set_cache_dir(const std::string& value)
 {
   m_cache_dir = value;
+#ifdef _WIN32
+  std::replace(m_cache_dir.begin(), m_cache_dir.end(), '\\', '/');
+#endif
   if (!m_temporary_dir_configured_explicitly) {
     m_temporary_dir = default_temporary_dir(m_cache_dir);
   }
--- a/src/Util.cpp
+++ b/src/Util.cpp
@@ -1100,7 +1100,15 @@ real_path(const std::string& path, bool return_empty_on_error)
     if (!ok) {
       return path;
     }
+#ifdef _WIN32
+    if(starts_with(buffer, "\\\\?\\")) {
+        resolved = buffer + 4; // Strip \\?\ from the file name.
+    } else {
+        resolved = buffer;
+    }
+#else
     resolved = buffer + 4; // Strip \\?\ from the file name.
+#endif
   } else {
     snprintf(buffer, buffer_size, "%s", c_path);
     resolved = buffer;
--- a/src/Win32Util.hpp
+++ b/src/Win32Util.hpp
@@ -18,6 +18,10 @@
 
 #pragma once
 
+#ifdef __MINGW32__
+#include <windows.h>
+#endif
+
 #include "system.hpp"
 
 #include <string>
--- a/src/ccache.cpp
+++ b/src/ccache.cpp
@@ -1511,7 +1511,25 @@ calculate_result_name(Context& ctx,
 
     // All other arguments are included in the hash.
     hash.hash_delimiter("arg");
+
+#ifdef _WIN32    
+    std::string hash_path = args[i];
+    int pos = hash_path.find(":") - 1; //Get the index of the beginning of the path string 
+    if (pos > 0) {
+      std::string prefix = hash_path.substr(0, pos);
+      hash_path = hash_path.substr(pos);
+      if (!ctx.config.base_dir().empty() && Util::is_absolute_path(hash_path)) {
+        hash_path = Util::get_relative_path(ctx.config.base_dir(), hash_path);
+      }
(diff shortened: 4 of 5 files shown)


src/Config.hpp
```diff
@@ -398,12 +398,18 @@ inline void
 Config::set_base_dir(const std::string& value)
 {
   m_base_dir = value;
+#ifdef _WIN32
+  std::replace(m_base_dir.begin(), m_base_dir.end(), '\\', '/');
+#endif
```
jrosdahl, line 403: `Config::set_base_dir` is only called by test code, so this change won't have any effect on the base dir value set via configuration file or environment variable. Is that what you intended?

I guess the idea is to normalize `base_dir` to only use forward slashes? But if we do that then it won't be possible to match paths (in `Util::make_relative_path`) that use backslashes. That is, if we normalize `base_dir`, shouldn't we also normalize the paths that are compared with `base_dir`?

src/Config.hpp
```diff
@@ -398,12 +398,18 @@ inline void
...
 }
 
 inline void
 Config::set_cache_dir(const std::string& value)
 {
   m_cache_dir = value;
+#ifdef _WIN32
+  std::replace(m_cache_dir.begin(), m_cache_dir.end(), '\\', '/');
+#endif
```
jrosdahl, line 412: Why is this needed?

src/Util.cpp
```diff
@@ -1100,7 +1100,15 @@ real_path(const std::string& path, bool return_empty_on_error)
     if (!ok) {
       return path;
     }
+#ifdef _WIN32
```
jrosdahl, line 1103: This `#ifdef` is not needed since `_WIN32` is checked on line 1084 above.

src/Win32Util.hpp
```diff
@@ -18,6 +18,10 @@
 
 #pragma once
 
+#ifdef __MINGW32__
+#include <windows.h>
+#endif
```
jrosdahl, line 23: `windows.h` is already included by `system.hpp` below, so why is this needed?

src/ccache.cpp
```diff
@@ -1511,7 +1511,25 @@ calculate_result_name(Context& ctx,
...
+      std::string prefix = hash_path.substr(0, pos);
+      hash_path = hash_path.substr(pos);
+      if (!ctx.config.base_dir().empty() && Util::is_absolute_path(hash_path)) {
+        hash_path = Util::get_relative_path(ctx.config.base_dir(), hash_path);
+      }
+      hash_path = prefix + hash_path;
+      hash.hash(hash_path);
+    } else {
+      hash.hash(args[i]);
+    }
```
jrosdahl, line 1528: Do I understand correctly that you propose to rewrite any Windows-style paths to relative for unknown options? I.e., if there is a `-do-something=C:\foo\bar` (and ccache doesn't know about `-do-something`) then look for the first `:` and, in this case, extract the path `C:\foo\bar` and potentially make it relative?

If so, I think there are two problems:

1. It only works for Windows paths. It would be better to add proper knowledge about options with path arguments so that the paths can be rewritten regardless of platform.
2. Path options can include `:` that are not part of a path, and then the code will probably not do the right thing.

src/ccache.cpp
```diff
@@ -1590,7 +1608,14 @@ calculate_result_name(Context& ctx,
...
     hash.hash_delimiter("inputfile");
-    hash.hash(ctx.args_info.input_file);
+
+    // If base_dir is set, use the relative path to generate hash
+    if (!ctx.config.base_dir().empty() && Util::is_absolute_path(ctx.args_info.input_file)) { 
+      std::string hash_path = Util::get_relative_path(ctx.config.base_dir(), ctx.args_info.input_file);
+      hash.hash(hash_path);
+    } else {
+        hash.hash(ctx.args_info.input_file);
+    }
```
jrosdahl, line 1618: This will make the expansion of the `__FILE__` macro incorrect (see line 1609). There used to be a way of opting out of hashing the input file (see 02bcef0d82ac8f1397c9b13bf1215cf770e6b7da for background) so it's a deliberate choice to include the full input path in the hash.

That said, I think I would be OK with making the input path relative if the "`file_macro` sloppiness" setting is reintroduced but with new semantics:

```
if file_macro sloppiness is set:
    input_file = make_relative(input_file)
hash(input_file)
```
That is, keep `input_file` as is by default but make it relative if `file_macro` sloppiness is set.

jrosdahl (commented):
Thanks for wanting to improve ccache. I'm however missing background and motivation for many of the proposed changes, so I would appreciate it if you could give more info.

> * fix the issues detected during build

Which issues?

> * Convert the "" to "/", then we avoid future issues

What future issues? (I guess you mean "\\", not ""?)

> * Fix BaseDir function:
>   * When base_dir option is set, get the relative path to calculate the file hash in cache.

Since there is no `BaseDir` function, do you mean the change you propose to `calculate_result_name`?

jrosdahl:
Closing this PR due to no response. Please feel free to reopen the issue if you have more information.

PR #640, opened 2020-08-23
Fix override usage

AlexanderLanin:
This change is required to make ccache compile with clangs trunk version. It seems they improved the "missing override" warning.

The change within fmt is reported as https://github.com/fmtlib/fmt/pull/1836
If we do not want to change fmt directly we could instead disable the warning via CMakeLists or upgrade to fmt trunk version.

### DIFF
--- a/src/ResultDumper.hpp
+++ b/src/ResultDumper.hpp
@@ -28,13 +28,13 @@ class ResultDumper : public Result::Reader::Consumer
 public:
   ResultDumper(FILE* stream);
 
-  void on_header(CacheEntryReader& cache_entry_reader) final;
+  void on_header(CacheEntryReader& cache_entry_reader) override;
   void on_entry_start(uint32_t entry_number,
                               Result::FileType file_type,
                               uint64_t file_len,
-                              nonstd::optional<std::string> raw_file) final;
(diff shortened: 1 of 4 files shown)


src/ResultDumper.hpp
```diff
@@ -28,13 +28,13 @@ class ResultDumper : public Result::Reader::Consumer
 public:
   ResultDumper(FILE* stream);
 
-  virtual void on_header(CacheEntryReader& cache_entry_reader);
-  virtual void on_entry_start(uint32_t entry_number,
+  void on_header(CacheEntryReader& cache_entry_reader) final;
```
jrosdahl, line 31: I'm using `override` in other interface implementations (e.g. in `ZstdCompressor.hpp`). How about using `override` here as well?

AlexanderLanin: You mean `override` instead of `final`? That would probably silence the warning too.

There is some (small) debate whether `final` is a good idea or not. I'm not really following. I personally usually set final since it's no effort to change it back to override when needed. In does allow further compiler optimizations, although I have not looked at the surrounding code and don't know if this is the case here / relevant here or not.

I can certainly change that if you prefer `override`.

jrosdahl:
Thanks.

Issue #639, opened 2020-08-23
Test fails on NFS and with old objdump


dyadav7:
### Description ###

This test " debug_prefix_map" fails when i compile ccache in NFS directory. It passes when i run ccache from /export (local disk) directory. It is better to fix it.

--logs--

Skipped test suite sanitize_blacklist [-fsanitize-blacklist not supported by compiler]
Running test suite debug_prefix_map.
FAILED

Test suite:     debug_prefix_map
Test case:      Mapping of debug info CWD
Failure reason: Relocation (dir) not found in test.o

ccache -s:

jrosdahl:
> This test " debug_prefix_map" fails when i compile ccache in NFS directory

Both the debug_prefix_map and the inode_cache suites work fine for me on NFS. I also don't see why those tests would be affected by the underlying file system type. Therefore the next thing to do is to investigate why they fail and fix the issues instead of just skipping them.

Since I can't reproduce the issue I'm afraid it will be up to you to do the debugging. A tip is to run something like `bash -x test/run debug_prefix_map` to get a trace of what the test is doing.

dyadav7:
I am getting this message "Failed to open inode cache" / " Inode cache not supported because the cache file would be located on nfs"  in NFS directory. When in run in local these messages are not there.

[2020-08-28T09:40:34.061851 21436] Apparent working directory: /home/deepaky/temp/ccache/orig/ccache/test/testdir/21363/run
[2020-08-28T09:40:34.062241 21436] Source file: test1.c
[2020-08-28T09:40:34.062256 21436] Object file: test1.o
[2020-08-28T09:40:34.066034 21436] Trying direct lookup
[2020-08-28T09:40:34.066377 21436] Failed to open inode cache /home/deepaky/temp/ccache/orig/ccache/test/testdir/21363/.ccache/tmp/inode-cache.v1: No such file or directory
[2020-08-28T09:40:34.066388 21436] Creating a new inode cache
[2020-08-28T09:40:34.070882 21436] Inode cache not supported because the cache file would be located on nfs: /home/deepaky/temp/ccache/orig/ccache/test/testdir/21363/.ccache/tmp/inode-cache.v1
[2020-08-28T09:40:34.073498 21436] Failed to open inode cache /home/deepaky/temp/ccache/orig/ccache/test/testdir/21363/.ccache/tmp/inode-cache.v1: No such file or directory
[...]

PR #643, opened 2020-08-25
Ignored 2 test on NFS 

dyadav7:
Fix for issue #639 - These 2 test fails in NF (network file system).


jrosdahl:
Two things:

* The list of commits looks very strange. Please consider rebasing to master instead of merging old commits.
* I don't think that it's a good idea to skip tests instead of fixing them. See discussion in #639.

PR #645, opened 2020-08-27
Command line parsing

AlexanderLanin:
Triggered by getopt complexity in #632 I had a look why it is even used.
It's only one small function. And it's used for parsing the command line.

We can do that in C++ instead! But let's not reinvent the wheel.
argh is one of the smallest command line parsers available, so let's start trying with that.
Doesn't sound terribly good to replace one dependecy with another, until you compare the complexity of the two options.

@jrosdahl: Any feedback so far?

Next terrifying thought was to apply that to calculate_result_name. But I guess for that a more powerful parser would be required.
Edit: ok, no. No way. Just not doable.
However one could carefully start wrapping args into something argh like.

e.g. instead of 
```
    if (Util::starts_with(args[i], "-specs=")
        || Util::starts_with(args[i], "--specs=")) {
      std::string path = args[i].substr(args[i].find('=') + 1);
```
something like
```
if(args[i].is({"-specs", "--specs"}) { 
  std::string path = args[i].value();
```


jrosdahl:
> Triggered by getopt complexity in #632 I had a look why it is even used.
> It's only one small function. And it's used for parsing the command line.

I find it slightly amusing that you just learned about the many decades old `getopt`/`getopt_long`. :slightly_smiling_face: (I'm probably just getting old.)

> @jrosdahl: Any feedback so far?

It sounds good to replace `getopt_long` with something simple if it emulate the current behavior in a satisfactory way.

jrosdahl:
I guess one blocker with argh could be that it doesn't keep the order of options? For instance, is it possible to make `ccache -M 10 -c -M 5` work as expected?

AlexanderLanin:
Haha I've replaced all command line parsing (with Args, not argh).
There is one single random test that is failing. Because it does something I did not expect: it calls `ccache -Cz`. I was not aware it's possible to combine arguments.

AlexanderLanin:
Well here it gets interesting since we do want to preserve current havior.
So first of all let's have a look at current behavior:

|ccache -M 10 -s -M 5|Executed left to right|
|ccache -sM 10|Executed left to right|
|ccache -sM=10|Error parsing "=10"|
|ccache -M=10|Error parsing "=10"|

jrosdahl:
> There is one single random test that is failing. Because it does something I did not expect: it calls `ccache -Cz`. I was not aware it's possible to combine arguments.

Maybe you have used `rm -rf dir` instead of `rm -r -f dir` or `tar -xf x.tar.gz` instead of `tar -x -f x.tar.gz`? That's the way many classical UNIX utilities work.

> ccache -M 10 -s -M 5 | Executed left to right

I should mention that I'm not aware of anybody using the options in this way, only that it's an edge case that we may want to care about. After some thought I think that it would be OK to not be 100% backward compatible in ccache 4.0. Basic common behavior still needs to work like before, though, for instance these:

```
ccache -M 10 -F 0
ccache -M10 -F0
ccache --max-size 10 --max-files 0
ccache --max-size=10 --max-files=0
```

It doesn't look like argh would be able to handle the second case above. On the other hand, since the `getopt_long` functionality is already in place maybe spend the effort on something else?

> ccache -Fa | Says "unset" but actually writes "a" into config. Completely bricking all consecutive ccache calls

I can't reproduce this. Which ccache version?

AlexanderLanin:
@jrosdahl: ccache version 3.4.1
So it's probably irrelevant ;-)

PR #647, opened 2020-08-30
Merge address and UB sanitizer

AlexanderLanin:
* Seems reasonable to have less builds.
* The concatenation of multiple sanitzers was actually wrong.
* I was going to enable memory sanitizer in the other job but failed due to https://github.com/onqtam/doctest/issues/414


jrosdahl:
Thanks.

PR #650, opened 2020-09-02
Add Visual Studio build directory to gitignore

AlexanderLanin:
Visual Studio 2019 builds into /out/ directory. The only way to change that is by adding yet another file named CMakeSettings.json. Adding the default to gitignore seems simpler.

Use Case: running unittests through Visual Studio due to its great debugger.


jrosdahl:
Thanks.

PR #654, opened 2020-09-03
const everything

AlexanderLanin:
Readability :-)

I'll leave it draft as some other pull requests seem to be merge-ready. I don't want to make them suffer through merging this.


jrosdahl:
I'm sorry, this is too intrusive.

AlexanderLanin:
I guess I can do it in smaller chunks and only on the relevant files ;-)

PR #658, opened 2020-09-12
Tests: Compare relevant part, not internal state

AlexanderLanin:
Assuming Args has (or gets 😉) an internal state, this is not what these tests should compare. They do not care about comparing the internal representation of Args, so it doesn't make sense to convert both sides to Args. Instead these tests care about what happens when Args are serialized to a string. Usually via to_argv, but to_string is way more convenient to use here.

Note: Alternatively operator== could be maintained to compare in such a way. But it always gets difficult then what exactly that operator compares as equal.

Note: This is a small innocent commit which improves the codebase insignificantly. It should not be difficult to review and should not have any disadvantage. It prepares for a bigger change of Args and Arg. Regardless of whether that last PR is accepted or not, this change in itself is a very very small improvement already.

### DIFF
--- a/unittest/test_argprocessing.cpp
+++ b/unittest/test_argprocessing.cpp
@@ -105,9 +105,9 @@ TEST_CASE("dependency_args_to_preprocessor_if_run_second_cpp_is_false")
   const ProcessArgsResult result = process_args(ctx);
 
   CHECK(!result.error);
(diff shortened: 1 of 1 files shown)


jrosdahl:
Sure, I have nothing against this!

PR #659, opened 2020-09-12
Test: introduce unit test for -Xclang argument

AlexanderLanin:
There is currently no unittest for testing parsing of -Xclang paramaters.
Expected output is reverse engineered form current implementation.


jrosdahl:
Thanks.

Issue #432, opened 2019-06-05
Support truly forcing which compiler to use


SebastianBoe:
### Description ###

Hi, ccache chooses the compiler to launch according to the following priority:

```
      # 1. Compiler specified on the command line with an absolute path
      # 2. Compiler specified with the CCACHE_CC env var
      # 3. Compiler specified on the command line with a relative path
```
But I need ccache to force a different compiler even though there is a compiler specified on the command line with an absolute path. So I need this priority:

```
      # 1. Compiler specified with the CCACHE_CC env var
      # 2. Compiler specified on the command line
```

Or a new variable and have this priority:

```
      # 0. Compiler specified with the CCACHE_CC_FORCED env var
      # 1. Compiler specified on the command line with an absolute path
      # 2. Compiler specified with the CCACHE_CC env var
      # 3. Compiler specified on the command line with a relative path
```

For background, the reason for needing this is that I am developing a CMake-based build system, and in CMake I can change the env vars (like CCACHE_CC) on a per-source file basis, but I can't change the compiler, and the compiler must have an absolute path.

I would like to know if a patch that supported either approach would be accepted.

jrosdahl:
(Note: `CCACHE_CC` is a deprecated alias of `CCACHE_COMPILER`.)

I think that it would be reasonable to let `CCACHE_COMPILER` have highest priority even in the absolute path case. Such a patch would likely be accepted, especially if it comes with a suitable test case. :slightly_smiling_face:

SebastianBoe:
>  Such a patch would likely be accepted, especially if it comes with a suitable test case. 

Great, thank you for the reply.

mumpitzstuff:
You could set the compiler path before cmake is started with: ccache --set-config=compiler=<path>/gcc

PR #661, opened 2020-09-15
Refactor: convert Util function to use string_view

AlexanderLanin:
The change is small, but once you start using string_view all those small functions have to use string_view as well. Otherwise a string needs to be created out of a string_view just to be discarded a second later.
(Compiler may be able to optimize it away if the function would accept a string object, but using string_view seems cleaner and is more in line to the other functions)

'+ Add some constness in the surrounding lines since we are touching the file anyway.

### DIFF
--- a/src/Util.cpp
+++ b/src/Util.cpp
@@ -1330,7 +1330,7 @@ strip_ansi_csi_seqs(string_view string)
 std::string
 strip_whitespace(string_view string)
 {
-  auto is_space = [](const int ch) { return std::isspace(ch); };
+  auto is_space = [](int ch) { return std::isspace(ch); };
   auto start = std::find_if_not(string.begin(), string.end(), is_space);
   auto end = std::find_if_not(string.rbegin(), string.rend(), is_space).base();
   return start < end ? std::string(start, end) : std::string();
--- a/src/Util.cpp
+++ b/src/Util.cpp
@@ -1312,7 +1312,7 @@ strip_ansi_csi_seqs(string_view string)
   std::string result;
 
   while (true) {
-    const auto seq_span = find_first_ansi_csi_seq(string.substr(pos));
+    auto seq_span = find_first_ansi_csi_seq(string.substr(pos));
     auto data_start = string.data() + pos;
     auto data_length =
       seq_span.empty() ? string.length() - pos : seq_span.data() - data_start;
--- a/src/Util.cpp
+++ b/src/Util.cpp
@@ -1328,7 +1328,7 @@ strip_ansi_csi_seqs(string_view string)
 }
 
 std::string
-strip_whitespace(nonstd::string_view string)
+strip_whitespace(string_view string)
 {
   const auto is_space = [](const int ch) { return std::isspace(ch); };
   const auto start = std::find_if_not(string.begin(), string.end(), is_space);
--- a/src/Util.cpp
+++ b/src/Util.cpp
@@ -1314,7 +1314,7 @@ strip_ansi_csi_seqs(string_view string)
   while (true) {
     const auto seq_span = find_first_ansi_csi_seq(string.substr(pos));
     auto data_start = string.data() + pos;
-    const auto data_length =
+    auto data_length =
       seq_span.empty() ? string.length() - pos : seq_span.data() - data_start;
     result.append(data_start, data_length);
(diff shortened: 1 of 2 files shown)


jrosdahl:
> [...] once you start using string_view all those small functions have to use string_view as well [...]

Yes, using `string_view` where appropriate is work in progress. Letting `Util::strip_whitespace` take a `string_view` is OK.

> + Add some constness in the surrounding lines since we are touching the file anyway.

Short answer: No thanks.

Longer answer: I'm OK with `const`-ifying local variables when modifying those lines (or very adjacent lines) for other purposes. As you're probably aware, the code currently doesn't have any `const`-ified parameter (value) types in function definitions – I have so far kept function declaration and definition parameter lists identical (except with dropped namespace prefix where applicable). If you want to propose a change of style I would appreciate a discussion instead of an implicit "by the way" change in a PR.

That said, I'm focusing on getting the code ready for release, and code style discussion is not what I would like to spend time on right now. Maybe you would like to help with getting closer to a release by having a look at #432?

AlexanderLanin:
Oh I was not aware that was intentional. I thought it's just old code. I'll split this as you suggested ASAP.

AlexanderLanin:
I've removed const from parameters. Will write an issue to discuss it after 4.0.

src/Util.cpp
```diff
@@ -1328,11 +1328,12 @@ strip_ansi_csi_seqs(string_view string)
 }
 
 std::string
-strip_whitespace(const std::string& string)
+strip_whitespace(nonstd::string_view string)
```
jrosdahl, line 1331:
```suggestion
strip_whitespace(string_view string)
```

src/Util.cpp
```diff
@@ -1312,9 +1312,9 @@ strip_ansi_csi_seqs(string_view string)
   std::string result;
 
   while (true) {
-    auto seq_span = find_first_ansi_csi_seq(string.substr(pos));
+    const auto seq_span = find_first_ansi_csi_seq(string.substr(pos));
     auto data_start = string.data() + pos;
-    auto data_length =
+    const auto data_length =
```
jrosdahl, line 1317:
```suggestion
    auto data_length =
```

src/Util.cpp
```diff
@@ -1312,9 +1312,9 @@ strip_ansi_csi_seqs(string_view string)
   std::string result;
 
   while (true) {
-    auto seq_span = find_first_ansi_csi_seq(string.substr(pos));
+    const auto seq_span = find_first_ansi_csi_seq(string.substr(pos));
```
jrosdahl, line 1315:
```suggestion
    auto seq_span = find_first_ansi_csi_seq(string.substr(pos));
```

src/Util.cpp
```diff
@@ -1328,11 +1328,12 @@ strip_ansi_csi_seqs(string_view string)
...
-strip_whitespace(const std::string& string)
+strip_whitespace(nonstd::string_view string)
 {
-  auto is_space = [](int ch) { return std::isspace(ch); };
-  auto start = std::find_if_not(string.begin(), string.end(), is_space);
-  auto end = std::find_if_not(string.rbegin(), string.rend(), is_space).base();
+  const auto is_space = [](const int ch) { return std::isspace(ch); };
+  const auto start = std::find_if_not(string.begin(), string.end(), is_space);
+  const auto end =
+    std::find_if_not(string.rbegin(), string.rend(), is_space).base();
```
jrosdahl, line 1336:
```suggestion
  auto is_space = [](const int ch) { return std::isspace(ch); };
  auto start = std::find_if_not(string.begin(), string.end(), is_space);
  auto end = std::find_if_not(string.rbegin(), string.rend(), is_space).base();
```

src/Util.cpp
```diff
@@ -1328,9 +1328,9 @@ strip_ansi_csi_seqs(string_view string)
 }
 
 std::string
-strip_whitespace(const std::string& string)
+strip_whitespace(string_view string)
 {
-  auto is_space = [](int ch) { return std::isspace(ch); };
+  auto is_space = [](const int ch) { return std::isspace(ch); };
```
jrosdahl, line 1333:
```suggestion
  auto is_space = [](int ch) { return std::isspace(ch); };
```

PR #662, opened 2020-09-17
Added logging of CPATH and other envvars affecting preprocessing & compilation

Str1ker17:
Setting CCACHE_LOGFILE environment variable allows to quickly track down compilation problems, especially on large projects with a plenty of Makefiles.
Use of ccache output like `Command line: ccache /usr/bin/gcc-5 ...` allows to re-run compilation of a single file manually, play with flags, run preprocessing instead of compilation, etc.

However, one day I encountered a problem on different project, where re-running gcc with all parameters from log.
After some research I found that env var CPATH was set, that GCC accepts as additional include path.

This fix just prints "Environment variable affects preprocessing" messages to log, if these variables are present. Concatenating them with a command line results in a successfull compilation.
This was useful for me. (Convicting the practise of using tricky envvars is beyond the scope of ccache)
I decided to share my experience with the community.
Hope this will help someone one day.
Thanks

### DIFF
--- a/src/ccache.c
+++ b/src/ccache.c
@@ -88,6 +88,15 @@ static const char USAGE_TEXT[] =
 	"\n"
 	"See also <https://ccache.dev>.\n";
 
+static const char *preproc_envvars[] = {
+	"CPATH",
+	"C_INCLUDE_PATH",
+	"CPLUS_INCLUDE_PATH",
+	"OBJC_INCLUDE_PATH",
+	"OBJCPLUS_INCLUDE_PATH", // clang
+	NULL
+};
+
 // Global configuration data.
 struct conf *conf = NULL;
 
@@ -2271,15 +2280,7 @@ calculate_object_hash(struct args *args, struct args *preprocessor_args,
 	struct file_hash *object_hash = NULL;
 	if (direct_mode) {
 		// Hash environment variables that affect the preprocessor output.
-		const char *envvars[] = {
-			"CPATH",
-			"C_INCLUDE_PATH",
-			"CPLUS_INCLUDE_PATH",
-			"OBJC_INCLUDE_PATH",
-			"OBJCPLUS_INCLUDE_PATH", // clang
-			NULL
-		};
-		for (const char **p = envvars; *p; ++p) {
+		for (const char **p = preproc_envvars; *p; ++p) {
 			char *v = getenv(*p);
 			if (v) {
 				hash_delimiter(hash, *p);
@@ -4065,6 +4066,17 @@ ccache(int argc, char *argv[])
 	cc_log_argv("Command line: ", argv);
 	cc_log("Hostname: %s", get_hostname());
 	cc_log("Working directory: %s", get_current_working_dir());
+
+    // Since some environment variables can affect preprocessing
+    // and make logs unreplayable, this is an outlet
+    // See https://gcc.gnu.org/onlinedocs/gcc/Environment-Variables.html
+	for(const char **p = preproc_envvars ; *p ; ++p) {
+        char *v = getenv(*p);
+        if (v) {
+            cc_log("Environment variable affects preprocessing: %s='%s'"
+                , *p, v);
+        }
+    }
 
 	conf->limit_multiple = MIN(MAX(conf->limit_multiple, 0.0), 1.0);
 


jrosdahl:
Thanks for your contribution. Unfortunately I don't want to apply this to the 3.7-maint branch since it's in "bug fixes only" mode.

It could potentionally be added to the master branch, but I'm not sure that it's a good idea to use the log file for the purpose of replaying actual compiler commands  – the log file is only meant for debugging and the format is not guaranteed to be stable across ccache versions. Also, the code that prints the command line to the log file doesn't quote characters, so arguments with spaces or quotes will lead to an incorrect command line.

I would prefer to see a feature where you can set e.g. `CCACHE_COMMANDLOG` to make ccache write only the commands in some format to a separate file.

AlexanderLanin:
This doesn't seem to be about a replayable command log in that sense. It just prints ENV variables to log. They are not quotes, but that seems fine since it's only supposed to be human readable.

jrosdahl:
> This doesn't seem to be about a replayable command log in that sense. [...] it's only supposed to be human readable

Then we interpret the description differently. From the PR description: _"Concatenating them with a command line results in a successfull compilation."_ From the diff: _"make logs unreplayable"._

AlexanderLanin:
Yes indeed it's a little confusing. I think the code change is fine. The commit message is fine. The comment in code is a little confusing. The PR is a little confusing.

I guess that "replay" is the source of confusion here. Maybe "reproduce" would be better suited.

src/ccache.c
```diff
@@ -4066,6 +4067,17 @@ ccache(int argc, char *argv[])
 	cc_log("Hostname: %s", get_hostname());
 	cc_log("Working directory: %s", get_current_working_dir());
 
+    // Since some environment variables can affect preprocessing
+    // and make logs unreplayable, this is an outlet
```
AlexanderLanin, line 4071:
```suggestion
    // and make logs unreproducible, this is an outlet
```

Str1ker17:
@AlexanderLanin, you first assumtion was right:

> I guess that "replay" is the source of confusion here. Maybe "reproduce" would be better suited.

I manually reproduce compilation commands using `CCACHE_LOGFILE` (this is convenient for me because ccache is enforced in all my compilations, and no other wrappers needed to log commands). Reproducibility was the intention of this code change.

On the other hand, ccache takes these env vars into account when detects cache hits/misses, but logs nothing about them. This may lead to surprising cache misses.

So, if @jrosdahl says that `CCACHE_LOGFILE` is not a good source for reproducing compiler commands (I personally do not agree with it), logging these env vars can be helpful to debug ccache misses in some circumstances.

AlexanderLanin:
@Str1ker17: maybe you could update PR and potentially other stuff to reflect that this is simply about logging ENV vars and not about creating some new thing which enables replay of compiler commands? I think that would eliminate some discussions here.

Str1ker17:
Ok, I'll update descriptions after some time.

jrosdahl:
OK. Clarifying my first message (since it seems it was partially misunderstood) and adding some more points:

1. I don't want this change on the 3.7-maint branch since I only want to apply bug fixes there. It's possible that there will be no more 3.7.x releases. @AlexanderLanin: Did you miss this point or are you of a different opinion since you continue to encourage changes in the 3.7-maint context?
2. I could be OK with this or something similar on the master branch. In that case I think that it would make sense to try to log all environment variables that affect the build, including `DEPENDENCIES_OUTPUT` and `SUNPRO_DEPENDENCIES`. Also, should ccache then log `LIBRARY_PATH` as well since it affects the linking and you may want to reconstruct a linker command from the log? Where to draw the line between "useful for ccache debugging" and "useful for reconstructing any command seen by ccache"?
3. On the other hand: to debug cache miss/hit behavior you're better off enabling the debug mode and comparing `<objectfile>.ccache-input-text` files since they contain the actual answer. I don't think that it makes sense to strive to include exactly everything that affects the compilation in the main log file.
4. Using the log file for reconstructing compiler commands ad hoc is of course perfectly OK given the limitations. You can do what you want with the log.
5. However, I don't want to advertise the logging of commands and environment variables as an official way to reconstruct compilation commands. I want to have the freedom to restructure the log file and its format without making compatibility guarantees.
6. After reading https://gcc.gnu.org/onlinedocs/gcc/Environment-Variables.html it seems like it would be a good idea to look into handling `GCC_COMPARE_DEBUG` and `SOURCE_DATE_EPOCH` as well.

jrosdahl:
@Str1ker17: Any thoughts on my points 2 and 3 (how much to log)? Would you like to work on adding logging to the master branch if we agree on what to log?

@AlexanderLanin: Any answer to my question in point 1?

> 6. After reading https://gcc.gnu.org/onlinedocs/gcc/Environment-Variables.html it seems like it would be a good idea to look into handling `GCC_COMPARE_DEBUG` and `SOURCE_DATE_EPOCH` as well.

Done now.

Str1ker17:
> Also, should ccache then log LIBRARY_PATH as well since it affects the linking and you may want to reconstruct a linker command from the log?

I don't think so, because link results are not cached by ccache in any way. Handling or not by ccache is a good place to "draw the line" in my opinion.

> to debug cache miss/hit behavior you're better off enabling the debug mode and comparing <objectfile>.ccache-input-text files

This will produce too many files (and information in them) and take much more time comparing to just printing to log.
Also, finding relevant `*.ccache-input-text` file for every input file in project tree is a little challenge.
This is perfect for debugging internals of ccache on small projects but seems messy with large projects when I'm not a ccache contributor.

> I don't want to advertise the logging of commands and environment variables as an official way to reconstruct compilation commands. I want to have the freedom to restructure the log file and its format without making compatibility guarantees.

I don't advertise log to be parsed by machine (but TBH I had this idea when implementing project file generator for C/C++ projects). In this PR logfile is supposed to be human-readable only. You are still free to restructure the log :)

> Would you like to work on adding logging to the master branch if we agree on what to log?

I can move on master branch with my logging PR, but the new version is in C++, not C (omg why did you do this?!?!) and require some time to adapt.

AlexanderLanin:
1) While this is a really small and unrisky change it's technically still a feature and not a bugfix. So it should be aimed at 3.8 or 4.0. Since there is no 3.8 planned and 4.0 is approaching I agree that it makes sense to target that for 4.0. Regardless of 3.7 or 3.8, we would have two active branches. While that is definitely nice in general, ccache does not have enough contributors to keep that up. If all goes as planned 3.x will be dead soon. So let's focus on 4.0 indeed. It's difficult enough to get 4.0 done and fix the most important bugs on both branches.
That said I personally wouldn't mind accepting such a **super small** PR into a stabilization branch (3.7) since it's already available and doesn't harm anyone (to my understanding). There is pretty much 0 risk in it having a new bug which is one of the main points of having such stabilization branches. It doesn't need any documentation since it's not an official "feature" - it's only unofficial log. It only helps to understand why ccache-input-text have differences. So the biggest problem I see is that upgrading to 4.0 will drop this "feature". And that we waste time on discussions such as these ;-)
I can understand both ways. E.g. saying let's not waste ANY effort on 3.x.

Now regarding the feature itself, but this goes way beyond this PR:
I guess the problem to be addressed here is that it's difficult to understand cache misses. Maybe there could be an env variable CCACHE_EXPLAIN_MISSES which helps somehow. It could implicitly enable debug mode. But I don't know how much can be determined beyond "that hash is not cached". Second idea would be that it would automatically compare ccache-input-text files from different build directories (same source directory).

jrosdahl:
Please rebase to master if you are still interested in continuing this.

> This is perfect for debugging internals of ccache on small projects but seems messy with large projects when I'm not a ccache contributor.

It's also perfect when you want to debug cache misses in projects. It's certainly not only for ccache developers.

Str1ker17:
Ok.
Is 4.x version stable enough to be used as a replacement for 3.7?

jrosdahl:
As far as I know: yes.

PR #663, opened 2020-09-17
Refactor: slightly improve naming in compopt

AlexanderLanin:
This is a small almost insignificant change.
I was wondering what those functions do, so I improved their names slightly.
+ Tests are cleaned up to less test cases + some border cases added.

I made separate commits for trivial review 👍

I was actually looking into the argument type those functions receive (string, string_view or char*), but that got a little out of hand so I wanted to get this trivial part merged first.

### DIFF
--- a/src/argprocessing.cpp
+++ b/src/argprocessing.cpp
@@ -318,9 +318,9 @@ process_arg(Context& ctx,
       && (args[i + 1] == "-emit-pch" || args[i + 1] == "-emit-pth"
           || args[i + 1] == "-include-pch" || args[i + 1] == "-include-pth"
           || args[i + 1] == "-fno-pch-timestamp")) {
-    if (compopt_affects_comp(args[i + 1])) {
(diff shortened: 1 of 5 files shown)


jrosdahl:
The first two commits are OK.

Regarding the third: Improving names is of course good, but I'm not sure if these changes are an improvement. :slightly_smiling_face:  `compopt_affects_cpp` tells if an option affects the **preprocessor output** (i.e., the output of `$compiler -E`), not the preprocessor arguments. Similarly, `compopt_affects_comp` tells if an option affects the **compiler output** (i.e., object or dependency file, etc).

Since you got confused about this, the names or the (nonexistent) documentation would clearly benefit from some change. How about `compopt_affects_{cpp,compiler}_output` and/or adding documentation comments?

AlexanderLanin:
That's kind of obvious after reading your comment. I named the functions after the effect of how they are used. But that's already logic of process_args.

PR #665, opened 2020-09-19
Allow ccache to be built from github's source tarballs

pfultz2:
When downloading the source tarballs, the build fails because its not a git repository. Enabling builds directly from the source tarballs, allows it to be installed from github with tools like `cget install ccache@master`. 

For now, it changes it to a warning message, and sets the version to 0.0.0.


afbjorklund:
I think that ccache still uses "make dist", to generate tarballs that work

AlexanderLanin:
Yeah there is a specific target to generate source tars. I guess this works as a quick fix. But I would really like to see how others have solved the issue.

afbjorklund:
For other software, they just require the version as an extra input parameter if not using git.

So I guess it is down to either _requiring_ a -DVERSION, or adding the version to the repo ?

AlexanderLanin:
I'd love those source tarballs to be buildable which this PR achieves (So it should be merged for now).

More than that it makes sense for those builds to report the correct version. But that's probably a separate discussion in a separate issue if you all agree since that's difficult to achieve in detail.

jrosdahl:
@pfultz2 wrote:
> When downloading the source tarballs, the build fails because its not a git repository.

Right, that's working as intended: my intention has been that users never should use those annoying GitHub source code archives since they are not proper source releases. If there was a way to disable the source code archives for tags I would have done that.

As I've seen it, there are two use cases: non-developer users build from an uploaded source code release archive while developers clone and build from the Git repo. Users who want to test out an in-development version also clone the repo.

> Enabling builds directly from the source tarballs, allows it to be installed from github with tools like cget install ccache@master.

I didn't know about cget, so that's kind of a new use case.

> For now, it changes it to a warning message, and sets the version to 0.0.0.

Sorry, I don't like that because then a user may choose the incorrect source code archive from the asset list on a release page (e.g. https://github.com/ccache/ccache/releases/tag/v3.7.11) and not see the CMake warning. The ccache binary then won't have a proper version.

Also, when building in-development versions my philosophy has been that the version in the resulting binary for bug reporting and general sanity reasons should say which Git hash it is built from, not just an anonymous "dev" version.

Would it be possible for cget to use some GitHub API to query the version or hash and pass it either as an environment variable or an explicit `-D` option to the CMake script?

@AlexanderLanin wrote:
> I'd love those source tarballs to be buildable

Could you perhaps expand on why you would love that? Do you have another use case than the ones mentioned above?

pfultz2:
> Users who want to test out an in-development version also clone the repo.

For users, they should be using the source tarballs and not cloning, since they wont be doing development. The source tarballs is much more efficient(as it can be more easily cached) and puts less strain on github's infrastructure(since it doesn't have to touch the Git layer), see [here](https://github.com/CocoaPods/CocoaPods/issues/4989#issuecomment-193801376). 

> Would it be possible for cget to use some GitHub API to query the version or hash and pass it either as an environment variable or an explicit -D option to the CMake script?

Cget just uses standard cmake build process, which is to have the version defined by the `project` command. The user can pass in additional cmake flags(such as `-DCCACHE_VERSION`) so it could be installed with `cget install ccache -DCCACHE_VERSION=0.0`. Of course, this relies on the user to set the version correct.

jrosdahl:
> For users, they should be using the source tarballs and not cloning, since they wont be doing development. The source tarballs is much more efficient(as it can be more easily cached) and puts less strain on github's infrastructure(since it doesn't have to touch the Git layer), see [here](https://github.com/CocoaPods/CocoaPods/issues/4989#issuecomment-193801376).

I think there's a difference between the tiny use case of a handful of non-developer users wanting to try out a development ccache version and the resources required to serve a large package dependency service like CocoaPods. The GitHub UI doesn't do anything to hint users to download non-release tarballs instead of cloning so I don't get the impression that GitHub thinks it's a problem in the general case. I fully understand why you want to use source archives for cget, though.

> The user can pass in additional cmake flags(such as `-DCCACHE_VERSION`) so it could be installed with `cget install ccache -DCCACHE_VERSION=0.0`. Of course, this relies on the user to set the version correct.

OK. Sounds doable but not optimal.

As commented in #667 I found a way to make a build from a source archive behave like I want, so closing this PR.

PR #666, opened 2020-09-19
Add support for hip language

pfultz2:
This adds support for the [hip](https://github.com/ROCm-Developer-Tools/HIP) language, which is supported in llvm. This also runs the preprocessor with `-E -o -` since by default clang writes to a .cui file when using `-x hip`, so this will redirect it to stdout.


AlexanderLanin:
Could you add some tests that show how this is working and also ensure it will keep working in case of changes.

pfultz2:
Is there documentation to explain how tests should be written?

AlexanderLanin:
Unfortunately not much. There are two levels mostly: Unittests via doctest and software tests via bash scripts.
Unfortunately there do not seem to be any unittest for the functions affected here. And it does look like a significant nightmare to get started on those ;-)

So my guest guess is:
Add a new bash file in tests by duplicating profiling_clang.bash (for how to filter for clang and how to write a file to be compiled). Use something like the "Base case" from base.bash to trigger compilation of that one file and to ensure it will actually increase 'cache hit'.

pfultz2:
Thanks for the info, A couple of questions:

1) This requires clang or hip/rocm to be installed. Is there a way to enable it conditionally?

2) Is there a way to pass a variable from cmake to the bash script? This needs to happen because the flags to compile come from the `hip::device` cmake targets.

AlexanderLanin:
1) that's what you'll find in profiling_clang.bash. It runs only for clang. It filters for clang, it doesn't actively force clang. That is when cc/cxx points to clang/clang++ the script runs. Half the CI runs with gcc, the other half with clang. You can test locally by running `CC=clang CXX=clang++ test/run yourtest`
2) not sure what you aim at here. Isn't it enough to call clang with  `-c file.hip`? Similar to what the "Base case" does for `-c test1.c` file?

pfultz2:
> not sure what you aim at here. Isn't it enough to call clang with -c file.hip? Similar to what the "Base case" does for -c test1.c file?

No, it also needs the arch and the device libs. I think by default it will look for device libs in /opt/rocm, so if its not installed there the probe can fail.

pfultz2:
Actually, I can add `-nogpulib` to skip linking the device libs.

jrosdahl:
Thanks!

PR #673, opened 2020-09-29
Add error detection to zstd download

AlexanderLanin:
I encountered problems downloading zstd due to a proxy and got no helpful information from the current cmake script whatsoever since it did not evaluate exit codes.


jrosdahl:
Thanks.

PR #676, opened 2020-10-02
Add secondary HTTP storage

gjasny:
Hello,

this is my attempt to add an HTTP backend-storage to ccache. Despite trying we did not had success enlarging memcached with extstore. While working with Bazel we had good experience with it's HTTP-based [remote caching](https://docs.bazel.build/versions/master/remote-caching.html). So I thought it might be easiest to go with a simple HTTP storage backend first. There we also won't have to care about splitting the cache content into 1 MByte chunks.

The good news is also that my employer LogMeIn allowed me to contribute code as part of my regular work. Legal also signed off.

Could you please take a look at the changes and tell me where things could be improved or completely reworked?

Thanks,
Gregor

PS: I really like the new CMake and C++11 code base!


AlexanderLanin:
Besides the actual change which I cannot comment on, I wonder why HTTPCACHE_BACKEND is a compile time flag. Is there any significant downside to include it always?

jrosdahl:
Hi @gjasny, I'll have to come back to this some time after the 4.0 release.

afbjorklund:
> Despite trying we did not had success enlarging memcached with extstore.

Normally memcached is scaled horizontally (more servers), not vertically...

The suggested solution was Couchbase, if wanting to store things on disk too.
Also looked into Hazelcast IMDG instead of Memcached, if staying in-memory.

But having HTTP (and maybe SQL) would be nice features to have for #414
With the right interface in place, it should be simple to switch storage later on...

We ended up using NFS for the secondary cache storage... "Good ole rock"

afbjorklund:
> I wonder why HTTPCACHE_BACKEND is a compile time flag. Is there any significant downside to include it always?

Seems to add a dependency on libcurl

AlexanderLanin:
This sounds great but has been stale for a while. Sounds like that could be a really huge performance boost to caching in company environments (10+ people working on same codebase, but compiling locally like in the 80s).

Potential improvements:
* Improved interfaces.. whatever that means in detail
* Most definitely tests. We need to provide such a backend (fake) as part of the test suite.
* Provide documentation how to set it all up.
* I'd suggest turning it on by default / if curl can be found. Issue a warning otherwise but keep compiling without it.
* Don't we need a login option for the backend? I don't expect it to accept cached object files from just about anybody.
* Split curl logic into a separate C++ class (maybe there is such a wrapper already somewhere on GitHub), don't use it directly from HttpBackend.
* Move "multiple caches" (even if it's only a fixed setup of local + optional backend) into separate class, out of ccache.cpp
* I personally think the discussion in #414 has been going on long enough. Obviously no one can manage some full blown (open) solution. So we should just go ahead with this one and be done with it.
* Is this already compatible to something? It definitely should be compatible to existing (google, aws, azure) solutions, otherwise it's useless. e.g. https://docs.bazel.build/versions/master/remote-caching.html#http-caching-protocol and https://docs.bazel.build/versions/master/remote-caching.html#google-cloud-storage

afbjorklund:
Was there any updates on this ? I updated to 4.3, and it seemed to be working still: 80d1726e965137649381ac704084ec0b06d477dc

> Sounds like that could be a really huge performance boost to caching in company environments (10+ people working on same codebase, but compiling locally like in the 80s).

It also helps when you have a build farm, like when doing continuous integration...

> Is this already compatible to something? It definitely should be compatible to existing (google, aws, azure) solutions, otherwise it's useless.

If you want it more compatible with those, then I think you need to add [CAS](https://en.wikipedia.org/wiki/Content-addressable_storage) support.

afbjorklund:
@AlexanderLanin : I made a small hack to the ResultDumper, just to show the SHA-256 of each file (for the CAS)

```
Magic: cCrS
Version: 1
Compression type: zstd
Compression level: 1
Content size: 1891
Embedded file #0: <stderr> (303 bytes)
4901f9161519178198e5028dc89ba88ed72b87a441ce9b48a78592b6c54b419e  -
Raw file #1: .o (1544 bytes)
e051f56a0aa7f79cf934efeff5621e71eaf8970cbb0bd87dafc7263f48939d5e  /home/anders/.ccache/7/c/25m5m1mv979qjgv59t10rn7pl02sa5m1W
```

To make it clear, the key for the file is the checksum for the file's content:

```
$ sha256sum /home/anders/.ccache/7/c/25m5m1mv979qjgv59t10rn7pl02sa5m1W
e051f56a0aa7f79cf934efeff5621e71eaf8970cbb0bd87dafc7263f48939d5e  /home/anders/.ccache/7/c/25m5m1mv979qjgv59t10rn7pl02sa5m1W
```

The next step would be to add that to the http backend, so that it stored the files under `/cas/` on the server ?

I'm not sure where the other files would go, but you could something similar to the Bazel `/ac/` perhaps ?
> Bazel supports remote caching via HTTP/1.1. The protocol is conceptually simple: Binary data (BLOB) is uploaded via PUT requests and downloaded via GET requests. Action result metadata is stored under the path /ac/ and output files are stored under the path /cas/.

In the current http code they use file suffixes (like .manifest and .result), instead of new directory prefixes.

gjasny:
At work we switched from the memcached backend to the plain http put / get storage backend and were able to replace a 300GB RAM equipped EC2 instance with a much smaller one with 40gb network interface and a 2TB local disk at 1/10th of the price. It's backed by a nginx with sendfile enabled. Performance is great and it's running without issues since December.

Our ccache is still at the 3.something memcache branch with my http patches ontop. I'll try to work on this issue during my non-feature Fridays.

ftrofin:
So memcached doesn't seem to be an efficient solution for shared cache. How about Redis? Did anybody try ccache with Redis?

afbjorklund:
> So memcached doesn't seem to be an efficient solution for shared cache.

I think you will need to do your own benchmarks, it depends on the setup.

But this particular issue is about a **http** backend, not memcached/redis...

> How about Redis? Did anybody try ccache with Redis?

I made a backend for it, but decided on the memcached.

Once the new API is up, I'm sure it would be simple to add one for Redis too.
Otherwise the main feature of Redis was greater support for advanced types.

see e.g. https://aws.amazon.com/elasticache/redis-vs-memcached/

It was similar with SQL, a backend was added - but decided to use NoSQL.
Again, having both backends available would make it easy to do comparisons.

afbjorklund:
For the discussion I added a similar backend also for redis: 0576d5437aa1c8e853f2d4579b7f72f77a7c3775

One benefit over memcached is that it allows 512M rather than 1M

One downside seems to be only 1 server, only 1 thread, no proxy ?

This naive implementation just loads everything into `std::string`

afbjorklund:
@gjasny : I fixed the tracing (as per "todo") in 23235c79339841d26207b7750b554966ff04982a, but also discovered that it (tracing) was broken on master...

It would be nice if we could have some kind of `StorageBackend` available on master, with only a "null" implementation ?

----

Did you have any thoughts about how configuration could scale to having multiple backends ?

Obviously having one variable for each is not great, when talking about 4 or 5 storage backends...

`CCACHE_HTTPCACHE_URL=http://localhost:8080/cache`

`CCACHE_REDISCACHE_URL=localhost:6379`

`CCACHE_FILECACHE_URL=/mnt/nfs/path/to/shared/cache`

`CCACHE_MYSQLCACHE_URL=localhost`

There was also the discussion about compile-time vs. run-time (plugins), for configuring them.

_Sorry for hi-jacking the original HTTP request, we could move the discussion somewhere else._

afbjorklund:
There is a proof-of-concept of the CAS result files now, storing the object files outside of the result file.
It is very similar to how the "raw" files work, but not tied to reflink/hardlink but always doing copying.

Initially the files are just kept in a local directory, from where they would be uploaded/downloaded.
You would lose compression and the faster/smaller checksum, but it would be more "compatible"...

----

Normally the file is kept _inside_ the result:

```
8,0K	/tmp/ccache/d/b/2bo74c56j5urn1rjb3r8mihnorku3r6M
432K	/tmp/ccache/0/e/10d8p8k28uai926eb5pdd76un0brovaR
```

```
Magic: cCrS
Version: 1
Compression type: zstd
Compression level: 1
Content size: 1209706
Embedded file #0: .o (1209672 bytes)
cd6304128a5100f22d21046cc1c61cac858c84020fafdc433a7b691c4c018f3a  -
```

Here, it would instead be stored _outside_ the result:

```
1,2M	/tmp/ccache/cas/cd6304128a5100f22d21046cc1c61cac858c84020fafdc433a7b691c4c018f3a
8,0K	/tmp/ccache/d/b/2bo74c56j5urn1rjb3r8mihnorku3r6M
4,0K	/tmp/ccache/0/e/10d8p8k28uai926eb5pdd76un0brovaR
```

```
Magic: cCrS
Version: 1
Compression type: zstd
Compression level: 1
Content size: 1209706
SHA-256 checksum #0: .o (1209672 bytes)
cd6304128a5100f22d21046cc1c61cac858c84020fafdc433a7b691c4c018f3a
```

Only the checksum remains, inside the result.

(the raw file only stores the length, not any key)

Romain-Geissler-1A:
Hi,

It seems several people in this thread are working in companies with 10+ employees who would benefit from any kind of shared ccache folder without using a shared filesystem (like for example NFS). I see topic related to non-filesystem backends being updated from time to time, as well as #414 but I guess for now we are still far from having some kind of official support (or even just API) in ccache, right ? I am wondering who in this thread is using what backends for now, since apparently some do in their own company. Do you have problems maintaining this out of tree patches and apply it to newer ccache ? Do you regularly update ccache or freeze it once and let it like this for a long time ? How "production ready" would you consider for now the different out of tree storage backends ?

We are wondering this since in our own company we do use ccache massively (1000+ developers with hundreds of machine, maybe thousands as well), we used to rely on NFS to share these, but would like to phase NFS out. So now we are looking for replacements. I wonder if other kind of mountable filesystem (like for example S3) would be a suitable replacement. Some people in our company seem to have also mentionned sccache from Mozilla, but none of us did try it so far. I am looking for advices if some people in other companies already have some experience in the different alternatives to NFS ;)

Cheers,
Romain

afbjorklund:
@Romain-Geissler-1A  : you might want to bring this topic to the Discussions ?

The plan in this PR is to use HTTP (which I suppose would also include any S3),
but one could consider using File backend with _any_ filesystem - possibly FUSE.

The Redis and SQL was mostly for discusssion and API, not really deployment.

afbjorklund:
When using the **Redis** backend, one compromise is to use a memory-mapped file instead.
Then it will stay in memory as long as it is available, but persist to disk just in case it isn't.

So one can expect slower "put" performance, but almost the same "get" performance...
There is a nice C++ project: [**Ardb**](https://github.com/yinqiwen/ardb), that provides Redis-protocol for in-memory databases:

https://maheshsenniappan.medium.com/disk-based-key-value-stores-5e0da4c9bdea

I personally prefer [LMDB](https://symas.com/lmdb/), but default is [RocksDB](http://rocksdb.org/). (See http://lmdb.tech/bench/inmem/)

* https://github.com/LMDB/lmdb
* https://github.com/facebook/rocksdb

So this can also be considered, as an alternative to using a memcached cluster for ccache ?

* moved to https://github.com/afbjorklund/memccache

There is also a "[nutcracker](https://github.com/twitter/twemproxy)" program, that performs the same proxy task as old "[moxi](https://github.com/couchbase/moxi)" did...

* see also https://redis.io/topics/benchmarks discussion

----

For these storage backends, we use the native ccache format and don't bother with sha256/gzip.
That is, it will use BLAKE3/Zstd for performance - rather than trying to stay compatible with Bazel.

I would be happy to break out the Redis backend into a separate PR, but GitHub doesn't allow
for dependencies and since it uses the same storage backend - I included it in this issue instead.

jrosdahl:
I wrote:
> Hi @gjasny, I'll have to come back to this some time after the 4.0 release.

Sorry for taking so long to come back to this. The 4.0 release has now happened and I've made a couple of (mostly) bug fix releases, so now the storage backend stuff has bubbled higher up on my list of things to think about.

Thanks for the PR – it's generous of your company to allow you to contribute code! :+1: 

Before summarizing my thoughts on this I have some replies to comments made by others.

jrosdahl:
@AlexanderLanin wrote:
> I personally think the discussion in #414 has been going on long enough.

I would say that the problem with #414 is that the real discussion did not even _start_, not that it has been going on long enough. But see my recent comment to #414.

> So we should just go ahead with this one and be done with it.

If you mean "just merge it": No we shouldn't.

> Is this already compatible to something?

Yes, plain HTTP GET/PUT is compatible with many servers.

jrosdahl:
@afbjorklund: Thanks for your prototyping work.

@afbjorklund wrote:

> If you want it more compatible with those, then I think you need to add CAS support.

I don't understand why you mention CAS support in this context. Could you expand on why we would need CAS support for the listed cloud storage solutions? I'm not familiar with APIs of all cloud storage providers, but I would be surprised if you can't store a key-value pair where key != HASH(value) in all of them.

Or did you just talk about the Bazel part? If so, you can AFAIK store any data you like under `/cas`. bazel-remote for instance accepts any data as long as the key _looks_ like a SHA-256 hash (i.e. is 64 base16 digits) and it's trivial to encode our keys into that format. Looks like it's possible to use `/ac` as well with `--disable_http_ac_validation` if you're uncomfortable violating the CAS-ness of `/cas` (but then you'll put invalid actions in the action cache instead). To clarify: there will be no risk of collision or anything like that in practice since ccache doesn't use SHA-256.

Storing files outside of the result files or doing CAS storage in general _by default_ does not sound appealing to me, spontaneously. It could be of interest for a deduplication feature, but as I see it that's not something a storage backend should be aware of. I have an unimplemented sketch for how to introduce deduplication and it makes use of special entries in the result files, similar to the raw entries. The lookup towards storage will then still be "get(key)" and "put(key, value)" so the storage can be agnostic about this.

jrosdahl:
@Romain-Geissler-1A wrote:
> I am wondering who in this thread is using what backends for now, since apparently some do in their own company.

I agree with @afbjorklund that such a discussion doesn't belong in the context of this PR. Please start a [discussion](https://github.com/ccache/ccache/discussions) instead. It's OK to invite people from this issue thread to it if you want.

jrosdahl:
@gjasny: Here is some feedback on the PR:

- It looks like `storeInBackendOnly` isn't needed since that information is available via the config and doesn't need to be stored in the backend instance.
- We should think more about generalizing the configuration settings. For instance, what should happen if another backend is configured as well? There needs to be some defined order in which lookups occur.
- It would be good to have statistics counters for hits and misses for secondary storage.
- Documentation needs to be updated.
- I would like to avoid a dependency on libcurl. The main reason is that it's so competent that it's also very large (when including dependencies) – the increased startup time just by _linking_ (dynamically) with libcurl is around 4-5 ms, which maybe doesn't sound much, but it's is a ≈300% increase (≈1.6 ms vs ≈6.4 ms on my system). Those milliseconds matter since fast cache hits take much less than 6.4 ms. (Linking statically libcurl is not feasible.) I think that it would be very interesting to try out <https://github.com/yhirose/cpp-httplib> as an alternative.

See also [my comment to #414](https://github.com/ccache/ccache/issues/414#issuecomment-855351889).

afbjorklund:
> > If you want it more compatible with those, then I think you need to add CAS support.
> 
> I don't understand why you mention CAS support in this context. Could you expand on why we would need CAS support for the listed cloud storage solutions?

It was in the context of the Bazel cache (as per the link), not in the context of cloud cache.

I think the cloud storage would be fine with anything, simple paths or simple (key) strings.

Let's leave the CAS out of this discussion, and have deduplication as a separate feature ?
Previously I haven't seen any big gains from it, would like some statistics to show it is common*.

* i.e. having two cache entries linking to the same object, especially _after_ preprocessing
I can see how changing a comment would lead to the same object, but that should be handled

jrosdahl:
> It was in the context of the Bazel cache (as per the link), not in the context of cloud cache.

Right. (I thought that you maybe referred to "cloud storage solutions" since you wrote "those" and not "that".)
 
> Agree that it would be nice if the HTTP storage backend was compatible with GCS/S3/etc.

Since cloud storage APIs don't just use simple GET/PUT but also require other things I would say that those are _different_ backends than the simple HTTP backend.

> Let's leave the CAS out of this discussion, and have deduplication as a separate feature ?

Yep!

For the record, I'm marking lots of comments as off topic (including my own) to keep discussion focused on the issue/PR at hand. It's not that I don't appreciate discussion and ideas, but please consider starting discussions under [Discussions](https://github.com/ccache/ccache/discussions) instead and keeping PR comments more focused on the actual implementation. Ideas about alternatives are valuable but I think it's better not to hide them in a comment thread to some PR that will be closed eventually.

gjasny:
I'll give the `cpp-httplib` a try and will check how it affects startup time.

What I like about that lib is that it contains both: client and server support. That will make integration testing much more easy.

gjasny:
I've got something working. Will polish and push tomorrow.

gjasny:
The numbers are in:

```
cmake .. -GNinja -DCMAKE_BUILD_TYPE=Release -DENABLE_IPO=ON -DCMAKE_C_COMPILER=clang-12 -DCMAKE_CXX_COMPILER=clang++-12 -DCMAKE_BUILD_WITH_INSTALL_RPATH=ON -DHTTPCACHE_BACKEND=ON
ninja
llvm-strip-12 ccache

ls -l ccache
time ./ccache --version
```

| Flavour          | Size           | Startup time  |
| ------------- |:-------------:| -----:|
| cpp-httplib      | 3021120 | 0,00s user 0,00s system 92% cpu 0,005 total |
| curl | 2551776      |   0,01s user 0,00s system 96% cpu 0,016 total |
| vanilla | 2533928      |    0,00s user 0,00s system 92% cpu 0,005 total |

jrosdahl:
Thanks for trying out cpp-httplib. Looks promising.

Related to this, I just bumped the required C++ version from 11 to 14 (4eec5d64e3bbf7aecbbf8fc58c9f44563fda187c) which as a side effect means that we can start using `std::regex` (which isn't supported by GCC 4.8).

By the way: Don't know if you've seen it already, but I'm currently working on configuration (#855), API and framework for storage backends (#856). I'll let you know when it's ready for usage by the HTTP backend (#858).

gjasny:
That's good news. Will std::filesystem be available across all operating systems now?

And yes I saw your tickets about config and storage backend. I don't intend to change any of my code until it's ready. Instead I wanted to focus on writing unit and integration tests. A test infrastructure for the (http) backend should hold some value regardless of the future storage interface.

jrosdahl:
> Will std::filesystem be available across all operating systems now?

No, unfortunately not. I'm not ready to require C++17.

> And yes I saw your tickets about config and storage backend. I don't intend to change any of my code until it's ready. Instead I wanted to focus on writing unit and integration tests. A test infrastructure for the (http) backend should hold some value regardless of the future storage interface.

Awesome, thanks! :+1:

jrosdahl:
The API and framework is now ready:

* [SecondaryStorage interface](https://github.com/ccache/ccache/blob/2b3368cfc2a076e3205e0d7f400b4f97db227dae/src/storage/SecondaryStorage.hpp)
* [Example implementation](https://github.com/ccache/ccache/commit/f8789e2f4156cd3413f72c0ce4ae0dc321f70084)

@gjasny: Would you like to work on an HTTP implementation of that interface?

gjasny:
Hi @jrosdahl,

I'll start working on it tomorrow. Thanks for all the preparational work!

-Gregor

gjasny:
I'll add an URL class and tests tomorrow.

gjasny:
Todo:
[ ] Test IPv6 capabilities of http client lib
[ ] Print error string instead of 4: Failed to put /edf0hot67e21geacg88dfjeceoug56vpk to http storage: error code: 4

jrosdahl:
Thanks for working on this!

The title says "WIP" and there are items left on the to do list, so perhaps mark the PR as a draft until ready for review?

Regarding authentication: I would like that to be supported (both via URL and via username/password attributes in the `secondary_storage` string), but that of course can be done as a separate task later.

Another thing that I would like to see (again: can be done later if wanted) is configurable timeouts (see https://github.com/ccache/ccache/issues/859#issuecomment-869208683).

gjasny:
Hello,

I believe this PR is ready for a review. There are some follow-up tasks:
* Configure and enforce HTTP timeouts
* Refactor Storage class to use Url class
* Authentication
* HTTPS support

I'll start working on those during the next days.

Thanks,
Gregor

src/storage/secondary/HttpStorage.cpp
```diff
@@ -0,0 +1,226 @@
...
+  } catch (Url::parse_error& e) {
+    throw Error(e.what());
+  }
+}
+
+std::string
+getUrlPath(const Url& url)
+{
+  // Url class is parsing the URL object lazily
+  // remove this try/catch block once Storage class verifies Url is valid
```
jrosdahl, line 96: Perhaps add the `Url` class and usage of it in a separate PR to be merged before this one?

test/suites/secondary_http.bash
```diff
@@ -0,0 +1,89 @@
+start_http_server() {
+    local HOST="127.0.0.1"
+    local PORT="8080"
+    local URL="http://${HOST}:${PORT}/"
+    export CCACHE_SECONDARY_STORAGE="${URL}"
+
+    trap 'terminate_all_children' EXIT # also cleanup after exceptional code flow
```
jrosdahl, line 7: Good idea to clean up after e.g. ctrl+c! I suggest moving the `trap` statement to `test/run` since there can be only one EXIT handler.

Maybe this also removes the need of a teardown mechanism? `test/run` could call `terminate_all_children` after each suite.

src/storage/secondary/HttpStorage.hpp
```diff
@@ -0,0 +1,57 @@
...
+  nonstd::expected<nonstd::optional<std::string>, Error>
+  get(const Digest& key) override;
+  nonstd::expected<bool, Error> put(const Digest& key,
+                                    const std::string& value,
+                                    bool only_if_missing) override;
+  nonstd::expected<bool, Error> remove(const Digest& key) override;
+
+private:
+  const std::string m_url_path;
+  std::unique_ptr<httplib::Client> m_http_client;
```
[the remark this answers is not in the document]
jrosdahl: Ah, you want to avoid including httplib.h from the include file. That makes sense.

Resolving.

doc/MANUAL.adoc
```diff
@@ -929,6 +930,28 @@ Optional attributes:
 * *update-mtime*: If *true*, update the modification time (mtime) of cache
   entries that are read. The default is *false*.
 
+=== HTTP storage backend
+
+URL format: `http://HOST:PORT/PATH`
+
+This backend stores data in a HTTP/WebDAV compatible server. The required HTTP
```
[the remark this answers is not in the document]
gjasny: You're right. It does not use any WebDAV verbs. I was confused by the necessity to use the nginx dav-plugin for "PUT".

src/storage/secondary/HttpStorage.cpp
```diff
@@ -0,0 +1,226 @@
...
+// FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public License for
+// more details.
+//
+// You should have received a copy of the GNU General Public License along with
+// this program; if not, write to the Free Software Foundation, Inc., 51
+// Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
+
+#ifndef WIN32_LEAN_AND_MEAN
+#  define WIN32_LEAN_AND_MEAN
+#endif
```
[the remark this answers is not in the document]
gjasny: I tried but got weird errors in the stat-code around error macros. I can re-try once this PR is merged.

src/system.hpp
```diff
@@ -91,8 +91,10 @@ const size_t READ_BUFFER_SIZE = 65536;
 #    error _WIN32_WINNT is undefined
 #  endif
 
-#  ifndef __MINGW32__
-typedef int64_t ssize_t;
+#  ifdef _WIN64
+typedef __int64 ssize_t;
+#  else
+typedef int ssize_t;
```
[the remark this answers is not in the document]
gjasny: Maybe it was needed prior to forward-declaring `httplib::Client`. The ccache typedefs conflicted with those of cxx-httplib:
https://github.com/yhirose/cpp-httplib/blob/v0.9.0/httplib.h#L111-L116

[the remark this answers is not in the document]
gjasny: The `HttpStorage.cxx` includes `ccache.hxx` for (`CCACHE_NAME, CCACHE_VERSION`) which includes `system.hpp`.

test/http-server.py
```diff
@@ -0,0 +1,93 @@
...
+    infos = socket.getaddrinfo(
+        *address,
+        type=socket.SOCK_STREAM,
+        flags=socket.AI_PASSIVE,
+    )
+    family, type, proto, canonname, sockaddr = next(iter(infos))
+    return family, sockaddr
+
+def run(HandlerClass, ServerClass, port, bind):
+    # HTTP/1.1 is needed for connection keep-alive but slows down the test by 20s
```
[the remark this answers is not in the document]
gjasny: I need to fire-up wireshark. Maybe some socket-shutdown waits for the keep-alive timeout. Need to check if it's client or server.

test/suites/secondary_http.bash
gjasny: The `TEST` / `reset_environment` removes the cwd from the HTTP server. Therefore I need to restart the server.

src/system.hpp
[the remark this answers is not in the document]
gjasny: I don't know how to proceed here. The "right thing to do":tm: would be to namespace all of the ccache code to not pollute the root namespace with those typedefs. I also tried to get rid of the `ssize_t` altogether by using `auto` but it does not work for all of the usages.

gjasny:
Ready for another review.

src/system.hpp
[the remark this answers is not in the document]
jrosdahl: > `HttpStorage.cxx` includes `ccache.hxx` for (`CCACHE_NAME, CCACHE_VERSION`) which includes `system.hpp`.

I went yak shaving and removed `system.hpp` in ca9ec9cfa8af99eb1e5c39c9dffdc1c611ed235c. I also tried to improve `ssize_t` typedefs in 4dfd7a958a8b203d66f83c3a59cd244ad310994d. Does this improve the situation?

doc/MANUAL.adoc
```diff
@@ -931,6 +932,27 @@ Optional attributes:
 * *update-mtime*: If *true*, update the modification time (mtime) of cache
   entries that are read. The default is *false*.
 
+=== HTTP storage backend
+
+URL format: `http://HOST:PORT/PATH`
```
jrosdahl, line 937: Maybe something like 


```suggestion
URL format: `http://HOST[:PORT][/PATH]`
```

to indicate optional parts?

doc/MANUAL.adoc
```diff
@@ -931,6 +932,27 @@ Optional attributes:
...
+
+This backend stores data in an HTTP compatible server. The required HTTP
+methods are `GET`, `PUT`, and `DELETE`.
```
jrosdahl, line 940:
```suggestion
methods are `GET`, `PUT` and `DELETE`.
```

(Not using Oxford commas elsewhere in the manual.)

doc/MANUAL.adoc
```diff
@@ -931,6 +932,27 @@ Optional attributes:
...
+Note that ccache will not perform any cleanup of the HTTP storage.
+
+Examples:
+
+* `http://localhost:8080/`
+* `http://example.org/cache`
+
+Known issues and limitations:
+
+* The cpp-httplib cannot handle URLs containing IPv6 addresses like: `http://[::1]/`
```
jrosdahl, line 951:
```suggestion
* URLs containing IPv6 addresses like `http://[::1]/` are not supported
```

src/storage/secondary/HttpStorage.cpp
```diff
@@ -0,0 +1,218 @@
...
+int
+get_url_port(const Url& url)
+{
+  if (!url.port().empty()) {
+    size_t idx = 0;
+    const auto port = std::stoi(url.port(), &idx);
+    if (idx != url.port().size()) {
+      throw Error("Cannot convert port to integer: {}", url.port());
+    }
+    return port;
```
jrosdahl, line 81: Did you find and consider using `Util::parse_unsigned`? It would look something like this:

```suggestion
    return Util::parse_unsigned(url.port(), 1, 65535, "port");
```

src/storage/secondary/HttpStorage.cpp
```diff
@@ -0,0 +1,218 @@
...
+get_url_path(const Url& url)
+{
+  auto path = url.path();
+  if (path.empty() || path.back() != '/') {
+    path += '/';
+  }
+  return path;
+}
+
+} // anonymous namespace
```
jrosdahl, line 100: For foolish consistency with other code:

```suggestion
} // namespace
```

(That is the comment that clang-format adds by default.)

src/storage/secondary/HttpStorage.hpp
```diff
@@ -0,0 +1,57 @@
...
+// You should have received a copy of the GNU General Public License along with
+// this program; if not, write to the Free Software Foundation, Inc., 51
+// Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
+
+#pragma once
+
+#include <storage/SecondaryStorage.hpp>
+#include <storage/types.hpp>
+
+#include <third_party/url.hpp>
```
jrosdahl, line 24: Could `Url` be forward-declared instead?

test/suites/secondary_http.bash
```diff
@@ -0,0 +1,87 @@
...
+SUITE_secondary_http_SETUP() {
+    unset CCACHE_NODIRECT
+
+    start_http_server
+
+    generate_code 1 test.c
+}
+
+SUITE_secondary_http_TEARDOWN() {
+    terminate_all_children
```
jrosdahl, line 30: Would it make sense to let `test/run` always call `terminate_all_children` after a test suite has been run?

src/storage/secondary/HttpStorage.cpp
gjasny: Nice!

test/suites/secondary_http.bash
gjasny: Added it behind the `_TEARDOWN` invocation.

gjasny:
Window build looks fine. But after removing my conflicting commit I cannot resolve the conversation, anymore.

LICENSE.adoc
```diff
@@ -674,6 +674,38 @@ The full license text can be found in LGPL-3.0.txt and at
 https://www.gnu.org/licenses/lgpl-3.0.html.
 
 
+src/third_party/httplib.h
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+cpp-httplib - A C++11 single-file header-only cross platform HTTP/HTTPS library.
+Copied from cpp-httplib 0.9.0 downloaded from <https://github.com/yhirose/cpp-httplib>.
+It has the following license text:
```
jrosdahl, line 682:
```suggestion
Copied from commit 6b9ffc8b of https://github.com/yhirose/cpp-httplib[cpp-httplib].
The library has the following license text:
```

doc/MANUAL.adoc
jrosdahl: Did you mean to apply this or did you not agree with my suggestion?

I don't think that we should mention cpp-httplib here since it's an implementation detail that only will confuse the reader.

test/suites/secondary_http.bash
```diff
@@ -0,0 +1,87 @@
...
+}
+
```
jrosdahl, line 32: I guess this is superfluous now that `test/run` runs `terminate_all_children`? Would you like to keep it for symmetry with `start_http_server`?

gjasny: I was under the assumption that each `TEST` invocation also cleans the server working dir. But that's not the case. I'll simplify the logic.

doc/MANUAL.adoc
gjasny: I simply overlooked it. I'm not too used to GitHub PR comments. At work we use BitBucket which handles force-pushes much better.

test/suites/secondary_http.bash
gjasny: I completely removed `TEARDOWN` and by using subdirs, still have the complete state archived on failure

jrosdahl:
Thanks!

PR #677, opened 2020-10-02
Disable some new clang-tidy features until they are fixed

AlexanderLanin:
We have a list of disabled clang-tidy checks. This has some upsides, but also a downside: that means that new checks of clang-tidy which are not yet fixed need to be disabled.

* [cppcoreguidelines-avoid-non-const-global-variables](http://clang.llvm.org/extra/clang-tidy/checks/cppcoreguidelines-avoid-non-const-global-variables.html) has been introduced with clang-tidy-11.
   > I.2: Avoid non-const global variables
   > Non-const global variables hide dependencies and make the dependencies subject to unpredictable changes.
   
   While this rule is extremely reasonable IMHO it will require quite some rework of global variables in ccache.

* [cppcoreguidelines-const-correctness](https://reviews.llvm.org/D54943#change-wOvYzQgFtVgt) will be introduced soonish.
   > ES.25: Declare an object const or constexpr unless you want to modify its value later on
   > That way you can't change the value by mistake. That way might offer the compiler optimization opportunities.
   
   While this rule is also extremely reasonable IMHO, the current codebase is far from const correct, compare #654.


jrosdahl:
Sure. Just some comments on future expectations:

> * [cppcoreguidelines-avoid-non-const-global-variables](http://clang.llvm.org/extra/clang-tidy/checks/cppcoreguidelines-avoid-non-const-global-variables.html) has been introduced with clang-tidy-11.
>   > I.2: Avoid non-const global variables
>   > Non-const global variables hide dependencies and make the dependencies subject to unpredictable changes.
>   
>   
>   While this rule is extremely reasonable IMHO it will require quite some rework of global variables in ccache.

It won't be possible to get rid of all global mutable variables because the signal handler only has access to global state. (Introducing a singleton instead of global variables doesn't count since singletons are global variables in disguise.)

> * [cppcoreguidelines-const-correctness](https://reviews.llvm.org/D54943#change-wOvYzQgFtVgt) will be introduced soonish.
>   > ES.25: Declare an object const or constexpr unless you want to modify its value later on
>   > That way you can't change the value by mistake. That way might offer the compiler optimization opportunities.
>   
>   
>   While this rule is also extremely reasonable IMHO, the current codebase is far from const correct, compare #654.

Please note that "const correct" here means "correct according to CppCoreGuidelines" which ccache doesn't necessarily follows.

PR #679, opened 2020-10-08
Remove -lgcov from compiler independent profiling script

chrisburr:
Fixes a failure seen in `make check` when updating ccache to 3.7.12 in conda-forge (https://github.com/conda-forge/ccache-feedstock/pull/17):

```
Running test suite sanitize_blacklist...
ld: library not found for -lgcov
clang-10: error: linker command failed with exit code 1 (use -v to see invocation)
./test/suites/profiling.bash: line 79: ./test: No such file or directory
error: data/*.profraw: No such file or directory
error: Could not read profile data/default.profdata: No such file or directory
1 error generated.
Running test suite profiling.....
FAILED
```

I didn't manage to figure out why this change was made, but I think the GCC specific code should have been isolated to `profiling_gcc.bash` instead of also being included in `profiling_.bash`. In the conda-forge update this patch has been used with both GCC 7.3 and Clang 10 without issues.

### DIFF
--- a/test/suites/profiling.bash
+++ b/test/suites/profiling.bash
@@ -74,7 +74,7 @@ SUITE_profiling() {
     expect_stat 'cache hit (direct)' 0
     expect_stat 'cache miss' 1
 
-    $COMPILER -fprofile-generate=data test.o -lgcov -o test
+    $COMPILER -fprofile-generate=data test.o -o test
 
     ./test
     merge_profiling_data data
--- a/test/suites/profiling.bash
+++ b/test/suites/profiling.bash
@@ -74,7 +74,7 @@ SUITE_profiling() {
     expect_stat 'cache hit (direct)' 0
     expect_stat 'cache miss' 1
 
-    $COMPILER -fprofile-generate=data test.o -o test
+    $COMPILER -fprofile-generate test.o -o test
 
     ./test
     merge_profiling_data data
--- a/test/suites/profiling_gcc.bash
+++ b/test/suites/profiling_gcc.bash
@@ -17,7 +17,7 @@ SUITE_profiling_gcc() {
     expect_stat 'cache hit (direct)' 0
     expect_stat 'cache miss' 1
 
-    $COMPILER -fprofile-generate test.o -lgcov -o test
+    $COMPILER -fprofile-generate test.o -o test
 
     ./test
 
@@ -44,7 +44,7 @@ SUITE_profiling_gcc() {
     expect_stat 'cache hit (direct)' 0
     expect_stat 'cache miss' 1
 
-    $COMPILER -fprofile-dir=data -fprofile-generate test.o -lgcov -o test
+    $COMPILER -fprofile-generate test.o -o test
 
     ./test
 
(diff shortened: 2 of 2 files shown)


jrosdahl:
The change was made in 3f77bbd2654297900fa15b2d09811c67542893c8 to fix test suite errors when run with GCC 4.4. Since it works for me with Clang 10 I didn't realize that adding `-lgcov` could be problematic. Looks like this might be because Clang detects and uses a GCC installation if available.

Anyway, the proposed change breaks the GCC 4.4 case again. But thinking more about it, it's strange that `-lgcov` is needed for the `-fprofile-generate=dir + -fprofile-use=dir` case but not for the `-fprofile-generate + -fprofile-use` case...

Ah, now I get it. GCC 4.4 seems to be buggy and wants `-fprofile-generate` to the linker, not `-fprofile-generate=dir`. Also, `-fprofile-use` isn't needed so I think this would be a better fix that works for all known cases:

```patch
--- a/test/suites/profiling.bash
+++ b/test/suites/profiling.bash
@@ -74,7 +74,7 @@ SUITE_profiling() {
     expect_stat 'cache hit (direct)' 0
     expect_stat 'cache miss' 1
 
-    $COMPILER -fprofile-generate=data test.o -lgcov -o test
+    $COMPILER -fprofile-generate test.o -o test
 
     ./test
     merge_profiling_data data
diff --git a/test/suites/profiling_gcc.bash b/test/suites/profiling_gcc.bash
index 814212289873..e67b6fa26eaf 100644
--- a/test/suites/profiling_gcc.bash
+++ b/test/suites/profiling_gcc.bash
@@ -17,7 +17,7 @@ SUITE_profiling_gcc() {
     expect_stat 'cache hit (direct)' 0
     expect_stat 'cache miss' 1
 
-    $COMPILER -fprofile-generate test.o -lgcov -o test
+    $COMPILER -fprofile-generate test.o -o test
 
     ./test
 
@@ -44,7 +44,7 @@ SUITE_profiling_gcc() {
     expect_stat 'cache hit (direct)' 0
     expect_stat 'cache miss' 1
 
-    $COMPILER -fprofile-dir=data -fprofile-generate test.o -lgcov -o test
+    $COMPILER -fprofile-generate test.o -o test
 
     ./test
 
@@ -71,7 +71,7 @@ SUITE_profiling_gcc() {
     expect_stat 'cache hit (direct)' 0
     expect_stat 'cache miss' 1
 
-    $COMPILER -fprofile-generate -fprofile-dir=data test.o -lgcov -o test
+    $COMPILER -fprofile-generate test.o -o test
 
     ./test
 
@@ -98,7 +98,7 @@ SUITE_profiling_gcc() {
     expect_stat 'cache hit (direct)' 0
     expect_stat 'cache miss' 1
 
-    $COMPILER -fprofile-dir=data2 -fprofile-generate=data test.o -lgcov -o test
+    $COMPILER -fprofile-generate test.o -o test
 
     ./test
 
```

What do you think?

chrisburr:
Thanks for the quick response @jrosdahl!

That looks reasonable to me and I've tested it for conda-forge and it's passes  (https://github.com/conda-forge/ccache-feedstock/pull/18).

PR #681, opened 2020-10-12
Fix expect_perm to work with SELinux

liljenzin:
Test case CCACHE_UMASK fails when running it in Fedora because /bin/ls
adds a trailing dot to the output. Thus truncate the output to expected
length.


jrosdahl:
Thanks. Looks like this is not specifc to Fedora (I run Fedora at work and haven't stumbled on this) but to SELinux. I'll change the commit message accordingly.

PR #685, opened 2020-10-19
Cache result even without object file

### LINKED ISSUE #683: Cache stderr even without object file
No object file is no reason for not caching.
This happens for example when using clang/clang++/g++ and using the `-fsyntax-only` option.
While some of the benefits of the cache are lost, so are some of the cost.
So it's likely still worthwhile to cache the result.

The current behavior was introduced in https://github.com/ccache/ccache/commit/131111cd58945ab20aea8b75a7d476376956eb55.

I'm not sure about any major use case besides clang itself (and probably gcc) having test suites which run hundreds of files with `-fsyntax-only`. Maybe some build systems test language support that way since it's quicker than actually compiling.
I guess this does not add any complexity to ccache and even the implementation will not touch more than 10 lines of code.

Discussion points:
* Is there any specific reason to not cache this result?
* Should this be configurable?
* Cache only if CCACHE_PREFIX is empty to avoid problems as addressed by https://github.com/ccache/ccache/commit/131111cd58945ab20aea8b75a7d476376956eb55?
* How to deal with nocpp2 as it triggers `warning: argument unused during compilation: '-fsyntax-only' [-Wunused-command-line-argument]`?

AlexanderLanin:
Proof of concept to resolve #683.
However several points (see #683) are currently unclear.

### DIFF
--- a/test/suites/base.bash
+++ b/test/suites/base.bash
@@ -1007,7 +1007,7 @@ EOF
     umask $saved_umask
 
     # -------------------------------------------------------------------------
-    TEST "No object file"
+    TEST "No object file due to bad prefix"
 
     cat <<'EOF' >test_no_obj.c
 int test_no_obj;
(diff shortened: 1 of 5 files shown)


test/suites/base.bash
```diff
@@ -964,22 +964,60 @@ EOF
...
-int test_no_obj;
-EOF
-    cat <<'EOF' >prefix-remove.sh
+    cat <<'EOF' >prefix-noop.sh
 #!/bin/sh
-"$@"
-[ x$2 = x-fcolor-diagnostics ] && shift
-[ x$2 = x-fdiagnostics-color ] && shift
-[ x$2 = x-std=gnu99 ] && shift
-[ x$3 = x-o ] && rm $4
```
AlexanderLanin, line 978: Not sure what these lines do, but I had to remove them to support gcc 4.8.5

jrosdahl: `-std=gnu99` was added in 1be74015e3970d4679beab8b8beb808bde6ff4c6. Some users had set `CC="gcc -std=gnu99"` in order to be able to compile ccache. `-fcolor-diagnostics` and `-fdiagnostics-color` were added in 396df7efdf26a, presumably for similar reasons.

Not running the compiler at all in the prefix program, like you suggest, indeed sounds better.

AlexanderLanin:
Rebased and squashed

test/suites/base.bash
```diff
@@ -1007,22 +1007,59 @@ EOF
...
+{
+  // Trigger warning by having no return statement.
+}
+EOF
+
+    $REAL_COMPILER -Wall -c stderr.c -fsyntax-only 2>reference_stderr.txt
+
+    # GCC does not report any warnings in -fsyntax-only mode.
+    # Note: By limiting this to GCC it's ensured that the file actually does
+    #       produce a warning, at least with Clang.
```
jrosdahl, line 1047: Both GCC and Clang seem to report warnings with `-fsyntax-only` if you put a `#warning` directive in the source file so I suggest doing that instead.

jrosdahl:
Thanks!

PR #690, opened 2020-10-20
CI: list all configurations explicitly

AlexanderLanin:
Currently CI runs with:
Part 1:
ubuntu-16.04-gcc-5.5.0
ubuntu-16.04-clang-9.0.1
ubuntu-18.04-gcc-7.5.0
ubuntu-18.04-clang-9.0.0
ubuntu-20.04-gcc-9.3.0
ubuntu-20.04-clang-10.0.0
macos-10.15-gcc-11.0.3 (or is it clang, apple clang or xcode?)
macos-10.15-clang-11.0.3 (or is it clang, apple clang or xcode?)

Part 2:
ubuntu-16.04-gcc-4.8.5
ubuntu-18.04-i686-w64-mingw32-gcc-posix (GCC) 7.3-posix
ubuntu-18.04-x86_64-w64-mingw32-gcc-posix (GCC) 9.3-posix

(Determined via adding `${{ matrix.config/compiler.CC }} --version` to build.yaml)

This is less then obvious from build-yaml. And while it runs lots of jobs it doesn't cover as much as it could by carefully selecting the exact os - compiler combinations.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -3,15 +3,15 @@ on:
   push:
   pull_request:
 
+env:
+  CMAKE_GENERATOR: Ninja
+  CTEST_OUTPUT_ON_FAILURE: ON
+  VERBOSE: 1
+
 jobs:
   build_and_test:
     name: ${{ matrix.config.os }}-${{ matrix.config.compiler }}-${{ matrix.config.version }}
     runs-on: ${{ matrix.config.os }}
-    env:
-      CMAKE_GENERATOR: Ninja
-      CTEST_OUTPUT_ON_FAILURE: ON
-      VERBOSE: 1
-
     strategy:
       fail-fast: false
       matrix:
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -4,41 +4,130 @@ on:
   pull_request:
 
 jobs:
-  # These test the standard build on several systems with GCC + Clang.
-  standard_tests:
-    name: ${{ matrix.os }} & ${{ matrix.compiler.CC }}
-    runs-on: ${{ matrix.os }}
+  build_and_test:
+    name: ${{ matrix.config.os }}-${{ matrix.config.compiler }}-${{ matrix.config.version }}
+    runs-on: ${{ matrix.config.os }}
+    env:
+      CMAKE_GENERATOR: Ninja
+      CTEST_OUTPUT_ON_FAILURE: ON
+      VERBOSE: 1
+
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-16.04, ubuntu-18.04, ubuntu-20.04, macos-10.15]
-        compiler:
-          - CC: gcc
-            CXX: g++
-          - CC: clang
-            CXX: clang++
+        config:
+          - os: ubuntu-16.04
+            compiler: gcc
+            version: "4.8" # results in 4.8.5
 
+          - os: ubuntu-16.04
+            compiler: gcc
+            version: "5"
+
+          - os: ubuntu-18.04
+            compiler: gcc
+            version: "6"
+
+          - os: ubuntu-18.04
+            compiler: gcc
+            version: "7"
+
+          - os: ubuntu-18.04
+            compiler: gcc
+            version: "8"
+
+          - os: ubuntu-18.04
+            compiler: gcc
+            version: "9"
+
+          - os: ubuntu-20.04
+            compiler: gcc
+            version: "10"
+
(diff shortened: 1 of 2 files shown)


AlexanderLanin:
While this adds ~100 LOC, most of it is a trivial list of compilers and it now supports previously unsupported compilers:
gcc 6 & 8 & 10, clang 3.4 & 3.5 & 5 & 6 & 7 & 8, xcode 11.3

It will be all green after #692 and #693 are merged.

.github/workflows/build.yaml
```diff
@@ -4,41 +4,144 @@ on:
...
-      - name: Install dependencies (Ubuntu 16 & 18)
-        if: startsWith(matrix.os, 'ubuntu') && matrix.os != 'ubuntu-20.04'
-        run: sudo apt-get install elfutils libzstd1-dev
+              # clang-3.4 is available from trusty
+              sudo add-apt-repository "deb http://dk.archive.ubuntu.com/ubuntu/ trusty main"
+              sudo add-apt-repository "deb http://dk.archive.ubuntu.com/ubuntu/ trusty universe"
+  
+              # Disable warnings on ancient compiler.
+              # doctest does not compile with clang-3.4.
+              echo "CMAKE_PARAMS=-DWARNINGS_AS_ERRORS=OFF -DENABLE_TESTING=OFF" >> $GITHUB_ENV
```
jrosdahl, line 106: Should this be removed since it's dead code?

.github/workflows/build.yaml
```diff
@@ -4,41 +4,144 @@ on:
...
+              echo "CC=clang-${{ matrix.config.version }}" >> $GITHUB_ENV
+              echo "CXX=clang++-${{ matrix.config.version }}" >> $GITHUB_ENV
 
-      - name: Install dependencies (Ubuntu 20)
-        if: matrix.os == 'ubuntu-20.04'
-        run: sudo apt-get install elfutils libzstd-dev
+              if [ "${{ matrix.config.version }}" = "3.5" ]; then
+                # Disable warnings on ancient compiler.
+                echo "CMAKE_PARAMS=-DWARNINGS_AS_ERRORS=OFF" >> $GITHUB_ENV
+              fi
```
jrosdahl, line 114: Should this be removed as well or do you plan to reintroduce a test for Clang 3.5?

.github/workflows/build.yaml
```diff
@@ -4,41 +4,144 @@ on:
...
+      - name: Install dependencies (Mac)
+        if: runner.os == 'macOS'
+        run: |
+          brew install ninja
+          if [ "${{ matrix.config.compiler }}" = "gcc" ]; then
+            brew install gcc@${{ matrix.config.version }}
+            echo "CC=gcc-${{ matrix.config.version }}" >> $GITHUB_ENV
+            echo "CXX=g++-${{ matrix.config.version }}" >> $GITHUB_ENV
+          else
+            ls -ls /Applications/
```
jrosdahl, line 130: Debug statement? Maybe remove?

jrosdahl (commented):
Good idea, I like it.

.github/workflows/build.yaml
AlexanderLanin: since #702 this is dead code, yes

.github/workflows/build.yaml
AlexanderLanin: I want to support clang 3.5 - 5 since all it requires is #693, that sounds reasonable.
However I moved this setting to cmake, since it's not specific to CI.

.github/workflows/build.yaml
AlexanderLanin: Found more xcode versions in that ls statement add added them to the job matrix 😄

AlexanderLanin:
new compilers tested:
gcc 6, 8, 10
clang 6, 7, 8
xcode 10.3, 11, 12.2

Next steps (after this PR):
* Add clang 3.5 - 5 in/after #693
* Add clang 3.4 run-time support test (not build time, i.e. compile with newer compiler)

jrosdahl (commented):
Looks good!

Just one question: Is it intentional to remove the `CTEST_OUTPUT_ON_FAILURE` and `VERBOSE` settings for the "specific_tests" and "check_format" jobs? Or a mistake since you added them to the "build_and_test" env section (which, unless I'm mistaken, doesn't affect "specific_tests" and "check_format")?

AlexanderLanin:
Hups, you are right. I was wrongly assuming it would affect everything.
I have to move it up 1-2 levels.

jrosdahl:
Thanks.

Issue #689, opened 2020-10-20
Build failure on OS X 10.8-10.11 due to _mm512_loadu_si512, _mm256_set1_epi8, etc.


ryandesign:
Hi, I'm the maintainer of [ccache in MacPorts](https://ports.macports.org/port/ccache). After updating to 4.0 it became apparent that it doesn't build anymore on some older macOS versions for various reasons. 4.0 does build on 10.13 and up. 3.7.12 and earlier built fine on older macOS versions down to 10.6 or maybe even earlier.

### How to reproduce ###

1. Compile ccache 4.0 on OS X 10.8 thru 10.11

### Actual behavior ###

On 10.10 and 10.11 there are many errors like:

https://build.macports.org/builders/ports-10.10_x86_64-builder/builds/125455/steps/install-port/logs/stdio
https://build.macports.org/builders/ports-10.11_x86_64-builder/builds/123825/steps/install-port/logs/stdio

```
src/third_party/blake3/blake3_avx512.c:18:10: warning: implicit declaration of function '_mm512_loadu_si512' is invalid in C99 [-Wimplicit-function-declaration]
  return _mm512_loadu_si512((const __m512i *)src);
         ^
src/third_party/blake3/blake3_avx512.c:18:10: error: returning 'int' from a function with incompatible result type '__m512i' (vector of 8 'long long' values)
  return _mm512_loadu_si512((const __m512i *)src);
         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

On 10.8 and 10.9 there are many errors like:

https://build.macports.org/builders/ports-10.8_x86_64-builder/builds/27906/steps/install-port/logs/stdio
https://build.macports.org/builders/ports-10.9_x86_64-builder/builds/129032/steps/install-port/logs/stdio

```
src/hashutil.cpp:128:18: warning: unknown attribute 'target' ignored [-Wattributes]
  __attribute__((target("avx2")));
                 ^
src/hashutil.cpp:139:9: error: unknown type name '__m256i'
  const __m256i first = _mm256_set1_epi8('_');
        ^
src/hashutil.cpp:139:25: error: use of undeclared identifier '_mm256_set1_epi8'; did you mean '_mm_set1_epi8'?
  const __m256i first = _mm256_set1_epi8('_');
                        ^~~~~~~~~~~~~~~~
                        _mm_set1_epi8
```

[...]

PR #692, opened 2020-10-20
Blake3: use C version on ancient clang

AlexanderLanin:
The .S file does not compile with clang 3.4 and 3.5.
It seems it's only some optimization and we can as well trivially fall back to the C version.

(Not tested with clang 4, so let's just apply the fallback to anything before 5.0)

This is probably relevant for #689 for the older MacOS versions since they use clang 3.4 by default.

### DIFF
--- a/src/third_party/blake3/CMakeLists.txt
+++ b/src/third_party/blake3/CMakeLists.txt
@@ -2,7 +2,9 @@ add_library(blake3 STATIC blake3.c blake3_dispatch.c blake3_portable.c)
 
 target_link_libraries(blake3 PRIVATE standard_settings)
 
-if(CMAKE_SYSTEM_NAME STREQUAL "Linux" AND CMAKE_SIZEOF_VOID_P EQUAL 8)
+if(CMAKE_SYSTEM_NAME STREQUAL "Linux" AND CMAKE_SIZEOF_VOID_P EQUAL 8
+   AND NOT (CMAKE_CXX_COMPILER_ID STREQUAL "Clang"
+            AND CMAKE_CXX_COMPILER_VERSION VERSION_LESS 5.0))
   set(blake_source_type asm)
   set(blake_suffix "_x86-64_unix.S")
 else()


AlexanderLanin:
@erijo maybe you have some feedback here?

erijo:
If it works I see no problem with it.

ryandesign:
Note that `CMAKE_CXX_COMPILER_ID` will only `STREQUAL "Clang"` if the compiler is open source clang *or* if (the compiler is Xcode clang *and* `CMAKE_POLICY_DEFAULT_CMP0025=OLD`) (which is its default value). If `CMAKE_POLICY_DEFAULT_CMP0025=NEW` then for Xcode clang `CMAKE_CXX_COMPILER_ID` will be `AppleClang`. Xcode clang and open source clang are sufficiently different programs and have different version numbering schemes that, if you want to compare version numbers as you're doing here, you'll want to have separate checks for Clang and AppleClang so you'll want to set `CMAKE_POLICY_DEFAULT_CMP0025=NEW`, if that's not already set by this build system.

ryandesign:
Just learned that the default for CMP0025 will be NEW if cmake_minimum_required specifies 3.0.0 or later, which ccache's CMakeLists.txt does.

But all of this is probably not relevant to macOS after all since the block in question only takes effect on Linux.

erijo:
I also thought that NEW was the default, but this text from the [manual](https://cmake.org/cmake/help/latest/policy/CMP0025.html) makes me question that:
> This policy was introduced in CMake version 3.0. Use the cmake_policy() command to set this policy to OLD or NEW explicitly. Unlike most policies, CMake version 3.19.0-rc1 does not warn by default when this policy is not set and simply uses OLD behavior. See documentation of the CMAKE_POLICY_WARNING_CMP0025 variable to control the warning.

AlexanderLanin:
> The cmake_minimum_required(VERSION) command implicitly invokes the cmake_policy(VERSION) command to specify that the current project code is written for the given range of CMake versions. All policies known to the running version of CMake and introduced in the <min> (or <max>, if specified) version or earlier will be set to use NEW behavior.

ccache has that set to >3 like any other modern cmake setup, so CMP0025 is NEW.


As mentioned by @ryandesign this is about linux only, so my comment regarding #689 was probably wrong. At most it shows where the patch has to be applied for apple.

So I feel this can be merged. It fixes compilation for ancient clang on linux.

jrosdahl:
Thanks.

PR #694, opened 2020-10-21
Require CLONE_NOOWNERCOPY to be defined for clone on mac

### LINKED ISSUE #688: macOS: error: use of undeclared identifier 'CLONE_NOOWNERCOPY'
Hi, I'm the maintainer of [ccache in MacPorts](https://ports.macports.org/port/ccache). After updating to 4.0 it became apparent that it doesn't build anymore on some older macOS versions like 10.12 for various reasons. 4.0 does build on 10.13 and up. 3.7.12 and earlier built fine on older macOS versions down to 10.6 or maybe even earlier.

### How to reproduce ###

1. Compile ccache 4.0 on macOS 10.12

### Actual behavior ###

https://build.macports.org/builders/ports-10.12_x86_64-builder/builds/129053/steps/install-port/logs/stdio

```
src/Util.cpp:259:44: error: use of undeclared identifier 'CLONE_NOOWNERCOPY'
  if (clonefile(src.c_str(), dest.c_str(), CLONE_NOOWNERCOPY) != 0) {
                                           ^
```

### Expected behavior ###

successful compile

### Environment ###

ccache version 4.0

macOS Sierra v10.12.6 (16G2136)
Xcode v9.2 (9C40b)
Apple LLVM version 9.0.0 (clang-900.0.39.2)

erijo:
Fixes #688.

### DIFF
--- a/src/Util.cpp
+++ b/src/Util.cpp
@@ -72,7 +72,9 @@ extern "C" {
 #ifdef __APPLE__
 #  ifdef HAVE_SYS_CLONEFILE_H
 #    include <sys/clonefile.h>
-#    define FILE_CLONING_SUPPORTED 1
+#    ifdef CLONE_NOOWNERCOPY
+#      define FILE_CLONING_SUPPORTED 1
+#    endif
 #  endif
 #endif
 


AlexanderLanin:
How can it be not defined?
It's right there, assuming the header looks something like this:
https://fergofrog.com/code/cbowser/xnu/bsd/sys/clonefile.h.html

erijo:
I assumed that flag wasn't added in 10.12.

AlexanderLanin:
Feels like you are right, although I cannot find any concrete reference. Anyway this does fix 10.12 and 10.12 is past end of support anyway. https://en.wikipedia.org/wiki/MacOS_version_history#Releases

erijo:
https://opensource.apple.com/source/xnu/xnu-3789.21.4/bsd/man/man2/clonefile.2.auto.html does not contain CLONE_NOOWNERCOPY so there has at least been done release without it.

jrosdahl:
Thanks.

PR #696, opened 2020-10-21
Use BLAKE3 machinery for detecting AVX2 support

erijo:
Instead of relying on the sometimes broken __builtin_cpu_support. This hopefully also fixes the build error on Mac OSX 10.11 reported in a comment in #689.

This exposes an internal detail of BLAKE3, but it's hopefully a small change that shouldn't cause problems when upgrading BLAKE3.


AlexanderLanin:
While technically this probably solves all problems and is a better solution I'm not sure ccache should be coupled to blake so tightly. That's for others to decide :-)

jrosdahl:
I agree that it would be preferable to use some more less coupled functionality, but I'm OK with this for now.

jrosdahl:
Thanks.

Issue #567, opened 2020-03-22
Support clang -ivfsoverlay option


manuyavuz-pointr:
Hi,

Is it possible to support `-ivfsoverlay` option for clang?
Option reference: https://clang.llvm.org/docs/ClangCommandLineReference.html#cmdoption-clang-ivfsoverlay-arg

This option is mainly used by Xcode when ObjC/C++ modules are enabled. I was trying out recent `-fmodules` addition (#482), but got `multiple input files` error because `ccache` thinks the value of `-ivfsoverlay` (`all-product-headers.yaml` file which is generated by Xcode) is an input file along with actual source file.

If it's only about updating `compopts` in `compopt.cpp` file, I can submit a PR.

Thanks!

jrosdahl:
> If it's only about updating compopts in compopt.cpp file, I can submit a PR.

If the file content (`all-product-headers.yaml` in your example) does not affect the compilation output, then it's only a matter of adding an entry to `compopt.cpp`, yes.

jrosdahl:
@manuyavuz-pointr: Are you happy with the answer or do you want more information or feedback?

PR #700, opened 2020-10-22
Ignore -ivfsoverlay parameter

steipete:
Ignores `-ivfsoverlay <filepath>` parameter that is used in Xcode with an optional new `ivfsoverlay` sloppiness parameter.

First of, congrats to ccache v4 release, it's amazing! I'm excited that `-fmodules` is supported and this will greatly speed up our compilation. I tried to enable everything and noticed that there's a different error, ccache detects multiple source files as input.

```
[2020-10-22T13:28:08.137445 43404] Hostname: steipete-mbp-i9.local
[2020-10-22T13:28:08.137484 43404] Working directory: /Users/steipete/Projects/PSPDFKit/iOS/PSPDFKit
[2020-10-22T13:28:08.137910 43404] Set profile directory to .
[2020-10-22T13:28:08.137983 43404] /Users/steipete/Builds/PSPDFKit-apwojecgwxmuoieabhwzqzmecixq/Index/DataStore is not a regular file, not considering as input file
[2020-10-22T13:28:08.138633 43404] Multiple input files: /Users/steipete/Builds/PSPDFKit-apwojecgwxmuoieabhwzqzmecixq/Build/Intermediates.noindex/PSPDFKit.build/Debug-iphonesimulator/PSPDFKit.framework.build/all-product-headers.yaml and /Users/steipete/Projects/PSPDFKit/iOS/PSPDFKit/Glyphs/PSPDFTextBlock.m
[2020-10-22T13:28:08.138868 43404] Failed; falling back to running the real compiler
[2020-10-22T13:28:08.139051 43404] Executing /usr/bin/clang -x objective-c -target x86_64-apple-ios12.0-simulator -fmessage-length=0 -fdiagnostics-show-note-include-stack (...) -ivfsoverlay /Users/steipete/Builds/PSPDFKit-apwojecgwxmuoieabhwzqzmecixq/Build/Intermediates.noindex/PSPDFKit.build/Debug-iphonesimulator/PSPDFKit.framework.build/all-product-headers.yaml -MMD -MT dependencies -MF /Users/steipete/Builds/PSPDFKit-apwojecgwxmuoieabhwzqzmecixq/Build/Intermediates.noindex/PSPDFKit.build/Debug-iphonesimulator/PSPDFKit.framework.build/Objects-normal/x86_64/PSPDFTextBlock.d --serialize-diagnostics /Users/steipete/Builds/PSPDFKit-apwojecgwxmuoieabhwzqzmecixq/Build/Intermediates.noindex/PSPDFKit.build/Debug-iphonesimulator/PSPDFKit.framework.build/Objects-normal/x86_64/PSPDFTextBlock.dia -c /Users/steipete/Projects/PSPDFKit/iOS/PSPDFKit/Glyphs/PSPDFTextBlock.m -o /Users/steipete/Builds/PSPDFKit-apwojecgwxmuoieabhwzqzmecixq/Build/Intermediates.noindex/PSPDFKit.build/Debug-iphonesimulator/PSPDFKit.framework.build/Objects-normal/x86_64/PSPDFTextBlock.o
[2020-10-22T13:28:08.139143 43404] Result: multiple source files
```

(I removed most parameters that are not relevant for this to improve readability)

The `all-product-headers.yaml` file contains following content:

```
{"case-sensitive":"false","roots":[{"contents":[{"external-contents":"/Users/steipete/Projects/PSPDFKit/iOS/PSPDFKit/Encryption/PSPDFAESCryptoDataProvider.h","name":"PSPDFAESCryptoDataProvider.h","type":"file"},{"external-contents":"/Users/steipete/Projects/PSPDFKit/iOS/PSPDFKit/Encryption/PSPDFAESCryptoDataSink.h","name":"PSPDFAESCryptoDataSink.h","type":"file"},{"exte
```

The ivsoverlay feature seems to be used when combining Objective-C and Swift code, to provide a virtual file system overlay. I'm not 100% certain in what way this affects caching, but ignoring the property works great and and produces correct results with cached content.

This is my first contribution, I read the guideline, but happy to change things or provide more info to help with this project.

### DIFF
--- a/src/argprocessing.cpp
+++ b/src/argprocessing.cpp
@@ -230,6 +230,17 @@ process_arg(Context& ctx,
     return nullopt;
   }
 
+  // Ignore Xcode -ivfsoverlay <path to yaml file> to not detect multiple input files
+  if (args[i] == "-ivfsoverlay") {
+    i++;
+    if (i == args.size()) {
+      log("-ivfsoverlay lacks an argument");
+      return Statistic::bad_compiler_arguments;
+    }
+    state.common_args.push_back(args[i]);
+    return nullopt;
+  }
+
   // Special case for -E.
   if (args[i] == "-E") {
     return Statistic::called_for_preprocessing;
--- a/src/argprocessing.cpp
+++ b/src/argprocessing.cpp
@@ -230,11 +230,18 @@ process_arg(Context& ctx,
     return nullopt;
   }
 
-  // Ignore Xcode -ivfsoverlay <path to yaml file> to not detect multiple input files
+  // Ignore Xcode -ivfsoverlay <path to yaml file>
+  // to not detect multiple input files.
   if (args[i] == "-ivfsoverlay") {
+    if (!(config.sloppiness() & SLOPPY_IVFOVERLAY)) {
+      LOG_RAW(
+        "You have to specify \"ivfsoverlay\" sloppiness when using"
+        " -ivfsoverlay to get hits");
+      return Statistic::bad_compiler_arguments;
+    }
     i++;
     if (i == args.size()) {
-      log("-ivfsoverlay lacks an argument");
+      LOG_RAW("-ivfsoverlay lacks an argument");
       return Statistic::bad_compiler_arguments;
     }
     state.common_args.push_back(args[i]);
--- a/src/Config.cpp
+++ b/src/Config.cpp
@@ -272,7 +272,9 @@ parse_sloppiness(const std::string& value)
       result |= SLOPPY_LOCALE;
     } else if (token == "modules") {
       result |= SLOPPY_MODULES;
-    } // else: ignore unknown value for forward compatibility
+    } else if (token == "ivfsoverlay") {
+      result |= SLOPPY_IVFOVERLAY;
+    }// else: ignore unknown value for forward compatibility
     start = value.find_first_not_of(", ", end);
   }
   return result;
@@ -311,6 +313,9 @@ format_sloppiness(uint32_t sloppiness)
   }
   if (sloppiness & SLOPPY_MODULES) {
     result += "modules, ";
+  }
+  if (sloppiness & SLOPPY_IVFOVERLAY) {
+    result += "ivfsoverlay, ";
   }
   if (!result.empty()) {
     // Strip last ", ".
--- a/src/Config.cpp
+++ b/src/Config.cpp
@@ -274,7 +274,7 @@ parse_sloppiness(const std::string& value)
       result |= SLOPPY_MODULES;
     } else if (token == "ivfsoverlay") {
       result |= SLOPPY_IVFOVERLAY;
-    }// else: ignore unknown value for forward compatibility
+    } // else: ignore unknown value for forward compatibility
     start = value.find_first_not_of(", ", end);
   }
   return result;
--- a/src/ccache.hpp
+++ b/src/ccache.hpp
@@ -51,6 +51,8 @@ const uint32_t SLOPPY_CLANG_INDEX_STORE = 1 << 7;
 const uint32_t SLOPPY_LOCALE = 1 << 8;
 // Allow caching even if -fmodules is used.
 const uint32_t SLOPPY_MODULES = 1 << 9;
+// Ignore virtual file system (VFS) overlay file.
+const uint32_t SLOPPY_IVFOVERLAY = 1 << 10;
 
 using FindExecutableFunction =
   std::function<std::string(const Context& ctx,


AlexanderLanin:
1) Do you think you could add a [unittest](https://github.com/ccache/ccache/blob/master/unittest/test_argprocessing.cpp) and even more importantly a [product test](https://github.com/ccache/ccache/blob/master/test/suites/base.bash#L3)?

2) From the description I cannot imagine that ignoring that flag is a good idea?!
> -ivfsoverlay<arg>
> Overlay the virtual filesystem described by file over the real file system

steipete:
Thanks for replying so quickly! re tests: I can definitely give it a try. 

In our case the file simply contains a list of headers that are imported, I don't see any definition where a different path would be set.

```
{
  "case-sensitive": "false",
  "roots": [
    {
      "contents": [
        {
          "external-contents": "/Users/steipete/Projects/PSPDFKit/iOS/PSPDFKit/Encryption/PSPDFAESCryptoDataProvider.h",
          "name": "PSPDFAESCryptoDataProvider.h",
          "type": "file"
        },
        {
          "external-contents": "/Users/steipete/Projects/PSPDFKit/iOS/PSPDFKit/Encryption/PSPDFAESCryptoDataSink.h",
          "name": "PSPDFAESCryptoDataSink.h",
          "type": "file"
        },
(... many many more ...)
        {
          "external-contents": "/Users/steipete/Projects/PSPDFKit/iOS/PSPDFKit/Categories/UIScreen+PSPDFAdditions.h",
          "name": "UIScreen+PSPDFAdditions.h",
          "type": "file"
        }
      ],
      "name": "/Users/steipete/Builds/PSPDFKit-apwojecgwxmuoieabhwzqzmecixq/Build/Products/Debug-iphonesimulator/PSPDFKit.framework/PrivateHeaders",
      "type": "directory"
    }
  ],
  "version": 0
}
```

I can't find any documentation for this format, but I don't see any overrides, so - at least for our setup - it seems ok to ignore the file.

steipete:
I asked on Twitter, it seems to be used for debug maps: https://twitter.com/pxpgraphics/status/1319293490941943809?s=21

steipete:
We will test this internally a bit to see if it causes any issues. If I don't report back in a week feel free to ping me.

AlexanderLanin:
I do not see any restriction on what it can be used for so far. So my guess is any file could be within that virtual file system, e.g. some headers or even system headers. Feels like ccache should eat ivfsoverlay + given parameter and simply fall back to preprocessed mode.

jrosdahl:
See also #567 and #130.

jrosdahl:
> Feels like ccache should eat ivfsoverlay + given parameter and simply fall back to preprocessed mode.

That's my gut feeling as well.

steipete:
This article goes a bit into detail as to what VFS overlays are used, it seems to be mainly for Swift importing modules: https://milen.me/writings/swift-module-maps-vfs-overlays-header-maps/

src/argprocessing.cpp
```diff
@@ -235,6 +235,17 @@ process_arg(Context& ctx,
     return nullopt;
   }
 
+  // Ignore Xcode -ivfsoverlay <path to yaml file> to not detect multiple input files
+  if (args[i] == "-ivfsoverlay") {
+    i++;
+    if (i == args.size()) {
+      log("-ivfsoverlay lacks an argument");
```
niw, line 242: Just tested this patch and found that probably this need to be `LOG` instead of `log`.

niw: Or probably `LOG_RAW`.

steipete: Did you get a compile error or a different issue? (Can easily change this, just curious)

We have the current fork in production now and it greatly speeds up our compile times without any adverse effects. (I promised to report back - still want to give this another week or two to see if it really doesn't cause surprises).

I wrote about the process in https://pspdfkit.com/blog/2020/faster-compilation-with-ccache/

niw: Yes, I got an error with `cmake`, like `error: no matching function for call to 'log'`.
Also I rebased the patch on top of current tip branch, that may be related, and probably prefer to do anyways?

NoWiseMan:
Hi, we made very similar changes in our installation. When will this change officially reintegrated?

steipete:
We run this in production now for a few weeks and didn't see any issues. Should I continue with writing test cases for this and try to get this merged?

jrosdahl:
> We run this in production now for a few weeks and didn't see any issues. Should I continue with writing test cases for this and try to get this merged?

Thanks for the follow-up. It's good to hear that it works in your use case to ignore the VFS overlay map.

However, is there any scenario where modifying the VFS overlay map file makes the compiler produce a different output? If we can't answer "no" to that question with very high confidence then ignoring the VFS overlay map by default is simply too risky.

There is an escape hatch, though: Make the behavior optional by introducing a sloppiness setting (e.g. `sloppiness = vfs_overlay`). Or potentially make the existing `modules` sloppiness also cover `-ivfsoverlay`; I don't know enough to tell if `-fmodules` and `-ivfsoverlay` are related enough for this to make sense.

(Regarding earlier comments about falling back to the preprocesor mode instead: that won't be a popular behavior since ccache then won't use the depend mode which is a prerequisite for `-fmodules` support.)

steipete:
@jrosdahl This is a great idea! Let me rework this as a sloppiness setting - I can't *guarantee* that this setting doesn't change output, even if I see it as unlikely, so that's indeed a better fit there.

I'll try to find time to look into this next week and follow up.

niw:
`-ivfsoverlay` is sometimes used for cross compile the source code that is on case-sensitive file system (e.g. the one used on linux,) and case-insensitive file system (e.g. HFS+) and fill the gap between the content in source code (e.g. `#include "meow.h"` but actual file is "Meow.h",) seems typical for handling Windows source code on Linux.
https://github.com/llvm-mirror/llvm/blob/master/cmake/platforms/WinMsvc.cmake may describe this use case.

For the Xcode and iOS and Apple platform development purpose, this is my guess though, HFS+ and APFS, is by default case-insensitive however, can be case-sensitive, therefore, as tool, it makes sense that Xcode gives all header files as "case-insensitive" in this file by default.

In any cases, probably ignoring wouldn't cause any issue on certain situations but not all cases and probably parsing this file and mapping file path for cache purpose would be the best solution.

steipete:
@jrosdahl I've updated this PR to add the new `ivfsoverlay` sloppiness configuration. It works locally. I also rebased to include everything from master. Am I on the right path (just missing tests?)

jrosdahl:
Thanks for the update, I'll have a look.

Yes, tests would be welcome, and the new sloppiness value should be added to the manual as well. :slightly_smiling_face:

jrosdahl:
@steipete: Have you had time to consider my suggestions? Do you want help?

Nezz:
I'd love to see this revived.

steipete:
We still use this and it works, I just struggle with adding all the needed work in to make it mergeable.  Anyone wanna take this over?

niw:
@steipete I want to have this and currently applying patch locally for our use case, I can take it if no one can work on it continuously.

Nezz:
The changes that were requested look pretty minor, but the tricky part could be setting up good tests for this.

steipete:
@niw that'd be amazing! I'm a bit drowning in tasks, if you don't have time I'll finish this eventually, but probably not this April.

niw:
Just posted #839 based on this patch, resolved merge conflicts with tip branch, addressed code review comments on this pull request.

jrosdahl:
Obsoleted by #839.

PR #704, opened 2020-10-25
Disable some AVX features on old Apple clang versions

erijo:
See #689.

### DIFF
--- a/cmake/GenerateConfigurationFile.cmake
+++ b/cmake/GenerateConfigurationFile.cmake
@@ -54,15 +54,16 @@ check_struct_has_member("struct statfs" f_fstypename sys/mount.h
 
 include(CheckCXXCompilerFlag)
 
-if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU"
-    AND CMAKE_CXX_COMPILER_VERSION VERSION_LESS "5.0")
-  # Old GCC versions don't have the required header support.
+# Old GCC versions don't have the required header support.
+# Old Apple Clang versions seems to support -mavx2 but not the target
+# attribute that's used to enable AVX2 for a certain function.
+if((CMAKE_CXX_COMPILER_ID STREQUAL "GNU" AND CMAKE_CXX_COMPILER_VERSION VERSION_LESS 5.0)
+   OR (CMAKE_CXX_COMPILER_ID STREQUAL "AppleClang" AND CMAKE_CXX_COMPILER_VERSION VERSION_LESS 8.0))
   message(STATUS "Detected unsupported compiler for HAVE_AVX2 - disabled")
   set(HAVE_AVX2 FALSE)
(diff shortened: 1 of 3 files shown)


src/third_party/blake3/CMakeLists.txt
```diff
@@ -17,6 +17,16 @@ include(CheckCCompilerFlag)
...
 function(add_source_if_enabled feature compile_flags)
   string(TOUPPER "have_${blake_source_type}_${feature}" have_feature)
+
+  # AVX512 support fails to compile with old Apple Clang versions even though
+  # the compiler accepts the -m flags.
+  if(${feature} STREQUAL "avx512"
+      AND CMAKE_CXX_COMPILER_ID STREQUAL "AppleClang"
+      AND CMAKE_CXX_COMPILER_VERSION VERSION_LESS 9.0)
+    message(STATUS "Detected unsupported compiler for ${have_feature} - disabled")
+    set(${have_feature} FALSE)
```
AlexanderLanin, line 27: Will it be overwritten ~4 lines later? I do not know.

```
  if(NOT DEFINED have_feature)
    #MESSAGE(STATUS "Running automatic feature detection...")
    if(${blake_source_type} STREQUAL "asm")
      check_asm_compiler_flag(${compile_flags} ${have_feature})
    else()
      check_c_compiler_flag(${compile_flags} ${have_feature})
    endif()
  endif()
```

erijo: The two `check_X_compiler_flag` functions doesn't do anything if the variable is already set (so that the test wont be re-run if you run cmake a second time). But I agree that it might not be obvious from the code. I'll change the code slightly.

AlexanderLanin: Ah ok, thanks. Maybe a comment will suffice. At least I did now know this about this behavior, so other will also not know it.

erijo: Hopefully better now.

cmake/GenerateConfigurationFile.cmake
```diff
@@ -54,16 +54,17 @@ check_struct_has_member("struct statfs" f_fstypename sys/mount.h
 
 include(CheckCXXCompilerFlag)
 
-if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU"
-    AND CMAKE_CXX_COMPILER_VERSION VERSION_LESS "5.0")
-  # Old GCC versions don't have the required header support.
+# Old GCC versions don't have the required header support.
+# Old Apple Clang versions seems to support -mavx2 but not the target
```
jrosdahl, line 58:
```suggestion
# Old Apple Clang versions seem to support -mavx2 but not the target
```

jrosdahl:
Thanks.

PR #705, opened 2020-10-25
Build and install man page by default

### LINKED ISSUE #684: man page does not build by default for ccache-4.0
I have `asciidoc` installed in my system. When I build cmake man page does not get built and installed by default:

```
$ git clone https://github.com/ccache/ccache
$ mkdir build
$ cd build
$ cmake ../cmake
$ make
...
$ ls doc/
CMakeFiles  Makefile  cmake_install.cmake

$ make doc-man-page
$ ls doc/
CMakeFiles  Ccache.1  MANUAL.xml  Makefile  cmake_install.cmake
```

A few issues here:
1. `Ccache.1` seems to be in in wrong case. `man ccache` would not find it.
1. Would it make sense to build (and install) `ccache.1` by default if `asciidoc` is detected at cmake time?

Thank you!

erijo:
Fixes #684.

### DIFF
--- a/doc/CMakeLists.txt
+++ b/doc/CMakeLists.txt
@@ -58,7 +58,11 @@ else()
       COMMAND ${A2X_EXE} --doctype manpage --format manpage MANUAL.xml
       MAIN_DEPENDENCY MANUAL.xml
     )
-    add_custom_target(doc-man-page DEPENDS ccache.1)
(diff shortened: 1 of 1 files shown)


AlexanderLanin:
How about a test which would call install and verify the resulting files?

erijo:
I'll leave it to someone more fluent in the ccache test machinery to create such a test.

jrosdahl:
Thanks.

PR #706, opened 2020-10-25
Fall back to version "unknown" when Git is not installed

AlexanderLanin:
I was just looking into clang 3.4 and noticed the following strange behavior: cmake command failed due to missing git.

How to reproduce:
1) change ubuntu-14.04/Dockerfile from clang to clang-3.4
2) `CC=clang CXX=clang++ CMAKE_PARAMS="-DZSTD_FROM_INTERNET=ON" misc/build-in-docker ubuntu-14.04`

It fails since git is not installed into the docker image. Why should it? I'm quite fine with not knowing the git hash within docker.

### DIFF
--- a/cmake/CcacheVersion.cmake
+++ b/cmake/CcacheVersion.cmake
@@ -12,8 +12,7 @@
 #    configuration.
 # 3. Building from a Git repository. In this case the version will be a proper
 #    version if building a tagged commit, otherwise "branch.hash(+dirty)". In
-#    case where git is not available, version will be 'unknown' (e.g. mounting
-#    your source directory into docker without installing git).
+#    case Git is not available, the version will be "unknown".
 
 set(version_info "$Format:%H %D$")
 
--- a/cmake/CcacheVersion.cmake
+++ b/cmake/CcacheVersion.cmake
@@ -11,7 +11,9 @@
 #    version_info has not been substituted). In this case we fail the
 #    configuration.
 # 3. Building from a Git repository. In this case the version will be a proper
-#    version if building a tagged commit, otherwise "branch.hash(+dirty)".
+#    version if building a tagged commit, otherwise "branch.hash(+dirty)". In
+#    case where git is not available, version will be 'unknown' (e.g. mounting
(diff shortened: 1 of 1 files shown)


jrosdahl:
It is by design to require Git, but I don't feel strongly about it so I'm fine with this.

cmake/CcacheVersion.cmake
```diff
@@ -11,7 +11,9 @@
 #    version_info has not been substituted). In this case we fail the
 #    configuration.
 # 3. Building from a Git repository. In this case the version will be a proper
-#    version if building a tagged commit, otherwise "branch.hash(+dirty)".
+#    version if building a tagged commit, otherwise "branch.hash(+dirty)". In
+#    case where git is not available, version will be 'unknown' (e.g. mounting
+#    your source directory into docker without installing git).
```
jrosdahl, line 16:
```suggestion
#    version if building a tagged commit, otherwise "branch.hash(+dirty)". In
#    case Git is not available, the version will be "unknown".
```

cmake/CcacheVersion.cmake
```diff
@@ -30,7 +32,8 @@ elseif(EXISTS "${CMAKE_SOURCE_DIR}/.git")
   # Scenario 3.
   find_package(Git QUIET)
   if(NOT GIT_FOUND)
-    message(SEND_ERROR "Could not find git")
+    set(VERSION "unknown")
+    message(WARNING "Could not find git")
```
jrosdahl, line 36: How about adding `return()` after this line to not execute the following commands with an unset `${GIT_EXCUTABLE}`?

AlexanderLanin: Good point. While it seems to do no harm, I'm actually not sure what happens. Maybe it tries to execute `describe`.

AlexanderLanin: I've moved the rest to 'else' to avoid a hidden return in the middle of that section.

jrosdahl:
Thanks.

PR #708, opened 2020-10-25
Support compilation on Clang 3.4

AlexanderLanin:
While the compiler is rather old it seems trivial enough to support clang 3.4 as it requires only a trivial change.
Reported to doctest at https://github.com/onqtam/doctest/pull/428

Note: it's also possible to add clang-3.4 to GitHub Actions. I'll do that once all the puzzle pieces that are required (#704, #693, #690 etc) are merged.
It's something like:
```
            if [ "${{ matrix.config.version }}" = "3.4" ]; then
              echo "CC=clang" >> $GITHUB_ENV
              echo "CXX=clang++" >> $GITHUB_ENV
              echo "CXXFLAGS=-stdlib=libc++" >> $GITHUB_ENV

              # clang-3.4 is available from trusty
              sudo add-apt-repository "deb http://dk.archive.ubuntu.com/ubuntu/ trusty main"
              sudo add-apt-repository "deb http://dk.archive.ubuntu.com/ubuntu/ trusty universe"
            fi
```

Note: this will need some small adjustment for AppleClang. But I simply want to get clang 3.4 completely working first.

### DIFF
--- a/src/third_party/doctest.h
+++ b/src/third_party/doctest.h
@@ -301,11 +301,15 @@ DOCTEST_MSVC_SUPPRESS_WARNING(26812) // Prefer 'enum class' over 'enum'
 #define DOCTEST_NOINLINE __declspec(noinline)
 #define DOCTEST_UNUSED
 #define DOCTEST_ALIGNMENT(x)
-#else // MSVC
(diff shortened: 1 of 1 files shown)


jrosdahl:
Thanks for getting the fix into upstream doctest. Upgraded in 015bfd80b526dd00708ae6930e65229a8f3c4fea, thus obsoleting this PR.

Issue #703, opened 2020-10-24
Failed to create temporary file - permission denied


ffparra303:
I saw this reply to an issue on failed to open a temporary file. I get the same error shown below. Does anyone know how to solve it? Thanks!

On the other hand, if `$CCACHE_DIR/tmp` is unwritable by the current user, the error message looks similar to this:

`ccache: error: Failed to create temporary file for /home/joel/.ccache/tmp/hashutil.stdout: Permission denied`

I.e., not the same as the one in this bug report.

_Originally posted by @jrosdahl in https://github.com/ccache/ccache/issues/154#issuecomment-462065851_

AlexanderLanin:
Feel free to tell us more about your setup ;-)
OS, ccache version, exact error message you get.

Romain-Geissler-1A:
Hi,

FYI I recently tried to migrate my company builds to ccache 4.0 (we used 3.7.12 before) and indeed several different unrelated teams started to report me such errors.

I am right now not 100% sure this actually comes from ccache 4.0, but time wise, the problem started to happen when I switched it on by default. Our setup is:
 - Linux OSes
 - shared NFS ccache directories. So in our cases, chmod/umask is important.

It apparently happens on one of my ccache folder two, where only 2 users that I trust are using. I can see the following folder/rights:

```
drwxrwxrwx 18 combld root  357 Oct 26 00:42 .
drwxrwxrwt 18 combld root  333 Oct 25 00:01 ..
drwxrwxrwx  2 combld root 3723 Oct 26 00:38 0
drwxrwxrwx  2 combld root 3723 Oct 26 00:39 1
drwxrwxrwx  2 combld root 3473 Oct 26 00:42 2
drwxrwxrwx  2 combld root 3573 Oct 26 00:42 3
drwxrwxrwx  2 combld root 2823 Oct 26 00:41 4
drwxrwxrwx  2 combld root 2723 Oct 26 00:42 5
drwxr-xr-x  2 combld root 2273 Oct 26 00:39 6
drwxr-xr-x  2 combld root 3273 Oct 26 00:39 7
drwxr-xr-x  2 combld root 3723 Oct 26 00:42 8
drwxr-xr-x  2 combld root 3223 Oct 26 00:42 9
drwxr-xr-x  2 combld root 3373 Oct 26 00:38 a
drwxrwxrwx  2 combld root 3473 Oct 26 00:42 b
drwxr-xr-x  2 combld root 4073 Oct 26 00:39 c
[...]

PR #709, opened 2020-10-26
Reproduce #710

AlexanderLanin:
Goal here is to:
* Make sure the problem #710 is properly understood
* Make life easier for someone who can actually fix the problem

This currently fails in:
```
Test case:      CCACHE_UMASK 000
Failure reason: File with unapplied umask detected (linker run)

### DIFF
--- a/test/suites/base.bash
+++ b/test/suites/base.bash
@@ -947,7 +947,7 @@ EOF
 
 
     # -------------------------------------------------------------------------
-    TEST "CCACHE_UMASK"
+    TEST "CCACHE_UMASK 002"
 
     saved_umask=$(umask)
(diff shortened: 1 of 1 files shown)


Romain-Geissler-1A:
@AlexanderLanin Yes you got exactly what I tried to describe in #703. My guess is that things go wrong in my case in the linker run as I think umask is not applied in this case (however I might be wrong).

Note to whoever propose a fix: I can easily test it on:
 - the small example I was mentionning in #703 
 - bigger examples mixing many builds in parallel using several tens of projects from Amadeus

jrosdahl:
Thanks for the repro! :+1:

PR #713, opened 2020-10-28
MinGW: Fix path normalization issues

orgads:
Based on MSYS2 patch with some improvements

For example, if base_dir is D:\mydir and the input is D:/mydir/foo it
wasn't transformed to a relative path.

### DIFF
--- a/src/Util.cpp
+++ b/src/Util.cpp
@@ -875,10 +875,6 @@ make_relative_path(const std::string& base_dir,
                    const std::string& apparent_cwd,
                    nonstd::string_view path)
 {
-  if (base_dir.empty() || !Util::starts_with(path, base_dir)) {
-    return std::string(path);
-  }
-
 #ifdef _WIN32
   std::string winpath;
   if (path.length() >= 3 && path[0] == '/') {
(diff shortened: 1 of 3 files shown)


orgads:
Unit tests still fail on Windows.

jrosdahl:
> Based on MSYS2 patch

For reference, which patch is that?

orgads:
https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-ccache/01-directory-separator.patch

orgads:
Not sure how hard it would be, but ideally it is best to have something like [Utils::FilePath in Qt Creator](https://code.qt.io/cgit/qt-creator/qt-creator.git/tree/src/libs/utils/fileutils.h#n66) instead of passing plain strings around.

The main idea is that it derives string privately, to prevent unintended conversions, and has the following layout:

```c++
class FilePath : private std::string
{
    static FilePath fromString(const std::string &str); // constructing from a normalized string
    static FilePath fromUserInput(const std::string &str); // constructing from user input, normalizing the slashes if needed
    const std::string &toString() const { return *this; }
    const std::string toUserOutput() const; // Translating the output for interaction, logs etc.
};
```

What do you say?

jrosdahl (commented):
Have you looked into the failing tests?

jrosdahl:
> Not sure how hard it would be, but ideally it is best to have something like [Utils::FilePath in Qt Creator](https://code.qt.io/cgit/qt-creator/qt-creator.git/tree/src/libs/utils/fileutils.h#n66) instead of passing plain strings around.
> [...]
> What do you say?

I don't know how feasible this would be. It will have to be driven by somebody else.

jrosdahl:
@orgads: Are you interested in continuing this work?

orgads:
Generally yes, but I can't commit to a specific time.

PR #732, opened 2020-11-26
MinGW: Default to static linking

### LINKED ISSUE #729: Please include necessary LIBs for Win compile again ...
### Description ###

In the newest version 4.1 for win (64-bit), the following DLLs are missing and ccache does not run correctly:

libstdc++6.dll
libwinpthread-1.dll
libgcc_s_seh-1.dll

If I compare the compiled file to version 4.0, I see that the file is now much smaller. So I assume, you forgot to include the LIBs statically.

PS: If I copy those DLL files from "c:\Program Files\Git\mingw64\bin\" to the "ccache directory" it runs again.

orgads:
The release contains only the executable without the dll dependencies.

It is easier to deploy a single self-contained executable anyway.

Fixes #729


jrosdahl:
Thanks.

PR #730, opened 2020-11-26
Unbreak FreeBSD build

khng300:
There are two things to let ccache build on FreeBSD. One is we need to take care of the warnings that is related to llvm's thread safety analysis. Another one is to expose isascii by matching _XOPEN_SOURCE=700 with _POSIX_C_SOURCE=200809L.

PR #733, opened 2020-11-26
Github: Call apt-get update before install

orgads:
Without update, apt tries to download outdated packages, which possibly
don't exist.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -261,7 +261,7 @@ jobs:
 
       - name: Run apt-get
         if: matrix.config.apt_get != ''
-        run: sudo apt-get install ${{ matrix.config.apt_get }}
+        run: sudo apt-get update && sudo apt-get install ${{ matrix.config.apt_get }}
 
       - name: Build and test
         env:
@@ -314,7 +314,7 @@ jobs:
         uses: actions/checkout@v2
 
       - name: Install codespell
-        run: sudo apt-get install codespell
+        run: sudo apt-get update && sudo apt-get install codespell
 
       - name: Run codespell
         run: codespell -q 7 -S ".git,LICENSE.adoc,./src/third_party/*" -I misc/codespell-allowlist.txt


orgads:
@jrosdahl Do you really need *all* versions of gcc? IMHO it should be enough to test the oldest supported version (4.8), and the latest one. What value do you get from testing all of them? Same goes for clang.

AlexanderLanin:
Hi @orgads, is there some concrete problem in CI or is it more the general approach?

gcc/clang: We added tests for many versions because there have been bugs where the oldest version works and the newest version works, but some version in between doesn't. In addition it doesn't slow CI down significantly since GitHub runs so many VMs in parallel.

I was actually going to change this to test multiple compilers on the same system in order to avoid some redundant steps. Maybe even test different systems together via docker. But I never got to it since it just doesn't hurt that much at the moment.

orgads:
Yes, there is a concrete problem. See for example https://github.com/ccache/ccache/pull/730/checks?check_run_id=1457652650

AlexanderLanin:
Found the same solution on:
https://github.community/t/libpulse0-fails-to-fetch-with-apt-repo-broken/18214/5
https://stackoverflow.com/questions/28975075/azure-how-to-fix-apt-get-issues-with-the-azure-ubuntu-archive
https://askubuntu.com/questions/417466/arm-ubuntu-fails-to-get-certain-packages
etc

AlexanderLanin (approved):

jrosdahl:
> Without update, apt tries to download outdated packages, which possibly don't exist.

:+1:

> @jrosdahl Do you really need all versions of gcc? IMHO it should be enough to test the oldest supported version (4.8), and the latest one. What value do you get from testing all of them? Same goes for clang.

It's @AlexanderLanin who has been striving to run the test suite on many versions, and I don't disagree. As he mentioned, the in-between versions actually behave differently in some aspects. I don't understand what running tests on many versions has to do with the package update problem, though.

orgads:
My question about GCC versions was unrelated to this change. Just noticed it when I tried to understand why the CI failed.

PR #476, opened 2019-10-14
Add AVX2 variant of check_for_temporal_macros

erijo:
### Description ###

This implements [SIMD-friendly algorithms for substring searching](http://0x80.pl/articles/simd-strfind.html#generic-sse-avx2) for finding `__DATE__` and `__TIME__` in the input source. At least on my machine (Intel(R) Core(TM) i5-7200U CPU @ 2.50GHz) this is significantly faster.

Here are some numbers (measured using [hyperfine](https://github.com/sharkdp/hyperfine)) when compiling a file with all sources from ccache except hashutil.cpp (and excluding the last commit that makes AVX2 usage controlled by what the CPU supports instead of an environment variable):
```
Benchmark #1: ./ccache g++ -I. -Isrc -I src/third_party -c src/all.cpp
  Time (mean ± σ):      22.3 ms ±   2.0 ms    [User: 16.2 ms, System: 6.0 ms]
  Range (min … max):    18.7 ms …  29.7 ms    150 runs
 
Benchmark #2: USE_AVX=1 ./ccache g++ -I. -Isrc -I src/third_party -c src/all.cpp
  Time (mean ± σ):      19.0 ms ±   0.9 ms    [User: 12.9 ms, System: 6.1 ms]
  Range (min … max):    17.7 ms …  21.4 ms    156 runs
 
Summary
  'USE_AVX=1 ./ccache g++ -I. -Isrc -I src/third_party -c src/all.cpp' ran
    1.17 ± 0.12 times faster than './ccache g++ -I. -Isrc -I src/third_party -c src/all.cpp'
```

PR #735, opened 2020-11-29
Improve simd detection

### LINKED ISSUE #734: Build failure on Apple Silicon arm64 Macs
### How to reproduce ###

1. Build ccache 4.1 with Xcode 12.2 on macOS 11.0.1 on an Apple Silicon (arm64) Mac

### Actual behavior ###

Build failure. [Here is the full log](https://build.macports.org/builders/ports-11_arm64-builder/builds/3355/steps/install-port/logs/stdio). The errors are all of this type:

```
In file included from src/third_party/blake3/blake3_sse41.c:3:
In file included from /Library/Developer/CommandLineTools/usr/lib/clang/12.0.0/include/immintrin.h:14:
/Library/Developer/CommandLineTools/usr/lib/clang/12.0.0/include/mmintrin.h:50:12: error: invalid conversion between vector type '__m64' (vector of 1 'long long' value) and integer type 'int' of different size
    return (__m64)__builtin_ia32_vec_init_v2si(__i, 0);
           ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/Library/Developer/CommandLineTools/usr/lib/clang/12.0.0/include/mmintrin.h:129:12: error: invalid conversion between vector type '__m64' (vector of 1 'long long' value) and integer type 'int' of different size
```
Despite the fact that the error messages appear to refer to files that are part of the development environment, I do not believe that there is a problem with the development environment.

### Expected behavior ###

Successful build.

### Environment ###

ccache version 4.1
macOS 11.0.1
Xcode 12.2
Apple Silicon Mac

erijo:
Improve detection of SSE/AVX by trying to compile a short code snippet with some intrinsic from the feature that is tested for. Fixes #734.

Add support for detecting Neon support on AArch64 and enable the support in BLAKE3.

### DIFF
--- a/cmake/GenerateConfigurationFile.cmake
+++ b/cmake/GenerateConfigurationFile.cmake
@@ -77,7 +77,18 @@ if((CMAKE_CXX_COMPILER_ID STREQUAL "GNU" AND CMAKE_CXX_COMPILER_VERSION VERSION_
   message(STATUS "Detected unsupported compiler for HAVE_AVX2 - disabled")
   set(HAVE_AVX2 FALSE)
 else()
-  check_cxx_compiler_flag(-mavx2 HAVE_AVX2)
+  set(CMAKE_REQUIRED_FLAGS "-mavx2")
+    check_c_source_compiles(
+      [=[
+        #include <immintrin.h>
+        int main()
+        {
+          _mm256_abs_epi8(_mm256_set1_epi32(42));
+          return 0;
+        }
+      ]=]
+      HAVE_AVX2)
+    unset(CMAKE_REQUIRED_FLAGS)
 endif()
 
 list(APPEND CMAKE_REQUIRED_LIBRARIES ws2_32)
--- a/cmake/GenerateConfigurationFile.cmake
+++ b/cmake/GenerateConfigurationFile.cmake
@@ -67,29 +67,19 @@ check_struct_has_member("struct stat" st_mtim sys/stat.h
 check_struct_has_member("struct statfs" f_fstypename sys/mount.h
                         HAVE_STRUCT_STATFS_F_FSTYPENAME)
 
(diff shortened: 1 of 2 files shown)


cmake/GenerateConfigurationFile.cmake
```diff
@@ -67,18 +67,19 @@ check_struct_has_member("struct stat" st_mtim sys/stat.h
...
-  message(STATUS "Detected unsupported compiler for HAVE_AVX2 - disabled")
-  set(HAVE_AVX2 FALSE)
-else()
-  check_cxx_compiler_flag(-mavx2 HAVE_AVX2)
-endif()
+include(CheckCXXSourceCompiles)
+check_cxx_source_compiles(
+  [=[
+    #include <immintrin.h>
+    void func() __attribute__((target("avx2")));
```
gjasny, line 74: Could you please explain why you need that attribute here? Theoretically a compiler could support AVX2 without also supporting the attribute.

erijo: The reason I put the attribute test here is because the attribute is used in [src/hashutil.cpp](https://github.com/ccache/ccache/blob/master/src/hashutil.cpp#L114) and thus the compiler needs to support it to be able to build that file. In #689 there is a report of a compiler that supports -mavx2 but not the target attribute. As we are now checking for the actual support that is needed, I've removed the compiler specific tests that were here before.

The reason that the attribute is used instead of building the whole file with -mavx2 can be seen in [this commit](https://github.com/ccache/ccache/pull/476/commits/5043c1e5273a0e7e5f22b3ad940c16885ba7b2f8). I'm sure there are ways around that if someone really wants to.

jrosdahl:
Thanks.

PR #736, opened 2020-11-30
Make TemporaryFile generate unique file names on Windows

nickhutchinson:
On Windows, multiple ccache process can race each other to create, rename and delete temporary files, because they would attempt to generate the same sequence of temporary file names (`tmp.cpp_stdout.iG2Kb7`, `tmp.cpp_stdout.P1kAlM`, `tmp.cpp_stdout.FzP5tM`, ...). I observed this behaviour in [Process Monitor](https://docs.microsoft.com/en-us/sysinternals/downloads/procmon) when running a multithreaded build of a C++ project.

Root cause appears to be that the offical Windows ccache builds use mingw-w64's [implementation of `mkstemp()`][1], which uses `rand()` to generate temporary file names.

This PR replaces `mkstemp()` with a safer implementation that uses the CRT's `rand_s()` function to generate temporary file names that are more likely to be unique. `rand_s()` uses `RtlGenRandom()` under the hood, which is basically the Windows equivalent of `/dev/urandom`.

See: #703

Example errors:

- Some ccache process is in the process of deleting a temporary file:

      ccache: error: Failed to create temporary file for C:\Users\someuser/.ccache/tmp/tmp.cpp_stdout.FzP5tM: Access is denied.

- Some ccache process has destination file open, so it can't be overwritten:
          
      ccache: error: failed to rename C:\Users\someuser/.ccache/tmp/tmp.cpp_stdout.iG2Kb7 to C:\Users\someuser/.ccache/tmp/tmp.cpp_stdout.iG2Kb7.ii: Access is denied.

- Source file has been deleted by some other ccache process:
      
      ccache: error: failed to rename C:\Users\someuser/.ccache/tmp/tmp.cpp_stdout.P1kAlM to C:\Users\someuser/.ccache/tmp/tmp.cpp_stdout.P1kAlM.ii: The system cannot find the file specified.

[1]: https://github.com/mirror/mingw-w64/blob/v8.0.0/mingw-w64-crt/misc/mkstemp.c

### DIFF
--- a/src/TemporaryFile.cpp
+++ b/src/TemporaryFile.cpp
@@ -20,6 +20,10 @@
 
 #include "Util.hpp"
 
+#ifdef _WIN32
+#  include "third_party/win32/mktemp.h"
+#endif
+
 using nonstd::string_view;
 
 namespace {
@@ -45,7 +49,16 @@ TemporaryFile::TemporaryFile(string_view path_prefix)
   : path(std::string(path_prefix) + ".XXXXXX")
 {
   Util::ensure_dir_exists(Util::dir_name(path));
+#ifdef _WIN32
+  // MSVC lacks mkstemp() and [mingw-w64's implementation][1] is problematic, as
+  // it can reuse the names of recently-deleted files unless the caller
+  // remembers to call srand().
+  // [1]:
+  // https://github.com/Alexpux/mingw-w64/blob/d0d7f784833bbb0b2d279310ddc6afb52fe47a46/mingw-w64-crt/misc/mkstemp.c
+  fd = Fd(bsd_mkstemp(&path[0]));
+#else
   fd = Fd(mkstemp(&path[0]));
+#endif
   if (!fd) {
     throw Fatal(
       "Failed to create temporary file for {}: {}", path, strerror(errno));
--- a/LICENSE.adoc
+++ b/LICENSE.adoc
@@ -638,6 +638,30 @@ The full license text can be found in LGPL-3.0.txt and at
 https://www.gnu.org/licenses/lgpl-3.0.html.
 
 
+src/third_party/win32/mktemp.*
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+This implementation of `mkstemp()` for Win32 was adapted from
+<https://github.com/openbsd/src/blob/99b791d14c0f1858d87a0c33b55880fb9b00be66/lib/libc/stdio/mktemp.c>
+and has the folowing license text:
+
+-------------------------------------------------------------------------------
+Copyright (c) 1996-1998, 2008 Theo de Raadt
+Copyright (c) 1997, 2008-2009 Todd C. Miller
+
+Permission to use, copy, modify, and distribute this software for any
+purpose with or without fee is hereby granted, provided that the above
+copyright notice and this permission notice appear in all copies.
+
+THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF
+MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+ANY SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN AN
+ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+-------------------------------------------------------------------------------
+
 src/third_party/xxh*
 ~~~~~~~~~~~~~~~~~~~~
 
--- a/src/third_party/CMakeLists.txt
+++ b/src/third_party/CMakeLists.txt
@@ -6,6 +6,10 @@ else()
   target_compile_definitions(third_party_lib PUBLIC -DSTATIC_GETOPT)
 endif()
 
+if (WIN32)
+    target_sources(third_party_lib PRIVATE win32/mktemp.c)
+endif ()
+
 if(ENABLE_TRACING)
   target_sources(third_party_lib PRIVATE minitrace.c)
 endif()
--- a/src/third_party/win32/mktemp.c
+++ b/src/third_party/win32/mktemp.c
@@ -1,0 +1,260 @@
+/*	$OpenBSD: mktemp.c,v 1.39 2017/11/28 06:55:49 tb Exp $ */
+/*
+ * Copyright (c) 1996-1998, 2008 Theo de Raadt
+ * Copyright (c) 1997, 2008-2009 Todd C. Miller
+ *
+ * Permission to use, copy, modify, and distribute this software for any
+ * purpose with or without fee is hereby granted, provided that the above
+ * copyright notice and this permission notice appear in all copies.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+ * WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF
+ * MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+ * ANY SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+ * WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN AN
+ * ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+ * OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+ */
+
+#ifdef _WIN32
+#ifndef _WIN32_WINNT
+#define _WIN32_WINNT 0x0600 // _WIN32_WINNT_VISTA
+#endif
+
+#ifndef _CRT_NONSTDC_NO_DEPRECATE
+#define _CRT_NONSTDC_NO_DEPRECATE
+#endif
+
+#ifndef _CRT_SECURE_NO_WARNINGS
+#define _CRT_SECURE_NO_WARNINGS
+#endif
+#endif
+
+#include <sys/types.h>
+#include <sys/stat.h>
+
+#include <ctype.h>
+#include <errno.h>
+#include <fcntl.h>
+#include <limits.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+
+#ifdef _WIN32
+#include <direct.h>
+#include <io.h>
+
+#define WIN32_LEAN_AND_MEAN
+#define NOMINMAX 1
+#define WIN32_NO_STATUS
+#include <windows.h>
+#undef WIN32_NO_STATUS
+#include <ntstatus.h>
+
+// Work-around wrong calling convention for RtlGenRandom in old mingw-w64
+#define SystemFunction036 __stdcall SystemFunction036
+#include <ntsecapi.h>
+#undef SystemFunction036
+#endif
+
+#ifdef _MSC_VER
+#define S_IRUSR	(_S_IREAD)
+#define S_IWUSR	(_S_IWRITE)
+#endif
+
+#define MKTEMP_NAME	0
+#define MKTEMP_FILE	1
+#define MKTEMP_DIR	2
+
+#define TEMPCHARS	"ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789"
+#define NUM_CHARS	(sizeof(TEMPCHARS) - 1)
+#define MIN_X		6
+
+#ifdef _WIN32
+#define MKOTEMP_FLAGS		(_O_APPEND|_O_NOINHERIT|_O_BINARY|_O_TEXT| \
+				 _O_U16TEXT|_O_U8TEXT|_O_WTEXT)
+#define MKTEMP_FLAGS_DEFAULT	(_O_BINARY)
+#else
+#define MKOTEMP_FLAGS		(O_APPEND|O_CLOEXEC|O_DSYNC|O_RSYNC|O_SYNC)
+#define MKTEMP_FLAGS_DEFAULT	(0)
+#endif
+
+#ifndef nitems
+#define nitems(_a)	(sizeof((_a)) / sizeof((_a)[0]))
+#endif
+
+#ifdef _WIN32
+static BOOL CALLBACK
+lookup_ntdll_function_once(
+    PINIT_ONCE init_once, PVOID parameter, PVOID *context)
+{
+	(void)init_once;
+	*context = (PVOID)GetProcAddress(
+	    GetModuleHandleA("ntdll.dll"), parameter);
+	return(TRUE);
+}
+
+static NTSTATUS
+GetLastNtStatus()
+{
+	static INIT_ONCE init_once = INIT_ONCE_STATIC_INIT;
+	typedef NTSTATUS(NTAPI * RtlGetLastNtStatus_t)(void);
+	RtlGetLastNtStatus_t get_last_nt_status = NULL;
+	InitOnceExecuteOnce(&init_once, lookup_ntdll_function_once,
+	    "RtlGetLastNtStatus", (LPVOID *)&get_last_nt_status);
+	return(get_last_nt_status());
+}
+
+static int
+normalize_msvcrt_errno(int ret)
+{
+	if (ret == -1 && errno == EACCES && _doserrno == ERROR_ACCESS_DENIED) {
+		/*
+		 * Win32 APIs return ERROR_ACCESS_DENIED for many distinct
+		 * NTSTATUS codes, even when it's arguably inappropriate to do
+		 * so, e.g. if you attempt to open a directory, or open a file
+		 * that's in the "pending delete" state. These are mapped to
+		 * EACCESS in the C runtime. We instead map these to EEXIST.
+		 */
+		NTSTATUS nt_err = GetLastNtStatus();
+		if (nt_err == STATUS_FILE_IS_A_DIRECTORY ||
+		    nt_err == STATUS_DELETE_PENDING) {
+			errno = EEXIST;
+		}
+	}
+	return(ret);
+}
+
+#define open(...)		(normalize_msvcrt_errno(open(__VA_ARGS__)))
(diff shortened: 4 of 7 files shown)


mumpitzstuff:
This patch does not make sense to me. Random != Unique. It may be that this implementation will reduce the chance to fail but it is not zero. Normally the unique filenname generation depends on systemtime + thread ID which would make much more sense to me. Check this example:

[https://github.com/oetiker/rrdtool-1.x/blob/master/src/mkstemp.c](url)

jrosdahl:
Thanks for working on this.

> On Windows, multiple ccache process can race each other to create, rename and delete temporary files, because they would attempt to generate the same sequence of temporary file names [...] mingw-w64's implementation of `mkstemp()`, which uses `rand()` to generate temporary file names.

It sounds good to reduce the risk of collisions originating from `rand()`. Would it be possible to do that by calling `srand()` with some good (time-based would probably be good enough) value instead?

However, even if `mkstemp` in parallel ccache invocations get the same sequence of values from `rand()`, the `mkstemp` algorithm is supposed to be safe since it only returns when it actually has created a uniquely owned file because of `O_CREAT | O_EXCL`, right? I don't understand the underlying issue, but I'm surprised that there is one. Is the Windows implementation/emulation of `O_CREAT | O_EXCL` not adequate?

How about calling `srand()` and, like @mumpitzstuff suggested, adding thread ID and some time component to the filename to make the underlying issue even less likely to occur? Thread ID was actually part of temporary file name in ccache 3.x which I guess is why this issue has been popping up for ccache 4.x for Windows users.

nickhutchinson:
> This patch does not make sense to me. Random != Unique. It may be that this implementation will reduce the chance to fail but it is not zero. Normally the unique filenname generation depends on systemtime + thread ID which would make much more sense to me. Check this example:
> 
> https://github.com/oetiker/rrdtool-1.x/blob/master/src/mkstemp.c

It's pretty standard to use a PRNG to implement mkstemp.

Examples;
- glibc implement mkstemp with a fast, low-quality PRNG by mixing bits from the current process id with bits from the system time [1] [2] [3]
- BSDs seem to use `arc4random()`, which has better security properties. [4]
- Boost's `unique_path()` unfortunately doesn't return the file handle itself, but does use the platform's secure PRNG (getrandom/arc4random/CryptGenRandom). [5]

I would definitely recommend against the implementation you reference. With the ad-hoc mixing of bits from thread id and GetSystemTime(), you're just creating a low-quality random-number generator. ;) On Windows, thread ids are aggressively reused, and GetSystemTime() gives you millisecond-resolution at best.

IMHO if you need randomness, you should reach for the secure PRNG on your platform unless you have a specific reason to do otherwise, e.g. you're doing expensive simulation and really do need a super-fast PRNG or a reproducible random sequence, or are operating in a constrained context like early in system startup where the secure PRNG may not have gathered enough entropy. On Windows, this means using rand_s/CryptGenRandom/BCryptGenRandom, which all call through to RtlGenRandom under the hood.

[1]: https://github.com/bminor/glibc/blob/glibc-2.32/include/random-bits.h#L25
[2]: https://github.com/bminor/glibc/blob/glibc-2.32/sysdeps/posix/tempname.c#L183
[3]: https://github.com/bminor/glibc/blob/glibc-2.32/misc/mkstemp.c#L31
[4]: https://github.com/openbsd/src/blob/aade1140d2232df1a355e691c0c629962f55c530/lib/libc/stdio/mktemp.c
[5]: https://github.com/boostorg/filesystem/blob/boost-1.74.0/src/unique_path.cpp

nickhutchinson:
> However, even if `mkstemp` in parallel ccache invocations get the same sequence of values from `rand()`, the `mkstemp` algorithm is supposed to be safe since it only returns when it actually has created a uniquely owned file because of `O_CREAT | O_EXCL`, right? I don't understand the underlying issue, but I'm surprised that there is one. Is the Windows implementation/emulation of `O_CREAT | O_EXCL` not adequate?

`O_CREAT | O_EXCL` does get translated to a disposition flag of `CREATE_NEW` in `CreateFile()`, which does the right thing. However that's not enough for ccache. From Process Monitor, I see appears ccache creates a temp file, later renames it with a ".i" or ".ii" suffix, then at process exit time calls `remove()` on not just the the new name but the original name as well. Maybe this was an oversight, or maybe there was an assumption that the spurious `remove()` calls are not a problem. But this assumption breaks if you have a `mkstemp()` implementation that can eagerly return the names of recently-deleted files, as mingw-w64's one does. When this happens you end up with ccache processes deleting each other's temp files, causing errors like those I mentioned.

Even if ccache were to fix the spurious deletes issue, it's still a problem that mingw-w64's `mkstemp()` reuses the names of recently-deleted files. On Windows, when a file has been marked for deletion but there are still open file handles to it, it enters a "pending delete" state where it's visible in the filesystem and yet attempting to do almost anything with it errors with an NT status code of `STATUS_DELETE_PENDING`, which gets surfaced as `ERROR_ACCESS_DENIED`/`EACCES`. mingw-w64's `mkstemp()` gets this wrong and will fail outright if `open()` fails with `EACCES` rather than `EEXIST`.

Calling `srand()` would probably be sufficient for ccache, so long as you chose a good seed (not the thread id -- they get aggressively reused on Windows!) You'll further need to be careful as `rand()`'s state is thread-local on Windows -- you'll need to make sure you call `srand()` at the start of every thread that calls `mkstemp()` or it will start generating the same sequence of file names again.

A cheap and cheerful solution might be to concatenate `GetCurrentProcessId()` with `QueryPerformanceCounter()` to generate a temp file suffix that's probably "unique enough". Is this what ccache 3.x did? You'd end up with files like `tmp.cpp_stdout.8788.1671475151673`.

mumpitzstuff:
Hmm okay if it is common to use random... My expectation was that only threads running at the same time on a multicore system could generate such a tempfile problem and one ccache call can never be done in a millisecond. Therefore one of these values (threadid or systemtime) should always be different.

nickhutchinson:
Replaced rand_s with concatenation of thread id + monotonic clock.

jrosdahl:
@nickhutchinson: Thanks for a good analysis.

> From Process Monitor, I see ccache create a temp file, later renames it with a ".i" or ".ii" suffix, then at process exit time call `remove()` on not just the the new name but the original name as well. Maybe this was an oversight, or maybe there was an assumption that the spurious `remove()` calls are not a problem.

Yes, the assumption is that the redundant `unlink()` calls are not a problem. Ccache makes sure to remove its temporary files even if interrupted by a signal, such as when the user presses Ctrl+c to interrupt a build. To be able to do this, the list of temporary files must be available in a global context since it's traversed by a signal handler. This global list of temporary files is only added to since it's less complex that way. It would of course be possible to also remove entries from the global list at appropriate times, but there will always be a race condition where a redundant `unlink()` call can be made anyway. I don't see the redundant `unlink()` calls as an issue (except in combination with a faulty `mkstemp()`).

>  it's still a problem that mingw-w64's `mkstemp()` reuses the names of recently-deleted files.

Indeed.

Let me try to summarize my current understanding on this:

* The Mingw-w64 implementation of `mkstemp()` is incorrect because it doesn't do the right thing when `EACCES` is returned from `open(..., O_CREAT | O_EXCL)`. Relying on `rand()` is not what makes it incorrect per se, but it makes it much worse, especially without a proper `srand()`.
* Ccache's current usage of `mkstemp()` is correct given a correct implementation of `mkstemp()`.

> A cheap and cheerful solution might be to concatenate `GetCurrentProcessId()` with `QueryPerformanceCounter()` to generate a temp file suffix that's probably "unique enough". Is this what ccache 3.x did? You'd end up with files like `tmp.cpp_stdout.8788.1671475151673`.

There was no time component in temporary file names, only process ID from `getpid()` (ccache is single-threaded except when doing explicit recompression, and then no temporary files are used), plus the unique string added by `mkstemp()`. Since `mkstemp()` generates unique files by itself, I removed the redundant PID part.

Although I would be OK with including a timestamp in the filename, I'm not sure what it would gain. Wouldn't it be enough to get around the bad `mkstemp()` behavior by just including thread/process ID in the filename, maybe in combination with calling `srand(some_granular_enough_time_stamp)` to reduce the risk of having to probe and skip (because of `EEXIST) any leftover files from previous builds?

Oh, I see that you pushed a new PR version now. I wrote most of this before seeing it.

src/TemporaryFile.cpp
```diff
@@ -42,10 +43,23 @@ get_umask()
...
 
 TemporaryFile::TemporaryFile(string_view path_prefix)
-  : path(std::string(path_prefix) + ".XXXXXX")
 {
-  Util::ensure_dir_exists(Util::dir_name(path));
+  Util::ensure_dir_exists(Util::dir_name(path_prefix));
+
+#ifdef _WIN32
+  LARGE_INTEGER timestamp{};
+  QueryPerformanceCounter(&timestamp);
```
jrosdahl, line 51: Can we be sure that the timestamp never is the same for two calls to `TemporaryFile` quickly after each other?

src/TemporaryFile.cpp
```diff
@@ -42,10 +43,23 @@ get_umask()
 } // namespace
 
 TemporaryFile::TemporaryFile(string_view path_prefix)
-  : path(std::string(path_prefix) + ".XXXXXX")
 {
-  Util::ensure_dir_exists(Util::dir_name(path));
+  Util::ensure_dir_exists(Util::dir_name(path_prefix));
+
+#ifdef _WIN32
```
jrosdahl, line 49: I think this special case warrants a comment explaining why `mkstemp()` isn't used.

nickhutchinson:
> Let me try to summarize my current understanding on this:
> 
> * The Mingw-w64 implementation of `mkstemp()` is incorrect because it doesn't do the right thing when `EACCES` is returned from `open(..., O_CREAT | O_EXCL)`. Relying on `rand()` is not what makes it incorrect per se, but it makes it much worse, especially without a proper `srand()`.
> * Ccache's current usage of `mkstemp()` is correct given a correct implementation of `mkstemp()`.

I don't _think_ `mkstemp()` is actually required to generate uniformly-random temp file names, so a conforming implementation could still return the names of recently-deleted files, and trigger bugs in ccache. However all the `mkstemp()` implementations I've seen try to return unpredictable names, which is why you're unlikely to see these issues on Linux/macOS.

mingw-w64's `mkstemp()` might get away with not handling `EACCES` if it didn't use `rand()`.  On Windows, `EACCES` from `open()` (`ERROR_PERMISSION_DENIED` from `CreateFile()`) can mean either 'permission denied', 'path is a directory', or 'file is in the process of being deleted and can no longer be opened'. (It's possibly to distinguish between these cases by calling `RtlGetLastNtStatus()` to get the real error code, though this isn't documented.)

Note that Visual Studio doesn't implement `mkstemp()` at all.

src/TemporaryFile.cpp
nickhutchinson: In theory, this is indeed a problem. In practice, anything that touches the filesystem on Windows is going to take a few dozen microseconds at least, so you'll get a different timestamp from each `QueryPerformanceCounter(); open()/CreateFile()` call.

That said, my preference would still be along the lines of the original PR, i.e. generate a random file name using a good PRNG, or generate a UUID and use that.

jrosdahl:
> I don't _think_ `mkstemp()` is actually required to generate uniformly-random temp file names,

Correct. :slightly_smiling_face: `mkstemp()` is not about returning a random name (i.e., _hoping_ that it is unique), it's about returning a filename that is _guaranteed_ to be unique, where unique in this case means that the caller has unique and exclusive access to the file. AFAICT, a conforming implementation could just probe for an unused filename using a sequential counter, but it's of course much better to make use of randomness to avoid the risk of having to probe many filenames before finding a free one. Generating unpredictable names is naturally also not bad for various security-related reasons.

> so a conforming implementation could still return the names of recently-deleted files

Yes, but then it is also guaranteed that the recently deleted file is not owned or used by anybody else.

> and trigger bugs in ccache.

I don't see why that would be the case, given a properly implemented `mkstemp()`.

> However all the `mkstemp()` implementations I've seen try to return unpredictable names, which is why you're unlikely to see these issues on Linux/macOS.

No, these issues are not seen because the implementations probe with `open(..., O_CREAT | O_EXCL)` until success, which guarantees exclusive ownership.

> Note that Visual Studio doesn't implement mkstemp() at all.

Right. I'm talking about POSIX semantics here (and Mingw-w64's emulation of it).

src/TemporaryFile.cpp
jrosdahl: > In theory, this is indeed a problem. [...] That said, my preference would still be along the lines of the original PR, i.e. generate a random file name using a good PRNG, or generate a UUID and use that.

That sounds like hoping, not guaranteeing. I would prefer an implementation that doesn't even have the theoretical problem. How about some of these alternatives?

Alternative A: Add PID to the filename and use `mkstemp()`. That should be enough since if `mkstemp()` returns a recently deleted file, it won't be open due to the PID in the name (the previous process with our PID no longer exists). Bonus: To improve probing efficiency, do something like `srand(timestamp)` at program start.

Alternative B: Your current proposal but retry until `fd >= 0 || (fd != EEXIST && fd != EACCES)` plus some `RtlGetLastNtStatus()` check or similar to distinguish between the `EACCES` cases.

I would prefer alternative A unless you see something wrong with it.

jrosdahl: On the other hand, I guess alternative B would be preferred to keep people who want to build ccache with MSVC happy?

nickhutchinson:
> > so a conforming implementation could still return the names of recently-deleted files
> 
> Yes, but then it is also guaranteed that the recently deleted file is not owned or used by anybody else.
> 
> > and trigger bugs in ccache.
> 
> I don't see why that would be the case, given a properly implemented `mkstemp()`.

I believe ccache is indeed relying on `mkstemp` generating uniformly-random temp files -- the "exclusive ownership" guarantee that `mkstemp()` promises (even mingw-w64's implementation) isn't good enough. One of the failure modes I saw with Process Monitor:

```
ccache process 1     | ccache process 2
---------------------------------------------------
create tmp1          |
rename tmp1->tmp1.ii |
                     | create tmp1
unlink tmp1          |
unlink tmp1.ii       |
                     | rename tmp1->tmp1.ii: FATAL! tmp1 no longer exists
```

src/TemporaryFile.cpp
nickhutchinson: I can code up both and see which you prefer. I understand your reluctance to merge a bunch of Windows-only code given that ccache has historically only supported POSIX platforms. If ccache is intended to be single-threaded forevermore and only support Windows builds via mingw, then calling `srand()` in `main()` should fix the issue and not introduce much Windows-specific code.

But how about a `int Util::unique_fd(char*)` that calls a custom implementation on Windows but `int mkstemp(char*)` on other platforms? IMHO supporting Windows means avoiding a dependency on mingw -- I'd be inclined to distrust mingw-w64 and its portability wrappers after seeing how poorly they implemented `mkstemp()` ;) It looks like ccache relying on `mkstemp` is actually only thing that is preventing it being built under MSVC. GitHub actions supports building with MSVC (or Clang using the MSVC stdlib). Happy to contribute the CI glue to add support for this.

jrosdahl:
> I believe ccache is indeed relying on mkstemp generating uniformly-random temp files [...]

Thank! I had kind of forgotten that the `.cpp_stdout` file indeed is a special case – I didn't pay enough attention to the fact that you were talking about a temporary file with an .i/.ii extension specifically and not temporary files in general. Relying on randomness in `mkstemp()` is not intentional but an unfortunate side effect of the solution in a07f46a3b0f7e452d50690b831bb4c99fbc6650f. I'll see if I fix that.

src/TemporaryFile.cpp
jrosdahl: > But how about a `int Util::unique_fd(char*)` that calls a custom implementation on Windows but `int mkstemp(char*)` on other platforms?

Sure.

> IMHO supporting Windows means avoiding a dependency on mingw -- I'd be inclined to distrust mingw-w64 and its portability wrappers after seeing how poorly they implemented mkstemp() ;)

I have no opinion on this myself, other than that I would like to continue being able to cross-compile with Mingw-w64 and run unit tests in Wine locally since I don't have a Windows machine myself.

> It looks like ccache relying on `mkstemp` is actually only thing that is preventing it being built under MSVC.

That's because there was a drive in #632 to make it build with MSVC (this is much newer than the Windows support via Mingw-w64, and via MSYS2 even earlier) and I removed the fallback to `mktemp()` recently in 899a1025c287c57df25ac638d159fc03f075fd3f.

> GitHub actions supports building with MSVC (or Clang using the MSVC stdlib). Happy to contribute the CI glue to add support for this.

Sure, that would be likely be valuable. It was mentioned in #632 as well but nobody followed up on it.

nickhutchinson:
@jrosdahl I've revised the patch, using a mkstemp implementation based on [FreeBSD](https://github.com/openbsd/src/blob/1556cebc05a5b4c31325db27377428fbd7d21f4c/lib/libc/stdio/mktemp.c#L45), slightly tweaked for C++, with `arc4random_buf()` swapped out for `RtlGenRandom()`, and with a new `Win32Util:: create_file_exclusive()` routine to emulate `open()` with `O_RDWR|O_CREAT|O_EXCL`. Let me know what you think.

jimis:
FYI we get the error "failed to rename [...] The system cannot find the file specified" reproducibly in the [Github Actions Windows build of Qt](https://github.com/qt/qtbase/actions). The latest ccache is installed using `choco`.

* Last build that succeeded: [Oct 19th](https://github.com/qt/qtbase/runs/1275407911?check_suite_focus=true) using ccache v3.7.12
* First build that failed: [Oct 19th](https://github.com/qt/qtbase/runs/1276484485?check_suite_focus=true) using ccache v4.0. It keeps failing from then on.

jrosdahl:
I wrote:

> I'll see if I can fix that.

Done in 4da9f2b474b7ee39cd54ab4261f90936f3c944ec.

@nickhutchinson wrote:

> I've revised the patch, using a mkstemp implementation based on [FreeBSD](https://github.com/openbsd/src/blob/1556cebc05a5b4c31325db27377428fbd7d21f4c/lib/libc/stdio/mktemp.c#L45), slightly tweaked for C++, with `arc4random_buf()` swapped out for `RtlGenRandom()`, and with a new `Win32Util:: create_file_exclusive()` routine to emulate `open()` with `O_RDWR|O_CREAT|O_EXCL`. Let me know what you think.

Looks good to me, except for one thing: it doesn't have [proper attribution](https://github.com/openbsd/src/blob/1556cebc05a5b4c31325db27377428fbd7d21f4c/lib/libc/stdio/mktemp.c#L6-L8). I would prefer either a "new" standalone implementation (not based specifically on existing code) or the modified OpenBSD version placed in `src/third_party` plus an entry in `LICENSE.adoc`.

jrosdahl:
I applied 824e019a4d15f93878571726c03f2527b05a8375 separately to master (77ad1f78a14caa1d4283a3b8df54e78f8d67a431), by the way. Thanks.

nickhutchinson:
> Looks good to me, except for one thing: it doesn't have [proper attribution](https://github.com/openbsd/src/blob/1556cebc05a5b4c31325db27377428fbd7d21f4c/lib/libc/stdio/mktemp.c#L6-L8). I would prefer either a "new" standalone implementation (not based specifically on existing code) or the modified OpenBSD version placed in `src/third_party` plus an entry in `LICENSE.adoc`.

I've reworked the FreeBSD-derived mkstemp replacement and moved it into into third_party.

jrosdahl:
Thanks!

PR #739, opened 2020-12-07
Disable -Wthread-safety-analysis on FreeBSD currently

khng300:
This option is enabled on FreeBSD when -Werror and -Weverything are
enabled. However, it gives false-positive results. As a workaround
it is disabled on FreeBSD as of writing this commit message.

See PR #730: https://github.com/ccache/ccache/pull/730

### DIFF
--- a/cmake/DevModeWarnings.cmake
+++ b/cmake/DevModeWarnings.cmake
@@ -107,6 +107,13 @@ if(CMAKE_CXX_COMPILER_ID STREQUAL "Clang")
     CCACHE_COMPILER_WARNINGS "-Wno-undefined-func-template")
   add_compile_flag_if_supported(
     CCACHE_COMPILER_WARNINGS "-Wno-return-std-move-in-c++11")
+
+  # -Wthread-safety-analysis (on FreeBSD when dev mode is enabled) gives false
+  # positive results currently. See PR #730.
+  if (CMAKE_SYSTEM_NAME STREQUAL "FreeBSD")
+    add_compile_flag_if_supported(
+      CCACHE_COMPILER_WARNINGS "-Wno-thread-safety-analysis")
+  endif()
 elseif(CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
   list(
     APPEND


khng300:
@AlexanderLanin Could you help review this? Thanks.

AlexanderLanin:
This would disable thread safety analysis globally. Which is probably ok.
On the other hand we could mark only the functions in question as NO_THREAD_SAFETY_ANALYSIS. However that seems clang specific and is a little ugly. We could create some CCACHE_NO_THREAD_SAFETY_ANALYSIS into ccache config file maybe.
https://clang.llvm.org/docs/ThreadSafetyAnalysis.html#no-thread-safety-analysis

I've not yet researched what 'people' say about this clang feature.

jrosdahl:
> We could create some CCACHE_NO_THREAD_SAFETY_ANALYSIS into ccache config file maybe.

Sounds reasonable.

cmake/DevModeWarnings.cmake
```diff
@@ -107,6 +107,13 @@ if(CMAKE_CXX_COMPILER_ID STREQUAL "Clang")
     CCACHE_COMPILER_WARNINGS "-Wno-undefined-func-template")
   add_compile_flag_if_supported(
     CCACHE_COMPILER_WARNINGS "-Wno-return-std-move-in-c++11")
+
+  # -Wthread-safety-analysis (on FreeBSD when dev mode is enabled) gives false
+  # positive results currently. See PR #730.
```
jrosdahl, line 112: I'm not sure if it makes sense to apply this just to compensate for the current state in FreeBSD. As you say, maybe just don't build with `-Werror` and/or `-Weverything`?

jrosdahl:
I'm starting to think that we maybe should give in and don't use `-Weverything` plus a bunch of `-Wno-...` but instead enable extra warnings more selectively. What do you think, @AlexanderLanin?

AlexanderLanin:
Yeah... reasonable.
Another idea would be a compromise: it doesn't make sense to have warnings from Weverything show up as errors. Even in developer builds. We "constantly" run in new compiler versions which support yet another warning. It's nice to stumble over it, but it's not nice to be blocked by it.

cmake/DevModeWarnings.cmake
khng300: Disabling -Weverything seems like a big sacrifice. Removing -Werror sounds better to me.

jrosdahl: I don't want to remove `-Werror` from the developer builds simply because I like the developer to notice good warnings right away.

From https://clang.llvm.org/docs/UsersManual.html#diagnostics-enable-everything: _"Since -Weverything enables every diagnostic, we generally don’t recommend using it. -Wall -Wextra are a better choice for most projects. Using -Weverything means that updating your compiler is more difficult because you’re exposed to experimental diagnostics which might be of lower quality than the default ones."_ [Others agree](https://quuxplusone.github.io/blog/2018/12/06/dont-use-weverything/).

jrosdahl:
Thanks for the PR but I've chosen to remove `-Weverything` instead, at least for now. We can maybe re-add it again as an opt-in configuration option (similar to how it originally was when @afbjorklund added it in fc89a7a86d56c21208414280ebf78008687d1ab6) if wanted, but it's almost as easy to just hand edit the CMake scripts yourself if one wants to try it out.

PR #745, opened 2020-12-11
Adapt used XCode versions to GitHub changes

AlexanderLanin:
Apparently GitHub has changes available XCode versions. 11.0.0 is no longer available. Surely we need a better concept. This is a quick fix.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -82,11 +82,11 @@ jobs:
 
           - os: macOS-latest
             compiler: xcode
-            version: "11.0.0"
+            version: "11.7"
 
           - os: macOS-latest
             compiler: xcode
-            version: "12.2"
+            version: "12.3"
     steps:
       - name: Install dependencies
         run: |


AlexanderLanin:
I'm confused. This shows master to already have XCode 11.7
https://github.com/ccache/ccache/blob/52d4ada236823033a4e1aaf84a9648065fe0eab2/.github/workflows/build.yaml#L79

However in the diff here it's clearly visible that it changes from 11.0.0

jrosdahl:
Looks like your commit is based on master as of 5 weeks ago.

AlexanderLanin:
It seems `AlexanderLanin wants to merge 1 commit into ccache:master from AlexanderLanin:xcode-versions` does not imply that that is the diff that is being shown. Which is strange. Looks like it's a diff of `AlexanderLanin:xcode-versions` to `AlexanderLanin:master` which doesn't make any sense to show.

AlexanderLanin:
Confirmed: The diff view shows a diff of `AlexanderLanin:xcode-versions` and the common ancestor of `ccache:master` and `AlexanderLanin:xcode-versions`. Otherwise it would show all changes on `ccache:master` which happened since I created my branch.
I did not know that. But it's kind of ok.
However what is really bad is that this is so unobvious and that that common ancestor is not visible anywhere here in the ui.

Anyway it seems the actual issue has been resolved weeks ago, we can close this.

jrosdahl:
I agree that is it not obvious.

PR #746, opened 2020-12-11
Add CodeQL

AlexanderLanin:
While I'm not 100% sure this makes a lot of sense, I have it up and running so it makes sense to at least create a pull request.
The only downside I see so far is that it runs for ~7 minutes which makes it the slowest Actions Job we have. I'm not really sure this will ever catch anything since ccache is not really that user facing and security relevant. On the other hand these days everything is security relevant and it might be a low price to pay if this ever actually catches anything.

Background info: https://github.blog/2020-09-30-code-scanning-is-now-available/

### DIFF
--- a/.github/workflows/codeql-analysis.yaml
+++ b/.github/workflows/codeql-analysis.yaml
@@ -1,0 +1,49 @@
+# More info:
+# https://docs.github.com/en/github/finding-security-vulnerabilities-and-errors-in-your-code/configuring-code-scanning
+
+name: "CodeQL"
+
+on:
+  push:
+    branches: [*]
+  pull_request:
+    # The branches below must be a subset of the branches above
+    branches: [*]
+    paths-ignore:
+      - '**/*.adoc'
+      - '**/*.bash'
+      - '**/*.md'
+  schedule:
+    # Full scan once a week
+    - cron: '0 14 * * 3'
+
+jobs:
+  analyze:
+    name: Analyze
+    runs-on: ubuntu-18.04
+
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v2
+      with:
+        # We must fetch at least the immediate parents so that if this is
+        # a pull request then we can checkout the head.
+        fetch-depth: 2
+
+    - name: Install dependencies
+      run: sudo apt-get update && sudo apt-get install ninja-build elfutils libzstd1-dev
+
+    - name: Initialize CodeQL
+      uses: github/codeql-action/init@v1
+      with:
+        languages: cpp
+        queries: +security-and-quality
+
+    - name: Build
+      run: ci/build
+      env:
+        RUN_TESTS: none
+        CMAKE_GENERATOR: Ninja
+
+    - name: Perform CodeQL Analysis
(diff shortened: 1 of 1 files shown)


jrosdahl:
Sounds OK to me.

Regarding slowness: It looks like the Autobuild step builds the code in serial. Perhaps it would be possible to speed up at least that part by specifying custom build commands for building in parallel?

AlexanderLanin:
Oh right, half the time is for autobuild. I'll fix that. Then it should be 2-3 minutes faster.

AlexanderLanin:
Unfortunately it's only ~15 seconds faster. Probably autobuild was already executing in parallel.
Anyway by using our own script we know better what is happening here.
And while ~15 sec is not much, it's not nothing.

.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,49 @@
+# More info:
```
jrosdahl, line 1: I suggest naming the file with a proper `.yaml` extension. (Yes, I'm tilting at windmills. :slightly_smiling_face:)

.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,49 @@
...
+# https://docs.github.com/en/github/finding-security-vulnerabilities-and-errors-in-your-code/configuring-code-scanning
+
+name: "CodeQL"
+
+on:
+  push:
+    branches: [master]
+  pull_request:
+    # The branches below must be a subset of the branches above
+    branches: [master]
```
jrosdahl, line 11: Would it be possible (with no disadvantages) to run for pull requests to all branches? It would be nice to not have to remember to add future release branches to this list.

.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,49 @@
...
+    paths-ignore:
+      - '**/*.bash'
+      - '**/*.md'
+      - '**/*.adoc'
```
jrosdahl, line 15:
```suggestion
      - '**/*.adoc'
      - '**/*.bash'
      - '**/*.md'
```

.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,49 @@
...
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v2
+      with:
+        # We must fetch at least the immediate parents so that if this is
+        # a pull request then we can checkout the head.
+        fetch-depth: 2
+
+    - name: Install dependencies
+      run: sudo apt-get install ninja-build elfutils libzstd1-dev
```
jrosdahl, line 34:
```suggestion
      run: sudo apt-get update && sudo apt-get install ninja-build elfutils libzstd1-dev
```

.github/workflows/codeql-analysis.yml
AlexanderLanin: We can certainly make it run on all branches.

AlexanderLanin:
Squashed and rebased.

jrosdahl:
Thanks.

PR #750, opened 2020-12-20
Add option to link statically with zstd from system

khng300:
ZSTD_SYSTEM_STATIC is added as an CMake option to let libzstd from system being linked statically.

The inspiration was to allow ccache to be linked with a static library version of zstd to prevent a potential chicken-and-egg problem, since ccache binary depends on libzstd.so automatically if not using downloaded version of libzstd.


gjasny:
The `find_library(ZSTD_LIBRARY zstd)` command does a cache lookup prior to looking for the library. That enables you as an user to pass `cmake ... -DZSTD_LIBRARY=/path/to/your/shiny/libz.a` without any changes to the ccache cmake code.

khng300:
Thanks for the hint! I would close this now.

jrosdahl:
I added @gjasny's hint to `doc/INSTALL.md`.

PR #753, opened 2020-12-25
Split ccache_lib

### LINKED ISSUE #660: Architecture
First of all, sorry about yet another non functional idea.

Status:
I've been fiddling around with ccache for a while and mostly I still do not have much of an idea how the source code behaves in itself.
That's mostly because there is no high level concept available which would catch me up (or I missed it).

Goal:
Make contributions to ccache simpler (more contributors + better product) by directing people better to the exact location where they need to change something.

Proposal:
Currently it goes from "ccache/src" to "here are 40 files". On that scale I would however expect some structure to be encoded either on purpose or it emerged on its own. Therefore in this issue I want to propose creating subdirectories under src. I do not know what they would look like, that's for the PR to figure out. There is probably some basic/common part, some part about compression, argument parsing, cache handling etc.

This is **not** about:
* changing code or making it more modular or whatever.
* creating a dozen subdirectories and struggling to fit every single file somewhere.

The idea is simply that beginners can more quickly find the part they want to affect in some way.

AlexanderLanin:
While many features and bugs are open and they should have high priority in fixing, it feels like architecture is being ignored in the process as described before, therefore this closes #660.

This will facilitate many other tasks like better interfaces and abstractions.
Hint: review commit by commit.

@jrosdahl: can you provide early feedback whether I should continue? After first attempts failed miserably it's actually quite trivial now doing this super slowly with 1 file = 1 commit. I guess review is trivial as well.


AlexanderLanin:
Maybe this is the wrong approach... an alternative would be to leave the current lib as is (or move everything to "core") and create new libs like compression, argparsing etc.

As visible in the "[move exceptions](https://github.com/ccache/ccache/pull/753/commits/087005b0023c03219fcdf1cd75a6daa8f02e6ab0)" commit, core is not only utility classes, but classes which have significant parts of ccache logic itself. Although on second review maybe I moved to much since only Statistic and not Statistics is required by exceptions. But even Statistic cannot be argued as util-like classes.

jrosdahl:
> therefore this closes #660.

Just a note: As you might have noticed, I have moved all discussions and support issues to [Discussions](https://github.com/ccache/ccache/discussions). Please discuss ccache development there (under the _Development_ category).

> @jrosdahl: can you provide early feedback whether I should continue?

Sorry about the delayed feedback. I'm still slowly working through my backlog from before Christmas.

Some quick comments:

* Documenting and/or improving source code structure (which is what I assume you mean by "architecture" in this case): Sure.
* Grouping related source code files into subdirectories: Sure, when it makes sense. Grouping compression-related files makes sense. Grouping the various utility functionality would also make sense, as would future backend and frontend abstractions.
* Splitting ccache_lib into several internal libraries: I don't see the benefit.
* I would appreciate a written down vision or plan before starting to move code around. It's hard to understand where small steps aim to lead otherwise.

AlexanderLanin:
I was performing the small steps to see what is possible and where it might lead. Chicken-egg-problem ;-)
I'll try this again some time with a more vision based approach as I was stuck in the details with this one.

PR #755, opened 2020-12-27
Ignore SOURCE_DATE_EPOCH under time_macros sloppiness

azat:
SOURCE_DATE_EPOCH will be passed from debhelpers, by extracting last
entry from d/changelog (or current time if there is entries)
And this will not allow to use cache.

### DIFF
--- a/src/ccache.cpp
+++ b/src/ccache.cpp
@@ -1272,17 +1272,24 @@ hash_common_info(const Context& ctx,
   hash.hash(Util::base_name(args[0]));
 
   // Hash variables that may affect the compilation.
-  const char* always_hash_env_vars[] = {
+  struct
+  {
+    const char* name;
+    uint32_t sloppiness;
+  } hash_env_vars[] = {
     // From <https://gcc.gnu.org/onlinedocs/gcc/Environment-Variables.html>:
-    "COMPILER_PATH",
-    "GCC_COMPARE_DEBUG",
-    "GCC_EXEC_PREFIX",
-    "SOURCE_DATE_EPOCH",
+    {"COMPILER_PATH", 0},
+    {"GCC_COMPARE_DEBUG", 0},
+    {"GCC_EXEC_PREFIX", 0},
+    {"SOURCE_DATE_EPOCH", SLOPPY_TIME_MACROS},
   };
-  for (const char* name : always_hash_env_vars) {
-    const char* value = getenv(name);
+  for (const auto& env : hash_env_vars) {
+    if (env.sloppiness && (ctx.config.sloppiness() & env.sloppiness)) {
+      continue;
+    }
+    const char* value = getenv(env.name);
     if (value) {
-      hash.hash_delimiter(name);
+      hash.hash_delimiter(env.name);
       hash.hash(value);
     }
   }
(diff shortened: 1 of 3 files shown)


src/ccache.cpp
```diff
@@ -1272,17 +1272,24 @@ hash_common_info(const Context& ctx,
...
+    {"COMPILER_PATH", 0},
+    {"GCC_COMPARE_DEBUG", 0},
+    {"GCC_EXEC_PREFIX", 0},
+    {"SOURCE_DATE_EPOCH", SLOPPY_SOURCE_DATE_EPOCH},
   };
-  for (const char* name : always_hash_env_vars) {
-    const char* value = getenv(name);
+  for (const auto& env : hash_env_vars) {
+    if (env.sloppiness && (ctx.config.sloppiness() & env.sloppiness)) {
+      continue;
```
AlexanderLanin, line 1288: Why not cache it always? The others are also always cached.
This invalidates otherwise valid cache entries if sloppiness is added after the first run?!

AlexanderLanin (changes requested):
Please also add this feature to `docs/MANUAL.adoc` to the sloppiness section.

src/ccache.cpp
azat: >Why not cache it always? The others are also always cached.

debhelpers passed it unconditionally, using the last entry from debian/changelog, or current time if there is no entries.

IIUC it is not set if there is already such env variable, however, it does not sounds right to set this value to 0 (or similar), since this will affect `__TIME__`/`__DATE__`, while ccache will not cache those translation units anyway (if `time_macros` sloppiness is not set)

>This invalidates otherwise valid cache entries if sloppiness is added after the first run?!

Yes, like other sloppiness, i.e. `locale`

jrosdahl:
Thanks for the pull request. It sounds good to make it possible to ignore SOURCE_DATE_EPOCH.

Since `SOURCE_DATE_EPOCH` only affects the expansion of time macros (`__DATE__` and `__TIME__`), have you considered extending the `time_macros` sloppiness value for this instead of adding a new setting? I think that it would make sense to do that since what you really are being sloppy about when ignoring `SOURCE_DATE_EPOCH` _is_ the expansion of those macros. I can't think of a situation where you would like to ignore `SOURCE_DATE_EPOCH` but not be sloppy about time macros. What do you think?

azat:
>Since SOURCE_DATE_EPOCH only affects the expansion of time macros (__DATE__ and __TIME__), have you considered extending the time_macros sloppiness value for this instead of adding a new setting? I think that it would make sense to do that since what you really are being sloppy about when ignoring SOURCE_DATE_EPOCH is the expansion of those macros. I can't think of a situation where you would like to ignore SOURCE_DATE_EPOCH but not be sloppy about time macros. What do you think?

Indeed, I though about this, but decided to add a new one, for flexibility, but extending `time_macros` looks better, will update the patch.

azat:
Rebased and updated the subject and description

jrosdahl:
Note: I refined the solution in cad2416291c042443cf0c045047c34a2e07e103a. There will be no need to set sloppiness at all in normal cases regardless of whether `SOURCE_DATE_EPOCH` is set.

PR #756, opened 2020-12-27
Fix running tests on macOS

nickhutchinson:
Bash tests were not actually being run on the macOS CI agents because the version of sed installed there does not understand the `-r` flag:

    sed: illegal option -- r
    usage: sed script [-Ealn] [-i extension] [file ...]
           sed [-Ealn] [-i extension] [-e script] ... [-f script_file] ... [file ...]

The PR:
- fixes scripts to use `sed -E` instead of `sed -r` as the latter isn't supported by BSD sed.

- exports `SDKROOT` in `test/run`. Otherwise it appears at least some some Apple toolchains (e.g. Xcode 10.3) will pick the _latest_ SDK installed on the host (e.g. `/Library/Developer/CommandLineTools/SDKs/MacOSX10.15.sdk`) instead of using the SDK bundled with the toolchain (e.g. `/Applications/Xcode_10.3.app/.../MacOSX10.14.sdk`).

  The 10.15 SDK is not compatible with Xcode 10.3:

      ld: unsupported tapi file type '!tapi-tbd' in YAML file '/Library/Developer/CommandLineTools/SDKs/MacOSX10.15.sdk/usr/lib/libSystem.tbd' for architecture x86_64
      clang: error: linker command failed with exit code 1

### DIFF
--- a/misc/format-files
+++ b/misc/format-files
@@ -46,7 +46,7 @@ for file in "$@"; do
         echo "Error: $file not formatted with Clang-Format"
         echo 'Run "make format" or apply this diff:'
         git diff $cf_color --no-index "$file" "$tmp_file" \
-            | sed -r -e "s!^---.*!--- a/$file!" \
+            | sed -E -e "s!^---.*!--- a/$file!" \
                      -e "s!^\+\+\+.*!+++ b/$file!" \
                      -e "/diff --/d" -e "/index /d" \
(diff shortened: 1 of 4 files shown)


AlexanderLanin (approved):

jrosdahl:
Thanks.

PR #757, opened 2020-12-28
CI: Add VS2019 build jobs

nickhutchinson:
Add VS2019 build jobs that use clang for the test suite. There are many test failures on Windows, but these are ignored for now.

Tweak CMake build scripts:
- Fix CI build type handling for MSVC (recognise `/NDEBUG` and not just `-DNDEBUG`)
- Fix incorrect warnings-as-errors flag for MSVC
- Suppress an additional conversion warning on MSVC


.github/workflows/build.yaml
```diff
@@ -258,13 +286,77 @@ jobs:
...
+          cmd /c "`"${vcvarsall}`" ${{ matrix.config.msvc_arch }} & set" | foreach {
+            if ($_ -match '^\w+=') {
+              $k, $v = $_.split('=', 2)
+              Set-Item -Force -Path "env:\${k}" -Value $v
+            }
+          }
+
+          # The test suite currently requires that the compiler specified by the "CC"
+          # env variable is on a path without spaces. Provide that by creating a
+          # junction from ~/opt/llvm to the Visual Studio path.
```
AlexanderLanin, line 317: Sounds like that should be fixed instead?

.github/workflows/build.yaml
```diff
@@ -258,13 +286,77 @@ jobs:
...
+          # Determine VS2019 install path
+          $vsInstallPath = (Get-VSSetupInstance | `
+            Select-VSSetupInstance -Product '*' -Latest -Version '[16.0,17.0)' `
+          ).InstallationPath
+
+          $vcvarsall = "${vsInstallPath}\VC\Auxiliary\Build\vcvarsall.bat"
+          if (-not (Test-Path -Path $vcvarsall)) {
+            Out-Error "Could not find Visual Studio setup script"
+            exit 1
+          }
```
AlexanderLanin, line 304: To my knowledge this is unnecessary complexity. It's simply `C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\VC\Auxiliary\Build\vcvarsall.bat`.

.github/workflows/build.yaml
```diff
@@ -258,13 +286,77 @@ jobs:
...
+
+          # Run VS environment script
+          $env:VSCMD_SKIP_SENDTELEMETRY = '1'
+          cmd /c "`"${vcvarsall}`" ${{ matrix.config.msvc_arch }} & set" | foreach {
+            if ($_ -match '^\w+=') {
+              $k, $v = $_.split('=', 2)
+              Set-Item -Force -Path "env:\${k}" -Value $v
+            }
+          }
```
AlexanderLanin, line 313: What does this do? I guess the comment is wrong?
This seems to do more than just running the script.

.github/workflows/build.yaml
```diff
@@ -258,13 +286,77 @@ jobs:
...
+            "Path=${env:Path}",
+            "VCINSTALLDIR=${env:VCINSTALLDIR}",
+            "VCToolsInstallDir=${env:VCToolsInstallDir}",
+            "VSINSTALLDIR=${env:VSINSTALLDIR}"
+          )
+          $Utf8NoBomEncoding = New-Object System.Text.UTF8Encoding $False
+          [System.IO.File]::AppendAllLines(
+            $env:GITHUB_ENV,
+            [string[]]$msvcEnv,
+            $Utf8NoBomEncoding)
```
AlexanderLanin, line 339: Another stupid idea: does it make sense to use e.g. https://github.com/marketplace/actions/enable-developer-command-prompt? Sounds like it could simplify setup a lot.

AlexanderLanin (changes requested):

.github/workflows/build.yaml
nickhutchinson: I'll have a look at the action and see how much of this can be replaced by it. Only disadvantage I can see is that it doesn't let you pick with Visual Studio major version to use, but it looks like the current GitHub runner only has 2019 installed so that's not really an issue.

.github/workflows/build.yaml
nickhutchinson: I think that'll be tricky unfortunately -- there seem to be loads of implicit assumptions in the `test/suites/*.bash` tests that all tools are in paths without spaces. Ccache itself seems to cope okay, with the exception of environment variables like CCACHE_COMPILERCHECK where you can't reference tools with spaces

.github/workflows/build.yaml
AlexanderLanin: Looks a lot simpler now. I love the action marketplace 👍

AlexanderLanin (approved):

jrosdahl:
Thanks.

PR #758, opened 2020-12-28
Fix scanning of headers with clang-tidy

AlexanderLanin:
By adding `.` as an include directory, cmake actually took it literally and files are included from e.g. `src/./AtomicFile.h`.
However in `.clang-tidy` headers with an additional slash (headers from subdirectory third_party) are excluded from reports.

This PR:
* gets rid of include via `.`
* fixes some warnings
* disables the rest


src/SignalHandler.hpp
```diff
@@ -20,7 +20,7 @@
 
 #include "system.hpp"
 
-#include "signal.h"
+#include <csignal>
```
jrosdahl, line 23: I don't think this change is correct. `SignalHandler` uses `sigaddset` and friends which are documented to be part of `signal.h`, not `<csignal>`.

PR #759, opened 2020-12-29
Faster compile times

AlexanderLanin:
Clean build time on my slow laptop (this will not be comparable to yours, but for me it's reproducible across builds)

|Before|With ccache|With ccache and linker selection|
|--|--|--|
|48.3s|8.5s|5.6s (ld.gold)<br>5.5s (ld.lld)|

The goal here is not only to improve ccache compile times, but also to provide a template to others how to improve compilation times.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -60,6 +60,8 @@ else()
 endif()
 message(STATUS "Ccache dev mode: ${CCACHE_DEV_MODE}")
 
+include(UseCcache)
+include(UseFastestLinker)
 include(StandardSettings)
 include(StandardWarnings)
 include(CIBuildType)
--- a/cmake/UseCcache.cmake
+++ b/cmake/UseCcache.cmake
@@ -5,7 +5,7 @@
 # Calls `message(VERBOSE msg)` if and only if VERBOSE is available (since CMake 3.15).
 # Call CMake with --loglevel=VERBOSE to view those messages.
 function(message_verbose msg)
-  if(NOT ${CMAKE_VERSION} VERSION_LESS "3.5")
+  if(NOT ${CMAKE_VERSION} VERSION_LESS "3.15")
     message(VERBOSE ${msg})
   endif()
 endfunction()
@@ -52,8 +52,7 @@ function(use_ccache)
         file(APPEND ${launcher} "export ${keyVal}\n")
       endforeach()
       file(APPEND ${launcher}
-        "exec \"${CCACHE_PROGRAM}\" "
-        "\"${CMAKE_${lang}_COMPILER}\" \"$@\"\n")
+        "exec \"${CCACHE_PROGRAM}\" \"${CMAKE_${lang}_COMPILER}\" \"$@\"\n")
       execute_process(COMMAND chmod a+rx ${launcher})
     endforeach()
     set(CMAKE_XCODE_ATTRIBUTE_CC ${CMAKE_BINARY_DIR}/launch-C PARENT_SCOPE)
--- a/cmake/UseCcache.cmake
+++ b/cmake/UseCcache.cmake
@@ -1,0 +1,69 @@
+# Note: Compiling ccache via ccache is fine, because this uses a stable version which
+# is installed on the system.
+
+
+# Calls `message(VERBOSE msg)` if and only if VERBOSE is available (since CMake 3.15).
+# Call CMake with --loglevel=VERBOSE to view those messages.
+function(message_verbose msg)
+  if(NOT ${CMAKE_VERSION} VERSION_LESS "3.5")
+    message(VERBOSE ${msg})
+  endif()
+endfunction()
+
+# Modified version of Craig Scott's "Professional CMake: A Practical Guide", 8th Edition
+function(use_ccache)
+  if(NOT CMAKE_CURRENT_SOURCE_DIR STREQUAL CMAKE_SOURCE_DIR)
+    message(WARNING "use_ccache() disabled, as it is not called from the project top level")
+    return()
+  endif()
+
+  find_program(CCACHE_PROGRAM ccache)
+  if(NOT CCACHE_PROGRAM)
+    message_verbose("Ccache program not found, not enabling ccache for faster recompilation")
+    return()
+  endif()
+
+  message_verbose("Ccache enabled for faster recompilation")
+
+  # This will override any config and environment settings.
+  # Worst case it's overriding better suited user defined values.
+  set(ccacheEnv
+    # Another option would be CMAKE_BINARY_DIR, however currently only one basedir is supported.
+    CCACHE_BASEDIR=${CMAKE_SOURCE_DIR}
+
+    # In case of very old ccache versions (pre 3.3)
+    CCACHE_CPP2=true
+
+    # This has been turned on by default in ccache 4.0
+    # CCACHE_COMPRESS=1
+  )
+
+  if(CMAKE_GENERATOR MATCHES "Ninja|Makefiles")
+    foreach(lang IN ITEMS C CXX OBJC OBJCXX CUDA)
+      set(CMAKE_${lang}_COMPILER_LAUNCHER
+        ${CMAKE_COMMAND} -E env ${ccacheEnv} ${CCACHE_PROGRAM}
+        PARENT_SCOPE)
+    endforeach()
+  elseif(CMAKE_GENERATOR STREQUAL Xcode)
+    foreach(lang IN ITEMS C CXX)
+      set(launcher ${CMAKE_BINARY_DIR}/launch-${lang})
+      file(WRITE ${launcher} "#!/bin/bash\n\n")
+      foreach(keyVal IN LISTS ccacheEnv)
+        file(APPEND ${launcher} "export ${keyVal}\n")
+      endforeach()
+      file(APPEND ${launcher}
+        "exec \"${CCACHE_PROGRAM}\" "
+        "\"${CMAKE_${lang}_COMPILER}\" \"$@\"\n")
+      execute_process(COMMAND chmod a+rx ${launcher})
+    endforeach()
+    set(CMAKE_XCODE_ATTRIBUTE_CC ${CMAKE_BINARY_DIR}/launch-C PARENT_SCOPE)
+    set(CMAKE_XCODE_ATTRIBUTE_CXX ${CMAKE_BINARY_DIR}/launch-CXX PARENT_SCOPE)
+    set(CMAKE_XCODE_ATTRIBUTE_LD ${CMAKE_BINARY_DIR}/launch-C PARENT_SCOPE)
+    set(CMAKE_XCODE_ATTRIBUTE_LDPLUSPLUS ${CMAKE_BINARY_DIR}/launch-CXX PARENT_SCOPE)
+  endif()
+endfunction()
+
+option(USE_CCACHE "Use ccache to speed up recompilation time" TRUE)
+if(USE_CCACHE)
+  use_ccache()
+endif()
--- a/cmake/UseFastestLinker.cmake
+++ b/cmake/UseFastestLinker.cmake
@@ -1,7 +1,7 @@
 # Calls `message(VERBOSE msg)` if and only if VERBOSE is available (since CMake 3.15).
 # Call CMake with --loglevel=VERBOSE to view those messages.
 function(message_verbose msg)
-  if(NOT ${CMAKE_VERSION} VERSION_LESS "3.5")
+  if(NOT ${CMAKE_VERSION} VERSION_LESS "3.15")
     message(VERBOSE ${msg})
   endif()
 endfunction()
--- a/cmake/UseFastestLinker.cmake
+++ b/cmake/UseFastestLinker.cmake
@@ -1,0 +1,33 @@
+# Calls `message(VERBOSE msg)` if and only if VERBOSE is available (since CMake 3.15).
+# Call CMake with --loglevel=VERBOSE to view those messages.
+function(message_verbose msg)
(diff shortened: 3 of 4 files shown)


AlexanderLanin:
squashed and rebased

cmake/UseCcache.cmake
```diff
@@ -0,0 +1,56 @@
...
+    # This will override any user settings
+    set(ccacheEnv
+            # Another option would be CMAKE_BINARY_DIR, however currently only one basedir is supported.
+            CCACHE_BASEDIR=${CMAKE_SOURCE_DIR}
+
+            # In case of very old ccache versions (pre 3.3)
+            CCACHE_CPP2=true
+
+            # In case of old ccache versions (pre 4.0)
+            CCACHE_COMPRESS=1
```
jrosdahl, line 28: I think that it would be better to leave this choice to the user. I have personally never used compression with ccache before 4.0.

cmake/UseCcache.cmake
```diff
@@ -0,0 +1,56 @@
+# Compiling ccache via ccache is fine, because this uses a stable version which
+# is installed on the system.
+
+# Modified version of Craig Scott's "Professional CMake: A Practical Guide", 7th Edition
+function(use_ccache)
+    if(NOT CMAKE_CURRENT_SOURCE_DIR STREQUAL CMAKE_SOURCE_DIR)
+        message(INFO " use_ccache() disabled, as not called at root of project")
```
jrosdahl, line 7:
```suggestion
        message(INFO " use_ccache() disabled, as it is not called from the project top level")
```

cmake/UseCcache.cmake
```diff
@@ -0,0 +1,56 @@
...
+        return()
+    endif()
+
+    find_program(CCACHE_PROGRAM ccache)
+    if(NOT CCACHE_PROGRAM)
+        message(INFO " use_ccache() disabled, as ccache program not found")
```
jrosdahl, line 13: Are the `message(INFO " ...")` calls temporary debug printouts that should be removed or should they be kept? If the latter I think that they should be more succinctly phrased and use `message(STATUS "...")` instead.

cmake/UseFastestLinker.cmake
```diff
@@ -0,0 +1,22 @@
+function(use_fastest_linker)
+  if(NOT CMAKE_CURRENT_SOURCE_DIR STREQUAL CMAKE_SOURCE_DIR)
+    message(INFO " use_fastest_linker() disabled, as not called at root of project")
```
jrosdahl, line 3:
```suggestion
    message(INFO " use_fastest_linker() disabled, as it is not called at the project top level")
```

cmake/UseCcache.cmake
```diff
@@ -0,0 +1,56 @@
+# Compiling ccache via ccache is fine, because this uses a stable version which
+# is installed on the system.
+
+# Modified version of Craig Scott's "Professional CMake: A Practical Guide", 7th Edition
+function(use_ccache)
+    if(NOT CMAKE_CURRENT_SOURCE_DIR STREQUAL CMAKE_SOURCE_DIR)
```
jrosdahl, line 6: Since we're indenting CMake scripts with two spaces elsewhere I think that it would make sense to do so here as well even though it's based on an example that uses four space indentation.

CMakeLists.txt
```diff
@@ -52,6 +52,8 @@ endif()
 # Settings
 #
 include(CcacheVersion)
+include(UseCcache)
+include(UseFastestLinker)
```
jrosdahl, line 56: Since the `if` block below is associated with `include(CcacheVersion)` above I think it makes sense to move these two lines below the `if` block.

jrosdahl (commented):
Sounds like a good idea.

cmake/UseCcache.cmake
AlexanderLanin: I've been using ccache 3.4.3 with compression for many years now. It works fine :-)

cmake/UseCcache.cmake
AlexanderLanin: Actually cmake does not have INFO according to documentation. This prints NOTICE with text "INFO use_ccache() disabled, as ccache program not found".
NOTICE is more "significant" (or something) than STATUS.
So I'll change this to STATUS.

https://cmake.org/cmake/help/latest/command/message.html

AlexanderLanin:
squashed and rebased

cmake/UseCcache.cmake
```diff
@@ -0,0 +1,59 @@
+# Compiling ccache via ccache is fine, because this uses a stable version which
+# is installed on the system.
+
+# Modified version of Craig Scott's "Professional CMake: A Practical Guide", 7th Edition
+function(use_ccache)
+  if(NOT CMAKE_CURRENT_SOURCE_DIR STREQUAL CMAKE_SOURCE_DIR)
+    message(STATUS " use_ccache() disabled, as it is not called from the project top level")
```
jrosdahl, line 7: The status messages still look a bit like debug messages to me: They contain an extra space prefix and some of them mention internal details like function name which is not relevant for a normal project user.

Idea: Add a `MESSAGE_DEBUG` function that calls `message(DEBUG ...)` for CMake ≥ 3.15 and nothing for older versions. Then:

```suggestion
    MESSAGE_DEBUG("use_ccache() disabled, as it is not called from the project top level")
```

What do you think?

cmake/UseCcache.cmake
```diff
@@ -0,0 +1,59 @@
...
+    return()
+  endif()
+
+  find_program(CCACHE_PROGRAM ccache)
+  if(NOT CCACHE_PROGRAM)
+    message(STATUS " use_ccache() disabled, as ccache program not found")
```
jrosdahl, line 13:
```suggestion
    message(STATUS "Ccache program not found, not enabling ccache")
```

cmake/UseCcache.cmake
```diff
@@ -0,0 +1,59 @@
...
+    return()
+  endif()
+
+  message(STATUS " use_ccache() enabled")
+
+  # This will override any environment settings, but not user settings in config file
```
jrosdahl, line 19: Isn't this incorrect? Environment variables have higher priority than config file settings.

cmake/UseCcache.cmake
```diff
@@ -0,0 +1,59 @@
...
+  set(ccacheEnv
+    # Another option would be CMAKE_BINARY_DIR, however currently only one basedir is supported.
```
jrosdahl, line 21: It's about time to implement this...

cmake/UseCcache.cmake
```diff
@@ -0,0 +1,59 @@
...
+      execute_process(COMMAND chmod a+rx ${launcher})
+    endforeach()
+    set(CMAKE_XCODE_ATTRIBUTE_CC ${CMAKE_BINARY_DIR}/launch-C PARENT_SCOPE)
+    set(CMAKE_XCODE_ATTRIBUTE_CXX ${CMAKE_BINARY_DIR}/launch-CXX PARENT_SCOPE)
+    set(CMAKE_XCODE_ATTRIBUTE_LD ${CMAKE_BINARY_DIR}/launch-C PARENT_SCOPE)
+    set(CMAKE_XCODE_ATTRIBUTE_LDPLUSPLUS ${CMAKE_BINARY_DIR}/launch-CXX PARENT_SCOPE)
+  endif()
+endfunction()
+
+option(USE_CCACHE "Use ccache to speed up compilation time" TRUE)
```
jrosdahl, line 56:
```suggestion
option(USE_CCACHE "Use ccache to speed up recompilation time" TRUE)
```

cmake/UseFastestLinker.cmake
```diff
@@ -0,0 +1,25 @@
+function(use_fastest_linker)
+  if(NOT CMAKE_CURRENT_SOURCE_DIR STREQUAL CMAKE_SOURCE_DIR)
+    message(STATUS " use_fastest_linker() disabled, as it is not called at the project top level")
```
jrosdahl, line 3:
```suggestion
    MESSAGE_DEBUG("use_fastest_linker() disabled, as it is not called at the project top level")
```

cmake/UseCcache.cmake
```diff
@@ -0,0 +1,59 @@
...
+    return()
+  endif()
+
+  find_program(CCACHE_PROGRAM ccache)
+  if(NOT CCACHE_PROGRAM)
+    message(STATUS " use_ccache() disabled, as ccache program not found")
+    return()
+  endif()
+
+  message(STATUS " use_ccache() enabled")
```
jrosdahl, line 17:
```suggestion
  message(STATUS "Ccache enabled")
```

cmake/UseFastestLinker.cmake
```diff
@@ -0,0 +1,25 @@
...
+    return()
+  endif()
+
+  find_program(HAS_LD_LLD ld.lld)
+  if(HAS_LD_LLD)
+    link_libraries(-fuse-ld=lld)
+    message(STATUS " use_fastest_linker() using ld.lld for faster linking")
```
jrosdahl, line 10:
```suggestion
    message(STATUS "Using lld linker")
```

cmake/UseFastestLinker.cmake
```diff
@@ -0,0 +1,25 @@
...
+  else()
+    find_program(HAS_LD_GOLD ld.gold)
+    if(HAS_LD_GOLD)
+      link_libraries(-fuse-ld=gold)
+      message(STATUS " use_fastest_linker() using ld.gold for faster linking (ld.lld not found)")
```
jrosdahl, line 15:
```suggestion
      message(STATUS "Using gold linker")
```

cmake/UseFastestLinker.cmake
```diff
@@ -0,0 +1,25 @@
...
+    else()
+      message(STATUS " use_fastest_linker() using default linker (ld.gold and ld.lld (preferred) not found)")
```
jrosdahl, line 17:
```suggestion
      message(STATUS "Using default linker")
```

cmake/UseFastestLinker.cmake
```diff
@@ -0,0 +1,25 @@
...
+    endif()
+  endif()
+endfunction()
+
+option(USE_FASTER_LINKER "Use lld or ld.gold instead of ld for faster linking" TRUE)
```
jrosdahl, line 22:
```suggestion
option(USE_FASTER_LINKER "Use the lld or gold linker instead of the default for faster linking" TRUE)
```

cmake/UseCcache.cmake
AlexanderLanin: I made them VERBOSE. See https://cmake.org/cmake/help/latest/command/message.html

jrosdahl: OK, but neither VERBOSE nor DEBUG exist for CMake < 3.15 so they will be always printed with a strange literal "VERBOSE" prefix for those versions.

AlexanderLanin: Argh

AlexanderLanin:
Squashed and rebased

cmake/UseCcache.cmake
```diff
@@ -0,0 +1,69 @@
+# Note: Compiling ccache via ccache is fine, because this uses a stable version which
+# is installed on the system.
+
+
+# Calls `message(VERBOSE msg)` if and only if VERBOSE is available (since CMake 3.15).
+# Call CMake with --loglevel=VERBOSE to view those messages.
+function(message_verbose msg)
+  if(NOT ${CMAKE_VERSION} VERSION_LESS "3.5")
```
jrosdahl, line 8:
```suggestion
  if(NOT ${CMAKE_VERSION} VERSION_LESS "3.15")
```

cmake/UseCcache.cmake
```diff
@@ -0,0 +1,69 @@
...
+  elseif(CMAKE_GENERATOR STREQUAL Xcode)
+    foreach(lang IN ITEMS C CXX)
+      set(launcher ${CMAKE_BINARY_DIR}/launch-${lang})
+      file(WRITE ${launcher} "#!/bin/bash\n\n")
+      foreach(keyVal IN LISTS ccacheEnv)
+        file(APPEND ${launcher} "export ${keyVal}\n")
+      endforeach()
+      file(APPEND ${launcher}
+        "exec \"${CCACHE_PROGRAM}\" "
+        "\"${CMAKE_${lang}_COMPILER}\" \"$@\"\n")
```
jrosdahl, line 56:
```suggestion
        "exec \"${CCACHE_PROGRAM}\" \"${CMAKE_${lang}_COMPILER}\" \"$@\"\n")
```

cmake/UseFastestLinker.cmake
```diff
@@ -0,0 +1,33 @@
+# Calls `message(VERBOSE msg)` if and only if VERBOSE is available (since CMake 3.15).
+# Call CMake with --loglevel=VERBOSE to view those messages.
+function(message_verbose msg)
+  if(NOT ${CMAKE_VERSION} VERSION_LESS "3.5")
```
jrosdahl, line 4:
```suggestion
  if(NOT ${CMAKE_VERSION} VERSION_LESS "3.15")
```

AlexanderLanin:
Hups. 3.15 indeed.

jrosdahl:
Thanks.

jrosdahl:
I noticed that the overhead of `cmake -E env` is ≈9 ms on my system. This means that ccache direct mode hits on average have become about twice as slow when building ccache itself on my system.

I improved this in 68d8481c173eb711624232c8f9c8e4507d747dc9 by using the standard `env` program if available. Its overhead is ≈1 ms.

PR #761, opened 2020-12-31
doc/MANUAL.adoc: don't use non-ASCII quotes

trofi:
Soome locales like `LANG=fr_FR.iso885915@euro make` can't
handle UTF-8 single- and double-quotes:

```
$ LANG=fr_FR.iso885915@euro make
...
asciidoc: FAILED: MANUAL.adoc: line 529: unexpected error:
...
  File "/usr/lib/python3.8/encodings/iso8859_15.py", line 19, in encode
    return codecs.charmap_encode(input,self.errors,encoding_table)[0]
UnicodeEncodeError: 'charmap' codec can't encode character '\u201c'
  in position 54: character maps to <undefined>
```

To avoid it the patch uses ASCII equivalents of symbols.

The patch is generated as:

```shell
$ sed \
    -e 's/\xE2\x80\x99/'\''/g' \
    -e 's/\xE2\x80\x9C/"/g' \
    -e 's/\xE2\x80\x9D/"/g' \
    -i doc/MANUAL.adoc
```

Reported-by: Christophe PEREZ
Bug: https://bugs.gentoo.org/762814

### DIFF
--- a/doc/MANUAL.adoc
+++ b/doc/MANUAL.adoc
@@ -522,9 +522,9 @@ Semantics of *compression_level*:
     levels are however useful when recompressing the cache with command line
     option *-X/--recompress*.
 *< 0*::
-    A negative value corresponds to Zstandard's “ultra-fast” compression
+    A negative value corresponds to Zstandard's ``ultra-fast'' compression
     levels, which are even faster than level 1 but with less good compression
-    ratios. For instance, level *-3* corresponds to “--fast=3” for the *zstd*
+    ratios. For instance, level *-3* corresponds to ``--fast=3'' for the *zstd*
     command line tool. In practice, there is little use for levels lower than
     *-5* or so.
 *0* (default)::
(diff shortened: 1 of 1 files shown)


AlexanderLanin (approved):

jrosdahl:
Sorry, this change will make the UTF-8 man page and the HTML version of the manual worse, so I can't accept it. In principle, I don't want to be constrained to ASCII, especially for the HTML manual, but since the only non-ASCII codepoints used so far are double quotes, I would accept a patch like this:

* `’` → `'`
*  `“` → ``` `` ```
* `”` → `''`

That way AsciiDoc will still render the same results. Would you like to update the patch like indicated?

trofi:
Sounds good! Updated the patch generated as:

```shell
$ sed \
    -e 's/\xE2\x80\x99/'\''/g' \
    -e 's/\xE2\x80\x9C/'\`\`'/g' \
    -e 's/\xE2\x80\x9D/'\'\''/g' \
    -i doc/MANUAL.adoc
```

PR #762, opened 2020-12-31
Always use 64bit to print time_t

kraj:
some 32bit architectures e.g. RISCV32 use 64bit time_t from beginning
which does not fit into long int size on LP32 systems

Upstream-Status: Pending

Signed-off-by: Khem Raj <raj.khem@gmail.com>


AlexanderLanin (approved):

jrosdahl:
Thanks.

PR #764, opened 2021-01-01
Extract enums into separate headers

AlexanderLanin:
`ccache.hpp` and `Statistics.hpp` are included by lots of other files.
`ccache` is basically the business logic, nothing should depend on it.
There is no need for something as central as exceptions to depend upon `Statistics`.

This PR  extracts the parts which are truly central and required by other files into separate headers.


src/Config.cpp
```diff
@@ -20,9 +20,9 @@
 
 #include "AtomicFile.hpp"
 #include "Compression.hpp"
+#include "SloppinessEnum.hpp"
```
jrosdahl, line 23: Having "Enum" in the filename feels a bit redundant. Would you be OK with naming it `Sloppiness.hpp`?

jrosdahl (commented):
The sloppiness enum is the final step of refactoring away stuff from `ccache.hpp`, so that's good.

> There is no need for something as central as exceptions to depend upon `Statistics`.

Yes, there is a need: since the `Failure` class embeds a statistics counter it needs to depend on statistics knowledge, and it still does that via `StatisticsEnum.hpp`, right? What do you think about moving the `Failure` class to `ccache.cpp` since that's the only place it's used? And maybe keep the statistics enum in `Statistics.hpp`?

AlexanderLanin:
> Yes, there is a need: since the `Failure` class embeds a statistics counter it needs to depend on statistics knowledge, and it still does that via `StatisticsEnum.hpp`, right? 
Well yes. But now it's only the enum. The trouble with Statistics is that that they require Counters, AtomicFile, Config, Lockfile, Logging etc.

> What do you think about moving the `Failure` class to `ccache.cpp` since that's the only place it's used?
That seems reasonable. ccache.cpp is (ab)using Failure for control flow and no one else need it. That's the only dependency upon Statistic (enum).

> And maybe keep the statistics enum in `Statistics.hpp`?
I'll give it a try to see how that looks.

src/Config.cpp
AlexanderLanin: Yeah that's strange, but there is a huge number of files in src and it's difficult to find what you are looking for. Originally I did this because of `Statistic` and `Statistics`.
Should I rename them?

AlexanderLanin: renamed. see latest version.

jrosdahl: > huge number of files in src

We must be coming from quite different realities, because to me ccache is a small project and has relatively few source code files. Compare with larger projects like (just spontaneously picking a few) CMake, Git, Linux and glibc. I'm not saying that it's bad to group source code files, I just don't think the number of files in the src directory is spectacular in any way.

src/Statistics.hpp
```diff
@@ -21,6 +21,7 @@
 #include "system.hpp"
 
 #include "Counters.hpp"
+#include "Statistic.hpp" // Any reasonable use of Statistics required Statistic enum.
```
jrosdahl, line 24:
```suggestion
#include "Statistic.hpp" // Any reasonable use of Statistics requires the Statistic enum.
```

src/ccache.cpp
```diff
@@ -181,6 +182,43 @@ const uint8_t k_max_cache_levels = 4;
 // stored in the cache changes in a backwards-incompatible way.
 const char HASH_PREFIX[] = "3";
 
+namespace {
+// Throw a Failure if ccache did not succeed in getting or putting a result in
```
jrosdahl, line 186:
```suggestion
namespace {

// Throw a Failure if ccache did not succeed in getting or putting a result in
```

src/ccache.cpp
```diff
@@ -181,6 +182,43 @@ const uint8_t k_max_cache_levels = 4;
...
+{
+  return m_exit_code;
+}
+
+inline Statistic
+Failure::statistic() const
+{
+  return m_statistic;
+}
+} // namespace
```
jrosdahl, line 220:
```suggestion
}

} // namespace
```

Issue #625, opened 2020-07-13
Unable to compile the latest ccache repository


dyadav7:
### How to reproduce ###

Using gcc version 5.4.0 (GCC) 
cmake version 3.17.2

```
setenv CC     /../../gcc/5.4.0-static/bin/gcc
setenv CXX   /../../gcc/5.4.0-static/bin/g++

rm -rf build
mkdir build
cd build
cmake -DZSTD_FROM_INTERNET=ON ..
make
```

### Actual behavior ###

Got error : unrecognized command line option "-mavx2"

```console
$ make
...
[  3%] Building ASM object src/third_party/blake3/CMakeFiles/blake3.dir/blake3_avx2_x86-64_unix.S.o
cc1: error: unrecognized command line option "-mavx2"
make[2]: *** [src/third_party/blake3/CMakeFiles/blake3.dir/build.make:127: src/third_party/blake3/CMakeFiles/blake3.dir/blake3_avx2_x86-64_unix.S.o] Error 1
make[1]: *** [CMakeFiles/Makefile2:792: src/third_party/blake3/CMakeFiles/blake3.dir/all] Error 2
make: *** [Makefile:183: all] Error 2
```

#### Another error message ####

Determining if the function pthread_create exists in the pthreads failed with the following output:

```
Change Dir: /home/deepaky/temp/ccache/orig/ccache/build/CMakeFiles/CMakeTmp

Run Build Command(s):/usr/bin/gmake cmTC_3cb18/fast && /usr/bin/gmake  -f CMakeFiles/cmTC_3cb18.dir/build.make CMakeFiles/cmTC_3cb18.dir/build
gmake[1]: Entering directory `/home/deepaky/temp/ccache/orig/ccache/build/CMakeFiles/CMakeTmp'
Building C object CMakeFiles/cmTC_3cb18.dir/CheckFunctionExists.c.o
/zin/tools/gcc/5.4.0-static/bin/gcc   -DCHECK_FUNCTION_EXISTS=pthread_create   -o CMakeFiles/cmTC_3cb18.dir/CheckFunctionExists.c.o   -c /zin/plat/local.zincon/cmake-3.17.2/share/cmake-3.17/Modules/CheckFunctionExists.c
Linking C executable cmTC_3cb18
/zin/plat/local.zincon/cmake-3.17.2/bin/cmake -E cmake_link_script CMakeFiles/cmTC_3cb18.dir/link.txt --verbose=1
[...]

PR #768, opened 2021-01-05
blake3: Remove asm checks for sse/avx

kraj:
This ends up passing on clang/linux wrongly when building for aarch64
the check in else part is good to detect the feature support and this
check can be removed, it was setting

HAVE_ASM_AVX* and HAVE_ASM_SSE* macros which are not used in the build
anyway

Signed-off-by: Khem Raj <raj.khem@gmail.com>

### DIFF
--- a/src/third_party/blake3/CMakeLists.txt
+++ b/src/third_party/blake3/CMakeLists.txt
@@ -25,8 +25,6 @@ function(add_source_if_enabled feature compile_flags intrinsic)
       AND CMAKE_CXX_COMPILER_VERSION VERSION_LESS 9.0)
     message(STATUS "Detected unsupported compiler for ${have_feature} - disabled")
     set(${have_feature} FALSE)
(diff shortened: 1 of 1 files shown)


jrosdahl:
Applying this would reintroduce issue #625 (see 81c5610b764aa8e9cc632748698fea0e43684e2f).

Please describe in further detail what problem you see.

> HAVE_ASM_AVX* and HAVE_ASM_SSE* macros which are not used in the build

It's correct that they are not used outside `src/third_party/blake3/CMakeLists.txt`, but they _are_ used inside said file: https://github.com/ccache/ccache/blob/389789fc613896e8f6f3677b22e8ae7a009d9407/src/third_party/blake3/CMakeLists.txt#L41

erijo:
I did an attempt to fix this in another way. @kraj: could you please test [this branch](https://github.com/erijo/ccache/tree/blake3-build) and let me know how it works for you?

PR #769, opened 2021-01-06
Fix Lockfile on Windows when parent directories do not exist

nickhutchinson:
Fix flipped condition test that meant the file lock wouldn't be acquired if parent directories needed to be created first.

I noticed this when looking at some of the Windows test failures, as stats files wouldn't be updated.


jrosdahl:
Thanks.

PR #775, opened 2021-01-14
Fix -x cuda with clang issue?

### LINKED ISSUE #772: Support for "cuda" language
At the moment, ccache doesn't work when building cuda code with [clang directly](https://llvm.org/docs/CompileCudaWithLLVM.html).

The documentation seems to imply that this should be supported, but it isn't.

I would appreciate this feature.

rgreenblatt:
I don't understand the broader implications of this change or if changes are needed elsewhere for this to be consistent.
However, this does fix #772 based on my minimal testing.


rgreenblatt:
see https://github.com/ccache/ccache/issues/772#issuecomment-760462748

PR #779, opened 2021-01-15
Refactor: drop out parameter from get_level_1_files

AlexanderLanin:
Trivial change which makes some code slightly more readable. No functional change.


jrosdahl:
Thanks.

PR #780, opened 2021-01-16
Test suite fixes and improvements for Windows

nickhutchinson:
Small tweaks to get more tests passing on Windows: account for `\r\n` line endings in `ccache --version` test, and skip some tests that can never succeed on Windows.

I noticed that having more tests running on Windows would increase CI test run times by a few minutes. This seems to be due to git-bash's emulation of fork/exec being extremely slow -- on my machine it's roughly 30ms to spawn `/usr/bin/true` from a bash script compared to 2ms on my macOS machine. So I've also made some tweaks to improve performance on Windows by avoiding unneeded subshells.

Also fixes some other issues with the test suite:
- having multiple arguments in `CC` wasn't handled properly in some cases
- on test failure we'd pass the test name directly to `printf`, which caused some confusing errors when the test started with a `-`.

### DIFF
--- a/test/run
+++ b/test/run
@@ -32,15 +32,15 @@ if [[ -t 1 ]]; then
 fi
 
 green() {
-    printf "$ansi_boldgreen$*$ansi_reset\n"
+    printf "$ansi_boldgreen%s$ansi_reset\n" "$*"
 }
 
 red() {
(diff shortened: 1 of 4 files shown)


jrosdahl:
Thanks.

PR #781, opened 2021-01-17
Blake3 build

erijo:
This is an attempt to fix #768 in a different way: run the assembler to determine if it can compile the files, instead of just checking that the compiler flags are supported. Running the assembler doesn't take long time so it shouldn't slow down the build much.

I also added all *.asm files from blake3 (and updated the existing ones slightly) and enabled the assembler variant for MSVC. I haven't actually tested that they work, but it seems to build correctly at least.


jrosdahl:
Thanks.

PR #788, opened 2021-01-23
Fix fast linker selection for old gcc versions

AlexanderLanin:
Turns out it would have been a good idea to add that to CI in the first place.
Current master fails to compile if ld.gold / lld are installed with gcc <9.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -91,10 +91,12 @@ jobs:
           if [ "${{ runner.os }}" = "Linux" ]; then
             sudo apt-get update
 
+            # Install ld.gold (binutils) and ld.lld on different runs.
+            # Binding to Ubuntu 20 has no special meaning.
             if [ "${{ matrix.config.os }}" = "ubuntu-20.04" ]; then
-              sudo apt-get install -y ninja-build elfutils libzstd-dev
+              sudo apt-get install -y ninja-build elfutils libzstd-dev lld
             else
-              sudo apt-get install -y ninja-build elfutils libzstd1-dev
+              sudo apt-get install -y ninja-build elfutils libzstd1-dev binutils
             fi
 
             if [ "${{ matrix.config.compiler }}" = "gcc" ]; then
--- a/cmake/UseFastestLinker.cmake
+++ b/cmake/UseFastestLinker.cmake
@@ -11,19 +11,20 @@ function(use_fastest_linker)
     message(WARNING "use_fastest_linker() disabled, as it is not called at the project top level")
     return()
   endif()
-
-  find_program(HAS_LD_LLD ld.lld)
-  if(HAS_LD_LLD)
-    link_libraries(-fuse-ld=lld)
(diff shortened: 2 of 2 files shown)


jrosdahl:
@AlexanderLanin: I was too quick merging this since it doesn't work: `CMAKE_CXX_LINK_EXECUTABLE` is only set in the function scope, not the parent scope, so the global value stays the default. And if that is fixed then the build fails like this:

```
[78/89] Linking CXX executable ccache
FAILED: ccache 
: && /usr/bin/ld.lld -g   CMakeFiles/ccache.dir/src/main.cpp.o -o ccache  src/libccache_lib.a  /usr/lib/x86_64-linux-gnu/libzstd.so  -pthread  src/third_party/libthird_party_lib.a  src/third_party/blake3/libblake3.a && :
ld.lld: error: unknown argument '-pthread'
```

Also, I'm getting second thoughts on this for another reason as well: Is it a good idea to override `CMAKE_CXX_LINK_EXECUTABLE` unconditionally if `gold` or `ld.lld` is found? I'm thinking that some non-GCC/Clang toolchains may have a different `CMAKE_CXX_LINK_EXECUTABLE` command line, or the user has overridden it, so that replacing the full command line will be a bad idea.

How about reverting to the previous `link_libraries` approach and making this opt-in instead? Or maybe even better, don't bother at all since the speedup is very small, at least on my system?

AlexanderLanin:
@jrosdahl: Oh great... so it works because it does nothing. But actually this cannot be, because I did measure a reproducible effect with this disabled, enabled, disabled and enabled again. It was 25% improvement for a bigger project (120 to 90 seconds cached clean build).
However if that fails at PARENT_SCOPE that's a bad sign.

I'll look into this further in the next 1-2 weeks, but not today. I guess it's not that urgent if the current function does nothing 😆 
This is currently opt-out since with opt-in nobody would use it. It would just be a well hidden option. Maybe it should be opt-in for non gcc/clang.
If I cannot fix it in a way that will not lead to any more additional followup effort, I will create a pull request to remove it.

jrosdahl:
> Oh great... so it works because it does nothing.

Reported in #794 for tracking purposes.

jrosdahl:
I wrote:

> Or maybe even better, don't bother at all since the speedup is very small, at least on my system?

I'll have to tone that down a little. The speedup is large enough that a edit-build-unittest cycle is slightly but noticeably faster, so it would be great to keep it if possible.

@AlexanderLanin wrote:

> Current master fails to compile if ld.gold / lld are installed with gcc <9.

I fail to reproduce this locally with g++ from GCC 7 or 8. I also [tried to revert to your initial implementation using the `link_libraries` approach while keeping the CI part](https://github.com/jrosdahl/ccache/commit/074c6b9e6db11be5a76f5260de5afff2a11b9016), but that also worked, for instance [on Ubuntu 18.04 with GCC 7 and the gold linker](https://github.com/jrosdahl/ccache/runs/1858178642?check_suite_focus=true).

Any hints on how to reproduce the problem?

PR #791, opened 2021-01-24
disable inode cache on OS without pthread_mutexattr_setpshared()

sthen:
The inode cache requires pthread_mutexattr_setpshared() and build fails on OS with sys/mman.h that do not have this function. Here's my attempt at checking for this in the CMake files and disabling the feature if needed.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -87,7 +87,7 @@ include(GNUInstallDirs)
 include(GenerateConfigurationFile)
 include(GenerateVersionFile)
 
-if(HAVE_SYS_MMAN_H)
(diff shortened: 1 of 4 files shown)


sthen:
Works as expected on OpenBSD which does not have pthread_mutexattr_setpshared(), I haven't checked on other OS.

sthen:
..and with the additional pushes, should work as expected on other OS too.

sthen:
Thanks for merging, and apologies for the noise while I fixed the problems with my diff.

jrosdahl:
No problems, thanks for the fix.

PR #802, opened 2021-02-17
Support -fcolor-diagnostics passed to cc1 with -Xclang

### LINKED ISSUE #801: Coloring diagnostics does not always work with clang
### How to reproduce ###

1. Create file `foo.cc` with below content:
```
int main() {
    return "";
}
```
2. Run: `ccache clang++ -Xclang -fcolor-diagnostics -c foo.cc 2>&1 | cat`

### Actual behavior ###

ccache strips colors from diagnostics, here is an interesting excerpt from running it under strace which shows that clang did produce colored diagnostics, but ccache removed the colors:
```
openat(AT_FDCWD, "/run/user/1000/ccache-tmp/tmp.stderr.yQUHE0", O_RDONLY) = 4
read(4, "\33[1mfoo.cc:2:8: \33[0m\33[0;1;31merr"..., 1024) = 187
close(4)                                = 0
write(2, "foo.cc:2:8: error: cannot initia"..., 145foo.cc:2:8: error: cannot initialize return object of type 'int' with an lvalue of type 'const char [1]'
return "";
       ^~
1 error generated.
) = 145
```

### Expected behavior ###

ccache should print colored diagnostics

### Environment ###

ccache version 4.2

Linux archlinux 5.10.16-arch1-1 SMP PREEMPT Sat, 13 Feb 2021 20:50:18 +0000 x86_64 GNU/Linux
clang version 11.0.1 (also reproduced with 10.0.0)

### Remarks ###
If I remove `-Xclang` or `-c` it works i.e. both `ccache clang++ -fcolor-diagnostics -c foo.cc 2>&1 | cat` and `ccache clang++ -Xclang -fcolor-diagnostics foo.cc 2>&1 | cat` give colored diagnostics.

foutrelis:
Before this change `clang++ -Xclang -fcolor-diagnostics -c foo.cc`
would result in diagnostics without color if stderr was not a TTY.

Fixes #801.

### DIFF
--- a/src/argprocessing.cpp
+++ b/src/argprocessing.cpp
@@ -700,6 +700,14 @@ process_arg(Context& ctx,
     return Statistic::unsupported_compiler_option;
   }
 
+  // In the "-Xclang -fcolor-diagnostics" form, -Xclang is skipped and the
+  // -fcolor-diagnostics argument which is passed to cc1 is handled below.
+  if (args[i] == "-Xclang" && i < args.size() - 1
+      && args[i + 1] == "-fcolor-diagnostics") {
+    state.compiler_only_args_no_hash.push_back(args[i]);
+    ++i;
+  }
+
   if (args[i] == "-fcolor-diagnostics" || args[i] == "-fdiagnostics-color"
       || args[i] == "-fdiagnostics-color=always") {
     state.color_diagnostics = ColorDiagnostics::always;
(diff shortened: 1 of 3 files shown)


foutrelis:
Hopefully this is on the right track. Please review in case there are possible issues with this approach.

jrosdahl:
Thanks.

PR #805, opened 2021-02-20
Build ccache with xlclang++ on AIX72

jonpetri:
- Update of config.h.in to build with xlclang++ on AIX72
- Little fixes in code chunks never built on usual platforms


jrosdahl:
Thanks.

PR #807, opened 2021-02-22
Fix "Debug option" test

pawelkrysiak:
Trivial fix for 'debug option' test from base testsuite.
Previously `reference_test1.o` was compared to itself, therefore not really testing anything.
Now it properly compares `reference_test1.o` to `test1.o` to check that CCACHE produces the same output as real compiler.

### DIFF
--- a/test/suites/base.bash
+++ b/test/suites/base.bash
@@ -64,7 +64,7 @@ base_tests() {
     expect_stat 'cache miss' 1
 
     $REAL_COMPILER -c -o reference_test1.o test1.c -g
-    expect_equal_object_files reference_test1.o reference_test1.o
+    expect_equal_object_files reference_test1.o test1.o
 
     # -------------------------------------------------------------------------
     TEST "Output option"


pawelkrysiak:
Well, seeing that `macOS-latest-xcode-12.2` check fails, it is not so trivial fix anymore.
Clearly it's good to have this UT fixed, as it discovers some discrepancy.
Now I'm unsure what should the process look like - creating a separate issue for fixing this macOS case or trying to resolve it within this PR (or yet another option)?

jrosdahl:
> Trivial fix for 'debug option' test from base testsuite.

Thanks.

> Well, seeing that `macOS-latest-xcode-12.2` check fails, it is not so trivial fix anymore.

That's interesting. I had a look at this, and it seems like `compiler -E file.c -o file.i; compiler -g file.i -o file.o` doesn't produce the same result as `compiler -g file.c -o file.o` on that platform: In the latter case more information is encoded in the object file.

> Now I'm unsure what should the process look like - creating a separate issue for fixing this macOS case or trying to resolve it within this PR (or yet another option)?

I went with option three and fixed it in 19fa063b109006373be9d00c2b1a5f0d32086c3f.

PR #808, opened 2021-02-22
Fix Util::read_file returning truncated results if size_hint is an underestimate

### LINKED ISSUE #803: Ccache 4.2 returns stale cache results on Windows
After switching to the offical cache 4.2 builds (built with mingw-w64) from my development build (built with MSVC 2019), I started getting frequent build failures that suggested cache was returning stale cache results.

Digging deeper, I believe the reason is that mingw-w64's implementation of `stat()` (which forwards through to the implementation in the legacy `\Windows\System32\msvcrt.dll`) can report the wrong value in the `st_size` field. Using Process Monitor, it's clear that it's getting the file size by querying using [`FindFirstFile`](https://docs.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-findfirstfilew), which returns a cached result, instead of actually opening the file and querying its attributes with [`GetFileInformationByHandle`](https://docs.microsoft.com/en-gb/windows/win32/api/fileapi/nf-fileapi-getfileinformationbyhandle?redirectedfrom=MSDN) or similar.

Here's an example where running `dir` to list a directory contents shows different file sizes for the same underlying file.

```
> type NUL > a
> mklink /H b a
> echo abcde >> a
> dir
17/02/2021  21:10    <DIR>          .
17/02/2021  21:10    <DIR>          ..
17/02/2021  21:10                 8 a
17/02/2021  21:09                 0 b
               2 File(s)              8 bytes
               2 Dir(s)  155,061,624,832 bytes free
> type a
abcde

> type b
abcde

> dir
17/02/2021  21:10    <DIR>          .
17/02/2021  21:10    <DIR>          ..
17/02/2021  21:10                 8 a
17/02/2021  21:10                 8 b
               2 File(s)             16 bytes
               2 Dir(s)  155,061,624,832 bytes free
```

This affects ccache because `Utils::read_file()` [will first call `stat()` on the file](https://github.com/ccache/ccache/blob/27f00db43a8330be87b5355210f1b553ef0122da/src/Util.cpp#L1103) to determine how many bytes it should read. Using Process Monitor, you can observe that ccache only reads the first 1024 bytes from the preprocessed source file.

### How to reproduce ###

On Windows in a MSVC command prompt:
```
> echo #include ^<__msvc_all_public_headers.hpp^> > test.cpp
> set CCACHE_DEBUG=1
> clang++ -E test.cpp -o test.cpp.preprocessed
> ccache clang++ -c test.cpp -o test.o
```

(Debug outputs from ccache are here: [ccache-repro.zip](https://github.com/ccache/ccache/files/5998933/ccache-repro.zip))

### Actual behavior ###
The `test.o.ccache-input-p` and `test.o.ccache-input-text` files created by ccache only appear to have the first 1024 bytes of `test.cpp.preprocessed`. (`test.cpp.preprocessed` is around 3.4 MB!)

### Expected behavior ###
The `test.o.ccache-input-p` and `test.o.ccache-input-text` files created by ccache are around the same size as `test.cpp.preprocessed`.

### Environment ###

ccache version 4.2 (official build)
Windows 10 20H2
LLVM 11.0.1

nickhutchinson:
If the size_hint passed to `Util::read_file()` was an underestimate, or the platform's `stat()` implementation gives an inaccurate file size (e.g. MinGW) then `Util::read_file()` would only issue a single `read()` call instead of reading the entire file.

Fixes #803.

### DIFF
--- a/src/Util.cpp
+++ b/src/Util.cpp
@@ -1180,7 +1180,7 @@ read_file(const std::string& path, size_t size_hint)
   result.resize(size_hint);
 
   while (true) {
-    if (pos > result.size()) {
+    if (pos == result.size()) {
       result.resize(2 * result.size());
     }
     const size_t max_read = result.size() - pos;
--- a/unittest/test_Util.cpp
+++ b/unittest/test_Util.cpp
@@ -820,6 +820,10 @@ TEST_CASE("Util::read_file and Util::write_file")
 #else
   CHECK(data == "carpet\n\n");
 #endif
+
+  Util::write_file("size_hint_test", std::string(8192, '\0'));
+  CHECK(Util::read_file("size_hint_test", 4096 /*size_hint*/).size() == 8192);
+
   CHECK_THROWS_WITH(Util::read_file("does/not/exist"),
                     "No such file or directory");
 


nickhutchinson:
As an aside, how dependent is ccache on `stat()` returning accurate results for st_mtime/st_size etc?

If these need to be correct rather than best effort, then I suggest writing a new implementation that queries using `GetFileInformationByHandleEx()`. The stat implementation in msvcrt.dll used by MinGW seems very flawed, and the one in ucrtbase.dll used by recent Visual Studio isn't much better.

jrosdahl:
> If the size_hint passed to `Util::read_file()` was an underestimate, or the platform's `stat()` implementation gives an inaccurate file size (e.g. MinGW) then `Util::read_file()` would only issue a single `read()` call instead of reading the entire file.

Oops.

> As an aside, how dependent is ccache on `stat()` returning accurate results for st_mtime/st_size etc?
>
> If these need to be correct rather than best effort, then I suggest writing a new implementation that queries using `GetFileInformationByHandleEx()`.

At the very least, statistics and cleanup behavior would be affected negatively without accurate `st_size` values. Stored results will likely also risk being stored in truncated form since `Result::Writer::do_finalize` reads `st_size` bytes from each file to store (it doesn't use `Util::read_file`). Inaccurate `st_mtime` values potentially affect cleanup negatively but I think they won't produce false cache hits or corruption.

> The stat implementation in msvcrt.dll used by MinGW seems very flawed, and the one in ucrtbase.dll used by recent Visual Studio isn't much better.

:slightly_frowning_face:

jrosdahl:
Thanks.

PR #816, opened 2021-03-10
Fix win32execute

asalwa:
I'm trying to make Ccache work (again) on Windows. What is broken: handling of "execute" with long command parameters (see win32execute).

In more details:
- parameter lpCommandLine was formatted incorrectly (it has to contain app name too)
- temporary file with parameters to the compiler was formatted incorrectly (it should NOT contain app name; if it contains backslashes, then these backslashes need to be escaped as double-backslashes)
- wrong location (directory) of temporary files in win32execute
- premature deletion of temporary files in win32execute

### DIFF
--- a/src/Win32Util.cpp
+++ b/src/Win32Util.cpp
@@ -67,8 +67,8 @@ argv_to_string(const char* const* argv, const std::string& prefix)
     for (size_t j = 0; arg[j]; ++j) {
       switch (arg[j]) {
       case '\\':
-        ++bs;
-        break;
+        //++bs;
+        //break;
       // Fallthrough.
       case '"':
         bs = (bs << 1) + 1;
--- a/src/Win32Util.cpp
+++ b/src/Win32Util.cpp
@@ -55,7 +55,7 @@ error_message(DWORD error_code)
 }
 
 std::string
-argv_to_string(const char* const* argv, const std::string& prefix)
+argv_to_string(const char* const* argv, const std::string& prefix, bool escape_backslashes)
 {
   std::string result;
   size_t i = 0;
@@ -67,9 +67,11 @@ argv_to_string(const char* const* argv, const std::string& prefix)
     for (size_t j = 0; arg[j]; ++j) {
       switch (arg[j]) {
       case '\\':
-        //++bs;
-        // break;
-      // Fallthrough.
+    	if (!escape_backslashes) {
+          ++bs;
+          break;
+    	}
+        // Fallthrough.
       case '"':
         bs = (bs << 1) + 1;
       // Fallthrough.
--- a/src/Win32Util.cpp
+++ b/src/Win32Util.cpp
@@ -68,7 +68,7 @@ argv_to_string(const char* const* argv, const std::string& prefix)
       switch (arg[j]) {
       case '\\':
         //++bs;
-        //break;
+        // break;
       // Fallthrough.
       case '"':
         bs = (bs << 1) + 1;
--- a/src/Win32Util.cpp
+++ b/src/Win32Util.cpp
@@ -55,7 +55,9 @@ error_message(DWORD error_code)
 }
 
 std::string
-argv_to_string(const char* const* argv, const std::string& prefix, bool escape_backslashes)
+argv_to_string(const char* const* argv,
+               const std::string& prefix,
+               bool escape_backslashes)
 {
   std::string result;
   size_t i = 0;
@@ -67,10 +69,10 @@ argv_to_string(const char* const* argv, const std::string& prefix, bool escape_b
     for (size_t j = 0; arg[j]; ++j) {
       switch (arg[j]) {
       case '\\':
-    	if (!escape_backslashes) {
+        if (!escape_backslashes) {
           ++bs;
           break;
-    	}
+        }
         // Fallthrough.
       case '"':
         bs = (bs << 1) + 1;
--- a/src/Win32Util.hpp
+++ b/src/Win32Util.hpp
@@ -30,7 +30,7 @@ std::string add_exe_suffix(const std::string& program);
 
 // Recreate a Windows command line string based on `argv`. If `prefix` is
 // non-empty, add it as the first argument.
-std::string argv_to_string(const char* const* argv, const std::string& prefix);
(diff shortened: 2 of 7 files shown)


src/execute.cpp
```diff
@@ -108,13 +114,16 @@ win32execute(const char* path,
   std::string args = Win32Util::argv_to_string(argv, sh);
   std::string full_path = Win32Util::add_exe_suffix(path);
   std::string tmp_file_path;
+  const char* lpAppName = full_path.c_str();
```
jrosdahl, line 117: I suggest continuing to use `full_path.c_str()` instead of introducing an alias. If another variable is still wanted for some reason, please follow the naming convention (i.e. name it something like `full_path_str`).

Also, perhaps the compiler path should be double quoted, both in the first and second argument to `CreateProcess`?

src/Win32Util.hpp
```diff
@@ -30,7 +30,9 @@ std::string add_exe_suffix(const std::string& program);
 
 // Recreate a Windows command line string based on `argv`. If `prefix` is
 // non-empty, add it as the first argument.
-std::string argv_to_string(const char* const* argv, const std::string& prefix);
+std::string argv_to_string(const char* const* argv,
+                           const std::string& prefix,
+                           bool escape_backslashes = false);
```
jrosdahl, line 35: Please describe the new parameter briefly in the comment above.

src/execute.cpp
```diff
@@ -108,13 +114,16 @@ win32execute(const char* path,
...
   if (args.length() > 8192) {
-    TemporaryFile tmp_file(path);
+    TemporaryFile tmp_file(FMT("{}/cmd_args", temp_dir));
```
jrosdahl, line 119: How about placing the temporary file in `%TEMP%` (with some appropriate fallback if that environment variable isn't set, I guess?) instead to avoid the need for a `temp_dir` parameter?

src/execute.cpp
asalwa: I'll remove this alias, sure. I was doing a lot of LOG() debugging and probably I added this variable for debugging.
As for quoting the path - I don't want to add it in this pull request. It works for me as it is. I can analyze it later (just to divide work into smaller chunks).

src/execute.cpp
asalwa: I don't think it is good idea - using system temp directory when Ccache already has its own temp directory. It is weird externally visible behavior just because we didn't want to pass extra parameter.
I think we need to find good solution for execv() macro issue (ugly `saved_temp_dir` variable I added in `cache_compilation()`).

We could try to get rid of direct call to execv(). Maybe adding such version of execute:
`int execute_noreturn(const char* const* argv, Fd&& fd_out, Fd&& fd_err, const std::string& temp_dir);`
I would call this new function in `cache_compilation` (instead of execv()).
There is still `temp_dir` parameter, but now it is passed explicitly to called function, no surprising macro magic (I mean execv() macro in system.hpp).
This `temp_dir` parameter won't be used by Linux version of `execute_noreturn` but I think this is not a big deal.

Implementation of execute_noreturn() for Linux will be trivial - just a wrapper for execv (only `argv` parameter is used).
Implementation for Windows is also simple - a wrapper for win32_execute, similar to existing `execute`.

What do you think? Is it acceptable approach?

asalwa: IMHO this approach is a step toward better internal architecture of Ccache. I mean well architectured portability - instead of pretending on Windows that it is Linux (by macros like this execv()) there should be explicit "porting layer". `execute_noreturn` says what it does - it executes without return. It has different implementation on Windows and Linux. I'll try to implement it.

src/Win32Util.cpp
```diff
@@ -67,9 +69,11 @@ argv_to_string(const char* const* argv, const std::string& prefix)
...
-        ++bs;
-        break;
-      // Fallthrough.
+        if (!escape_backslashes) {
+          ++bs;
+          break;
+        }
+        // Fallthrough.
       case '"':
         bs = (bs << 1) + 1;
```
asalwa, line 78: @jrosdahl Could you explain the logic of `bs` variable? I made the change, it works for me, but the be honest I don't understand well why `bs` is manipulated in such complex way.

asalwa:
I've removed "Draft" tag. I believe this branch is in good shape to merge. It works well for big project on Windows with a lot of very long compilation commands. I've resolved issues we discussed above according to my taste/experience. Feel free to suggest required changes.

src/execute.cpp
jrosdahl: Sure, steps towards improving internal architecture are welcome. The Windows port has been bolted on by several different contributors during the years. I have accepted the contributions in their current form since (a) I feel they have been unintrusive enough and (b) I know it's a large effort to restructure code that is/was inherently Unix-only (n.b.: not _Linux_-only) from the start, so I can't require more elaborate restructuring from contributors. I also often simply don't have time to do it myself. It's just how it is.

That said, I see nothing weird with using a predetermined temporary directory for temporary files. That's what most tools do all the time. The `temporary_dir` setting is actually legacy from how ccache worked before version 3.0 (released 2010). The setting could simply be removed, but I guess it still may have _some_ value since potentially large preprocessed output files are written there, so one might still want to be able to configure it. Maybe. I don't expect argument lists to be that large, though.

src/Win32Util.cpp
jrosdahl: > Could you explain the logic of `bs` variable?

No, sorry.

> What is the reasoning behind it?

I have no idea since it wasn't written by me. It was added in 4d7b4810bd3b8b53832605927c100a5ba4a60fb1 with a reference to <http://msdn.microsoft.com/en-us/library/17w5ykft.aspx>. The Windows parts of ccache are in practice maintained by ccache community members who care about Windows, now including you. :slightly_smiling_face:

jrosdahl:
Thanks.

asalwa:
Thank you for merging and releasing 4.2.1.

src/Win32Util.cpp
asalwa: Thanks for MSDN reference. I'll keep it for further analysis in hypothetical free time :-)

PR #819, opened 2021-03-17
Implement a better Stat::stat(), Stat::lstat() for Windows

nickhutchinson:
`stat()` as implemented in msvcrt.dll (used by mingw-w64) has a number of flaws:
- `st_ino` is always 0 (so `Stat::same_inode_as` gives inaccurate results)
- other file attributes are queried from the parent directory instead of the file itself, and so can be incorrect/stale, e.g. if a file is modified via another hard link.

The implementation in the UCRT is slightly better -- it prefers to query attributes from the file itself if possible, but it still doesn't populate `st_ino` and can silently fall back to querying from the parent directory and giving stale results.

I don't believe the msvcrt.dll source is available (you have to infer its behaviour from e.g. Process Monitor) but the ucrtbase.dll source code is available in the Windows 10 SDK.

Write an implementation of `stat`/`lstat` that sidesteps these issues by using Win32 APIs to query file information. A few implementation notes:

- We do not fall back to querying the parent directory (as the UCRT `stat()` implementation and [Python's `stat()`][1] do) if we can't get a handle to the actual file. If we don't have `FILE_READ_ATTRIBUTES` permissions, we fail with `EACCES` instead of returning incomplete/stale results fetched from the parent directory.
- Windows-specific `Stat::file_attributes()` / `Stat::reparse_tag()` are added to provide the Win32 file attributes and file reparse tag.
- `lstat()` sets the `S_IFLNK` bit in `st_mode` for symlinks. On Windows, there are other symlink-like things ("name surrogate reparse points") such as directory junctions and mountpoints. To identify these, the caller can check the reparse tag to identify a directory junction/mountpoint.
- We use Python's mapping of Win32 error codes to C errno.

[1]: https://github.com/python/cpython/blob/6086ae7fd4aeb4089282189673f9bd0cc33abf9b/Modules/posixmodule.c#L1884

### DIFF
--- a/LICENSE.adoc
+++ b/LICENSE.adoc
@@ -662,6 +662,63 @@ ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
 OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
 -------------------------------------------------------------------------------
 
+src/third_party/win32/winerror_to_errno.h
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The implementation of `winerror_to_errno()` was adapted from
+<https://github.com/python/cpython/blob/1a79785e3e8fea80bcf6a800b45a04e06c787480/PC/errmap.h>
+and has the folowing license text:
+
+-------------------------------------------------------------------------------
+PYTHON SOFTWARE FOUNDATION LICENSE VERSION 2
+
+1. This LICENSE AGREEMENT is between the Python Software Foundation
+("PSF"), and the Individual or Organization ("Licensee") accessing and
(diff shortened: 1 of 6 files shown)


jrosdahl:
I'm very happy that you care about making ccache work robustly on Windows. Thanks!

At the same time I'm sad and surprised that so much Windows-specific code is needed due to bad portability layer implementations.

nickhutchinson:
> I'm very happy that you care about making ccache work robustly on Windows. Thanks!
> 
> At the same time I'm sad and surprised that so much Windows-specific code is needed due to bad portability layer implementations.

I hear you. Perhaps this will improve if/when you're able to move ccache to C++17? Visual Studio's C standard library is pretty awful as we've seen, but its `std::filesystem` is pretty good, and having `std::filesystem`would probably remove the need for the `Stat` class entirely. (No idea about mingw-w64 though.)

PR #835, opened 2021-04-11
test: Use `env` instead of `compgen -e`

MetaDark:
`compgen` isn't available in non-interactive bash, so this change just updates `test/run` to use `env` instead.

### DIFF
--- a/test/run
+++ b/test/run
@@ -346,11 +346,11 @@ expect_perm() {
 }
 
 reset_environment() {
-    while IFS= read -r name; do
+    while IFS='=' read -r name value; do
         if [[ $name =~ ^CCACHE_[A-Z0-9_]*$ ]]; then
             unset $name
         fi
-    done < <(compgen -e)
+    done < <(env)
 
     unset GCC_COLORS
     unset TERM


jrosdahl:
Thanks for the pull request.

One elegant thing with using `compgen -e` is that it handles values with newlines correctly in contrast to the method of parsing output from `env`:

```console
% export EXAMPLE="foo
bar"
% env | grep ^bar
bar
```

I guess that's just a minor thing, though – if it fixes a real problem I'm fine with that. Another potential way to retrieve the environment variable names could be to run `export` but I'm not sure if the output format can vary between bash versions.

Regarding the real problem: I can't reproduce any problem with using `compgen -e`. On systems I have available (e.g. Ubuntu 20.04 with bash 5 and 14.04 with bash 4) it seems to be available in non-interactive mode as well. Any hints on how to reproduce it or ideas on why things are different in NixOS? How did you determine that it doesn't work?

MetaDark:
Oh, sorry @jrosdahl. I should have phrased it better. On NixOS `nixpkgs.bash` refers to a build without `readline` support enabled, whereas `nixpkgs.bashInteractive` refers to a build with readline support enabled https://github.com/NixOS/nixpkgs/blob/71e2a4f48b3069bf306aad2cd481872f244d9f4b/pkgs/shells/bash/4.4.nix#L61.

The build with `readline` support has the compgen function:
```bash
>/nix/store/bbp2qxxsihmx87w30qvi1c020vggkdqn-bash-interactive-4.4-p23/bin/bash -c 'compgen -e'
AR
AS
CC
CONFIG_SHELL
CXX
...
```

but the build without it doesn't:
```bash
/nix/store/jdi2v7ir1sr6vp7pc5x0nhb6lpcmg6xg-bash-4.4-p23/bin/bash -c 'compgen -e'
/nix/store/jdi2v7ir1sr6vp7pc5x0nhb6lpcmg6xg-bash-4.4-p23/bin/bash: compgen: command not found
```

It looks like Ubuntu doesn't make this distinction, and only provides a bash package with `readline` support. NixOS uses the non-interactive build for building packages, but provides the interactive version for user shells.

I didn't consider that there would be a problem with newlines though, so I think it would make sense to stick to `compgen -e`. We can just update the Nix package to use `bashInteractive` for the test.

Thank you!

PR #838, opened 2021-04-16
Fix performance impact from 'Canonicalize path for comparison; Clang …

binli71:
…uses ./header.h'

Problem Summary and RCA:
The 'ctx.included_files.find(path)' was done before 'Canonicalize path for comparison; Clang uses ./header.h' which using './header.h'.
The following operation, such as 'ctx.included_files.emplace(path, d)' all after 'Canonicalize path' which using 'header.h'.
This mismatch will cause './header.h' being added over and over which leading to performance hit.

Fix summary:
Using 'Canonicalize path' in 'ctx.included_files.find(path)'.

### DIFF
--- a/src/ccache.cpp
+++ b/src/ccache.cpp
@@ -398,14 +398,14 @@ do_remember_include_file(Context& ctx,
     return true;
   }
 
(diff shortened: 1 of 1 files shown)


jrosdahl:
Thanks.

PR #840, opened 2021-04-27
Fix detection of pthread features

liljenzin:
Add the -pthread flag to the compiler when building the test
program. Otherwise the compiler will build with wrong defines.

Fixes failing unit tests in ubuntu 18.04.


liljenzin:
Interesting that all tests seem to have passed here also before my change, while the inode_cache unit test failed on my local virtual machine running ubuntu 18.04. Perhaps depending on the exact version of system headers and libraries? Not passing -pthread to the compiler is wrong anyway.

jrosdahl:
Thanks.

PR #842, opened 2021-04-30
CMake: make build of documentation optional

ynezz:
So we don't need to support corner cases as for example one fixed in
commit f6202db308e3 ("doc/MANUAL.adoc: Don't use non-ASCII quotes
(#761)") when the documentation is actually not needed at all as ccache
is used as a build tool only.

Signed-off-by: Petr Štetiar <ynezz@true.cz>

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -133,7 +133,10 @@ target_link_libraries(ccache PRIVATE standard_settings standard_warnings ccache_
 #
 # Documentation
 #
-add_subdirectory(doc)
+option(ENABLE_DOCUMENTATION "Enable documentation" ON)
+if(ENABLE_DOCUMENTATION)
+  add_subdirectory(doc)
+endif()
 
 #
 # Installation


ynezz:
ccache is being used in OpenWrt and I find it quite strange to backport [fixes for documentation](https://patchwork.ozlabs.org/project/openwrt/patch/mailman.1136.1619731704.1230.openwrt-devel@lists.openwrt.org/) in order to use ccache as a build tool, where the documentation is not needed at all.

jrosdahl:
Sure, thanks.

AlexanderLanin:
@ynezz why do you backport changes and not use ccache/ccache directly? Do you have some OpenWrt specific adaptions? Why not merge them here?

ynezz:
> @ynezz why do you backport changes and not use ccache/ccache directly?

@AlexanderLanin OpenWrt [prefers](https://github.com/openwrt/openwrt/blob/master/tools/ccache/Makefile) to use ccache releases, so we use latest 4.2.1 for current development/master branch, 4.1 for 21.02 upcoming stable release and 3.7.2 for current 19.07 stable release. AFAIK we would need to backport [proposed fix](https://patchwork.ozlabs.org/project/openwrt/patch/mailman.1136.1619731704.1230.openwrt-devel@lists.openwrt.org/) as it's not part of any release yet.

Every such patch then becomes maintenance burden and eventually might break some exotic setups, so we try avoid doing such backports into supported releases, disabling the documentation altogether seems like more robust and future proof solution.

> Do you have some OpenWrt specific adaptions? Why not merge them here?

So far it's just single patch for [GCC_HONOUR_COPTS](https://github.com/openwrt/openwrt/blob/master/tools/ccache/patches/100-honour-copts.patch) and IIRC this is project specific and can't be upstreamed.

jrosdahl:
> AFAIK we would need to backport proposed fix as it's not part of any release yet.

The fix is part of ccache 4.2 and newer, but there will be no ccache 4.1.1 release with the fix if that's what you mean.

Bastian-Krause:
@jrosdahl 
> The fix is part of ccache 4.2 and newer, but there will be no ccache 4.1.1 release with the fix if that's what you mean.

You mean 4.3, not 4.2, right?

jrosdahl:
> You mean 4.3, not 4.2, right?

No, I do mean 4.2, released three months ago. The fix was made in PR #761 (commit f6202db308e3).

Bastian-Krause:
Thanks for the clarification!

PR #843, opened 2021-04-30
Support `-specs file.specs` and `--specs file.specs`

joshtriplett:
ccache currently supports specs files supplied via `-specs=file.specs`
and `--specs=file.specs`, but using a space instead of an `=` will cause
ccache to error out on the subsequent .specs file with
"unsupported source language".

Add support for `-specs file.specs` and `--specs file.specs`.


src/ccache.cpp
```diff
@@ -1652,8 +1652,21 @@ calculate_result_name(Context& ctx,
     }
 
     if (Util::starts_with(args[i], "-specs=")
-        || Util::starts_with(args[i], "--specs=")) {
-      std::string path = args[i].substr(args[i].find('=') + 1);
+        || Util::starts_with(args[i], "--specs=")
+        || args[i] == "-specs"
+        || args[i] == "--specs") {
```
jrosdahl, line 1657: A hint to clang-format:

```suggestion
        || (args[i] == "-specs" || args[i] == "--specs")) {
```

jrosdahl:
Thanks.

PR #844, opened 2021-05-04
doc: allow disabling docs/man page generation

Bastian-Krause:
The assumption that HTML documentation and manual pages should be generated if the required tools (asciidoc) are present is not always true. So add a cmake option that allows disabling the docs/man page generation explicitly. The default is to generate docs/man pages like before.


jrosdahl:
Thanks for the PR.

A similar option has already been implemented in #842. Does that look OK to you?

Bastian-Krause:
#842 looks fine to me, closing this.

jrosdahl:
> Edit: looks like I can't close my own PR.

That was due to a bug in the bot, sorry about that. Should be fixed now.

PR #850, opened 2021-05-21
Add support for clang's --config option

tstellar:
This is more or less equivalent to gcc's --specs option except that the
contents of the file have different syntaxes.

### DIFF
--- a/src/ccache.cpp
+++ b/src/ccache.cpp
@@ -1657,7 +1657,8 @@ calculate_result_name(Context& ctx,
 
     if (Util::starts_with(args[i], "-specs=")
         || Util::starts_with(args[i], "--specs=")
-        || (args[i] == "-specs" || args[i] == "--specs")) {
+        || (args[i] == "-specs" || args[i] == "--specs")
+        || args[i] == "--config") {
       std::string path;
       size_t eq_pos = args[i].find('=');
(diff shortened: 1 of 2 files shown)


jrosdahl:
Thanks.

isbadawi:
Based on https://clang.llvm.org/docs/UsersManual.html#configuration-files, it seems to me that `--config` should be processed more like @-files (which are expanded with `Args::from_atfile` and substituted back into the arg list to process the contents). The config file might contain flags that ccache wants to treat specially, for example by hashing additional input files (such as with `-fplugin=`). Maybe it could be implemented by as a new `Args::AtFileFormat` -- there are some differences like supporting `#` comments.

jrosdahl:
> Based on [clang.llvm.org/docs/UsersManual.html#configuration-files](https://clang.llvm.org/docs/UsersManual.html#configuration-files), it seems to me that `--config` should be processed more like @-files (which are expanded with `Args::from_atfile` and substituted back into the arg list to process the contents). The config file might contain flags that ccache wants to treat specially, for example by hashing additional input files (such as with `-fplugin=`). Maybe it could be implemented as a new `Args::AtFileFormat` -- there are some differences like supporting `#` comments.

Seems reasonable. Please open a new issue instead of leaving a comment here.

PR #852, opened 2021-05-24
Enable minitrace tracing and stop it from crashing

### LINKED ISSUE #851: ccache internal tracing is not working
### How to reproduce ###

1. build ccache with `ENABLE_TRACING=ON` 
2. `export CCACHE_INTERNAL_TRACE=1`
3. ccache cc -c foo.c

### Actual behavior ###

The generated file is empty (invalid JSON)

### Expected behavior ###

It should contain JSON data, for tracing

### Environment ###

ccache version 4.3 (`v4.3`)

g++ (Ubuntu 9.3.0-17ubuntu1~20.04) 9.3.0

----

The MTR_ENABLED define is never passed to the minitrace library (i.e. minitrace.c), so it is effectively disabled.

What is worse, is that when enabled it causes a segfault! This is because all the strings have been deallocated...

```
Program received signal SIGSEGV, Segmentation fault.
__strlen_avx2 () at ../sysdeps/x86_64/multiarch/strlen-avx2.S:96
96	../sysdeps/x86_64/multiarch/strlen-avx2.S: No such file or directory.
(gdb) bt
#0  __strlen_avx2 () at ../sysdeps/x86_64/multiarch/strlen-avx2.S:96
#1  0x00005555556965bc in mtr_flush () at ../src/third_party/minitrace.c:262
#2  0x000055555560c643 in MiniTrace::~MiniTrace (this=0x555555851a80, __in_chrg=<optimized out>) at ../src/MiniTrace.cpp:83
#3  0x00005555555947e4 in std::default_delete<MiniTrace>::operator() (this=0x7fffffffdb28, __ptr=0x555555851a80) at /usr/include/c++/9/bits/unique_ptr.h:81
#4  0x00005555555916ac in std::unique_ptr<MiniTrace, std::default_delete<MiniTrace> >::~unique_ptr (this=0x7fffffffdb28, __in_chrg=<optimized out>)
    at /usr/include/c++/9/bits/unique_ptr.h:292
#5  0x00005555555ef850 in Context::~Context (this=0x7fffffffd510, __in_chrg=<optimized out>) at ../src/Context.cpp:43
#6  0x0000555555573437 in cache_compilation (argc=4, argv=0x7fffffffdda8) at ../src/ccache.cpp:2320
#7  0x0000555555578057 in ccache_main (argc=4, argv=0x7fffffffdda8) at ../src/ccache.cpp:2843
#8  0x0000555555560ead in main (argc=4, argv=0x7fffffffdda8) at ../src/main.cpp:24
```

afbjorklund:
Need to make sure to pass the -DMTR_ENABLED also when compiling
the library and not only in the ccache internal config.h file.

Also the strings used need to be still around when the destructor
is called, so avoid using local strings and any strdup strings.

Closes #851

### DIFF
--- a/src/third_party/CMakeLists.txt
+++ b/src/third_party/CMakeLists.txt
@@ -12,6 +12,7 @@ endif ()
 
 if(ENABLE_TRACING)
   target_sources(third_party_lib PRIVATE minitrace.c)
+  set_source_files_properties(minitrace.c PROPERTIES COMPILE_DEFINITIONS "MTR_ENABLED;_POSIX_C_SOURCE=200809L")
 endif()
 
 set(xxhdispatchtest [=[
--- a/src/MiniTrace.cpp
+++ b/src/MiniTrace.cpp
@@ -72,7 +72,8 @@ MiniTrace::MiniTrace(const ArgsInfo& args_info)
   m_tmp_trace_file = tmp_file.path;
 
   mtr_init(m_tmp_trace_file.c_str());
-  MTR_INSTANT_C("", "", "time", FMT("{:f}", time_seconds()).c_str());
+  m_start_time = FMT("{:f}", time_seconds());
+  MTR_INSTANT_C("", "", "time", m_start_time.c_str());
   MTR_META_PROCESS_NAME("ccache");
   MTR_START("program", "ccache", m_trace_id);
 }
--- a/src/MiniTrace.hpp
+++ b/src/MiniTrace.hpp
@@ -38,6 +38,7 @@ private:
   const ArgsInfo& m_args_info;
   const void* const m_trace_id;
   std::string m_tmp_trace_file;
+  std::string m_start_time;
 };
 
 #endif


jrosdahl (commented):
Thanks for the fix.

My guess is that you might be the only user of the minitrace tracing mode in practice... My intention was nevertheless to keep it in the code as long as it isn't too much in the way. It's at least tested to build without failure in CI. :slightly_smiling_face: Even better would be an automated test case that verifies some basic sanity of the output, otherwise it might very well break again.

src/third_party/minitrace.h
```diff
@@ -139,8 +139,8 @@ void internal_mtr_raw_event_arg(const char *category, const char *name, char ph,
 
 // Metadata. Call at the start preferably. Must be const strings.
 
-#define MTR_META_PROCESS_NAME(n) internal_mtr_raw_event_arg("", "process_name", 'M', 0, MTR_ARG_TYPE_STRING_COPY, "name", (void *)(n))
-#define MTR_META_THREAD_NAME(n) internal_mtr_raw_event_arg("", "thread_name", 'M', 0, MTR_ARG_TYPE_STRING_COPY, "name", (void *)(n))
+#define MTR_META_PROCESS_NAME(n) internal_mtr_raw_event_arg("", "process_name", 'M', 0, MTR_ARG_TYPE_STRING_CONST, "name", (void *)(n))
+#define MTR_META_THREAD_NAME(n) internal_mtr_raw_event_arg("", "thread_name", 'M', 0, MTR_ARG_TYPE_STRING_CONST, "name", (void *)(n))
```
jrosdahl, line 143: Could you please explain this part of the fix? Why doesn't it work to copy the process string?

I would prefer not to modify third party source code if possible.

afbjorklund: Not sure, actually. It used to work before (C), but it seems the copy happens "too late" in the process (C++).
My initial feeling was that it should be totally the opposite, "copy" would be safe and "const" dangerous...

```C
        case MTR_ARG_TYPE_STRING_CONST: ev->a_str = (const char*)arg_value; break;
        case MTR_ARG_TYPE_STRING_COPY: ev->a_str = strdup((const char*)arg_value); break;
```

It's like those "strdup" strings are deallocated as part of the shutdown process, before destructor is called.
Maybe there is a way to deallocate it more explicitly, that would avoid having to patch third party code.

Or do some other fix, that could go upstream.

afbjorklund:
There were some partial changes added to enable tracing for the http backend, which was why I was looking at it...

Some end-to-end documentation and a test case (but with the special build option) would probably be a good idea.

src/third_party/minitrace.h
afbjorklund: Adding an explicit `mtr_ flush` didn't help at all, and valgrind says that both strings are leaked (never freed).

So I don't know what could cause `strlen` to segfault like that, when called on a string literal or a `c_str()` ?

afbjorklund:
The strings are not modified after `strdup`, but still they are crashing `strlen`

```
==249303== 6 bytes in 1 blocks are definitely lost in loss record 1 of 2
==249303==    at 0x483B7F3: malloc (in /usr/lib/x86_64-linux-gnu/valgrind/vgpreload_memcheck-amd64-linux.so)
==249303==    by 0x4D3450E: strdup (strdup.c:42)
==249303==    by 0x24AC48: internal_mtr_raw_event_arg (minitrace.c:391)
==249303==    by 0x129260: do_cache_compilation(Context&, char const* const*) (ccache.cpp:2451)
==249303==    by 0x1273FD: cache_compilation(int, char const* const*) (ccache.cpp:2329)
==249303==    by 0x12C0AA: ccache_main(int, char const* const*) (ccache.cpp:2855)
==249303==    by 0x114EAC: main (main.cpp:24)
==249303== 
==249303== 7 bytes in 1 blocks are definitely lost in loss record 2 of 2
==249303==    at 0x483B7F3: malloc (in /usr/lib/x86_64-linux-gnu/valgrind/vgpreload_memcheck-amd64-linux.so)
==249303==    by 0x4D3450E: strdup (strdup.c:42)
==249303==    by 0x24AC48: internal_mtr_raw_event_arg (minitrace.c:391)
==249303==    by 0x1C05C1: MiniTrace::MiniTrace(ArgsInfo const&) (MiniTrace.cpp:78)
==249303==    by 0x145671: std::unique_ptr<MiniTrace, std::default_delete<MiniTrace> > std::make_unique<MiniTrace, ArgsInfo&>(ArgsInfo&) (StdMakeUnique.hpp:28)
==249303==    by 0x1250BD: initialize(Context&, int, char const* const*) (ccache.cpp:2085)
==249303==    by 0x127369: cache_compilation(int, char const* const*) (ccache.cpp:2322)
==249303==    by 0x12C0AA: ccache_main(int, char const* const*) (ccache.cpp:2855)
==249303==    by 0x114EAC: main (main.cpp:24)
==249303== 
```

The workaround was to not call strdup at all, but maybe it can be made to work...

afbjorklund:
Some kind of bug with `strdup`.

For "const" strings it goes fine:

```
(gdb) print((char *) arg_value)
$1 = 0x55555584efb0 "1622365915.738774"
(gdb) print(ev->a_str)
$2 = 0x55555584efb0 "1622365915.738774"
```

For "copy" strings, it goes horribly wrong:
```
(gdb) print((char *) arg_value)
$3 = 0x5555557d016b "ccache"
(gdb) print(ev->a_str)
$4 = 0x55851d70 <error: Cannot access memory at address 0x55851d70>
```

Even without the rest.

afbjorklund:
Apparently it was another missing define, causing the function to return `int`.

```
../src/third_party/minitrace.c:391:45: error: implicit declaration of function ‘strdup’; did you mean ‘strcmp’? [-Werror=implicit-function-declaration]
  391 |  case MTR_ARG_TYPE_STRING_COPY: ev->a_str = strdup((const char*)arg_value); break;
      |                                             ^~~~~~
      |                                             strcmp
```

But all those warnings were hidden by default (`-Wno-implicit-function-declaration`)

Apparently `strdup` requires `_XOPEN_SOURCE >= 500` in order to be defined...

src/third_party/CMakeLists.txt
```diff
@@ -12,6 +12,7 @@ endif ()
 
 if(ENABLE_TRACING)
   target_sources(third_party_lib PRIVATE minitrace.c)
+  set_source_files_properties(minitrace.c PROPERTIES COMPILE_DEFINITIONS "MTR_ENABLED;_POSIX_C_SOURCE=200809L")
```
jrosdahl, line 15: I think that it would be better to apply `config.h` instead of adding this special case, but I'll make that as a follow-up commit.

jrosdahl (commented):
> But all those warnings were hidden by default (`-Wno-implicit-function-declaration`)

Right, I unfortunately missed this in my review of #573 – it makes no sense to hide such warnings.

jrosdahl:
Thanks.

PR #854, opened 2021-06-01
Add stats log file, logging input file and result

### LINKED ISSUE #262: Environment variable or configuration option to store statistics in a different location

afbjorklund:
This is a reimplementation of the original feature, 4e2799df1d3b617525235fa43f8ab048397ddbae

It adds an optional stats log, that will log the input files and their results (only)

```
# foo.c
cache_miss
# bar.c
cache_miss
# foo.c
direct_cache_hit
# bar.c
direct_cache_hit
```
_all lines starting with `#` are regarded as comments, and ignored as results_

It also adds a new command `--log-stats`, that will parse and present this file.

```
cache hit (direct)                     2
cache miss                             2
cache hit rate                     50.00 %
```

Closes #262

### DIFF
--- a/doc/MANUAL.adoc
+++ b/doc/MANUAL.adoc
@@ -153,6 +153,11 @@ compiler's documentation.
     (environment variable, configuration file or compile-time default) in
     human-readable format.
 
+*`--show-log-stats`*::
+
+    Print statistics counters from the stats log in human-readable format.
+    See <<config_stats_log,*stats_log*>>.
+
 *`-s`*, *`--show-stats`*::
 
     Print a summary of configuration and statistics counters in human-readable
@@ -848,6 +853,7 @@ information.
 
     If set to a file path, ccache will write statistics counter updates to the
     specified file. This is useful for getting statistics for individual builds.
+    To show a summary of the current stats log, use: `ccache --show-log-stats`.
 +
 NOTE: Lines in the stats log starting with a hash sign (`#`) are comments.
 
--- a/doc/MANUAL.adoc
+++ b/doc/MANUAL.adoc
@@ -844,6 +844,13 @@ information.
     If true, ccache will update the statistics counters on each compilation.
     The default is true.
 
+[[config_stats_log]] *stats_log* (*CCACHE_STATSLOG*)::
+
+    If set to a file path, ccache will write statistics counter updates to the
+    specified file. This is useful for getting statistics for individual builds.
++
+NOTE: Lines in the stats log starting with a hash sign (`#`) are comments.
+
 [[config_temporary_dir]] *temporary_dir* (*CCACHE_TEMPDIR*)::
 
     This option specifies where ccache will put temporary files. The default is
--- a/src/Statistics.cpp
+++ b/src/Statistics.cpp
@@ -28,9 +28,13 @@
 #include "exceptions.hpp"
 #include "fmtmacros.hpp"
 
-const unsigned FLAG_NOZERO = 1; // don't zero with the -z option
-const unsigned FLAG_ALWAYS = 2; // always show, even if zero
-const unsigned FLAG_NEVER = 4;  // never show
+#include <fstream>
+#include <unordered_map>
+
+const unsigned FLAG_NOZERO = 1;     // don't zero with the -z option
+const unsigned FLAG_ALWAYS = 2;     // always show, even if zero
+const unsigned FLAG_NEVER = 4;      // never show
+const unsigned FLAG_NOSTATSLOG = 8; // don't show for statslog
 
 using nonstd::nullopt;
 using nonstd::optional;
@@ -175,11 +179,14 @@ const StatisticsField k_statistics_fields[] = {
   STATISTICS_FIELD(bad_output_file, "could not write to output file"),
   STATISTICS_FIELD(no_input_file, "no input file"),
   STATISTICS_FIELD(error_hashing_extra_file, "error hashing extra file"),
-  STATISTICS_FIELD(cleanups_performed, "cleanups performed", FLAG_ALWAYS),
-  STATISTICS_FIELD(files_in_cache, "files in cache", FLAG_NOZERO | FLAG_ALWAYS),
+  STATISTICS_FIELD(
+    cleanups_performed, "cleanups performed", FLAG_NOSTATSLOG | FLAG_ALWAYS),
+  STATISTICS_FIELD(files_in_cache,
+                   "files in cache",
+                   FLAG_NOZERO | FLAG_NOSTATSLOG | FLAG_ALWAYS),
   STATISTICS_FIELD(cache_size_kibibyte,
                    "cache size",
-                   FLAG_NOZERO | FLAG_ALWAYS,
+                   FLAG_NOZERO | FLAG_NOSTATSLOG | FLAG_ALWAYS,
                    format_size_times_1024),
   STATISTICS_FIELD(obsolete_max_files, "OBSOLETE", FLAG_NOZERO | FLAG_NEVER),
   STATISTICS_FIELD(obsolete_max_size, "OBSOLETE", FLAG_NOZERO | FLAG_NEVER),
@@ -212,6 +219,34 @@ read(const std::string& path)
     counters.set_raw(i, value);
     ++i;
     str = end;
+  }
+
+  return counters;
+}
+
+Counters
+read_log(const std::string& path)
+{
+  Counters counters;
+
+  std::unordered_map<std::string, Statistic> m;
+  for (const auto& field : k_statistics_fields) {
+    m[field.id] = field.statistic;
+  }
+
+  std::ifstream in(path);
+  std::string line;
+  while (std::getline(in, line, '\n')) {
+    if (line[0] == '#') {
+      continue;
+    }
+    auto search = m.find(line);
+    if (search != m.end()) {
+      Statistic statistic = search->second;
+      counters.increment(statistic, 1);
+    } else {
+      LOG("Unknown statistic: {}", line);
+    }
   }
 
   return counters;
@@ -310,6 +345,16 @@ zero_all_counters(const Config& config)
 }
 
 std::string
+format_stats_log(const Config& config)
+{
+  std::string result;
+
+  result += FMT("{:36}{}\n", "stats log", config.stats_log());
+
+  return result;
+}
+
+std::string
 format_config_header(const Config& config)
 {
   std::string result;
@@ -323,7 +368,9 @@ format_config_header(const Config& config)
 }
 
 std::string
-format_human_readable(const Counters& counters, time_t last_updated)
+format_human_readable(const Counters& counters,
+                      time_t last_updated,
+                      bool from_log)
 {
   std::string result;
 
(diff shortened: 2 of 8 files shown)


src/Logging.cpp
```diff
@@ -176,6 +188,36 @@ bulk_log(string_view message)
...
+      path = FMT("{}/stats{}.log", path, getpid() % 256);
+    }
+    Lockfile lock(path);
+    if (!lock.acquired()) {
+      LOG("Failed to acquire lock for {}", path);
+      return;
+    }
+    File file(path, "a");
+    if (file) {
+      (void)fprintf(*file, "# %.*s\n", (int)input.size(), input.data());
```
jrosdahl, line 213: Tip: Use something like `PRINT(*file, "{:{}}\n", input.data(), input.size());`.

src/Logging.cpp
```diff
@@ -141,6 +148,11 @@ init(const Config& config)
   }
 #endif
 
+  if (!config.stats_log().empty()) {
+    use_statslog = true;
+    statslog_path = config.stats_log();
```
jrosdahl, line 153: This line won't be reached if `config.log_file() == "syslog"`, so move it before the syslog stuff?

src/ccache.cpp
```diff
@@ -146,6 +146,8 @@ Options for scripting or debugging:
     -k, --get-config KEY       print the value of configuration key KEY
         --hash-file PATH       print the hash (160 bit BLAKE3) of the file at
                                PATH
+        --log-stats            print statistics counters from the stats log
```
jrosdahl, line 149: I would appreciate if  documentation of the new command line option and the configuration option could be added to the manual. Also, some test cases would be awesome!

jrosdahl (commented):
Using a log instead of incrementing statistics counters sounds fine!

src/ccache.cpp
```diff
@@ -2207,6 +2209,9 @@ finalize_stats_and_trigger_cleanup(Context& ctx)
     if (result) {
       LOG("Result: {}", *result);
     }
+    if (Logging::results()) {
```
jrosdahl, line 2212: The `!config.log_file().empty() || config.debug()` condition above makes stats logging enabled only if a log file or debug is specified. Is that intentional? I would expect stats logging to not depend on those configuration options.

In fact, how about not putting stats log code in Logging at all since it isn't really related?

Here's a sketch on how I think it could look like:

```c++
if (result && !config.stats_log.empty()) {
  const auto result_id = Statistics::get_result_id(); // new function: optional<string> Statistics::get_result_id();
  ASSERT(result_id);
  File stats_log(config.stats_log(), "a");
  PRINT(*stats_log, "# {}\n", ctx.args_info.input_file);
  PRINT(*stats_log, "{}\n", *result_id);
}
```

afbjorklund: The stats log was supposed to be small enough to run _without_ debugging, so that was not intentional.

src/Logging.cpp
```diff
@@ -176,6 +188,36 @@ bulk_log(string_view message)
...
+}
+
+void
+log_result(const std::string& input, nonstd::optional<std::string> result)
+{
+  if (result) {
+    auto path = statslog_path;
+    if (Util::ends_with(path, ".d")) {
+      Util::ensure_dir_exists(path);
+      path = FMT("{}/stats{}.log", path, getpid() % 256);
```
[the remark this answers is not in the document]
afbjorklund: That sounds OK, it might have been a case of premature optimization (or perhaps paranoia, when I added the locks)

Removing the lockfile while also removing the directory sounds safe enough. I think it was mostly "unsafe" for longer lines ?

src/ccache.cpp
```diff
@@ -2207,6 +2209,9 @@ finalize_stats_and_trigger_cleanup(Context& ctx)
...
+      Logging::log_result(ctx.args_info.input_file, result);
```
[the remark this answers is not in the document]
afbjorklund: The original idea was to keep it human readable, but that was also before I added the parser.

Logging the source file was supposed to help with finding those files that _always_ misses cache.
Might take that out as well, and only log the statistics ? One could do a special trouble-shooting build.

i.e. use the statslog for the release build, and do the regular log for the debug build (to find the files)

afbjorklund:
It is probably obvious, but if anyone else is new to this: the stats log is a _complement_ to the regular cache counters

They keep the statistics per **build** (or similar, you decide) while the regular counters keep the statistics per **cache**.

src/Logging.cpp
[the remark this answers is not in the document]
jrosdahl: >  I think it was mostly "unsafe" for longer lines ?

Correct.

src/ccache.cpp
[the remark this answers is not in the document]
jrosdahl: > The original idea was to keep it human readable, but that was also before I added the parser.

Right, but the IDs are also human-readable. It's the ones you get with `ccache --print-stats`.

> Logging the source file was supposed to help with finding those files that always misses cache.

Yes, I think that's a good idea and I don't mind those comments.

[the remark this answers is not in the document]
afbjorklund: Ok, let's use the IDs and leave the comments.

For some special purposes we even logged the complete command lines. Those are usually enormous and would not be atomic and so on, but were sometimes useful when trying to investigate simple cache misses without diving too much into detail. Usually it was someone adding a global macro with a timestamp, or something similar. "Instant cache invalidation". Having the source file in the stats log is a good compromise, it should give an indication of where to look at least.

afbjorklund:
I think most of the review comments have been addressed now, left to do is to squash and rebase.

afbjorklund:
As mentioned earlier, if we want to make this more "efficient" we could use statsd similar to syslog ?

In order to **not** have to open and append to any files, but just send some UDP packets. (see ~~#410~~)

* https://github.com/romanbsd/statsd-c-client
* https://github.com/jbuchbinder/statsd-c

Available as an option, that is.

Files are OK too, as the default.

jrosdahl:
> As mentioned earlier, if we want to make this more "efficient" we could use statsd similar to syslog ?

As mentioned in https://github.com/ccache/ccache/pull/410#issuecomment-490255532, that's not a solution to #262 but to another use case, right?

src/Counters.hpp
```diff
@@ -30,6 +30,7 @@ class Counters
 {
 public:
   Counters();
+  Counters(bool from_log);
```
jrosdahl, line 33: Suggestion: Add a `bool from_log` parameter to `Statistics::format_human_readable` instead. Then we can let `Counters` remain unaware of its origin.

src/Statistics.cpp
```diff
@@ -306,6 +388,14 @@ format_human_readable(const Config& config)
       continue;
     }
 
+    // don't show cache directory info if reading from a log
+    if (counters.from_log()
+        && (statistic == Statistic::cleanups_performed
+            || statistic == Statistic::files_in_cache
+            || statistic == Statistic::cache_size_kibibyte)) {
```
jrosdahl, line 395: I suggest adding a flag in `k_statistics_counters` for this instead. Maybe named `FLAG_NOSTATSLOG`?

unittest/test_Statistics.cpp
```diff
@@ -80,6 +80,17 @@ TEST_CASE("Read future counters")
   }
 }
 
+TEST_CASE("Read log")
```
jrosdahl, line 83: Thanks for writing tests!

doc/MANUAL.adoc
```diff
@@ -858,6 +863,13 @@ information.
     If true, ccache will update the statistics counters on each compilation.
     The default is true.
 
+[[config_stats_log]] *stats_log* (*CCACHE_STATSLOG*)::
+
+    If set to a file path, ccache will write results to the specified file.
+    This is useful for getting statistics for individual builds.
```
jrosdahl, line 869: Perhaps add a hint here on how to summarize the stats log?

src/Statistics.cpp
```diff
@@ -215,6 +217,36 @@ read(const std::string& path)
...
+  data = Util::read_file(path);
+
+  std::stringstream ss(data);
+  std::string line;
+  while (getline(ss, line, '\n')) {
+    if (line[0] == '#') {
+      continue;
+    }
+    Statistic statistic = Statistic::none;
+    for (const auto& field : k_statistics_fields) {
```
jrosdahl, line 235: Not sure if it's a major problem in practice, but it would be nice to avoid having to loop once per logged result.

Here's an idea: What if we change the log format to something like

```
# direct_cache_hit path/to/file.c
22
```

or

```
# path/to/file.c
22 # direct_cache_hit
```

(where 22 is the index in `Counters` for `direct_cache_hit`) instead and then call `counters.increment_raw(index)` for each logged result?

jrosdahl (commented):
Nice, starting to look good!

src/Statistics.cpp
jrosdahl: This is just a minor thing; I'm OK with keeping the current format. Perhaps it's even better not to leak implementation details like array index into the log.

doc/MANUAL.adoc
```diff
@@ -118,6 +118,11 @@ Common options
 
     Print a summary of command line options.
 
+*`--log-stats`*::
```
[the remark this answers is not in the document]
afbjorklund: Ok

src/Statistics.cpp
```diff
@@ -320,6 +410,55 @@ format_human_readable(const Config& config)
...
+}
+
+static inline bool
+should_count(const StatisticsField& field)
+{
+  switch (field.statistic) {
+  case Statistic::stats_zeroed_timestamp:
+  case Statistic::cleanups_performed:
+  case Statistic::files_in_cache:
+  case Statistic::cache_size_kibibyte:
```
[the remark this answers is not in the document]
afbjorklund: I can remove the counter

src/Statistics.cpp
afbjorklund: We can make a lookup table for the parsing, the idea was to keep the stats log human readable (and version-proof)

But the number is probably readable anyway, and as long as you don't remove indexes it would be "safe" as well.

It would be possible to remove the file name and write only the result code (like in the `stats` files, their order)

Maybe will try with some bigger builds first, and see how big the stats log gets for those. Like, if it is a problem ?

afbjorklund:
> As mentioned in [#410 (comment)](https://github.com/ccache/ccache/pull/410#issuecomment-490255532), that's not a solution to #262 but to another use case, right?

Right, it was just a minor reflection that it would be easier to accomplish with this new code base...

src/Statistics.cpp
jrosdahl: Right, let's leave it as is and maybe optimize it with a lookup table later if needed. :+1:

src/Statistics.cpp
```diff
@@ -215,6 +223,37 @@ read(const std::string& path)
   return counters;
 }
 
+Counters
+read_log(const std::string& path)
+{
+  Counters counters;
+
+  std::string data;
+  data = Util::read_file(path);
```
jrosdahl, line 232: Only one minor thing left on my wish list: I suggest reading the file in a streaming fashion instead (using `std::ifstream` instead of `Util::read_file` + `std::stringstream`).

jrosdahl (commented):
Thanks for the lookup table!

jrosdahl:
Thanks!

PR #861, opened 2021-06-17
Explicitly select C++ when using -fcxx-modules

r-burns:
Typically, clang is able to auto-detect C++ files and switch to C++
mode, even when `argv[0]` is simply `clang` (without the `++`).

However, Nixpkgs provides `clang` as a wrapper script over the
underlying compiler binary, and only enables C++ standard library
include paths when it thinks it's being invoked as a C++ compiler:

https://github.com/NixOS/nixpkgs/blob/961cae37edeb34b5dec8060e49e5abbde3af9c00/pkgs/build-support/cc-wrapper/cc-wrapper.sh#L28

https://github.com/NixOS/nixpkgs/blob/961cae37edeb34b5dec8060e49e5abbde3af9c00/pkgs/build-support/cc-wrapper/cc-wrapper.sh#L54-L55

https://github.com/NixOS/nixpkgs/blob/961cae37edeb34b5dec8060e49e5abbde3af9c00/pkgs/build-support/cc-wrapper/cc-wrapper.sh#L135-L138

So for the benefit of less clever compiler toolchains which
can't always see that these invocations are using C++ mode,
we explicitly specify `-x c++` so they do the right thing.

See also https://github.com/NixOS/nixpkgs/pull/119176#issuecomment-862853452

### DIFF
--- a/test/suites/modules.bash
+++ b/test/suites/modules.bash
@@ -34,10 +34,10 @@ SUITE_modules() {
     # -------------------------------------------------------------------------
     TEST "fall back to real compiler, no sloppiness"
 
-    $CCACHE_COMPILE -fmodules -fcxx-modules -c test1.cpp -MD
+    $CCACHE_COMPILE -x c++ -fmodules -fcxx-modules -c test1.cpp -MD
     expect_stat "can't use modules" 1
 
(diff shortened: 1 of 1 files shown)


jrosdahl:
Sure, why not. Thanks.

PR #863, opened 2021-06-18
[WIP] Add seek and size function to File

gjasny:
Hello,

for the HTTP cache upload I need to be able to get the filesize of a `FILE` handle as well as seek within.
Due to portability issues I'd like hide the implementation details somewhere.

What do you thing of those two methods? Should they stay in `File` or should I move them right next to `fallocate`?

Thanks,
Gregor

### DIFF
--- a/src/File.hpp
+++ b/src/File.hpp
@@ -37,6 +37,15 @@ public:
   void open(const std::string& path, const char* mode);
   void close();
 
+  enum class Seek {
+    Set = SEEK_SET,
+    Cur = SEEK_CUR,
+    End = SEEK_END,
+  };
+
+  int seek(int64_t offset, Seek origin);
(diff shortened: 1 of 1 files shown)


jrosdahl:
> Due to portability issues I'd like hide the implementation details somewhere.
>
> What do you thing of those two methods? Should they stay in `File` or should I move them right next to `fallocate`?

I think it would make sense to add a `static Stat::fstat(FILE*, OnError)` method, and maybe also a convenience `Stat File::stat(Stat::OnError)` method that calls the former. However:

> for the HTTP cache upload I need to be able to get the filesize of a `FILE` handle as well as seek within.

Unless I change my mind we won't need this – my plan is to let the secondary backends operate on in-memory data blobs, not files, at least for now.

PR #866, opened 2021-06-20
Test suite repair

Meinersbur:
A set of compatibility fixes for running the tests.

 * Add `CCACHE_SLOPPINESS=include_file_mtime` to ivfsoverlay test. This is possibly due to me using my self-compiled clang since it does not fail with the version of Clang that comes with Ubuntu 20.04.

 * Add detection to `nvcc_ldir` of whether nvcc works with chosen host compiler.  For instance, `nvcc` aborts with
```
#error -- unsupported clang version! clang version must be less than 9 and greater than 3.2
```
if the version of clang or gcc is too new. The test should be skipped in this case.

 * Add treatment of Windows drive letters in paths to additional functions.
 
 * Fix for `make_relative_path`. The result of this depends on the existence of directories in the root folder which is tested with inodes, a concept that does not exist in the Win32 API. The "Linux MinGW 64-bit" test is actually failing, presumably it actually provides inode numbers under the Wine emulation layer.

### DIFF
--- a/unittest/test_Util.cpp
+++ b/unittest/test_Util.cpp
@@ -590,7 +590,7 @@ TEST_CASE("Util::make_relative_path")
   SUBCASE("Path matches neither actual nor apparent CWD")
   {
 #ifdef _WIN32
-    CHECK(make_relative_path("C:/", "C:/a", "C:/b", "C:/x") == "C:/x");
+    CHECK(make_relative_path("C:/", "C:/a", "C:/b", "C:/x") == "../x");
 #else
     CHECK(make_relative_path("/", "/a", "/b", "/x") == "/x");
 #endif
--- a/src/Util.cpp
+++ b/src/Util.cpp
@@ -743,8 +743,12 @@ get_relative_path(string_view dir, string_view path)
     // Drive letters differ.
     return std::string(path);
   }
-  dir = dir.substr(2);
-  path = path.substr(2);
+  if (dir.length() >= 2 && dir[0] != '/' && dir[1] == ':') {
+    dir = dir.substr(2);
+  }
+  if (path.length() >= 2 && path[0] != '/' && path[1] == ':') {
+    path = path.substr(2);
+  }
 #endif
 
   std::string result;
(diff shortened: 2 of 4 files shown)


unittest/test_Util.cpp
```diff
@@ -590,7 +590,7 @@ TEST_CASE("Util::make_relative_path")
   SUBCASE("Path matches neither actual nor apparent CWD")
   {
 #ifdef _WIN32
-    CHECK(make_relative_path("C:/", "C:/a", "C:/b", "C:/x") == "C:/x");
+    CHECK(make_relative_path("C:/", "C:/a", "C:/b", "C:/x") == "../x");
```
jrosdahl, line 593: Hm, this doesn't look right. The (I guess badly named) test case is about verifying what happens when neither the actual nor the apparent current working directory _exists_ (so in the non-Windows case, ccache cannot prove that `../x` would end up in `/x` from `/a` or `/b`).

Since the whole thing about apparent CWD isn't relevant on Windows, I think we could just as well remove the `_WIN32` version of the test case if it doesn't make sense. (Note: The actual logic of rewriting paths when CWD does exist is verified in the `Util::get_relative_path` tests.)

jrosdahl (commented):
> A set of compatibility fixes for running the tests.

Thanks.

If it's not too much trouble for you I would appreciate splitting the fixes into different commits. (Different PRs are also OK, and in a way better, but that's maybe too much overhead.)

> * Add `CCACHE_SLOPPINESS=include_file_mtime` to ivfsoverlay test. This is possibly due to me using my self-compiled clang since it does not fail with the version of Clang that comes with Ubuntu 20.04.

Sounds good.

> * Add detection to `nvcc_ldir` of whether nvcc works with chosen host compiler.

:+1:

> * Add treatment of Windows drive letters in paths to additional functions.

Could you describe the problem in more details? Perhaps by adding test cases?

> * Fix for `make_relative_path`. The result of this depends on the existence of directories in the root folder which is tested with inodes, a concept that does not exist in the Win32 API. The "Linux MinGW 64-bit" test is actually failing, presumably it actually provides inode numbers under the Wine emulation layer.

I thought this was solved by #819?

Meinersbur:
> I thought this was solved by #819?

It only hides the problems (I wrote this patch before #819). I still think `make_relative_path` is quote broken. After sort, the first item is `".."` which is only not returned because with #819 `..` and `path` are now detected as not equal. However, I think, `..` should never have been a path to be considered in the first place.

Part of the problem comes from what `Util.cpp` considers an absolute path. Under strict Win32, `/x` is not an absolute path, when using it is completed using the cwd's drive. However, `is_absolute_path("/")` returns true, motivated by the test case
```
  CHECK(Util::is_absolute_path("/C:\\foo/fie")); // MSYS/Cygwin path
```
However, as a msys2 user I have never seen such paths. An actually path would be `/c/foo/fie` because msys2 automatically mounts the `C:` drive as `/c`. But in Win32 this points to the `c` directory of the current drive's root (e.g. `C:\c`, which actually exists on my system).

There are also more ways to describe a path under Windows, such as `\\NETWORK\SHARE` and `\\?\C:\foo\fie`
See https://docs.microsoft.com/en-us/dotnet/standard/io/file-path-formats.

The other changes also stem from the false dichotomy of windows and posix absolute paths. I am not using ccache in msys2 myself, so I will not further pursue this patch (other than #868 and that #867 I moved to other PRs). I think the long-term fix is to use a normalized form that is always converted to before processing paths, like `std::filesystem::path` does (instead of trying to detect Win32 paths in every function on the fly).

jrosdahl:
> Part of the problem comes from what `Util.cpp` considers an absolute path. Under strict Win32, `/x` is not an absolute path, when using it is completed using the cwd's drive. [...] false dichotomy of windows and posix absolute paths.

Thanks for providing more details. Yes, path handling for various win32 flavors has been bolted on during the years, sometimes by drive-by contributors and sometimes by me by guesswork. I don't use ccache on Windows in any form myself so I haven't had any personal motivation to improve things in this area.

> I think the long-term fix is to use a normalized form that is always converted to before processing paths, like `std::filesystem::path` does (instead of trying to detect Win32 paths in every function on the fly).

Yep, I agree.

PR #867, opened 2021-06-20
Skip nvcc_ldir test with non-working nvcc.

Meinersbur:
Add detection to nvcc_ldir of whether nvcc works with chosen host
compiler. For instance, nvcc aborts with

    #error -- unsupported clang version! clang version must be less than 9 and greater than 3.2

if the version of clang (or gcc) is too new. The test should be skipped in
this case.


jrosdahl:
Thanks.

PR #868, opened 2021-06-20
Add sloppy include_file_mtime to ivfsoverlay test

Meinersbur:
Add CCACHE_SLOPPINESS=include_file_mtime to ivfsoverlay test. This is
possibly due to me using my self-compiled clang since it does not fail
with the version of Clang that comes with Ubuntu 20.04.


jrosdahl:
Thanks.

PR #869, opened 2021-06-20
Put compile option with args into quotes

Meinersbur:
According to
https://cmake.org/cmake/help/latest/command/target_compile_options.html,
compile options are de-duplicated individually, presumably leading to
any non-first "/FI" to be removed. Put flag and option into quotes to interpret as a
single option.

Observed with CMake 3.20.3. I don't know why this does not happen with
-include or what the first /FI option could be.


jrosdahl:
Thanks.

PR #873, opened 2021-06-28
fix: AppleClang detection

gjasny:
Apple Clang, when called via `gcc`, emits a `Configured with` line
on stderr which confuses the compiler detection. As a consequence
lots of tests were skipped.

To work-around default to clang on macOS.

### DIFF
--- a/test/run
+++ b/test/run
@@ -470,7 +470,7 @@ HOST_OS_FREEBSD=false
 HOST_OS_WINDOWS=false
 HOST_OS_CYGWIN=false
 
-compiler_version="`$COMPILER --version 2>&1 | head -1`"
+compiler_version="`$COMPILER --version 2>/dev/null | head -1`"
 case $compiler_version in
     *gcc*|*g++*|2.95*)
         COMPILER_TYPE_GCC=true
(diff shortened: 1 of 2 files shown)


gjasny:
If we'd export what CMake already knows (`CMAKE_C_COMPILER` and `CMAKE_C_COMPILER_ID`) we could guessing in the test suite.

jrosdahl:
Thanks for the PR.

If the compiler detection is broken, wouldn't it be better to fix the detection instead (or as well)? With the workaround I guess the detection will still be broken when running the test suite with `CC=gcc`?

That said, I'm OK with changing the default for the test suite to Clang on macOS since I guess Clang is more often used than GCC on macOS. But the compiler to use for the test suite is in principle not tied to the compiler used to build ccache.

gjasny:
Do you remember why 258bee6250839ff83dfa10f9ed7293fde5447688 added the stderr (`2>&1`) redirection:
https://github.com/ccache/ccache/blob/2c13bce2609a02b70c92e84cdd177a6072f44d5e/test/run#L472

Could we just omit that or redirect stderr to `/dev/null`?

jrosdahl:
> Do you remember why [258bee6](https://github.com/ccache/ccache/commit/258bee6250839ff83dfa10f9ed7293fde5447688) added the stderr (`2>&1`) redirection:
> https://github.com/ccache/ccache/blob/2c13bce2609a02b70c92e84cdd177a6072f44d5e/test/run#L472
> 
> Could we just omit that or redirect stderr to `/dev/null`?

I don't remember, but redirecting stderr to /dev/null sounds OK to me.

jrosdahl:
Thanks.

PR #875, opened 2021-06-28
Add optional Redis secondary storage backend

### LINKED ISSUE #859: Storage backends: Redis
- Implement a Redis storage backend according to the API in #856.

afbjorklund:
Requires the "hiredis" library.

https://github.com/redis/hiredis

Tests also requires redis-server and redis-cli.

Closes #859.


afbjorklund:
Unfortunately there are some issues with the headers. The included "HiredisConfig.cmake" is broken, returning wrong include dir:

```cmake
find_path(HIREDIS_INCLUDE_DIRS hiredis/hiredis.h HINTS "/usr/include")
find_library(HIREDIS_LIB_HIREDIS NAMES hiredis HINTS "/usr/lib")
```

Since you are supposed to include \<hiredis.h\> (as per the example code), the include directory needs to be "/usr/include/hiredis"

However, when you do - you get an error from the pedantic C++ treating warnings as errors. (When parsing system header)

Final straw is that the different clang-format versions seem to argue about include order:

 
```diff
 
-#include <hiredis/hiredis.h>
 #include <storage/SecondaryStorage.hpp>
 #include <storage/types.hpp>
 
+#include <hiredis/hiredis.h>
+
```

When it comes to system headers, as compared to local headers (such as storage)

test/suites/secondary_redis.bash
```diff
@@ -0,0 +1,63 @@
...
+        echo "no redis-server"
+        return
+    fi
+    if [ "`which redis-cli`" = "" ]; then
+        echo "no redis-cli"
+        return
+    fi
+}
+
+SUITE_secondary_redis_SETUP() {
```
gjasny, line 12: My http storage backend also spawns a server in the `_SETUP` function. Unfortunately this will spawn multiple servers for multiple `TEST` cases. I believe we need a `_TEARDOWN` method.

CMakeLists.txt
```diff
@@ -104,6 +104,26 @@ endif()
 option(ZSTD_FROM_INTERNET "Download and use libzstd from the Internet" ${ZSTD_FROM_INTERNET_DEFAULT})
 find_package(zstd 1.1.2 REQUIRED)
 
+option(REDIS_STORAGE_BACKEND "Enable redis secondary storage")
+if(REDIS_STORAGE_BACKEND)
+  find_package(Hiredis 0.13.3)
+  if(DEFINED HIREDIS_INCLUDE_DIRS)
```
gjasny, line 110: the `Hiredis_FOUND` should be already set by find_package call

gjasny:
> Since you are supposed to include <hiredis.h> (as per the [example code](https://github.com/redis/hiredis/tree/v1.0.0/examples)), the include directory needs to be "/usr/include/hiredis"
> 
> However, when you do - you get an error from the pedantic C++ treating warnings as errors. (When parsing system header)
> 
> `/usr/include/hiredis/sds.h:48:10: error: ISO C++ forbids flexible array member ‘buf’ [-Werror=pedantic]`

You could try making that include dir as a `SYSTEM` include. That usually shuts up the compiler warnings.

Another option is surrounding the include in your code with:

```
#if defined(__clang__)
#pragma clang diagnostic push
#pragma clang diagnostic ignored "-Wall"
#endif

#include <hiredis.h>

#if defined(__clang__)
#pragma clang diagnostic pop
#endif
```

But with multiple compilers that gets ugly pretty fast.

CMakeLists.txt
gjasny: But the `HiredisConfig.cmake` is utterly broken.

afbjorklund:
Clang honors the GCC pragmas as well.

CMakeLists.txt
afbjorklund: I gave up on it, but I find this generally to be the case... All different.

test/suites/secondary_redis.bash
afbjorklund: Did a workaround, i.e. copied it to the main test instead. Also for some reason it _never_ shut down for the Mac builds ?

CMakeLists.txt
afbjorklund: Sadly it seems that cmake behaves diffently on Mac, the LIBDIR is gone

afbjorklund:
No idea why Redis won't shut down on Mac, would be thankful if someone could check it ?

`ctest -V -R 'secondary.*'`

Here (on Ubuntu) it runs _faster_ than the file test, and shuts down happily once test is done.

```
34: PASSED
1/2 Test #34: test.secondary_file ..............   Passed    0.73 sec

35: PASSED
35: 3765400:signal-handler (1625083732) Received SIGTERM scheduling shutdown...3765400:signal-handler (1625083732) Received SIGTERM scheduling shutdown...
35: 
35: 3765400:M 30 Jun 2021 22:08:52.794 # User requested shutdown...
35: 3765400:M 30 Jun 2021 22:08:52.794 # Redis is now ready to exit, bye bye...
2/2 Test #35: test.secondary_redis .............   Passed    0.56 sec
```

Added a 10 second timeout, that should kill the server.

jrosdahl:
Thanks for working on this!

The title says "WIP" but the PR is not marked as a draft. Do you want me to start reviewing or is it still WIP?

> Unfortunately there are some issues with the headers. The included "HiredisConfig.cmake" is broken, returning wrong include dir:
> 
> ```cmake
> find_path(HIREDIS_INCLUDE_DIRS hiredis/hiredis.h HINTS "/usr/include")
> find_library(HIREDIS_LIB_HIREDIS NAMES hiredis HINTS "/usr/lib")
> ```
> 
> Since you are supposed to include <hiredis.h> (as per the [example code](https://github.com/redis/hiredis/tree/v1.0.0/examples)), the include directory needs to be "/usr/include/hiredis"

I would spontaneously say that setting up the include path so that you `#include <hiredis/hiredis.h>` sounds better (a library shouldn't require a special `-I` IMO). What downside do you see what that variant?

> Final straw is that the different clang-format versions seem to argue about include order:

I think that's not different clang-format versions but different configuration since I changed how include grouping works on master very recently.

> No idea why Redis won't shut down on Mac, would be thankful if someone could check it ?

I have no idea either, but if it blocks you I'd say just disable the test on macOS for now.

afbjorklund:
> Thanks for working on this!
> 
> The title says "WIP" but the PR is not marked as a draft. Do you want me to start reviewing or is it still WIP?

Sorry, some other github projects use "WIP" with a bot that sets a "work-in-progress-do-not-merge" label...

It is ready for review now, but could possibly need some squashing and rebasing before eventually merging ?
 
> > Since you are supposed to include <hiredis.h> (as per the [example code](https://github.com/redis/hiredis/tree/v1.0.0/examples)), the include directory needs to be "/usr/include/hiredis"
> 
> I would spontaneously say that setting up the include path so that you `#include <hiredis/hiredis.h>` sounds better (a library shouldn't require a special `-I` IMO). What downside do you see what that variant?

There are no immediate downsides, it should work on "both" platforms (Ubuntu and Homebrew)

It is just that it is documented (examples and pkg-config) as `<hiredis.h>`, I suppose ?

And we would need to add the homebrew prefix, if **not** relying on pkg-config for the include dir.

I seem to recall similar issues earlier when using glib, with <glib/glib.h> versus <glib.h> ?

> > No idea why Redis won't shut down on Mac, would be thankful if someone could check it ?
> 
> I have no idea either, but if it blocks you I'd say just disable the test on macOS for now.

The tests do pass, the only downside is that the test takedown takes an extra ~9 (10 - 1) seconds...

Instead of shutting down immediately when test is done (1s), it waits for the timeout (10s) to happen

afbjorklund:
It seems that there is a <module>_PREFIX variable set up in cmake when using `pkg_check_modules`

So if we don't want to use $HIREDIS_INCLUDEDIR, we could use $HIREDIS_PREFIX/include ?

I have no idea why CMake adds two similar variables, and they don't seem to match the documentation:

deb
```
HIREDIS_INCLUDEDIR:INTERNAL=/usr/include/hiredis
HIREDIS_INCLUDE_DIRS:INTERNAL=/usr/include/hiredis
HIREDIS_LIBDIR:INTERNAL=/usr/lib
HIREDIS_LIBRARY_DIRS:INTERNAL=
HIREDIS_PREFIX:INTERNAL=/usr
```

brew
```
HIREDIS_INCLUDEDIR:INTERNAL=/home/anders/.linuxbrew/Cellar/hiredis/1.0.0/include/hiredis
HIREDIS_INCLUDE_DIRS:INTERNAL=/home/anders/.linuxbrew/Cellar/hiredis/1.0.0/include/hiredis
HIREDIS_LIBDIR:INTERNAL=/home/anders/.linuxbrew/Cellar/hiredis/1.0.0/lib
HIREDIS_LIBRARY_DIRS:INTERNAL=/home/anders/.linuxbrew/Cellar/hiredis/1.0.0/lib
HIREDIS_PREFIX:INTERNAL=/home/anders/.linuxbrew/Cellar/hiredis/1.0.0
```

CMakeLists.txt
```diff
@@ -104,6 +104,14 @@ endif()
 option(ZSTD_FROM_INTERNET "Download and use libzstd from the Internet" ${ZSTD_FROM_INTERNET_DEFAULT})
 find_package(zstd 1.1.2 REQUIRED)
 
+option(REDIS_STORAGE_BACKEND "Enable redis secondary storage")
+if(REDIS_STORAGE_BACKEND)
+  find_package(PkgConfig)
```
gjasny, line 109:
```suggestion
  find_package(PkgConfig REQUIRED)
```

CMakeLists.txt
```diff
@@ -104,6 +104,14 @@ endif()
...
+  pkg_check_modules(HIREDIS REQUIRED hiredis>=0.13.3)
+  include_directories(${HIREDIS_INCLUDEDIR})
```
gjasny, line 111: The `_INCLUDEDIR` is not [documented](https://cmake.org/cmake/help/v3.4/module/FindPkgConfig.html). 
```suggestion
  include_directories(${HIREDIS_INCLUDE_DIRS})
```
(Also the [cmake developer guide](https://cmake.org/cmake/help/latest/manual/cmake-developer.7.html#standard-variable-names) recommends `_INCLUDE_DIRS`)

CMakeLists.txt
```diff
@@ -104,6 +104,14 @@ endif()
...
+  add_definitions(${HIREDIS_CFLAGS_OTHER})
+endif()
```
gjasny, line 113: If we'd bump-up minimum CMake to 3.6 we could use:
```cmake
pkg_check_modules(HIREDIS REQUIRED IMPORTED_TARGET hiredis>=0.13.3)
...
target_link_libraries(ccache_lib PUBLIC PkgConfig::HIREDIS)
```
and omit everything else.

src/storage/secondary/RedisStorage.hpp
```diff
@@ -0,0 +1,64 @@
...
+  nonstd::expected<nonstd::optional<std::string>, Error>
+  get(const Digest& key) override;
+  nonstd::expected<bool, Error> put(const Digest& key,
+                                    const std::string& value,
+                                    bool only_if_missing) override;
+  nonstd::expected<bool, Error> remove(const Digest& key) override;
+
+private:
+  std::string m_url;
+  redisContext* m_context;
```
gjasny, line 50: I'd forward declare the `redisContext` and avoid leaking stuff from `<hiredis.h>` into other consumers of this header. In my http-branch the httplib pulls in `windows.h` things fall apart everywhere.

test/suites/secondary_redis.bash
```diff
@@ -0,0 +1,95 @@
+SUITE_secondary_redis_PROBE() {
+     if [ "`which redis-server`" = "" ]; then
```
gjasny, line 2: On some Ubuntu docker images `which` is not available by default. Better use `[[ -x "$(command -v redis-server)" ]]`.

test/suites/secondary_redis.bash
```diff
@@ -0,0 +1,95 @@
...
+        echo "no redis-server"
+        return
+    fi
+    if [ "`which redis-cli`" = "" ]; then
```
gjasny, line 6: Please use the shell-builtin `[[` instead of `[`. Maybe run [shellcheck](https://github.com/koalaman/shellcheck) on this file.

test/suites/secondary_redis.bash
```diff
@@ -0,0 +1,95 @@
...
+SUITE_secondary_redis_SETUP() {
+    unset CCACHE_NODIRECT
+    export CCACHE_SECONDARY_STORAGE="redis://localhost:7777"
+
+    generate_code 1 test.c
+}
+
+expect_number_of_cache_entries() {
+    local expected=$1
+    local host=$2
```
gjasny, line 21: Since redis v4 the `redis-cli` accepts an `-u <url>` command. Mabye you could expose that URL as part of the `_SETUP` function and avoid the `localhost 7777` arguments to `expect_number_of_cache_entries`.

CMakeLists.txt
afbjorklund: INCLUDEDIR and LIBDIR are documented (under pkg_check_modules), but we could change variables.
Just that we gave up on using find_package and vars, because the provided files were not correct anyway.

CMakeLists.txt
afbjorklund: Yes, I reverted that in order to stay compatible.

CMakeLists.txt
afbjorklund: Changed to use the prefix instead, as per discussion above.

src/storage/secondary/RedisStorage.hpp
afbjorklund: Not sure I would want to expose it being a struct even, but I don't see other ccache headers trying to hide implementation.

Only looked at zstd, though.

test/suites/secondary_redis.bash
afbjorklund: I think I will keep them as separate variables, for now. Maybe there should be some variables, though.

afbjorklund:
> > Final straw is that the different clang-format versions seem to argue about include order:
> 
> I think that's not different clang-format versions but different configuration since I changed how include grouping works on master very recently.

Changed it, and it feel apart. So there is definitely some difference between `format` target and "Code formatting" ?

https://github.com/ccache/ccache/pull/875/checks?check_run_id=2974781282

jrosdahl:
> > I would spontaneously say that setting up the include path so that you #include <hiredis/hiredis.h> sounds better (a library shouldn't require a special -I IMO). What downside do you see what that variant?
> 
> There are no immediate downsides, it should work on "both" platforms (Ubuntu and Homebrew)
> 
> It is just that it is documented (examples and pkg-config) as <hiredis.h>, I suppose ?

One more data point is that [the hiredis-config.cmake file in hiredis itself](https://github.com/redis/hiredis/blob/master/hiredis-config.cmake.in) also uses the package include installation directory (e.g. `/usr/include`) as the include path for hiredis, so it doesn't seem like something only cooked up by Debian.

> > I think that's not different clang-format versions but different configuration since I changed how include grouping works on master very recently.
>
> Changed it, and it fell apart. So there is definitely some difference between format target and "Code formatting" ?

There shouldn't be.

Note that GitHub CI jobs for merge requests run on `merge(your_branch, master)`, not just `your_branch` as you maybe would expect. You can see that on https://github.com/ccache/ccache/pull/875/checks?check_run_id=2974781282#step:2:468 where it says that GitHub actually tested a merge of your commit (363078ff690f1f74e333791c8ce2e04ccf43fb9b) with latest master at the time (5530530f7e6ce17b40d656a8a7b7584f40270c9b).

doc/MANUAL.adoc
```diff
@@ -794,6 +794,7 @@ Examples:
 +
 * `file:///shared/nfs/directory`
 * `file:///shared/nfs/one|read-only file:///shared/nfs/two`
+* `redis://server`
```
jrosdahl, line 797:
```suggestion
* `redis://example.com`
```

doc/MANUAL.adoc
```diff
@@ -929,6 +930,33 @@ Optional attributes:
...
+This backend stores data in a Redis (or Redis-compatible) server.
+There are implementations for both memory and disk based storage.
+
+See <https://redis.io>
+
+Note that ccache will not perform any cleanup of the Redis storage.
+But you can configure LRU eviction: <https://redis.io/topics/lru-cache>
+
+In addition to the in-memory database, there are also persistent NoSQL.
+See <https://github.com/yinqiwen/ardb> (using https://rocksdb.org/[RocksDB] or similar)
```
jrosdahl, line 946: This is great information, but I feel that it's a bit too specific and volatile to be part of the documentation. When you read the ccache 4.4 man page three years from now it's maybe some other Redis-compatible software that is the one to prefer, URLs may have changed, etc.

One idea I had when sketching the configuration format was to put this kind of information on the ccache wiki instead: 
https://github.com/ccache/ccache/commit/ed1f6622123315977e95e44a1ecb39202852b88d#diff-ecf215ac3df4ba49cf18dc85a8c137f0567539bd19e17270063d59e15f412667R926-R927

What do you think, does that make sense?

doc/MANUAL.adoc
```diff
@@ -929,6 +930,33 @@ Optional attributes:
...
+
+Examples:
+
+* `redis://localhost:6379`
+* `redis:///var/run/redis/redis-server.sock`
+
+Optional attributes:
+
+* *connect-timeout*: Timeout (in ms) for network connection.
+* *operation-timeout*: Timeout (in ms) for Redis commands.
```
jrosdahl, line 956: What do you think about my [suggested default values](https://github.com/ccache/ccache/issues/859#issuecomment-869208683)?

I think that we should have (quite small) default values since it makes sense to me to time out quickly for a misbehaving network or remote server. I also think we don't need to support "no timeout".

src/storage/secondary/RedisStorage.cpp
```diff
@@ -0,0 +1,353 @@
...
+    }
+    LOG("Redis reconnect err: {}", m_context->errstr);
+    redisFree(m_context);
+    m_context = nullptr;
+  }
+
+  const nonstd::string_view prefix = "redis://";
+  ASSERT(Util::starts_with(m_url, prefix));
+  const auto suffix = m_url.substr(prefix.size());
+  std::string host, port, sock;
```
jrosdahl, line 137: Hopefully we can use the `Url` class added in #676 for this instead.

src/storage/secondary/RedisStorage.cpp
```diff
@@ -0,0 +1,353 @@
...
+    m_connected = false;
+  }
+}
+
+nonstd::expected<nonstd::optional<std::string>, SecondaryStorage::Error>
+RedisStorage::get(const Digest& digest)
+{
+  if (!connect()) {
+    return nonstd::make_unexpected(Error::error);
+  }
```
jrosdahl, line 237: This is exactly what I had in mind, thanks! :+1:

However, a heads-up: when working on statistics for secondary storage, I've realized that I need to remember which backends are in an error state anyway to be able to only increment "secondary storage error" once. Therefore I plan to instead let the framework create a backend only when needed and remove it from the list of backends after the first error/timeout. This should also simplify backends so that you can connect in the constructor and don't have to remember the error state yourself.

src/storage/secondary/RedisStorage.cpp
```diff
@@ -0,0 +1,353 @@
...
+  } else if (reply->type == REDIS_REPLY_NIL) {
+    missing = true;
+  }
+  freeReplyObject(reply);
+  if (found) {
+    return value;
+  } else if (missing) {
+    return nonstd::nullopt;
+  } else {
+    return nonstd::make_unexpected(Error::error);
```
jrosdahl, line 261: Is it possible to distinguish timeout from error in `redisReply`? If so, return `Error::timeout` for timeout.

src/storage/secondary/RedisStorage.hpp
jrosdahl: Sounds good to me to forward declare.

I've written a related to do item for zstd.h as well.

test/suites/secondary_redis.bash
```diff
@@ -0,0 +1,96 @@
...
+SUITE_secondary_redis() {
+    if $HOST_OS_APPLE; then
+         # no coreutils on darwin by default, perl rather than gtimeout
+         function timeout() { perl -e 'alarm shift; exec @ARGV' "$@"; }
+    fi
+    timeout 10 redis-server --bind localhost --port 7777 &
+    sleep 0.1 # wait for boot
+    redis-cli -p 7777 ping
+    # shellcheck disable=SC2064
+    trap "kill $!" EXIT
```
jrosdahl, line 40: This could use the `terminate_all_children` mechanism from #676.

test/suites/secondary_redis.bash
```diff
@@ -0,0 +1,96 @@
...
+
+    # -------------------------------------------------------------------------
+    TEST "Base case"
```
jrosdahl, line 43: Integration tests make me happy, thanks! :+1:

doc/MANUAL.adoc
afbjorklund: I think that is a domain, but added it to the example

doc/MANUAL.adoc
afbjorklund: I will keep the reference to "LRU eviction" (should be easy enough to find, even if documentation is rearranged)

src/storage/secondary/RedisStorage.cpp
afbjorklund: It should be possible, when using hiredis 1.0 and above. But not when using a `bool` for success

jrosdahl: > But not when using a bool for success

You'd return `nonstd::make_unexpected(Error::timeout)` in that case, not a bool.

CMakeLists.txt
```diff
@@ -104,6 +104,14 @@ endif()
 option(ZSTD_FROM_INTERNET "Download and use libzstd from the Internet" ${ZSTD_FROM_INTERNET_DEFAULT})
 find_package(zstd 1.1.2 REQUIRED)
 
+option(REDIS_STORAGE_BACKEND "Enable redis secondary storage")
```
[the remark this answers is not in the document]
afbjorklund: I don't think it is available on Windows, but otherwise that should be doable

src/storage/secondary/RedisStorage.cpp
afbjorklund: Right, mostly internal refactoring

afbjorklund:
> Note that GitHub CI jobs for merge requests run on `merge(your_branch, master)`, not just `your_branch` as you maybe would expect.

Makes sense, and you actually did mention that were some recent changes to how headers are sorted...

CMakeLists.txt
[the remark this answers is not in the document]
afbjorklund: Never mind, it was hiredis_ssl that was not on Windows and hiredis was just untested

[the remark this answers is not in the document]
afbjorklund: I enabled it by default, chances are that it will need some kind of HIREDIS_FROM_INTERNET

test/suites/secondary_redis.bash
afbjorklund: The aim was to use the same test as for file backend

src/storage/secondary/RedisStorage.cpp
gjasny: I have a preparational PR to add the Url class in #881

CMakeLists.txt
```diff
@@ -104,6 +104,12 @@ endif()
 option(ZSTD_FROM_INTERNET "Download and use libzstd from the Internet" ${ZSTD_FROM_INTERNET_DEFAULT})
 find_package(zstd 1.1.2 REQUIRED)
 
+option(REDIS_STORAGE_BACKEND "Enable redis secondary storage" ON)
+if(REDIS_STORAGE_BACKEND)
+  find_package(PkgConfig REQUIRED)
+  pkg_check_modules(HIREDIS REQUIRED hiredis>=0.13.3)
```
gjasny, line 110: CMake 3.10 is now required on master branch.

src/storage/secondary/RedisStorage.cpp
```diff
@@ -0,0 +1,374 @@
...
+#include <Digest.hpp>
+#include <Logging.hpp>
+#include <UmaskScope.hpp>
+#include <Util.hpp>
+#include <assertions.hpp>
+#include <fmtmacros.hpp>
+#include <util/file_utils.hpp>
+#include <util/string_utils.hpp>
+
+#include <third_party/nonstd/string_view.hpp>
```
jrosdahl, line 31:
```suggestion
#include <Digest.hpp>
#include <Logging.hpp>
#include <fmtmacros.hpp>
```

src/storage/secondary/RedisStorage.cpp
```diff
@@ -0,0 +1,374 @@
...
+}
+
+static std::string
+timeval_to_string(struct timeval tv)
+{
+  return FMT("{:.3f}s", tv.tv_sec + tv.tv_usec / 1000000.0);
+}
+
+static nonstd::optional<struct timeval>
+parse_connect_timeout(const AttributeMap& attributes)
```
jrosdahl, line 55: Since `parse_connect_timeout` and `parse_operation_timeout` differ only in the attribute name, I suggest replacing them with a single `parse_timeout_attribute(const AttributeMap& attributes, const std::string& name)` function.

src/storage/secondary/RedisStorage.cpp
```diff
@@ -0,0 +1,374 @@
...
+{
+  const auto it = attributes.find("operation-timeout");
+  if (it == attributes.end()) {
+    return nonstd::nullopt;
+  }
+  return milliseconds_to_timeval(it->second);
+}
+
+static nonstd::optional<std::string>
+parse_username(const AttributeMap& attributes)
```
jrosdahl, line 75: Similar here, maybe introduce a single `parse_string_attribute` function?

src/storage/secondary/RedisStorage.cpp
```diff
@@ -0,0 +1,374 @@
...
+}
+
+RedisStorage::RedisStorage(const Url& url, const AttributeMap& attributes)
+  : m_url(url),
+    m_connect_timeout(parse_connect_timeout(attributes)),
+    m_operation_timeout(parse_operation_timeout(attributes)),
+    m_username(parse_username(attributes)),
+    m_password(parse_password(attributes))
+{
+  m_prefix = "ccache"; // TODO: attribute
```
jrosdahl, line 101: Ah, that sounds like a good idea.

src/storage/secondary/RedisStorage.cpp
```diff
@@ -0,0 +1,374 @@
...
+    m_invalid = true;
+    return REDIS_ERR;
+  }
+
+  if (!m_context) {
+    LOG("Redis connect {} err NULL", m_url.str());
+    m_invalid = true;
+    return REDIS_ERR;
+  } else if (m_context->err) {
+    LOG("Redis connect {} err: {}", m_url.str(), m_context->errstr);
```
jrosdahl, line 167: I tried building with a locally compiled libhiredis 1.0.0 (and also with latest hiredis master) and I always get an error with `m_context->err` and `m_context->errstr` being different garbage values. Have you tried hiredis 1.0.0? I can debug this when I get time, maybe tomorrow.

src/storage/secondary/RedisStorage.cpp
```diff
@@ -0,0 +1,374 @@
...
+    m_connected = true;
+
+    if (m_operation_timeout) {
+      LOG("Redis timeout {}", timeval_to_string(*m_operation_timeout));
+      if (redisSetTimeout(m_context, *m_operation_timeout) != REDIS_OK) {
+        LOG_RAW("Failed to set timeout");
+      }
+    }
+
+    // TODO: break out AUTH to separate function
```
jrosdahl, line 187: Sounds good!

src/storage/secondary/RedisStorage.cpp
```diff
@@ -0,0 +1,374 @@
...
+    if (m_password) {
+      std::string username = m_username ? *m_username : "default";
+      LOG("Redis AUTH {} {}", username, "*****"); // don't log m_password !!!
```
jrosdahl, line 190: Note that the password will be logged as part of the configuration dump anyway. Unless we do something about that, of course, but my [current suggestion is to not do it](https://github.com/ccache/ccache/pull/881#discussion_r664074292).

src/storage/secondary/RedisStorage.hpp
```diff
@@ -0,0 +1,61 @@
...
+// Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
+
+#pragma once
+
+#include "storage/SecondaryStorage.hpp"
+#include "storage/types.hpp"
+
+#include <third_party/url.hpp>
+
+typedef struct redisContext redisContext;
```
jrosdahl, line 26:
```suggestion
struct redisContext;
```

test/suites/secondary_redis.bash
jrosdahl, line 40: You should be able to delete these two lines since the test framework now (as of a28ad9db64f39b49652bf2ad8ab8d3259786d33b) kills child processes started by a test suite.

src/storage/secondary/RedisStorage.cpp
jrosdahl: This is most likely false alarm – I seem to somehow have used a mix-up of header and library files of different versions.

CMakeLists.txt
afbjorklund: The reported include directory in the pkg-config file is for using \<hiredis.h\>:

```
prefix=/home/anders/.linuxbrew/Cellar/hiredis/1.0.0
exec_prefix=${prefix}
libdir=/home/anders/.linuxbrew/Cellar/hiredis/1.0.0/lib
includedir=/home/anders/.linuxbrew/Cellar/hiredis/1.0.0/include/hiredis
```

So we need the prefix, or something like `../`, to use \<hiredis/hiredis.h\> ?

For the deb package we got "lucky" since /usr/include is already searched.

src/storage/secondary/RedisStorage.cpp
afbjorklund: This is most likely optional (but default), it is when sharing the cache server with some other purpose.

src/storage/secondary/RedisStorage.cpp
afbjorklund: I have tried with 0.13.3 (18.04), 0.14.0 (20.04) and 1.0.0 (brew). But maybe some more asserts etc, are needed ?

test/suites/secondary_redis.bash
afbjorklund: Sounds good. Maybe can add that when adding a test for the unix socket path variant (only on non-Windows)

CMakeLists.txt
afbjorklund: To use `find_package`, we need to add a cmake file (for where it is missing).
It was easier to use the standard pkg-config, especially since it was broken...

But I made a change to add HIREDIS_FROM_INTERNET, and added one.
However, it didn't have any static library build target - only a shared library

src/storage/secondary/RedisStorage.cpp
afbjorklund: I think it should be masked in case someone decides to attach the log as part of a bug report or something, but your call.

CMakeLists.txt
jrosdahl: > The reported include directory in the pkg-config file is for using <hiredis.h>:

\*nod*

> [...] So we need the prefix, or something like `../`, to use <hiredis/hiredis.h> ?

Right, to use `#include <hiredis/hiredis.h>` we need `-I<pkg_config_prefix>/include` instead of `-I<pkg_config_includedir>`. But what I didn't understand was "adding the **homebrew** prefix" which sounded like it was something else. Did you mean "adding the **pkg-config** prefix"? If so I understand. :slightly_smiling_face:

> To use `find_package`, we need to add a cmake file (for where it is missing).

I realize that I was very unclear, sorry about that. What I should have written instead of "if using just `find_package`"  is something like "if using `find_package` and a `Findhiredis.cmake` file similar to how it's done for zstd". That is, in our own cmake code, use `find_library` and `find_path` to find the library and include directory in standard locations (or non-standard by e.g. setting `CMAKE_PREFIX_PATH`).

> It was easier to use the standard pkg-config, especially since it was broken...

I understand. If it's not too hard I would like to include Redis in the Windows builds as well, which is why I'm exploring alternatives here.

> But I made a change to add HIREDIS_FROM_INTERNET, and added one.

Now I don't follow again. I see no `HIREDIS_FROM_INTERNET`, so are you talking about some not yet published changes?

src/storage/secondary/RedisStorage.cpp
jrosdahl: I agree that (all other things being equal) it would be better to mask passwords, but I don't want to let `Config` know about what is a password and not. I guess we could enhance the backend API somehow and let backends parse configuration early, communicating back to Config, but my initial reaction to this was to not solve this problem right now since there are many other things up in the air already. But I'll sleep on it.

CMakeLists.txt
afbjorklund: Yeah, there were some merge conflicts. I will get it up soon, but a bit unsure about the shared lib.

afbjorklund: Added as 366c46461586222d1409170ad306a6901b83dcb9

afbjorklund: > But what I didn't understand was "adding the homebrew prefix" which sounded like it was something else. Did you mean "adding the pkg-config prefix"? If so I understand. slightly_smiling_face

I think I mixed them up a bit, since homebrew adds symlink into the individual installations from a global directory.
To confuse things further, this global directory is `/usr/local` on Mac - but I have it under `~/.linuxbrew` instead.

afbjorklund: Hmm, running from the package. Partly because I copied "lib" from zstd, partly because there is no "hiredis" directory.

-I/home/runner/work/ccache/ccache/build/hiredis-1.0.0/lib

build/hiredis-1.0.0/hiredis.h

So I'm not sure it is possible to use the build directory, without adding some workaround like symlinking hiredis to . ?

afbjorklund:
Maybe the HIREDIS_FROM_INTERNET should be reverted, it mostly affects Windows (without a package manager)

```
Install the project...
-- Install configuration: "Debug"
-- Installing: /tmp/ccache/opt/ccache/bin/ccache
-- Set runtime path of "/tmp/ccache/opt/ccache/bin/ccache" to "/opt/ccache/lib"
-- Installing: /tmp/ccache/opt/ccache/lib/libhiredis.so.1.0.0
-- Installing: /tmp/ccache/opt/ccache/share/man/man1/ccache.1
```

Since there is no static target in the CMakeLists.txt, we can't really build a static library (without patching hiredis)

And it seems that I missed that it couldn't find the header, since I had hiredis/hiredis.h in a default search path...

afbjorklund:
Looks like hiredis doesn't build on Windows after all.

afbjorklund:
There are still test flakes on Mac, the server starts up in 3 us but it still cannot contact Redis ?

Most likely the "sleep" command doesn't support fractions, so that sleep 0.1 means sleep 0 ?

CMakeLists.txt
jrosdahl: If I understand correctly you made some changes after you wrote the above to make it work (copying headers to a hiredis directory)? Looks good to me.

jrosdahl:
> Looks like it doesn't build on Windows after all.
> 
> EDIT: Seems that I had just missed to include a hiredis <win32.h> header.
> 
> Except that the whole thing is guarded by `#ifdef _MSC_VER` (Microsoft C++)
> 
> ```c++
> #include <winsock2.h> /* for struct timeval */
> ```

I think you should add

```c++
#ifdef HAVE_SYS_TIME_H
#  include <sys/time.h>
#endif
```

to `RedisStorage.hpp` instead (and remove the forward declaration of `struct timeval`). That should fix the "invalid application of ‘sizeof’ to incomplete type ‘storage::secondary::timeval’" errors you see as well.

afbjorklund:
Right, I included the local headers _before_ the system headers.. \<DOH\>

jrosdahl:
> we can't really build a static library (without patching hiredis)

Right, we need a static library with `-DHIREDIS_FROM_INTERNET` since I don't think it's reasonable to have to install a custom shared library alongside ccache. (In fact, I'm inclined to default to static linking for dependencies, including libzstd and libhiredis in all use cases.)

> Since there is no static target in the CMakeLists.txt

It seems like there is a [static library target on hiredis master](https://github.com/redis/hiredis/blob/5850a8ecd2fb4ab39d80773e3017f02aff097ec4/CMakeLists.txt#L49) but it's not included in a release yet.

Maybe we should just give up and bundle dependencies instead... :thinking:

afbjorklund:
> Maybe we should just give up and bundle dependencies instead... thinking

For this library we could go back to making it optional, and disable it on Windows.

jrosdahl:
> I guess I also need config.h in there ?

`config.h` is always included nowadays: https://github.com/ccache/ccache/blob/39c258908c9fbea414941e0d5db2b748052cb166/cmake/StandardSettings.cmake#L7-L10

jrosdahl:
> For this library we could go back to making it optional, and disable it on Windows.

I'm OK with it being optional (as in "opt out"), but it's should be easy enough to create a static library of our own for the Windows build.

But I think you can leave the Redis backend disabled for the Windows builds for now so that it doesn't block your PR progress. Then I can have a look at improving this later. Sounds OK?

afbjorklund:
Sounds good, then I can remove some of the other hacks as well. Maybe leave `find_package`, but clean it up.

Maybe this was apparent, but the aim was to do a squash and rebase before merging this work-in-progress :-D

CMakeLists.txt
afbjorklund: Since there are so many problems with it, I am going to leave HIREDIS_FROM_INTERNET and calling CMake.

jrosdahl:
> Maybe this was apparent, but the aim was to do a squash and rebase before merging this work-in-progress :-D

Sure. And you don't have to squash unless you want to – I always do that as part of merging anyway (unless it's obvious that the PR author has separated stuff in clean commits).

afbjorklund:
I mostly meant the merges and the reverts, but there were some other minor fixes that could be squashed together.

test/suites/secondary_redis.bash
afbjorklund: Went with url after all, thanks for the suggestion!

src/storage/secondary/RedisStorage.cpp
afbjorklund: It's a boolean now, so it could go either way in the future depending on what you come up with.

doc/MANUAL.adoc
afbjorklund: I'm not sure about the values to use, but the code should be ready for it (tested it locally, even if there are no tests)

afbjorklund: Added 100 ms connection timeout, and 10 second operation timeout (in case of an overloaded shared server)

jrosdahl:
Thanks!

src/storage/secondary/RedisStorage.cpp
jrosdahl: I wrote:

> I agree that (all other things being equal) it would be better to mask passwords [...] But I'll sleep on it.

Addressed in 873384667a5e8ba0e688303e18f34196ab3b57c6.

PR #878, opened 2021-07-02
Use generator expression for ccache location

gjasny:
With multi-config generators like Xcode the binary is stored within an extra Debug / Release directory.

Tested with CMake 3.4.3


jrosdahl:
Thanks.

PR #881, opened 2021-07-04
Use URL class for URL parsing

gjasny:


test/suites/secondary_url.bash
```diff
@@ -0,0 +1,41 @@
...
+    $CCACHE_COMPILE -c test.c 2> "${ABS_TESTDIR}/stderr.log"
+    expect_contains "${ABS_TESTDIR}/stderr.log" "URL scheme must not be empty"
+
+    # -------------------------------------------------------------------------
+    TEST "Redact user info"
+
+    export CCACHE_SECONDARY_STORAGE="file://username:secret@host$PWD/secondary"
+    $CCACHE_COMPILE -c test.c
+    expect_contains "${CCACHE_LOGFILE}" "username:REDACTED"
+    #expect_not_contains "${CCACHE_LOGFILE}" "secret" # --show-config leaks the secret
```
gjasny, line 40: Fixing this requires moving the parsing code into the configuration class. Should we do this now or could we postpone it?

test/suites/secondary_url.bash
```diff
@@ -0,0 +1,41 @@
...
+SUITE_secondary_url() {
+    # -------------------------------------------------------------------------
+    TEST "Reject empty url (without config attributes)"
+
+    export CCACHE_SECONDARY_STORAGE="|"
+    $CCACHE_COMPILE -c test.c 2> "${ABS_TESTDIR}/stderr.log"
+    expect_contains "${ABS_TESTDIR}/stderr.log" "must provide an URL"
+
+    # -------------------------------------------------------------------------
+#    TEST "Reject empty url (but with config attributes)"
```
gjasny, line 14: The `Util::split_into_views` omits empty entries. That way `|key=value` gets parsed into an vector of one element and the `key=value` argument is interpreted as URL.

Should I make it configurable or could we change default behavior?

gjasny: Thanks for the quick addition.

test/suites/secondary_url.bash
```diff
@@ -0,0 +1,41 @@
+SUITE_secondary_url_SETUP() {
+    generate_code 1 test.c
+}
+
+SUITE_secondary_url() {
+    # -------------------------------------------------------------------------
+    TEST "Reject empty url (without config attributes)"
+
+    export CCACHE_SECONDARY_STORAGE="|"
+    $CCACHE_COMPILE -c test.c 2> "${ABS_TESTDIR}/stderr.log"
```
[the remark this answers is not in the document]
gjasny: Bear with me. I'm still new to the ccache test framework. I was not aware that `ABS_TESTDIR` is the `$PWD`

src/storage/Storage.cpp
```diff
@@ -179,24 +182,46 @@ namespace {
...
+
 } // namespace
 
 static ParseStorageEntryResult
 parse_storage_entry(const nonstd::string_view& entry)
 {
   const auto parts = Util::split_into_views(entry, "|");
 
+  if (parts.empty() || parts.front().empty()) {
+    throw Error("secondary storage config must provide an URL: {}", entry);
```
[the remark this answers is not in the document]
gjasny: I learned something today: https://www.quora.com/Which-is-correct-A-URL-or-An-URL :)

test/suites/secondary_url.bash
gjasny: Sounds reasonable. I was inspired by [RFC3986](https://datatracker.ietf.org/doc/html/rfc3986#section-3.2.1) which recommends:
> Applications should not render as clear text any data after the first colon (":") character found within a userinfo

gjasny:
All comments were addressed. I believe this PR is ready for another review.

jrosdahl:
Thanks.

test/suites/secondary_url.bash
jrosdahl: Redacting secrets in logs has been addressed in 873384667a5e8ba0e688303e18f34196ab3b57c6.

PR #886, opened 2021-07-05
Update dockerfiles with supported distributions

gjasny:
I used alpine-3.8 as the new lower bar for CMake support (ships with 3.11.1). Unfortunately bundled clang 5.0.1 cannot find `<immintrin.h>`.


src/storage/secondary/FileStorage.hpp
```diff
@@ -21,6 +21,8 @@
 #include <storage/SecondaryStorage.hpp>
 #include <storage/types.hpp>
 
+#include <sys/stat.h> // for mode_t
```
gjasny, line 24: Those are needed on alpine with its musl libc.

dockerfiles/debian-11/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM debian:9
+FROM debian:bullseye
```
jrosdahl, line 1: Isn't Debian 10 supported?

gjasny: Maybe not until it's finally released: https://hub.docker.com/_/debian

jrosdahl: Ah, I did mean 10, not 11, since I mistook the diff as if you had moved the lower bound from Debian 9 to 11 instead of 10 – I forgot that there already is a debian-10 directory.

jrosdahl:
Thanks.

PR #890, opened 2021-07-11
Add secondary HTTPS storage

gjasny:


jrosdahl:
Looks like a solid implementation.

However, I'm not yet sure that we actually should support HTTPS in this way, so I would like to discuss our options first. See #894.

gjasny:
Depends on #895

test/CMakeLists.txt
```diff
@@ -54,6 +54,9 @@ addtest(readonly_direct)
 addtest(sanitize_blacklist)
 addtest(secondary_file)
 addtest(secondary_http)
+if(HTTPS_STORAGE_BACKEND)
```
gjasny, line 57: This condition is ignored by direct invocation of `run` as called by `build-and-verify-package`. One way to fix it would be to teach ccache about all its optional features (may reported by `--version`) and then have `_PROBE` skipping the test suite if a required feature is not present.

jrosdahl: Yes, adding a list of optional features to `--version` sounds good.

gjasny: Merged #897

afbjorklund:
This needs renaming from "secondary" to "remote"

afbjorklund:
Should this PR be updated with master changes?

woggioni:
why not merge this? it can be a feature that is disabled by default so that it won't affect the startup time for the existing userbase, at the same time it can be useful to people like me that want to have a public readonly cache hosted on a public cloud where only the CI/CD pipeline is allowed to write

jrosdahl:
This PR will not be merged due to reasons mentioned in #894. Instead the plan is to implement HTTPS support in #1214.

jrosdahl:
HTTPS support is now available on master using a separate storage helper process: see #1672 and <https://ccache.dev/storage-helpers.html>.

jrosdahl:
HTTPS support is now available in ccache 4.13!

PR #891, opened 2021-07-11
Only run redis test if backend is enabled

gjasny:


jrosdahl:
I converted this to a PR with only the Redis test fix since the `value_or` improvements were made obsolete by b1569ec8572d26a689ea9a2888727cd4a4c56c41.

By the way, if it's not too much trouble I would prefer different PRs for commits that don't belong together logically.

jrosdahl:
Thanks.

PR #893, opened 2021-07-12
Add Kyoto Tycoon secondary storage backend

### LINKED ISSUE #435: Support kyototycoon backend
https://github.com/alticelabs/kyoto/blob/master/README.md

afbjorklund:
See https://github.com/afbjorklund/kyoto for the requirements.

It is possible to run `ktserver` both in memory or on disk.

Mostly for performance benchmarks, not necessarily merging.

Closes #435


afbjorklund:
Added a discussion topic in #900, if this secondary storage backend should be included in ccache (at all).

jrosdahl:
I agree. Thanks for having a look, though.

PR #902, opened 2021-07-18
Add Redis SSL support

afbjorklund:
See #894 for discussion about TLS/SSL.


src/CMakeLists.txt
```diff
@@ -91,6 +91,13 @@ if(REDIS_STORAGE_BACKEND)
     PUBLIC standard_settings standard_warnings HIREDIS::HIREDIS
            third_party_lib)
 endif()
+if(REDISS_STORAGE_BACKEND)
+  target_compile_definitions(ccache_lib PRIVATE -DHAVE_REDISS_STORAGE_BACKEND)
+  target_link_libraries(
+    ccache_lib
+    PUBLIC standard_settings standard_warnings HIREDIS::HIREDIS HIREDIS::HIREDIS_SSL
```
gjasny, line 98: The standard* and third party libraries are already added elsewhere.

The same applies to the redis backend.

src/storage/secondary/RedisStorage.cpp
```diff
@@ -89,6 +102,8 @@ RedisStorage::RedisStorage(const Url& url, const AttributeMap& attributes)
   : m_url(url),
     m_prefix("ccache"), // TODO: attribute
     m_context(nullptr),
+    m_ca_cert(parse_string_attribute(attributes, "cacert").value_or("")),
```
gjasny, line 105: I'd store the optional directly as a member. That way you avoid treating the empty string as special value.

src/CMakeLists.txt
afbjorklund: It is theoretically possible to disable redis, but enable rediss. The rest was just copy/paste, I guess

src/storage/secondary/RedisStorage.cpp
afbjorklund: Storing the `const char *` that is required by the library would be another option. Then it can be passed straight through.

src/CMakeLists.txt
gjasny: I meant, the `REDIS_STORAGE_BACKEND` block above also links to standard and third party libs.

afbjorklund: It should be fixed, also it was calling the tests always. But that's not in _this_ backend

test/suites/secondary_rediss.bash
```diff
@@ -0,0 +1,176 @@
+SUITE_secondary_rediss_PROBE() {
+    if ! $CCACHE --version | fgrep -q -- redis-storage &> /dev/null; then
```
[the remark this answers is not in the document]
afbjorklund: It was not available, only said "redis-storage"

test/CMakeLists.txt
```diff
@@ -55,6 +55,9 @@ addtest(sanitize_blacklist)
 addtest(secondary_file)
 addtest(secondary_http)
 addtest(secondary_redis)
+if(REDISS_STORAGE_BACKEND)
```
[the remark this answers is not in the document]
afbjorklund: Actually it only looked at the supporting programs, as you noticed the ccache feature reporting didn't check it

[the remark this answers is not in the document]
afbjorklund: Think I missed the FEATURE_TEXT

test/suites/secondary_rediss.bash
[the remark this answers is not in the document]
afbjorklund: Think I missed the FEATURE_TEXT

sschleemilch:
Is there any update regarding that SSL support implementation?
What is missing? I would be very happy if that would made it into the master and I would be happy to support if necessary and if I am capable of doing so ;-)

afbjorklund:
I think the code has rotten a bit (tests failing) since it is not merged, but can try to update it

afbjorklund:
@sschleemilch updated the branch, and fixed the broken tests and added static build for ssl

jrosdahl:
> Is there any update regarding that SSL support implementation? What is missing?

What's missing is some solution of the problems mentioned in #894, especially the first one.

jrosdahl:
This PR will not be merged due to reasons mentioned in #894. Instead the plan is to implement REDISS support in #1214.

jrosdahl:
A ccache storage helper for Redis is now available: https://github.com/ccache/ccache-storage-redis

Issue #747, opened 2020-12-14
Document how to set settings from CMake


AlexanderLanin:
### Description ###
In google there are lots of instructions like this one:

```
    set(ENV{CCACHE_COMPRESS} "true")
    set(ENV{CCACHE_BASEDIR} "/home")
    set_property(GLOBAL PROPERTY RULE_LAUNCH_COMPILE ${CCACHE_PROGRAM})
```

However this does not work. Document how to set cmake variables from cmake.

AlexanderLanin:
The best I can figure out at the moment is to create a ccache-set-defaults.sh file like this one:
```
#!/bin/bash
#
# Sets reasonable ccache defaults for our project.


# Compress cache by default (default is false in old ccache 3.x)
ccache --show-config | grep "^(default) compression = false$" >/dev/null && ccache --set-config=compression=true

# Attempt to share cache for source directories within /home dir
# 'ccache --set-config=base_dir=/home' does not work in old ccache 3.x
# So instead locate the config file and write to it:
ccache --show-config | grep "^(default) base_dir = $" >/dev/null && echo "base_dir = /home" >> "$(ccache -s | grep 'primary config' | tr ' ' '\n' | tail -n1)"
```

And in cmake:
```
find_ccache(ENABLE_CCACHE)
if(ENABLE_CCACHE)
    execute_process(COMMAND bash -c "${CMAKE_CURRENT_SOURCE_DIR}/tools/ccache-set-defaults.sh")
    set_property(GLOBAL PROPERTY RULE_LAUNCH_COMPILE ${CCACHE_PROGRAM})
endif()
```
`find_ccache` is a little wrapper for `find_program` which checks ccache version ranges.

However that seems a little crazy?

Note: the goal is here is to provide reasonable defaults. I do not want to overwrite user settings in case something is configured. Usually users know best what they need.

AlexanderLanin:
Another option is to use this one by @craigscott-crascit.
In contrast to the above it overrides local user settings.

> Permission is given to use any of the code samples in this work without restriction. Attribution is not required.

https://github.com/CppCon/CppCon2020/blob/7a1b35024b833f55697ed420d68045d294eafe3e/Presentations/quickly_testing_qt_desktop_applications_with_approval_tests/approvals_live_demo-2020-09-cppcon/CMake/UseCompilerCache.cmake

In addition to the problem above it shows some other problems and how to solve them.
It sums up to a crazy amount of code. But that is a different issue.

[...]

PR #906, opened 2021-07-25
Export CMake configuration

gjasny:
Hello,

would you consider exporting the `EnableCcache.cmake` file useful?

The currect change would allow CMake-users of ccache to run:

```cmake
find_package(ccache CONFIG)
if(ccache_FOUND)
  enable_ccache()
endif()
```

Todo:

- [ ] export ccache version
- [ ] handle user-defined environment variables
- [ ] update documentation

Fixes: #747


jrosdahl:
I'm a bit torn on this feature (even for the ccache project itself). I actually find it bit annoying that the build runs ccache explicitly when I have set up ccache as masquerading the compiler already since this means that ccache is executed twice. That said, I can see why some users would like the project to set up usage of ccache even if the user hasn't set up ccache themselves.

Regarding `CCACHE_BASEDIR`: I don't think that the decision to use a base directory is for the project to decide – forcing a base directory on all users, especially of other projects, is too intrusive, for instance since it makes CMake's Unix generator not work well (hence the [absolute_paths_in_stderr](https://ccache.dev/manual/4.3.html#config_absolute_paths_in_stderr) hack).

The `CCACHE_CPP2` part seems unnecessary nowadays.

PR #909, opened 2021-07-30
 Disable preprocessor hits for PCHs with clang again

ahasselbring:
I added this in 3ec58dbcaa1e7903d59dce4c77baff7bc3783269 as I thought clang only saves the timestamps of included files, so as long as they are turned off by -Xclang -fno-pch-timestamp, preprocessed hits should be possible. However, clang also saves the size and possibly a hash of the un-preprocessed source file, so preprocessed hits don't make much sense.

As long as only the timestamp of a PCH source file changed, e.g. through git operations, there should still be direct hits (as long as -Xclang -fno-pch-timestamp is used).


jrosdahl:
Thanks.

PR #910, opened 2021-08-01
docs: Add scheme to link to fonts.googleapis.com

orgads:


jrosdahl:
The scheme is not missing – it's a [protocol-relative URL](https://en.wikipedia.org/wiki/URL#prurl). Did you experience some problem with it?

orgads:
Yes. When opening locally, the protocol is `file://`, and obviously this path is not found...

jrosdahl:
OK, then I'll treat it as an improvement instead of a bug which I first thought it was. :slightly_smiling_face:

jrosdahl:
Thanks.

PR #911, opened 2021-08-01
docs: Fix links to section anchors

orgads:
asciidoc uses #_section_header, but github uses #section-header.
Replace all the explicit refs with the section titles, and let
the engine handle it.


orgads:
I compared the output HTML, and it is equivalent.

jrosdahl:
Thanks!

PR #912, opened 2021-08-05
Add support for cache sharding on secondary storage

### LINKED ISSUE #908: Support consistent hashing of multiple secondary storage servers
It is possible to use more than one secondary storage backend, separated by space.

But currently that means that **all** keys will be stored in **all** servers.

It would be a nice improvement if one could use [consistent hashing](https://en.wikipedia.org/wiki/Consistent_hashing), such as [ketama](https://github.com/RJ/ketama).

There, which server to use is calculated based on the current key.

https://github.com/RJ/ketama/blob/master/libketama/ketama.h

```C
/** \brief Maps a key onto a server in the continuum.
  * \param key The key that you want to map to a specific server.
  * \param cont Pointer to the continuum in which we will search.
  * \return The mcs struct that the given key maps to. */
mcs* ketama_get_server( char*, ketama_continuum );
```

Each server has a "weight", normally the number of megs allocated.

----

The current workaround for ccache is to use a local proxy, to do the remote distribution.
e.g. twemproxy for redis: https://blog.twitter.com/developer/en_us/a/2012/twemproxy

https://redis.io/topics/partitioning

>Different implementations of partitioning
> ...
> * Client side partitioning ...
> * Proxy assisted partitioning ... 

That might still be a good idea when it comes to different larger deployment scenarios...
But still useful to have the option, and it is available in some other clients (like [libredis](https://github.com/toymachine/libredis))

jrosdahl:
This adds support for a shards attribute with a comma-separated list of names for sharding (partitioning) the cache entries using Rendezvous hashing, typically to spread the cache over a server cluster. When set, the storage URL must contain an asterisk (`*`), which will be replaced by one of the shard names to form a real URL. A shard name can optionally have an appended weight within parentheses to indicate how much of the key space should be associated with that shard. A shard with weight **w** will contain **w/S** of the cache, where **S** is the sum of all shard weights. A weight could for instance be set to represent the available memory for a memory cache on a specific server. The default weight is **1**.

For example,

    redis://cache-*.example.com|shards=a(3),b(1),c(1.5)

will put 55% (3/5.5) of the cache on redis://cache-a.example.com, 18% (1/5.5) on redis://cache-b.example.com and 27% (1.5/5.5) on redis://cache-c.example.com.

Closes #908.

Question: Is "shard" a good enough name for the concept? I guess the term is mostly used for databases, not caches? Other words I've considered are "partition", "slice", "node" and "server". "partition" feels more or less equivalent in clarity to "shard". "node" and "server" are good if they specify hostnames, but since I think it's nice to able to specify parts of a hostname those feel less good. Are there other alternatives?

### DIFF
--- a/doc/MANUAL.adoc
+++ b/doc/MANUAL.adoc
@@ -893,6 +893,28 @@ Optional attributes available for all secondary storage backends:
 
 * *read-only*: If *true*, only read from this backend, don't write. The default
   is *false*.
+* *shards*: A comma-separated list of names for sharding (partitioning) the
+  cache entries using
+  https://en.wikipedia.org/wiki/Rendezvous_hashing[Rendezvous hashing],
+  typically to spread the cache over a server cluster. When set, the storage URL
+  must contain an asterisk (`+*+`), which will be replaced by one of the shard
+  names to form a real URL. A shard name can optionally have an appended weight
+  within parentheses to indicate how much of the key space should be associated
+  with that shard. A shard with weight *w* will contain *w*/*S* of the cache,
+  where *S* is the sum of all shard weights. A weight could for instance be set
+  to represent the available memory for a memory cache on a specific server. The
+  default weight is *1*.
++
+Examples:
++
+--
+* `+redis://cache-*.example.com|shards=a(3),b(1),c(1.5)+` will put 55% (3/5.5)
+  of the cache on `+redis://cache-a.example.com+`, 18% (1/5.5) on
+  `+redis://cache-b.example.com+` and 27% (1.5/5.5) on
+  `+redis://cache-c.example.com+`.
+* `+http://example.com/*|shards=alpha,beta+` will put 50% of the cache on
+  `+http://example.com/alpha+` and 50% on `+http://example.com/beta+`.
+--
 
 These are the available backends:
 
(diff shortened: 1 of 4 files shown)


afbjorklund:
The only confusion with sharding in this context, is that _sometimes_ the term is used when splitting up large files into smaller fragments. Like you say, the term "sharding" is quite widely used in databases (like Mongo). Even if Redis also calls it "partitioning".

I think you can go with "shards", if there ever is a need to split files they could be called "chunks" or something. Since the default size of Redis is something like 512M, there should never be a _need_ to split values (like there was with Memcached's 1M), only an option.

afbjorklund (approved):
Looks good to me! I think one could use DNS aliases (and math...) if wanting to emulate the server list that the proxy-assisted partioning (i.e. Twemproxy) was using for distriibution.

jrosdahl:
Thanks for having a look!

> I think one could use DNS aliases (and math...) if wanting to emulate the server list that the proxy-assisted partioning (i.e. Twemproxy) was using for distriibution.

Could you explain the part about DNS aliases and math to emulate a server list? Why would that be needed or wanted?

afbjorklund:
> Could you explain the part about DNS aliases and math to emulate a server list? Why would that be needed or wanted?

Just if you are converting from an external proxy, to using the built-in method. Not if starting out from scratch, no.

What I _meant_ to say was that both methods should accomplish the same thing, just with slightly different syntax.

PR #918, opened 2021-08-22
Fix linker and IPO handling on various platforms

gjasny:
Enabling IPO seems to work best when the default linker is used. On mingw IPO seems to be utterly broken (out-of-diskspace and lacking linker module support), therefore IPO support is being removed.


jrosdahl:
Thanks.

PR #925, opened 2021-08-29
fix(test): Pass all three required arguments to expect_file_count

gjasny:
I got a strange `find: illegal option -- t` error due to `$dir` evaluating to nothing.


jrosdahl:
Thanks.

PR #927, opened 2021-09-05
fix(http-storage): Disable broken connection keep-alive

gjasny:
If compilation takes very long and the remote end closes
the http connection between the GETs and the PUTs the
ccache process dies with a SIGPIPE. Better disable keep-alive
handling until httplib is fixed.


gjasny:
I'll try to see why https://github.com/yhirose/cpp-httplib/blob/master/httplib.h#L5608 does not work as expected.

jrosdahl:
OK, thanks.

> I'll try to see why https://github.com/yhirose/cpp-httplib/blob/master/httplib.h#L5608 does not work as expected.

:+1:

PR #928, opened 2021-09-05
fix: Always store configuration origin value

gjasny:
The std::unordered_map::emplace function only stores the value if the map did not contain the key, yet.

But maybe a simple `m_origins[key] = origin;` would be good enough.


jrosdahl:
Thanks.

PR #933, opened 2021-09-11
Use shell builtin pwd for basedir test

### LINKED ISSUE #916: Test suite fails if PWD is a symlinked directory
While working on the homebrew ccache packag I also wanted to run the ccache test suite after compiling. But two tests always fail:

```
The following tests FAILED:
	  3 - test.basedir (Failed)
	 21 - test.nocpp2 (Failed)
```

### How to reproduce ###

1. Build ccache on MacOS
2. `mkdir /tmp/testcase`
3. `cd /tmp/testcase`
4. `CCACHE=~/src/ccache/_build/ccache ~/src/ccache/test/run`

### Actual behavior ###

```
Test suite:     basedir (line 291)
Test case:      Absolute paths in stderr
Failure reason: reference.stderr and ccache.stderr differ
```

The diff is the following:

```diff
diff -u ./testdir/84366/run.real/reference.stderr ./testdir/84366/run.real/ccache.stderr   
--- ./testdir/84366/run.real/reference.stderr	2021-08-22 13:25:29.000000000 +0200
+++ ./testdir/84366/run.real/ccache.stderr	2021-08-22 13:25:29.000000000 +0200
@@ -1,8 +1,8 @@
-In file included from /tmp/testcase/testdir/84366/run.real/test.c:1:
-/tmp/testcase/testdir/84366/run.real/test.h:1:2: warning: test.h [-W#warnings]
+In file included from /private/tmp/testcase/testdir/84366/run.real/test.c:1:
+/private/tmp/testcase/testdir/84366/run.real/test.h:1:2: warning: test.h [-W#warnings]
 #warning test.h
  ^
-/tmp/testcase/testdir/84366/run.real/test.c:2:2: warning: test.c [-W#warnings]
+/private/tmp/testcase/testdir/84366/run.real/test.c:2:2: warning: test.c [-W#warnings]
 #warning test.c
  ^
 2 warnings generated.
```

That might happen because `/tmp` is actually a symlink to `private/tmp`.

Thanks,
Gregor

kira-bruneau:
`/bin/pwd` won't exist on some systems (eg. NixOS). The only difference between the shell builtin and the binary is that the binary uses `-P` by default. By passing `-P` we can be sure that we get the actual path, and not the symlinked path, which was needed for #916.

### DIFF
--- a/test/suites/basedir.bash
+++ b/test/suites/basedir.bash
@@ -281,7 +281,7 @@ EOF
 EOF
     backdate test.h
 
-    pwd="$(/bin/pwd -P)"
+    pwd="$(pwd -P)"
     $REAL_COMPILER -c $pwd/test.c 2>reference.stderr
 
     CCACHE_ABSSTDERR=1 CCACHE_BASEDIR="$pwd" $CCACHE_COMPILE -c $pwd/test.c 2>ccache.stderr


jrosdahl:
Thanks.

Issue #804, opened 2021-02-18
-frecord-gcc-switches support


llchan:
`-frecord-gcc-switches` is not correctly handled by ccache.

### How to reproduce ###
```
$ echo 'int main() { return 0; }' > main.c
$ mkdir foo bar
$ ccache gcc -frecord-gcc-switches -I foo main.c -c -o main.foo.o
$ ccache gcc -frecord-gcc-switches -I bar main.c -c -o main.bar.o
$ readelf -p .GCC.command.line main.foo.o

String dump of section '.GCC.command.line':
  [     0]  -I foo
  ...

$ readelf -p .GCC.command.line main.bar.o

String dump of section '.GCC.command.line':
  [     0]  -I foo
  ...

```

### Actual behavior ###
The recorded gcc switches are stale for `main.bar.o` (includes a `-I foo` rather than `-I bar`).

### Expected behavior ###
At the most conservative, when `-frecord-gcc-switches` is on, the full command line should be saved in the cache entry and any change in the command line should point to a different cache entry.  This may lead to a significant drop in cache hits when the flag is enabled, so there may need to be some sloppiness flags to dial it back.

### Environment ###
ccache version 3.7.11

jrosdahl:
Thanks for the bug report.

Another option would be to simply declare `-frecord-gcc-switches` unsupported (and fall back to just running the original compiler), but I think that it makes sense to still support caching by including the full command line in the hash, like you suggest.

I won't add a sloppiness flag as part of the fix, but I could be OK with a future PR adding such support on popular demand.

PR #936, opened 2021-09-20
Hash relative path even when using -frecord-gcc-switches

chtcvl:
Absolute path to input file is added to hash when `-frecord-gcc-switches` is set because all original arguments are appended to `extra_args_to_hash` in https://github.com/ccache/ccache/blob/bb03159d9b8e5ba44818a44b4038f181e61f6836/src/argprocessing.cpp#L1314 . This causes cache misses when compiling in different directories and using absolute path even when CCACHE_BASEDIR is set.
Fix by updating original argument whenever find it to be an input file. 

Fixes  [#937 ](https://github.com/ccache/ccache/issues/937)

### DIFF
--- a/src/argprocessing.cpp
+++ b/src/argprocessing.cpp
@@ -928,6 +928,11 @@ process_arg(Context& ctx,
 
   // Rewrite to relative to increase hit rate.
   args_info.input_file = Util::make_relative_path(ctx, args[i]);
+  if (i < ctx.orig_args.size()) {
+    // Rewrite to relative in original args in case hash_full_command_line
+    // is set (-frecord-gcc-switches) - to increase hit rate
+    ctx.orig_args.replace(i, args_info.input_file);
+  }
 
   return nullopt;
 }
--- a/src/Args.cpp
+++ b/src/Args.cpp
@@ -228,3 +228,9 @@ Args::replace(size_t index, const Args& args)
     insert(index, args);
   }
 }
+
+void
+Args::replace(size_t index, const std::string& arg)
+{
+  m_args[index] = arg;
+}
--- a/src/Args.hpp
+++ b/src/Args.hpp
@@ -85,6 +85,9 @@ public:
   // Replace the argument at `index` with all arguments in `args`.
   void replace(size_t index, const Args& args);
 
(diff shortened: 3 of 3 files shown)


src/argprocessing.cpp
```diff
@@ -928,6 +928,11 @@ process_arg(Context& ctx,
 
   // Rewrite to relative to increase hit rate.
   args_info.input_file = Util::make_relative_path(ctx, args[i]);
+  if (i < ctx.orig_args.size()) {
+    // Rewrite to relative in original args in case hash_full_command_line
+    // is set (-frecord-gcc-switches) - to increase hit rate
+    ctx.orig_args.replace(i, args_info.input_file);
```
jrosdahl, line 934: Sorry, but `process_arg` is not allowed to modify `ctx.orig_args` like that. One reason is that `ctx.orig_args` needs to stay exactly the same so that ccache can execute the original command line on failure.

Also, won't we have the same problem if there are other absolute paths than the input file on the command line, for instance `-I/home/user/test1/include`? I'm thinking there should be a more generic solution to this.

A simple and generic fix could be to add a [sloppiness option](https://ccache.dev/manual/4.4.1.html#config_sloppiness) for not hashing the full command line when `-frecord-gcc-switches`, as mentioned in https://github.com/ccache/ccache/issues/804#issuecomment-784493246. What do you think about that solution?

jrosdahl (changes requested):

src/argprocessing.cpp
chtcvl: Thanks for the explanation on `ctx.orig_args` usage to why it's not appropriate to modify it.
> Also, won't we have the same problem if there are other absolute paths than the input file on the command line, for instance -I/home/user/test1/include? I'm thinking there should be a more generic solution to this.
Agreed, didn't check this.
> A simple and generic fix could be to add a sloppiness option for not hashing the full command line when -frecord-gcc-switches, as mentioned in #804 (comment). What do you think about that solution?
Sloppiness option looks more clean and will resolve both input and include absolute paths being cached. Will put a new merge request with sloppiness option instead of this one.

PR #938, opened 2021-09-27
feat: Add full_command_line sloppiness

### LINKED ISSUE #937: Cache miss due to absolute path to include file is hashed when `-frecord-gcc-switches` is used
### How to reproduce ###

1. Create two directories with the same file
```
mkdir /home/user/test1
mkdir /home/user/test2
cp example.cpp /home/user/test1/test.cpp
cp /home/user/test1/test.cpp /home/user/test2/
```
2. Run compilation with absolute paths:
```
CCACHE_BASEDIR=/home/user ccache gcc -frecord-gcc-switches -c /home/user/test1/test.cpp
CCACHE_BASEDIR=/home/user ccache gcc -frecord-gcc-switches -c /home/user/test2/test.cpp
```

### Actual behavior ###
Second run is a cache miss as absolute path to input file is added to the hash when `-frecord-gcc-switches` is used.

### Expected behavior ###
Second run is expected to be a cache hit (at least when CCACHE_BASEDIR is set, but think that in other cases too).

### Environment ###

ccache version 4.4.1

compiled with gcc10 on CentOS 7

chtcvl:
Add a sloppiness option `full_command_line` option for not hashing the full command line when `-frecord-gcc-switches` is set.

Closes #937 , mentioned in #804

### DIFF
--- a/doc/MANUAL.adoc
+++ b/doc/MANUAL.adoc
@@ -918,6 +918,11 @@ Examples:
 *time_macros*::
     Ignore `+__DATE__+`, `+__TIME__+` and `+__TIMESTAMP__+` being present in the
     source code.
+*full_command_line*::
+    By default, ccache will add full command line to hash if `-frecord-gcc-
+    switches` is used, resulting in cache miss when compiling in
+    different directories. With this sloppiness set ccache will only hash
+    processed command line (see _<<Compiling in different directories>>_).
 --
 +
 See the discussion under _<<Troubleshooting>>_ for more information.
--- a/src/argprocessing.cpp
+++ b/src/argprocessing.cpp
@@ -1310,7 +1310,7 @@ process_args(Context& ctx)
   if (config.run_second_cpp()) {
     extra_args_to_hash.push_back(state.dep_args);
   }
-  if (state.hash_full_command_line) {
+  if (state.hash_full_command_line && !ctx.config.sloppiness().is_enabled(core::Sloppy::full_command_line)) {
     extra_args_to_hash.push_back(ctx.orig_args);
   }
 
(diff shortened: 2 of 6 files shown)


doc/MANUAL.adoc
```diff
@@ -918,6 +918,11 @@ Examples:
 *time_macros*::
     Ignore `+__DATE__+`, `+__TIME__+` and `+__TIMESTAMP__+` being present in the
     source code.
+*full_command_line*::
```
jrosdahl, line 921: I think that this name could be improved. Naming of sloppiness options should follow "I'm OK if ccache is sloppy about \<name>", so perhaps `record_gcc_switches`?

Also, please keep the list of sloppiness names sorted alphabetically (also in other places).

src/argprocessing.cpp
```diff
@@ -1310,7 +1310,8 @@ process_args(Context& ctx)
   if (config.run_second_cpp()) {
     extra_args_to_hash.push_back(state.dep_args);
   }
-  if (state.hash_full_command_line) {
+  if (state.hash_full_command_line
+      && !ctx.config.sloppiness().is_enabled(core::Sloppy::full_command_line)) {
```
jrosdahl, line 1314: I would prefer it if you check the sloppiness option when setting the flag instead:

https://github.com/ccache/ccache/blob/6f9c35810004832751465fd8f4d88a4090c02e32/src/argprocessing.cpp#L805-L807

jrosdahl (changes requested):

test/suites/basedir.bash
```diff
@@ -327,6 +327,34 @@ EOF
     expect_stat preprocessed_cache_hit 0
     expect_stat cache_miss 1
 
+    # -------------------------------------------------------------------------
+    TEST "-frecord-gcc-switches"
```
jrosdahl, line 331: Since this test (of the sloppiness option) doesn't really have anything to do with basedir, I suggest to add a few cases to the existing -frecord-gcc-switches test in base.bash instead.

jrosdahl (changes requested):

Issue #841, opened 2021-04-29
Support of Tasking compiler toolchain


vsplesk:
Hi,

I would like to use ccache in combination with TASKING VX-toolset. This is a commercial C/C++ compiler which command line syntax is unfortunately not fully GCC compatible. See https://www.tasking.com/support/tricore/ctc_user_guide_v6.2r2.pdf for its documentation. Another toolchain coming close is from Green Hills which is listed as semi-supported on your features page.

In theory, I could hack the support into ccache but the code looks quite complicated. It would be good if someone could highlight the relevant locations to add a new compiler. Or I can offer testing a few things if needed. Please note that this compiler includes its own caching functionality already but I would like to benchmark this and ccache against each other.

jrosdahl:
Hi,

There are no specific locations where you can add support for a new compiler since there is no such compiler abstraction in the code. Instead, the current (relatively small variation in) compiler-specific behavior is handled in a few different places. You can search for `config.compiler_type()` to find them.

I know next to nothing about the TASKING compiler, but if you want to work on a proof of concept we could certainly have a look at it or give guidance about specific issues.

louiscaron:
Hello @vsplesk , were you able to add support for tasking ccache ? I have tried using the tasking internal cache mechanism (--cache), but the time gain I get is so small that I am evaluating switching to ccache where I would get a better control over the cache search algorithm.  Thanks for any advice!

vsplesk:
I am sorry, I couldn't find time to work on this project. I have also found --cache option to be not very effective most of the time, i.e. greatly varying between 5% and >> 50% speedup. It's simply not reliable and not as fast as ccache used to be in regular FOSS projects (not sure about our code base, this is yet to be investigated).

In case you start implementing this soon, please send me a note about your fork/branch and I might collaborate there.

louiscaron:
This is nice of you!  I will check how I can proceed, since the --cache option only gives me a 10% speedup (very far from the 30x improve time mentioned in ccache measurements).  On your side, do you care much about preprocessing mode or will the direct mode suffice?
    Le dimanche 19 septembre 2021, 11:57:24 UTC+2, vsplesk ***@***.***> a écrit :  
 
 


I am sorry, I couldn't find time to work on this project. I have also found --cache option to be not very effective most of the time, i.e. greatly varying between 5% and >> 50% speedup. It's simply not reliable and not as fast as ccache used to be in regular FOSS projects (not sure about our code base, this is yet to be investigated).

[...]

PR #942, opened 2021-10-06
Support Tricore compiler from Tasking

louiscaron:
First draft before full support.  This should allow more people to collaborate and use the feature, without forking from my own copy.

This is related to the following issue: https://github.com/ccache/ccache/issues/841


louiscaron:
@jrosdahl could you approve the workflows? Actually, I start having several changes that are independent from each other (tasking support vs windows fixes).  I have separated them clearly in commits, but do you want me to have indididual pull requests in order to get them in the mainline? thanks

louiscaron:
@jrosdahl : thanks to the compiled unit tests, I discovered a new path format (/Z:/path/to/dir), I fixed the path substitution to match the three formats I have seen (/path/to/dir, /Z:/path/to/dir and Z:/path/to/dir).  Could you run the workflows once more? thanks

jrosdahl:
Thanks for working on this, and especially for taking the time to write tests!

As mentioned in https://github.com/ccache/ccache/pull/506#issuecomment-958103940 and #956, I have some reservations regarding adding support for yet another new compiler by following the pre-existing pattern with logic sprinkled here and there. Since this PR brings even more complexity to core functionality, I'm currently not interested in merging this as is. Sorry.

Regarding the tests: They are of course good to have for a developer that has access to a Tasking compiler, but if the test suite isn't run by CI and not by any maintainer (i.e. me) then they won't protect against regression in reality and have a large risk of becoming broken. See #957 for an idea on how this could be improved.

louiscaron:
@jrosdahl thanks a lot for the long response, if i understand correctly, there is no chance that I get this merged in the mainline as is.  But since I hate wasted work, I am ready to do anything that can address your concers, so could I:
- rework my patch to completely duplicate the portions that are compiler specific so that it does not make the core argument parsing more complex
- add fake compilers (as python scripts eventually) so the regression tests can pass in CI
let me know if you have other ideas, thanks

jrosdahl:
> @jrosdahl thanks a lot for the long response, if i understand correctly, there is no chance that I get this merged in the mainline as is.

I wouldn't go as far as saying that there is no chance, but I currently just don't have time to engage with this low priority (sorry) topic. Just wanted to say that I haven't forgotten about it and that I have nothing against supporting Tricore/Tasking in principle, given a better designed framework.

> Let me know if you have other ideas to get it to be merged, thanks

That could be helping with designing or thinking about #956 and #957, or helping out with other maintenance aspects of ccache.

vsplesk:
Is there anything we can still do? I merged @louiscaron changes with v4.6 release (mostly mechanistic chanes, not sure about argprocessing.cpp).

See https://github.com/vsplesk/ccache/tree/tricoremerge

YYGCui:
Hi, is any update for this PR? 

@louiscaron I want to use ccache for tasking compiler，pls help me check is it available or not？

louiscaron:
@YYGCui : as per a discussion with @jrosdahl , it will not be merged in the mainline for various reasons (adding too much complexity in the compiler options handling, only a commercial compiler so no non regression tests can be added to the main branch etc...)

Nevertheless, you can use my fork which has added support for this compiler.

PR #943, opened 2021-10-06
Make the tests work under Windows

rvogg:
With these changes I made some adaptations that the tests run under Windows again. 
Tests that fail and cannot be fixed quickly have been deactivated for the moment. 
Additionally I added mingw-w64 builds under Windows, because the msvc builds contain other bugs.

In some places I had to give the shell scripts a '.sh' extension to make them run under Windows. It looks like this does not affect the other operating systems.
( CCACHE_DETECT_SHEBANG does not work correctly  -> The tests produce a "could_not_find_compiler" error if the .sh extension is missing.)

I hope this change helps avoid breaking existing behavior on Windows and make ccache even better on Windows. 

Note: The secondary_http test fails sporadically on Windows. (Perhaps it should also be disabled)

### DIFF
--- a/test/suites/modules.bash
+++ b/test/suites/modules.bash
@@ -1,9 +1,11 @@
 SUITE_modules_PROBE() {
-    if ! $COMPILER_TYPE_CLANG; then
+    if ! $COMPILER_TYPE_CLANG || $COMPILER_USES_MSVC; then
         echo "-fmodules/-fcxx-modules not supported by compiler"
     else
-        touch test.c
-        $COMPILER -fmodules test.c -S || echo "compiler does not support modules"
+        cat <<EOF >testmodules.cpp
+#include <string>
+EOF
+        $COMPILER -x c++ -fmodules testmodules.cpp -S || echo "compiler does not support modules"
     fi
 }
 
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -202,7 +202,6 @@ jobs:
           - name: Windows VS2019 32-bit
             os: windows-2019
             msvc_arch: x64_x86
-            allow_test_failures: true  # For now, don't fail the build on failure
             CC: cl
             CXX: cl
             ENABLE_CACHE_CLEANUP_TESTS: 1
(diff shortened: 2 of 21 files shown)


louiscaron:
This looks good to me, actually, i had to implement some very small portions of it in order to implement the new testcases for the new compiler I would like to add support for (https://github.com/ccache/ccache/pull/942).  

If, by chance, your patch gets merged soon I would rebase my own patch. thanks

test/suites/modules.bash
```diff
@@ -34,38 +36,38 @@ SUITE_modules() {
     # -------------------------------------------------------------------------
     TEST "fall back to real compiler, no sloppiness"
 
-    $CCACHE_COMPILE -x c++ -fmodules -fcxx-modules -c test1.cpp -MD
+    $CCACHE_COMPILE -x c++ -std=c++20 -fmodules -fcxx-modules -c test1.cpp -MD
```
llunak, line 39: Why would you need to force C++20 here? In case you're confusing this for C++20 modules, this is not that, it's Clang modules.

jrosdahl:
> With these changes I made some adaptations that the tests run under Windows again.

Note that the Windows tests are work in progress and have never worked before, so it's not the case that they have become broken.

test/suites/modules.bash
rvogg: I agree, that was a mistake by me.

rvogg:
> > With these changes I made some adaptations that the tests run under Windows again.
> 
> Note that the Windows tests are work in progress and have never worked before, so it's not the case that they have become broken.

I thought that because there were many `if ! $HOST_OS_WINDOWS` are spread over the tests, it was less broken than now. 
Currently not a single test runs on Windows because they all fail on the `expect_stat` function.

My initial motivation for this change was: I started to fix the basedir functionality for Windows, because it doesn't really work with all the Windows path mess. 
After a few `#ifdef __Win32` changes I realized that I broke a lot more but couldn't really check what. So I first tried to get to a state where I could verify my changes. 

I didn't fix all tests, because in some cases ccache really doesn't work on Windows or the effort to get it running in the git-bash was too high. And I don't expect to fix all bugs for ccache under Windows, but only that no additional bugs get in by mistakes.

rvogg:
I added the commit _'handle newlines on Windows properly'_ from #954. With this I was able to remove some hacks. 

I did not add the commit _'handle properly also stdout in addition to stderr'_ because it breaks the correct ccache behavior. 
With `export CCACHE_NOCPP2=1` the preprocessor output is written to stdout.

orgads:
@rvogg ping

orgads:
Rebased and created a new PR #1133

PR #952, opened 2021-10-31
MSVC features

llunak:
These changes makes various MSVC-related features work. This PR is based on #954.


orgads:
Should this be closed? Rebased?

cristianadam:
> Should this be closed? Rebased?

Should be rebased. The features are very welcome!

orgads:
Rebased and pushed as #987.

cristianadam:
Regarding the depend mode, here is what the [ninja](https://ninja-build.org/manual.html#ref_headers) manual has to say to their corresponding feature:

>It turns out that for large projects (and particularly on Windows, where the file system is slow) loading 
> these dependency files on startup is slow.
>
> Ninja 1.3 can instead process dependencies just after they’re generated and save a compacted form 
> of the same information in a Ninja-internal database.
>
> Ninja supports this processing in two forms.
>
> 1. `deps = gcc` specifies that the tool outputs gcc-style dependencies in the form of Makefiles. Adding this to 
> the above example will cause Ninja to process the depfile immediately after the compilation finishes, 
> then delete the .d file (which is only used as a temporary).
> 2. `deps = msvc` specifies that the tool outputs header dependencies in the form produced by Visual Studio’s 
> compiler’s `/showIncludes` flag. Briefly, this means the tool outputs specially-formatted lines to its stdout. 
> Ninja then filters these lines from the displayed output. No depfile attribute is necessary, but the localized 
> string in front of the the header file path. For instance `msvc_deps_prefix = Note: including file:` for a English 
> Visual Studio (the default). Should be globally defined.

The user doesn't have to add `/showIncludes` by themselves, their build tool (ninja) requires it, and the build tool generator takes care of that (cmake).

ccache will tap in and get the dependency information, since it's already there, but not in the form of a `.d` file.

jrosdahl:
@cristianadam wrote:

> Regarding the depend mode, here is what the [ninja](https://ninja-build.org/manual.html#ref_headers) manual has to say to their corresponding feature:

Thanks for the clarification.

@orgads wrote:

> The last commit is in https://github.com/ccache/ccache/pull/992. @llunak please close this PR and move the discussion there.

I'll close it!

PR #953, opened 2021-10-31
fix: Correct win32 log messages about file locks

llunak:


louiscaron:
@jrosdahl : Can I approve this change?

PR #954, opened 2021-11-01
Fix and simplify handling of options for MSVC

llunak:
These are various fixes (on top) of #506 .

### DIFF
--- a/src/argprocessing.cpp
+++ b/src/argprocessing.cpp
@@ -242,6 +242,15 @@ process_arg(const Context& ctx,
     return nullopt;
   }
 
+  bool changed_from_slash = false;
+  if (ctx.config.compiler_type() == CompilerType::cl
+      && util::starts_with(args[i], "/")) {
+    // MSVC understands both /option and -option, so convert all /option to
+    // -option to simplify our handling.
+    args[i][0] = '-';
+    changed_from_slash = true;
+  }
+
   // Ignore clang -ivfsoverlay <arg> to not detect multiple input files.
   if (args[i] == "-ivfsoverlay"
       && !(config.sloppiness().is_enabled(core::Sloppy::ivfsoverlay))) {
@@ -252,7 +261,11 @@ process_arg(const Context& ctx,
   }
 
   // Special case for -E.
-  if (args[i] == "-E" || args[i] == "/E") {
+  if (args[i] == "-E") {
+    return Statistic::called_for_preprocessing;
+  }
+  // MSVC -P is -E with output to a file.
+  if (args[i] == "-P" && ctx.config.compiler_type() == CompilerType::cl) {
     return Statistic::called_for_preprocessing;
   }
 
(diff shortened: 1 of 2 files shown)


src/argprocessing.cpp
```diff
@@ -242,6 +242,13 @@ process_arg(const Context& ctx,
     return nullopt;
   }
 
+  if (ctx.config.compiler_type() == CompilerType::cl
+      && util::starts_with(args[i], "/")) {
+    // MSVC understands both /option and -option, so convert all /option to
+    // -option to simplify our handling.
+    args[i][0] = '-';
```
rvogg, line 249: With this change, you break absolute paths without drive letters. Depending on what build system you are using, it is quite possible that a posix style path is used without a drive letter. 
Also, the feature should work when cross compiling from Linux to Windows with clang-cl.

llunak: That doesn't seem to be the case:
C:\lo\ccache.git\build>cl -nologo -c /lo/ccache.git/build/a.cpp
cl : Command line warning D9002 : ignoring unknown option '/lo/ccache.git/build/a.cpp'
cl : Command line error D8003 : missing source filename

llunak: But fair enough with clang-cl, I've added setting the slash back when the code gets to treating it as a file.

rvogg: Ok I didn't try it, I just looked it up in the Microsoft doc. (https://docs.microsoft.com/en-us/cpp/build/reference/cl-filename-syntax?view=msvc-160)
But I think it is very dangerous to just change the character here. Even if it does not work currently. 

What if the path is in quotes? Ccache should then receive the path without quotes.

jrosdahl:
I'd appreciate it if you could submit _"handle newlines on Windows properly"_ and _"handle properly also stdout in addition to stderr"_ as separate PRs since they are independent.

llunak:
> I'd appreciate it if you could submit _"handle newlines on Windows properly"_ and _"handle properly also stdout in addition to stderr"_ as separate PRs since they are independent.

They are not, they both modify the block in ccache.cpp around line 950.

jrosdahl:
> They are not, they both modify the block in ccache.cpp around line 950.

I did not mean that the commits as currently written are independent, I meant that the fix and the feature are logically independent of "MSVC fixes". The first can be introduced separately to fix `send_to_stderr` on Windows. This point is quite minor and I'm OK with not doing that one separately if you think it's hard or tedious to do so. The second one however is a generic "we can now cache stdout for any compiler" feature, not an MSVC fix per se.

rvogg:
The commit 60a6d2d51bcd6fb6e618c1be7aba933ae08d0a06 breaks the correct behavior of ccache. 
With `export CCACHE_NOCPP2=1` the preprocessor output is written to stdout.

llunak:
> I did not mean that the commits as currently written are independent, I meant that the fix and the feature are logically independent of "MSVC fixes". The first can be introduced separately to fix `send_to_stderr` on Windows. This point is quite minor and I'm OK with not doing that one separately if you think it's hard or tedious to do so. The second one however is a generic "we can now cache stdout for any compiler" feature, not an MSVC fix per se.

Done in #962.

> The commit [60a6d2d](https://github.com/ccache/ccache/commit/60a6d2d51bcd6fb6e618c1be7aba933ae08d0a06) breaks the correct behavior of ccache. With `export CCACHE_NOCPP2=1` the preprocessor output is written to stdout.

Fixed there.

llunak:
I've separated these commits into more PRs, so this one is now only one commit.

It makes the code properly handle both -option and /option way of writing MSVC options, which also fixes several places where they weren't both listed properly. It is done by converting all /option arguments to -option (rather than listing both cases everywhere and risking mistakes), and if the handling gets to passing the argument as a file it gets changed back.

PR #955, opened 2021-11-01
fix: Correct debug dir handling on Windows

llunak:
The drive letter removal was incorrect.

### DIFF
--- a/src/util/path.cpp
+++ b/src/util/path.cpp
@@ -58,4 +58,15 @@ to_absolute_path(nonstd::string_view path)
   }
 }
 
+std::string
+to_absolute_path_no_drive(nonstd::string_view path)
+{
+  std::string abs_path = to_absolute_path(path);
+#ifdef _WIN32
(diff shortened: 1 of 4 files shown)


louiscaron (approved):

louiscaron:
I had done a simular fix on my tricore support change but yours in much more in the "ccache" spirit. I will rebase as soon as yours gets merged.  In the meantime, could you allow the worklows on my PR: https://github.com/ccache/ccache/pull/942

src/util/path.cpp
```diff
@@ -58,4 +58,15 @@ to_absolute_path(nonstd::string_view path)
   }
 }
 
+std::string
+to_absolute_path_no_drive(nonstd::string_view path)
```
jrosdahl, line 62: It would be great it you could add some tests of this to `unittest/test_util_path.cpp`.

llunak: Added.

jrosdahl:
Thanks.

PR #960, opened 2021-11-05
fix: (Util.cpp) use static_cast<int> on FICLONE, port downstream patch form alpine linux

MaxPeal:
port downstream patch form alpine linux
enable support for file cloning on ppc64le
"musl uses an `int` instead of a `unsigend long` for the ioctl function
prototype, contrary to glibc, since POSIX mandates the former. This
causes a spurious error on ppc64le which can be silenced by casting to
int explicitly.

See https://www.openwall.com/lists/musl/2020/01/20/2"
https://gitlab.alpinelinux.org/alpine/aports/-/blob/de8bce08376a189ad14809d0272bb4c02e74cdf0/main/ccache/ioctl.patch

### DIFF
--- a/src/Util.cpp
+++ b/src/Util.cpp
@@ -253,7 +253,7 @@ clone_file(const std::string& src, const std::string& dest, bool via_tmp_file)
     }
   }
 
-  if (ioctl(*dest_fd, (int)FICLONE, *src_fd) != 0) {
+  if (ioctl(*dest_fd, static_cast<int>FICLONE, *src_fd) != 0) {
     throw core::Error(strerror(errno));
   }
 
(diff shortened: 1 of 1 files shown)


jrosdahl:
Looks good to me. We're generally using `static_cast<int>`(...) rather than `(int)...`, though.

MaxPeal:
@jrosdahl PR updated as suggested,
must i squash this PR?

jrosdahl:
> must i squash this PR?

No, that's not necessary.

----

I had a closer look now and I'm not yet convinced that the cast is appropriate. The original patch description says "spurious error". Does that mean a real error or a compile-time warning?

On all glibc-based Linux dists the second argument to `ioctl` should be `unsigned long`, not `int`. Is there a risk of errors or warnings in a non-musl case if the value is cast to `int`?

If there is need to cast to `int` for musl it spontaneously feels like it should be done only when compiling with musl.

andypost:
btw Linux using `unsigned int` https://github.com/torvalds/linux/blob/master/fs/ioctl.c#L823

PR #962, opened 2021-11-14
Add support for caching stdout from MSVC

llunak:
#954 redone without depending on MSVC support in #506 and with stdout in nocpp2 case fixed.


llunak:
Since #506 has been meanwhile merged, I've updated this to again be based on that and remove the output hacks from that PR.

src/ResultRetriever.cpp
```diff
@@ -141,9 +143,12 @@ ResultRetriever::on_entry_start(uint32_t entry_number,
...
 ResultRetriever::on_entry_data(const uint8_t* data, size_t size)
 {
-  ASSERT(!(m_dest_file_type == FileType::stderr_output && m_dest_fd));
+  ASSERT(!((m_dest_file_type == FileType::stdout_output
+            || m_dest_file_type == FileType::stderr_output)
+           && m_dest_fd));
 
-  if (m_dest_file_type == FileType::stderr_output
+  if ((m_dest_file_type == FileType::stdout_output
+       || m_dest_file_type == FileType::stderr_output)
```
rvogg, line 151: Here you could omit the parenthesis.
Logically it is following: 
- Is it stdout_output ? 
- Or is it stderr_output ? 
- Or is it dependency with non-empty dest path ? 

But from my point of view you can leave it like this.

src/ccache.cpp
```diff
@@ -789,18 +789,24 @@ static bool
...
   const auto stderr_stat = Stat::stat(stderr_path, Stat::OnError::log);
-  if (!stderr_stat) {
+  if (!stdout_stat || !stderr_stat) {
     return false;
   }
 
   if (stderr_stat.size() > 0) {
     result_writer.write(Result::FileType::stderr_output, stderr_path);
   }
+  // Write stdout only after stderr (better with MSVC).
```
rvogg, line 806: Why is the order important? Can you explain that a little bit more?

rvogg (approved):

rvogg (changes requested):
The commit 193979f417e5e18e5e29e25f873492b2a8e1f83a is still broken. If you run the test `./test/run cpp1`, then the preprocessor output is written to the console.

src/ccache.cpp
llunak: MSVC output is a bit of a mess. Error messages go to stdout, and the banner that can be disabled with -nologo goes to stderr. So in the logical order you'd get the banner at the end.

llunak:
> The commit [193979f](https://github.com/ccache/ccache/commit/193979f417e5e18e5e29e25f873492b2a8e1f83a) is still broken. If you run the test `./test/run cpp1`, then the preprocessor output is written to the console.

In what setup?

rvogg:
> > The commit [193979f](https://github.com/ccache/ccache/commit/193979f417e5e18e5e29e25f873492b2a8e1f83a) is still broken. If you run the test `./test/run cpp1`, then the preprocessor output is written to the console.
> 
> In what setup?

- On Windows with x86_64-w64-mingw32-gcc 11. 2.0 (used for build and test)
- On Ubuntu with gcc 9.3.0

src/ccache.cpp
rvogg: I took another look at it: 
Isn't that where it is written to the disk? 
The order doesn't matter, because they are two different files. 
I think it's better if you remove the comment here because it is more confusing.

llunak: ResultRetriever processes items in the order it reads them from the disk.

rvogg: > ResultRetriever processes items in the order it reads them from the disk.

Ok then I propose: 
```suggestion
  // Write stdout only after stderr (better with MSVC), because the ResultRetriever
  // processes the items in the order it reads them from disk afterwards. 
```

jrosdahl:
Sorry if I was unclear in https://github.com/ccache/ccache/pull/954#issuecomment-962625533, but as I see it, adding support for caching stdout from the compiler is a generic new feature (not a fix of anything since nothing is broken). In contrast, handling newlines properly on Windows sounds like a bug fix. I would appreciate new features in standalone pull requests.

rvogg:
> Sorry if I was unclear in [#954 (comment)](https://github.com/ccache/ccache/pull/954#issuecomment-962625533), but as I see it, adding support for caching stdout from the compiler is a generic new feature (not a fix of anything since nothing is broken). In contrast, handling newlines properly on Windows sounds like a bug fix. I would appreciate new features in standalone pull requests.

I've open a standalone PR for the newline bugfix: #968
After integrating that, this one can be changed to the feature PR for _"feat: Add stdout caching"_.

src/ccache.cpp
```diff
@@ -1106,6 +1103,7 @@ get_result_key_from_cpp(Context& ctx, Args& args, Hash& hash)
   if (!ctx.config.run_second_cpp()) {
     // If we are using the CPP trick, we need to remember this stderr data and
     // output it just before the main stderr from the compiler pass.
+    ctx.cpp_stdout = stdout_path;
```
rvogg, line 1106: At this point the `stdout_path` contains the output of the preprocessor and must not be printed to the console.  

The variable `stdout_path` should perhaps also be renamed so that no further confusion is caused here. (It can also be the path to a *.i file.)

With MSVC it is not clear to me what happens with the flag /E. Are then also still error texts in the stdout which must be printed here?

rvogg: I just don't understand why the msvc build here doesn't also write the preprocessor output to the console. @llunak can you explain or imagine why this is?

jrosdahl:
> I've open a standalone PR for the newline bugfix: #968 After integrating that, this one can be changed to the feature PR for _"feat: Add stdout caching"_.

OK. Since I don't have much time for ccache right now (as you probably have noticed), I'll do some tweaks myself to speed up the turnaround time.

jrosdahl:
I've now rebased the branch to latest master. I hope I didn't make any mistakes, but please check.

Some quick thoughts on the PR:

* The _Compiler produced stdout_ part of the manual needs to be updated since the description is no longer true. It would probably be OK to just remove it.
* It would be very good with some integration tests for the new feature. Then you'll be able to catch the problem that @rvogg points out: We shouldn't cache stdout from the preprocessor.

src/ccache.cpp
llunak: I've just submitted #970 which makes this moot.

llunak:
> The commit [193979f](https://github.com/ccache/ccache/commit/193979f417e5e18e5e29e25f873492b2a8e1f83a) is still broken. If you run the test `./test/run cpp1`, then the preprocessor output is written to the console.

That should work now.

>     * The _Compiler produced stdout_ part of the manual needs to be updated since the description is no longer true. It would probably be OK to just remove it.

I've added a note to it.
 
>     * It would be very good with some integration tests for the new feature. Then you'll be able to catch the problem that @rvogg points out: We shouldn't cache stdout from the preprocessor.

The commit now includes a test that there's no extra stdout output.

jrosdahl:
Now when I read the PR more closely I realize that it only adds support for caching stdout from MSVC, not from any compiler like I originally envisioned. But that's OK, I'll look into that myself.

PR #971, opened 2021-11-28
fix: Handle MSVC flags -MD[d], -MT[d] and -LT[d] properly

llunak:
This depends on #954 .

### DIFF
--- a/src/argprocessing.cpp
+++ b/src/argprocessing.cpp
@@ -570,7 +570,8 @@ process_arg(const Context& ctx,
     return nullopt;
   }
 
-  if (util::starts_with(args[i], "-MQ") || util::starts_with(args[i], "-MT")) {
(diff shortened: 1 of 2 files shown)


src/argprocessing.cpp
```diff
@@ -579,6 +593,21 @@ process_arg(const Context& ctx,
     return nullopt;
   }
 
+  // MSVC -MD[d], -MT[d] and -LT[d] options are something different than GCC's
+  // -MD etc.
+  if (config.compiler_type() == CompilerType::cl
+      && (util::starts_with(args[i], "-MD")
+          || util::starts_with(args[i], "-MDd")
```
jrosdahl, line 600: This (and similar below) is redundant since an argument starting with `-MDd` also starts with `-MD`. I'll fix it.

PR #973, opened 2021-11-28
Fix upload-redis after cache entry format change

afbjorklund:
There were some assertions in the file upload, that didn't match the new file format.

Updated the helper script to work with both versions, so that it would match either...


jrosdahl:
How about removing the asserts instead?

orgads:
ping

PR #979, opened 2021-12-09
Fix copying of binary files on Windows

rvogg:
The Secondary Storage creates corrupted files when they are copied to the Primary Storage.

The flag `O_BINARY` fix that issue.


jrosdahl:
Thanks.

Issue #982, opened 2021-12-29
Bogus `unsupported code directive` (.incbin false positive)


asheplyakov:
### How to reproduce ###

```bash
cat > test.cpp <<-EOF
struct A {
   void incbin(); 
};
void f() {
  A a;
  a.incbin(); 
}
EOF 
env CCACHE_DEBUG=1 ccache g++ -c test.cpp
grep 'incbin directive' test.o.ccache-log'
```

### Actual behavior ###

ccache rejects to cache the compilation result:
```
Found unsupported .incbin directive in source code
Failed; falling back to running the real compiler
Executing /usr/bin/g++ -c test.cc
Result: unsupported code directive
```

### Expected behavior ###

ccache caches the compilation result

### Environment ###

ccache version master.1045a6f0

PR #983, opened 2021-12-29
fix: avoid .incbin false positive

asheplyakov:
With this patch accepts C++ code like `obj.incbin(x)`.
Previously it complained `Found unsupported .incbin directive`.

Fixes: #982


jrosdahl:
Thanks.

PR #987, opened 2022-01-07
MSVC Features

orgads:
This is a rebased version of #952.

Hope I got it right.

### DIFF
--- a/src/ArgsInfo.hpp
+++ b/src/ArgsInfo.hpp
@@ -61,6 +61,9 @@ struct ArgsInfo
   // Is the compiler being asked to output dependencies?
   bool generating_dependencies = false;
 
+  // Is the compiler being asked to output includes (MSVC -showIncludes)?
+  bool generating_includes = false;
+
   // Seen -MD or -MMD?
   bool seen_MD_MMD = false;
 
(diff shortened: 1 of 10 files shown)


cristianadam:
My commit should already be merged. I don't know why it's being referenced. See https://github.com/ccache/ccache/pull/506

orgads:
> My commit should already be merged. I don't know why it's being referenced. See #506

It looks like there are some changes here (additional flags supported with / and not only -). Maybe this was a preliminary revision of your commit? Should I omit this commit?

cristianadam:
> > My commit should already be merged. I don't know why it's being referenced. See #506
> 
> It looks like there are some changes here (additional flags supported with `/` and not only `-`). Maybe this was a preliminary revision of your commit? Should I omit this commit?

It creates a bit of confusion. But having both `/` and `-` is a good thing. The commit should have a different commit message.

orgads:
Done

cristianadam:
With this PR we can then update the page at https://ccache.dev/platform-compiler-language-support.html.

- cl is supported
- clang-cl is also supported

jrosdahl:
Thanks for driving the MSVC features forward.

However, as mentioned in https://github.com/ccache/ccache/pull/952#issuecomment-1009285845, I would prefer not to review and merge the features in a single PR.

> Also note It looks like there are some changes here (additional flags supported with / and not only -)

The special handling of `/` version of the options is obsolete as of #954.

orgads:
Regarding the `/`, I thought so, but didn't know where to look. Thanks for the reference :)

I'll split the other commits.

PR #991, opened 2022-01-11
support MSVC /Yu option (compiling with PCH)

orgads:
This more or less maps to PCH support for GCC/Clang, just handle
the /Yu and /Fp options.

Note that MSVC is apparently unable to generate the same PCH
from the same sources, just like GCC, so a rebuild that also rebuilds
the PCH will get cache misses. So either somebody can add support
for -Yc (if caching those huge PCHs is a good idea), or
CCACHE_PCH_EXTSUM can be used (with a checksum based on preprocessed
PCH input, i.e. -E instead of -c -Yc).

### DIFF
--- a/src/argprocessing.cpp
+++ b/src/argprocessing.cpp
@@ -50,6 +50,8 @@ struct ArgumentProcessingState
   bool found_S_opt = false;
   bool found_pch = false;
   bool found_fpch_preprocess = false;
+  bool found_Yu = false;
+  bool found_valid_Fp = false;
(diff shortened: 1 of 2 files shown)


orgads:
@llunak I hope you don't mind I'm pushing your commits (I didn't reset-author, so they're still on your name).

llunak:
You mean you've rebased #952 and are now cherry-picking and submitting them one by one? Fine by me, thank you.

PR #992, opened 2022-01-11
feat: Support depend mode for MSVC

orgads:
Based on -showIncludes, which prints included files on stdout
with a certain text prefix. Otherwise pretty similar to depend mode
handling for GCC.

This makes MSVC building way faster.


orgads:
Quoting from #952

* Missing documentation.
* Unit tests would be appreciated.
* Since the feature doesn't use dependency files, I don't think that it makes sense to piggy back on the name "depend mode". We could consider renaming it to something more generic that doesn't lock us into implementation details, though (not sure what).
* Also, do I understand correctly that the user is expected to add `/showIncludes` themselves? If so, is that really reasonable since  stderr will be littered with header file lists intermixed with errors and warnings?

orgads:
@jrosdahl Can you please reply?

orgads:
Thank you.

jrosdahl:
@llunak wrote:

> I think these two apply to most of the MSVC commits, or they at least need checked. As for tests, do MSVC-based tests already work?

I was referring to the unit tests, i.e. the tests in the [unittest](https://github.com/ccache/ccache/tree/master/unittest) directory, not the integration tests in the [test](https://github.com/ccache/ccache/tree/master/test) directory. The `ShowIncludesParser` code is a good candidate for unit testing since it does not depend on a real compiler.

> It really is the same as the GCC-based depend mode, it just uses a different option and output format, but otherwise it's the same, including the usage. The build system will usually add the option automatically, just like it does for GCC, and it should also filter it out from MSVC's output (and commented in https://github.com/ccache/ccache/pull/952 that's the case for ninja and e.g. LibreOffice's build system works similarly).

OK, thanks. Since e.g. Ninja already makes use of it out of the box I agree that it makes sense to support it.

> I think even the name depend mode still fits, it's a mode that uses dependencies generated by the compiler, in whatever format the compiler provides them.

OK, I can live with that name if it's documented in a good way.

justinpryzby:
Docs seem to say that cl.exe writes to stderr, not stdout.
Does it depend on the version or ??

https://docs.microsoft.com/en-us/cpp/build/reference/showincludes-list-include-files?view=msvc-170
> The /showIncludes option emits to stderr, not stdout.

orgads:
> Docs seem to say that cl.exe writes to stderr, not stdout. Does it depend on the version or ??
> 
> https://docs.microsoft.com/en-us/cpp/build/reference/showincludes-list-include-files?view=msvc-170
> 
> > The /showIncludes option emits to stderr, not stdout.

It's a lie! 😈

I just tested, and it outputs to stdout.

orgads:
Reported now. https://github.com/MicrosoftDocs/cpp-docs/issues/3785

justinpryzby:
> Reported now. [MicrosoftDocs/cpp-docs#3785](https://github.com/MicrosoftDocs/cpp-docs/issues/3785)

It hardly matters, since if they change it, you'd still have to assume that the output could go to either.  (without assuming that every user is using the latest release of the latest version).

justinpryzby:
> They won't change it, since that will break existing tools. I reported so they fix the docs.

Will you plan to make another version of the patch to check both stdout and stderr ?

orgads:
> > They won't change it, since that will break existing tools. I reported so they fix the docs.
> 
> Will you plan to make another version of the patch to check both stdout and stderr ?

The docs were updated. stderr is no longer mentioned.

rkitover:
@orgads Hi, could you please rebase on master? I'm getting some compile errors due to recent changes like the bump to c++17.

rkitover:
@orgads Builds rebased on master for me with this patch:

```diff
diff --git a/src/ShowIncludesParser.cpp b/src/ShowIncludesParser.cpp
index 9e7365f3..e99ce817 100644
--- a/src/ShowIncludesParser.cpp
+++ b/src/ShowIncludesParser.cpp
@@ -23,8 +23,8 @@

 namespace ShowIncludesParser {

-std::vector<nonstd::string_view>
-tokenize(nonstd::string_view file_content, nonstd::string_view prefix)
+std::vector<std::string_view>
+tokenize(std::string_view file_content, std::string_view prefix)
 {
   // -showIncludes output is written to stdout together with other messages.
   // Every line of it is '<prefix> <spaces> <file>', prefix is 'Note: including
@@ -34,17 +34,17 @@ tokenize(nonstd::string_view file_content, nonstd::string_view prefix)
     prefix = "Note: including file:";
   }

-  std::vector<nonstd::string_view> result;
+  std::vector<std::string_view> result;
   // This will split at each \r or \n, but that simply means there will be empty
   // "lines".
-  for (nonstd::string_view line :
+  for (std::string_view line :
        Util::split_into_views(file_content, "\r\n")) {
     if (util::starts_with(line, prefix)) {
       size_t pos = prefix.size();
       while (pos < line.size() && isspace(line[pos])) {
         ++pos;
       }
-      nonstd::string_view include = line.substr(pos);
+      std::string_view include = line.substr(pos);
       if (!include.empty()) {
         result.push_back(include);
       }
diff --git a/src/ShowIncludesParser.hpp b/src/ShowIncludesParser.hpp
index 6a774641..89559c98 100644
--- a/src/ShowIncludesParser.hpp
+++ b/src/ShowIncludesParser.hpp
@@ -18,13 +18,12 @@

 #pragma once

-#include "third_party/nonstd/string_view.hpp"
-
+#include <string_view>
 #include <vector>

 namespace ShowIncludesParser {

-std::vector<nonstd::string_view> tokenize(nonstd::string_view file_content,
-                                          nonstd::string_view prefix);
+std::vector<std::string_view> tokenize(std::string_view file_content,
+                                       std::string_view prefix);

 } // namespace ShowIncludesParser
diff --git a/src/argprocessing.cpp b/src/argprocessing.cpp
index 18d1f7c8..44da048c 100644
--- a/src/argprocessing.cpp
+++ b/src/argprocessing.cpp
@@ -648,7 +648,7 @@ process_option_arg(const Context& ctx,
   if (args[i] == "-showIncludes") {
     args_info.generating_includes = true;
     state.dep_args.push_back(args[i]);
-    return nullopt;
+    return std::nullopt;
   }

   if (args[i] == "-fprofile-arcs") {
diff --git a/src/ccache.cpp b/src/ccache.cpp
index 5db23aac..e7db244b 100644
--- a/src/ccache.cpp
+++ b/src/ccache.cpp
@@ -678,7 +678,7 @@ result_key_from_depfile(Context& ctx, Hash& hash)

 // Extract the used includes from -showIncludes output in stdout. Note that we
 // cannot distinguish system headers from other includes here.
-static optional<Digest>
+static std::optional<Digest>
 result_key_from_includes(Context& ctx,
                          Hash& hash,
                          const std::string& stdout_path)
@@ -688,10 +688,10 @@ result_key_from_includes(Context& ctx,
     file_content = Util::read_file(stdout_path);
   } catch (const core::Error& e) {
     LOG("Cannot open stdout tmp file {}: {}", stdout_path, e.what());
-    return nullopt;
+    return std::nullopt;
   }

-  for (string_view token : ShowIncludesParser::tokenize(
+  for (std::string_view token : ShowIncludesParser::tokenize(
          file_content, ctx.config.msvc_dep_prefix())) {
     if (!ctx.has_absolute_include_headers) {
       ctx.has_absolute_include_headers = util::is_absolute_path(token);
```

orgads:
Thank you, sorry for the delay. Done.

rkitover:
Awesome, thank you! Hope we can get this merged soon :thumbsup:

orgads:
Rebased and adapted to build. @jrosdahl Please review.

justinpryzby:
Confirmed that this makes building postgres about 2x faster when the cache is empty.
Normally that takes ~12 minutes.  Since cache misses are so slow (3x-4x slower than without using ccache), that could easily mean the difference between ccache being a win or a loss for our CI.
https://cirrus-ci.com/task/5577311092211712
https://cirrus-ci.com/task/5588532281475072

jrosdahl:
@orgads: The implementation looks reasonable to me after a cursory glance. That leaves _"Missing documentation"_ and _"Unit tests would be appreciated"_. Is those something that you would like to address?

orgads:
@llunak can you take care of these issues?

jrosdahl:
@justinpryzby wrote:

> cache misses are so slow (3x-4x slower than without using ccache)

That doesn't make sense to me. What on earth makes them so much slower?

> Would it make sense to suppress (or allow suppressing) the output from -ShowIncludes, during cache hits ? [cirrus-ci.com/task/5401686842802176](https://cirrus-ci.com/task/5401686842802176)

I think that it could be a good idea for ccache to add `/showIncludes` by itself if `depend_mode = true` (and `/showIncludes` isn't already on the command line) and then filter out the `Note: including file:` lines before emitting the stdout to the user. Unless I'm missing something, that would improve the use case for build systems that aren't based on Ninja. Does this make sense?

orgads:
> > Would it make sense to suppress (or allow suppressing) the output from -ShowIncludes, during cache hits ? [cirrus-ci.com/task/5401686842802176](https://cirrus-ci.com/task/5401686842802176)
> 
> I think that it could be a good idea for ccache to add `/showIncludes` by itself if `depend_mode = true` (and `/showIncludes` isn't already on the command line) and then filter out the `Note: including file:` lines before emitting the stdout to the user. Unless I'm missing something, that would improve the use case for build systems that aren't based on Ninja. Does this make sense?

Sounds good to me.

orgads:
> @justinpryzby wrote:
> 
> > cache misses are so slow (3x-4x slower than without using ccache)
> 
> That doesn't make sense to me. What on earth makes them so much slower?

I can reproduce that. When running without `-showIncludes`, or with depend_mode disabled, ccache runs the preprocessor, and it takes as long as (and sometimes more than) the compiler itself.

On my tests, compilation without ccache takes ~1.7s, while with ccache it can reach even 5s, average is around 3-4s.

With depend mode and `-showIncludes`, performance is quite close to without ccache. 1.7s -> 2s.

orgads:
Auto mode pushed in #1158.

orgads:
Please review the documentation. I'll try to add unit tests tomorrow.

doc/MANUAL.adoc
```diff
@@ -1549,7 +1549,8 @@ The direct mode will be disabled if any of the following holds:
 If the depend mode is enabled, ccache will not use the preprocessor at all. The
 hash used to identify results in the cache will be based on the direct mode
 hash described above plus information about include files read from the
-dependency file generated by the compiler with `-MD` or `-MMD`.
+dependency list generated by MSVC with -showIncludes, or the dependency file
```
jrosdahl, line 1552:
```suggestion
dependency list generated by MSVC with /showIncludes, or the dependency file
```

doc/MANUAL.adoc
```diff
@@ -1566,15 +1567,16 @@ Disadvantages:
   to some types of changes of compiler options and source code changes.
 * If -MD is used, the manifest entries will include system header files as
   well, thus slowing down cache hits slightly, just as using -MD slows down
-  make.
+  make. This is also the case for MSVC with -showIncludes.
```
jrosdahl, line 1570:
```suggestion
  make. This is also the case for MSVC with `/showIncludes`.
```

doc/MANUAL.adoc
```diff
@@ -1566,15 +1567,16 @@ Disadvantages:
...
 * If -MMD is used, the manifest entries will not include system header files,
   which means ccache will ignore changes in them.
 
 The depend mode will be disabled if any of the following holds:
 
 * <<config_depend_mode,*depend_mode*>> is false.
 * <<config_run_second_cpp,*run_second_cpp*>> is false.
-* The compiler is not generating dependencies using `-MD` or `-MMD`.
+* The compiler is not generating dependencies using `-MD` or `-MMD` (for MSVC,
+  -showIncludes).
```
jrosdahl, line 1579:
```suggestion
  `/showIncludes`).
```

jrosdahl:
> Please review the documentation. I'll try to add unit tests tomorrow.

Thanks.

Regarding the manual I'm missing documentation of the `msvc_dep_prefix`/`CCACHE_MSVC_DEP_PREFIX` configuration option.

Another thing that comes to mind when I read the diff is that it would be good to use `msvc_dep_prefix` in `rewrite_stdout_from_compiler` as well:

https://github.com/ccache/ccache/blob/e9ddf6800eaed6ce45bcb01f976995ef423c6ae2/src/ccache.cpp#L936-L941

doc/MANUAL.adoc
orgads: Done

doc/MANUAL.adoc
orgads: Done

doc/MANUAL.adoc
orgads: Done

orgads:
Done. I moved the doc and tests to a separate commit.

justinpryzby:
On Mon, Sep 12, 2022 at 11:01:12PM -0700, Orgad Shaneh wrote:
> > @justinpryzby wrote:
> > 
> > > cache misses are so slow (3x-4x slower than without using ccache)
> > 
> > That doesn't make sense to me. What on earth makes them so much slower?
> 
> I can reproduce that. When running without `-showIncludes`, or with depend_mode disabled, ccache runs the preprocessor, and it takes as long as (and sometimes more than) the compiler itself.
> On my tests, compilation without ccache takes ~1.7s, while with ccache it can reach even 5s, average is around 3-4s.
> With depend mode and `-showIncludes`, performance is quite close to without ccache. 1.7s -> 2s.

I wonder if it's because of the "file tracker" used by MSVC to support
incremental compilation.  Even when it's ostensibly disabled, I still
see that "tlog" files are written.  If the file tracker is active, it
seems like it'd be similar to running ccache (and the compiler) under
strace - obviously not good for performance.

jrosdahl:
Thanks.

rkitover:
@orgads congrats on getting this merged! Some of us have been pulling this PR into our local copies.

orgads:
👍

Issue #958, opened 2021-11-02
Improved compiler identification


jrosdahl:
Ccache currently guesses the compiler type based on the name of the executable. This was the easy thing to do when the need first arose (d875edef19d0ab157b2f6408330bb1392fd38f60), and the same method has subsequently been used for other compiler types. However, in some cases complex workarounds are needed since ccache can't fully trust the classification as it's just a guess. This method is inherently brittle since the compiler could be an anonymous `/usr/bin/cc` or a wrapper script with a custom name. Also, to implement features that require very specific functionality from compilers (e.g. #815), guessing compiler from just filename simply won't do.

To improve this, we could cache the output of `$compiler --version` or similar and deduce compiler type and version from that output when ccache is called for compilation.

Implementation sketch:

```python
# To determine the key to use for caching "compiler --version":
if config.compiler_check() is "none":
    compiler_check_key = None
else:
    compiler_check_key = CCACHE_VERSION + basename(compiler)
    if config.compiler_check() is a command string:
        compiler_check_key += execute(config.compiler_check())
    elif config.compiler_check() is "string:value":
        compiler_check_key += value
    elif config.compiler_check() is "content":
        compiler_check_key += read_file(compiler)
    else: # config.compiler_check() is "mtime"
        compiler_check_key += get_mtime(compiler) + get_size(compiler)

# To determine compiler type and version:
compiler_version_output = None
if compiler_check_key:
    compiler_version_output = get_cached_compiler_version(compiler_check_key)
if not compiler_version_output:
    compiler_version_output = execute(compiler --version)
    # Note: Another methods could be used based on the compiler name.

    if not compiler_version_output:
        statistics.increment(failed_to_get_compiler_version)
        fall back to running the original compiler
    if compiler_check_key:
        put_cached_compiler_version(compiler_check_key, compiler_version_output)

compiler_info = classify_compiler(compiler_version_output)

# To add compiler information to the input hash:
input_hash.hash(basename(compiler) + compiler_version_output)

# TODO:
# - If config.compiler_type() is set, use that value instead.
# - Allow compiler_type to be "type version" instead just "type"?
```

Issue #956, opened 2021-11-02
Compiler abstraction


jrosdahl:
Ccache started with support for GCC and compilers that were "similar enough". During the years, more support for GCC-like compilers like Clang was added with well-placed `if`/`else` statements. Then NVCC support was added with even more `if`s and `else`s. I've accepted the current state since both Clang and NVCC support come with reasonably extensive test suites so that it will hopefully be easy to refactor compiler support in the future without breaking stuff.

However, since there are wishes for supporting compilers that don't behave like GCC at all (e.g. MSVC in #506 and Tasking in #841/#942) there is a desperate need of better code structure - the code base simply can't handle more accidental complexity. There should be an API that captures compiler specifics such as how to process arguments, how to preprocess code, how to retrieve included files (for the direct mode), etc. Compiler-specific behavior should then be put in implementations of that API.

PR #993, opened 2022-01-11
Detect clang-cl as a compiler

orgads:
Without this, clang-cl is detected as clang and then ccache barfs
because clang-cl produces stdout output.

### DIFF
--- a/src/Config.cpp
+++ b/src/Config.cpp
@@ -241,6 +241,8 @@ parse_compiler_type(const std::string& value)
 {
   if (value == "clang") {
     return CompilerType::clang;
+  } else if (value == "clang-cl") {
+    return CompilerType::clangcl;
   } else if (value == "gcc") {
     return CompilerType::gcc;
   } else if (value == "msvc") {
@@ -451,6 +453,7 @@ compiler_type_to_string(CompilerType compiler_type)
     return "auto";
 
     CASE(clang);
+    CASE(clangcl);
     CASE(gcc);
     CASE(msvc);
     CASE(nvcc);
--- a/src/argprocessing.cpp
+++ b/src/argprocessing.cpp
@@ -201,7 +201,7 @@ process_profiling_option(const Context& ctx,
     new_profile_path = arg.substr(arg.find('=') + 1);
   } else if (arg == "-fprofile-generate" || arg == "-fprofile-instr-generate") {
     args_info.profile_generate = true;
-    if (ctx.config.compiler_type() == CompilerType::clang) {
+    if (ctx.config.is_compiler_group_clang()) {
       new_profile_path = ".";
     } else {
       // GCC uses $PWD/$(basename $obj).
@@ -273,8 +273,7 @@ process_arg(const Context& ctx,
   }
 
   bool changed_from_slash = false;
-  if (ctx.config.compiler_type() == CompilerType::msvc
-      && util::starts_with(args[i], "/")) {
+  if (ctx.config.is_compiler_group_cl() && util::starts_with(args[i], "/")) {
     // MSVC understands both /option and -option, so convert all /option to
     // -option to simplify our handling.
     args[i][0] = '-';
@@ -295,7 +294,7 @@ process_arg(const Context& ctx,
     return Statistic::called_for_preprocessing;
   }
   // MSVC -P is -E with output to a file.
-  if (args[i] == "-P" && ctx.config.compiler_type() == CompilerType::msvc) {
+  if (args[i] == "-P" && ctx.config.is_compiler_group_cl()) {
     return Statistic::called_for_preprocessing;
   }
 
@@ -307,7 +306,7 @@ process_arg(const Context& ctx,
       ++argpath;
     }
     auto file_args =
-      Args::from_atfile(argpath, config.compiler_type() == CompilerType::msvc);
+      Args::from_atfile(argpath, ctx.config.is_compiler_group_cl());
     if (!file_args) {
       LOG("Couldn't read arg file {}", argpath);
       return Statistic::bad_compiler_arguments;
@@ -450,8 +449,7 @@ process_arg(const Context& ctx,
   }
 
   // MSVC -Fo with no space.
-  if (util::starts_with(args[i], "-Fo")
-      && config.compiler_type() == CompilerType::msvc) {
+  if (util::starts_with(args[i], "-Fo") && config.is_compiler_group_cl()) {
     args_info.output_obj =
       Util::make_relative_path(ctx, string_view(args[i]).substr(3));
     return nullopt;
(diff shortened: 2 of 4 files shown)


orgads:
This commit has a minor dependency with #991 so I stacked them. I can reorder them if you find it easier to approve this one (make this PR stand-alone, and the other one will depend on it).

orgads:
Rebased. Should be easier to review now.

src/ccache.cpp
```diff
@@ -228,7 +228,9 @@ guess_compiler(string_view path)
 #endif
 
   const string_view name = Util::base_name(compiler_path);
-  if (name.find("clang") != nonstd::string_view::npos) {
+  if (name.find("clang-cl") != nonstd::string_view::npos) {
+    return CompilerType::clangcl;
```
jrosdahl, line 232: This is starting to get a bit complicated. I don't want to end up in a situation where there is a plethora of compiler types, each describing details of compilers. That has to be handled by other means. See also #958 for what I envision will be in future change in how compiler detection works.

What would happen if clang-cl was just characterized as cl instead? (On a side note, a better name would be "msvc" which I plan to rename "cl" to. Compiler type is _type_, not necessarily the actual compiler name.) Put differently, is there a case where clang-cl behaves different enough from cl that we need to handle it specially?

src/Config.cpp
```diff
@@ -247,6 +247,12 @@ parse_compiler_type(const std::string& value)
     return CompilerType::nvcc;
   } else if (value == "other") {
     return CompilerType::other;
+  } else if (value == "cl") {
+    return CompilerType::cl;
+  } else if (value == "clangcl" || value == "clang-cl") {
+    // Allow also "clang-cl" because that's how the compiler is called
+    // but clang-cl is not a valid C++ name.
```
jrosdahl, line 254: No aliases here, please. This is what a human would enter in the configuration so one is enough.

Also, documentation is missing.

src/ccache.cpp
orgads: clang-cl supports also (some of) clang's flags like `-W*` and many `-f*`. See https://clang.llvm.org/docs/UsersManual.html#id10

src/Config.cpp
orgads: So do you prefer just `clangcl` and not matching the compiler name, or `clang-cl`, and not using `CASE(clangcl)` in `compiler_type_to_string()`?

src/ccache.cpp
jrosdahl: Right, but that doesn't mean that we have to explicitly _not_ handle `-W*` and `-f*` for cl/MSVC. For example, there is no logic for handling `-fdirectives-only` only if we know that we're dealing with GCC, or handle `-F` only for cl/MSVC. Compiler type only needs to be considered when ccache has to treat an option differently for different compilers. I would like to avoid adding more complexity in this area (see also #956).

So I think my question _"What would happen if clang-cl was just characterized as cl instead?"_ is still relevant.

src/Config.cpp
jrosdahl: I prefer `clang-cl` if we must since that seems to be its name, but I'd rather classify `clang-cl` as `cl` if possible as discussed above.

src/ccache.cpp
orgads: @llunak Can you justify it?

orgads: Done

src/argprocessing.cpp
```diff
@@ -1302,17 +1303,22 @@ process_args(Context& ctx)
   // Since output is redirected, compilers will not color their output by
   // default, so force it explicitly.
   nonstd::optional<std::string> diagnostics_color_arg;
-  if (config.compiler_type() == CompilerType::clang) {
+  switch (config.compiler_type()) {
+  case CompilerType::clang:
+  case CompilerType::msvc:
```
jrosdahl, line 1308: This looks like a case where we can't let `CompilerType::msvc` do "clang-cl things" since the code as currently written will add `-fcolor-diagnostics` to the command line, which won't work for MSVC.

I suggest not support color diagnostics for clang-cl for now.

src/argprocessing.cpp
```diff
@@ -201,7 +201,8 @@ process_profiling_option(const Context& ctx,
     new_profile_path = arg.substr(arg.find('=') + 1);
   } else if (arg == "-fprofile-generate" || arg == "-fprofile-instr-generate") {
     args_info.profile_generate = true;
-    if (ctx.config.compiler_type() == CompilerType::clang) {
+    if (ctx.config.compiler_type() == CompilerType::clang
+        || ctx.config.compiler_type() == CompilerType::msvc) {
```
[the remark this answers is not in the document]
orgads: clang-cl 12 supports `-fprofile-generate` and `-fprofile-instr-generate`. If I read it correctly, this section here is about these 2 arguments only. `-fbranch-probabilities` is not supported (neither with clang nor clang-cl).

src/argprocessing.cpp
orgads: Good point, but it's too bad not to supported.

Removed for now, but I think it will be better to have some sort of indication for clang after all.

src/argprocessing.cpp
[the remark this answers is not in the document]
jrosdahl: I see. I didn't try `-fprofile-generate` but some other options, so this sounds logically OK then. But it's indeed strange to include clang-cl behavior in `CompilerType::msvc`.

src/argprocessing.cpp
jrosdahl: Perhaps I should try to explain why I'm spontaneously pushing back on this (having clang-cl as a separate compiler type). The main reason is that it's not clear to me how to represent this somewhat strange mix of two compiler types (i.e., clang-cl) when implementing #956, and I don't want to add logic now that makes such code refactoring even harder. I'd rather wait with such features.

Also, what I agreed to in https://github.com/ccache/ccache/pull/506#issuecomment-962452481 was to add basic support for MSVC, with the original patch being small and well-contained. But now there have are been several MSVC-related fixes and features on top of that and also clang-cl support, which just takes more steps on the road of accidental complexity which I have to maintain. This is demotivating to me.

orgads: I see your point.

I don't use clang-cl personally, I just wanted to have good support for msvc.

Regarding the abstraction, I think the original patch was more accurate.

When refactoring, you can have feature sets, and each compiler can be associated with one or more of them. So gcc will have gcc-like, clang will have gcc-like + clang, and clang-cl will have msvc + clang.

Then arguments can be supported by querying the feature sets of the compiler.

This is a raw idea, and it needs more processing. What do you think? Do you have a design already?

jrosdahl: No, I haven't begun thinking about that since I don't have time, which is why I'm by default a bit sceptical of change right now. :raised_eyebrow: That said, I don't like the current PR version's `ctx.config.compiler_type() == CompilerType::msvc` check in `process_profiling_option` either since it's "lying".

I wanted an answer to _"What would happen if clang-cl was just characterized as cl instead?"_ and now I've got it. Let's go with the original approach for now. (No guarantees that it will stay that way in the future, though.)

PR #994, opened 2022-01-11
do not handle -o<file> for MSVC

orgads:
While MSVC technically has that option as deprecated, it's a source
of trouble with the way ccache curently handles the option, and since
it broke ccache's debug output, it was hard to find.


orgads:
Rebased and removed dependency.

jrosdahl:
Thanks.

PR #995, opened 2022-01-11
MSVC's -Zs is GCC's -fsyntax-only

orgads:


src/argprocessing.cpp
```diff
@@ -1166,6 +1174,8 @@ process_args(Context& ctx)
   if (!state.found_c_opt && !state.found_dc_opt && !state.found_S_opt) {
     if (args_info.output_is_precompiled_header) {
       state.common_args.push_back("-c");
+    } else if (state.found_Zs) {
+      // -Zs does not need -c
```
jrosdahl, line 1178: This seems to be true for `-fsyntax-only` as well, so I suggest merging the `-fsyntax-only` and `-Zs` cases as they behave identically.

jrosdahl (changes requested):

src/argprocessing.cpp
orgads: Good point. Done.

PR #999, opened 2022-01-16
doc: Document cl as a supported compiler_type

orgads:


jrosdahl:
This would document something that isn't true since [`compiler_type` currently cannot be set to `cl`](https://github.com/ccache/ccache/blob/3e4203920624fc8597e615ba3eca8a34e6906428/src/Config.cpp#L239-L254).

I'll fix it.

jrosdahl:
Obsoleted by 1672be839374.

PR #1000, opened 2022-01-23
feat: Enable HTTP keep-alive by default

jrosdahl:
The issue that made us not enable HTTP keep-alive by default (yhirose/cpp-httplib#1041) was resolved in cpp-httplib 0.10.0 (upgraded in e22e934dcaf1).

@gjasny: Does this sound OK to you?

### DIFF
--- a/doc/MANUAL.adoc
+++ b/doc/MANUAL.adoc
@@ -1105,13 +1105,7 @@ Optional attributes:
 
 * *connect-timeout*: Timeout (in ms) for network connection. The default is 100.
 * *keep-alive*: If *true*, keep the HTTP connection to the storage server open
-  to avoid reconnects. The default is *false*.
-+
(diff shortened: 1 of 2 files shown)


gjasny:
Testing on Linux and macOS ...

gjasny:
MacOS looks good. I wasted a long time to check why ccache resets the connection even before the first data was sent. Reason was the pretty short connection timeout. I'll work on improving the error message in that case.

But I have troubles accessing my Linux box right now. Testing will have to wait a few days.

gjasny:
Reported as yhirose/cpp-httplib#1171

gjasny:
Linux looks good as well.

gjasny (approved):

jrosdahl:
Thanks for testing!

PR #1001, opened 2022-01-27
feat: Add option to use Bearer token with HTTP backend

### LINKED ISSUE #997: Storage backend: Google Cloud Storage
Hi!

I'm trying to improve compilation time of builds running in CI pipeline. Currently, we are downloading and re-uploading the whole `ccache` directory between each build. This has two downsides:
* It's quite slow and sub-optimal as it download files not needed, and upload files unchanged.
* It causes possible cache erasure is two builds are run in parallel and finish and end one after the other.

I looked at the documentation and figured out that recent versions of CCache support remote secondary backend which is kind of mind blowing! :exploding_head: 

I would like to know if adding support for [Cloud Storage](https://cloud.google.com/storage/docs/introduction) was a possibility. It provides a [XML API](https://cloud.google.com/storage/docs/xml-api/reference-methods) using HTTP which implements the `GET`, `POST` and `DELETE` methods required by CCache storage backend. However, Cloud Storage requires Bearer Authentication instead of the Basic one. As far as I know, that's why it's not currently possible to use Cloud Storage as a CCache backend.

What do you think about detecting `gs://` scheme in `secondary_storage` or adding Bearer Authentication support to the already existing HTTP Backend?

If this is something you would accept in CCache, I can eventually try to open a PR. :+1:

Delgan:
This a proposal to add support for Bearer Authorization to the HTTP secondary storage backend.

I tested it locally and it worked perfectly: files were uploaded on GCloud during first build, and second build was much faster (after cleaning the cache).

Let me know if you prefer a different name than `"bearer-token"` or for any other improvement. Maybe we should redact the token in the logs (I don't know if it's logged) as it is done for the password?

Fixes #997.

### DIFF
--- a/doc/MANUAL.adoc
+++ b/doc/MANUAL.adoc
@@ -1131,6 +1131,7 @@ values.
 +
 The default is *subdirs*.
 * *operation-timeout*: Timeout (in ms) for HTTP requests. The default is 10000.
+* *bearer-token*: Bearer token used to authorize the HTTP requests.
 
 
 === Redis storage backend
--- a/doc/MANUAL.adoc
+++ b/doc/MANUAL.adoc
@@ -1103,6 +1103,7 @@ Examples:
 
 Optional attributes:
 
+* *bearer-token*: Bearer token used to authorize the HTTP requests.
 * *connect-timeout*: Timeout (in ms) for network connection. The default is 100.
 * *keep-alive*: If *true*, keep the HTTP connection to the storage server open
   to avoid reconnects. The default is *false*.
@@ -1131,7 +1132,6 @@ values.
 +
 The default is *subdirs*.
 * *operation-timeout*: Timeout (in ms) for HTTP requests. The default is 10000.
-* *bearer-token*: Bearer token used to authorize the HTTP requests.
 
 
 === Redis storage backend
--- a/src/storage/secondary/HttpStorage.cpp
+++ b/src/storage/secondary/HttpStorage.cpp
@@ -132,6 +132,8 @@ HttpStorageBackend::HttpStorageBackend(const Params& params)
       }
     } else if (attr.key == "operation-timeout") {
       operation_timeout = parse_timeout_attribute(attr.value);
+    } else if (attr.key == "bearer-token") {
+      m_http_client.set_bearer_token_auth(attr.value.c_str());
     } else if (!is_framework_attribute(attr.key)) {
       LOG("Unknown attribute: {}", attr.key);
     }
--- a/src/storage/secondary/HttpStorage.cpp
+++ b/src/storage/secondary/HttpStorage.cpp
@@ -116,7 +116,9 @@ HttpStorageBackend::HttpStorageBackend(const Params& params)
   auto operation_timeout = k_default_operation_timeout;
 
   for (const auto& attr : params.attributes) {
-    if (attr.key == "connect-timeout") {
+    if (attr.key == "bearer-token") {
+      m_http_client.set_bearer_token_auth(attr.value.c_str());
+    } else if (attr.key == "connect-timeout") {
       connect_timeout = parse_timeout_attribute(attr.value);
     } else if (attr.key == "keep-alive") {
       m_http_client.set_keep_alive(attr.value == "true");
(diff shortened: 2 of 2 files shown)


doc/MANUAL.adoc
```diff
@@ -1131,6 +1131,7 @@ values.
 +
 The default is *subdirs*.
 * *operation-timeout*: Timeout (in ms) for HTTP requests. The default is 10000.
+* *bearer-token*: Bearer token used to authorize the HTTP requests.
```
jrosdahl, line 1134: Please keep the attribute list sorted alphabetically.

src/storage/secondary/HttpStorage.cpp
```diff
@@ -132,6 +132,8 @@ HttpStorageBackend::HttpStorageBackend(const Params& params)
       }
     } else if (attr.key == "operation-timeout") {
       operation_timeout = parse_timeout_attribute(attr.value);
+    } else if (attr.key == "bearer-token") {
+      m_http_client.set_bearer_token_auth(attr.value.c_str());
```
jrosdahl, line 136: Similar here, I suggest checking `bearer-token` first.

jrosdahl:
> Let me know if you prefer a different name than "bearer-token"

Sounds good to me.

> Maybe we should redact the token in the logs (I don't know if it's logged) as it is done for the password?

Yes, that can be done by updating `params.attributes` in [`HttpStorage::redact_secrets`](https://github.com/ccache/ccache/blob/15fecbd7d540393e4dade71be78d7c10c69535cc/src/storage/secondary/HttpStorage.cpp#L277).

Delgan:
@jrosdahl Thanks for the review. I addressed your comments. I checked the logs and the token appears to be properly redacted. :+1:

jrosdahl (approved):

jrosdahl:
Thanks.

klusark:
@Delgan How do you actually use this with GCP? All the bucket endpoints seem to require https, which ccache doesn't support.

Delgan:
@klusark It works fine using `http` for me.

Here is the script I'm using on CI:

```bash
# Generate GCloud authentication token.
echo "$GCLOUD_API_KEYFILE" >~/.gcloud-api-key.json
gcloud auth activate-service-account --key-file ~/.gcloud-api-key.json
GCLOUD_BEARER_TOKEN=$(gcloud auth print-access-token)

# Create CCache folder.
mkdir -p ~/.ccache

# Configure ccache.
cat <<EOF >~/.ccache/ccache.conf
hash_dir = false
compiler_check = content
sloppiness = include_file_ctime,include_file_mtime
secondary_storage = http://ccache-remote-storage.storage.googleapis.com|bearer-token=${GCLOUD_BEARER_TOKEN}|share-hits=false
EOF
```

GMNGeoffrey:
I'm having the same problem as @klusark. I can't get GCS to respond to any http requests.

> [2022-11-23T16:33:40.303666 595755] Failed to put /20/87k38onedtcm5lt5gvkk1qlqdjn5rja to http storage: status code: 400
[2022-11-23T16:33:40.304393 595755] Added result key to manifest 4f42r6gnev3j5ii0se2lvglovsikvsj3q
[2022-11-23T16:33:40.304397 595755] Using default compression level 1
[2022-11-23T16:33:40.304541 595755] Not putting in http://iree-sccache.storage.googleapis.com since it failed earlier

(I tried http://storage.googleapis.com/iree-sccache as well)

Were you able to figure this out?

Notably, if I manually upload the file from the local storage with the right hash to GCS, it does successfully retrieve the file on a later run (though then complains "Failed to get result from 2087k38onedtcm5lt5gvkk1qlqdjn5rja: Bad magic value: 0x7f45"). So I think fetching over http works fine. It's writing over http that I can't get to work.

I was excited to learn that ccache now had support for remote storage. I'd really love to use it instead of sccache here (the other option I'm considering).

@Delgan did you set anything special on your bucket?

Is there a way to get ccache to show the response from the server with the 400 error?

Delgan:
@GMNGeoffrey The bucket isn't configured in any special way I'm aware of.

I'm using CCAche 4.6, maybe there is a regression in latest versions?

GMNGeoffrey:
I'll double-check, but I doubt that's it. I was initially using 4.6, but when it didn't work I tried upgrading to the latest before doing deep debugging.

GMNGeoffrey:
I wonder if it could be an organization setting. There are a lot of those in our cloud organization (this is a Google project)

GMNGeoffrey:
Yeah I confirmed I get the same error with ccache 4.6.3. Also built ccache from source and added extra logging. GCS returns `SecureConnectionRequired`. I'm betting this is some cloud organization security policy :-/ Let me try to check on that

GMNGeoffrey:
Ah (sorry for the multi-posting), I've been able to make a PUT from within a GCP VM. So it seems like there's something different about running locally as yourself vs running as a service account. I wasn't using an exported service account key because they're banned/by-exception-only. So I think that explains the difference. I'll look into what setting makes it so you can't use your own credentials without a secure connection. I guess because you're passing your own key and it could get sniffed

PR #1004, opened 2022-02-03
CMake: Added support for cmake upstream zstd package configuration

emmenlau:
I would like to use the `zstd` library with the name that is used upstream, not the cmake alias that `FindZSTD.cmake` creates. Sadly the names are slightly different. However its not very difficult to check if ZSTD was found via the upstream name or the CMake alias name, and use either or.

This change should be 100% backwards compatible. It only adds more broad support for finding ZSTD, but should not limit anything.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -98,6 +98,11 @@ endif()
 
 option(ZSTD_FROM_INTERNET "Download and use libzstd from the Internet" ${ZSTD_FROM_INTERNET_DEFAULT})
 find_package(zstd 1.1.2 REQUIRED)
+if(TARGET zstd::libzstd_shared)
+  set(ZSTD_LIBRARIES zstd::libzstd_shared)
+else()
+  set(ZSTD_LIBRARIES ZSTD::ZSTD)
+endif()
 
 option(REDIS_STORAGE_BACKEND "Enable Redis secondary storage" ON)
 if(REDIS_STORAGE_BACKEND)
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -64,7 +64,7 @@ set(THREADS_PREFER_PTHREAD_FLAG ON)
 find_package(Threads REQUIRED)
 target_link_libraries(
   ccache_framework
-  PRIVATE standard_settings standard_warnings ZSTD::ZSTD Threads::Threads third_party
+  PRIVATE standard_settings standard_warnings ${ZSTD_LIBRARIES} Threads::Threads third_party
 )
 
 target_include_directories(ccache_framework PRIVATE ${CMAKE_BINARY_DIR} ${CMAKE_CURRENT_SOURCE_DIR})


jrosdahl:
Thanks for the PR.

However, I don't understand the purpose of it. Why does the name used internally in ccache's CMake scripts matter?

emmenlau:
Dear @jrosdahl thanks for considering this! Sorry if the motivation was not too clear. So the problem is that I'd like to set the cmake option [CMAKE_FIND_PACKAGE_PREFER_CONFIG](https://cmake.org/cmake/help/latest/variable/CMAKE_FIND_PACKAGE_PREFER_CONFIG.html) in all my builds. Setting this to true makes sure that cmake will use universally the upstream package config, instead of some custom logic of the FindXXX-scripts. The main benefit can be that package configurations like compiler flags are more likely to be correctly transferred from upstream packages. Now, the problem comes: If `find_package(ZSTD)` is called with a preference on ZSTD's cmake config, the name of the target changes! The target will not be in variable `ZSTD_LIBRARIES`, and it will not be called `ZSTD::ZSTD`.

This is what the PR fixes, by looking for the alternative target name.

Ideally cmake should fix the naming to unify ZSTD target names. But until this may happen, this PR is a solution to the same issue.

jrosdahl:
Thanks for the explanation! I'm not sure I got less confused, though...

> Setting this to true makes sure that cmake will use (universally) the upstream package config,

What is upstream in this case?

> Now, the problem comes: If find_package(ZSTD) is called with a preference on ZSTD's cmake config, the name of the target changes! The target will not be in variable ZSTD_LIBRARIES, and it will not be called ZSTD::ZSTD.

Right, differing target names depending on MODULE/CONFIG mode (as I learned about today) sounds really bad.

> This is what the PR fixes, by looking for the alternative target name.

OK. However, isn't there a similar issue with `hiredis`?

> Ideally cmake should fix the naming to unify ZSTD target names.

Do you mean that CMake has special naming for libzstd specifically? Or do you refer to some more general issue with target names?

> But until this may happen, this PR is a solution to the same issue.

I don't like special cases, especially those I don't understand. :slightly_smiling_face: Is there an open CMake issue or effort to fix this? Or is it something that you just personally think should happen?

emmenlau:
> Thanks for the explanation! I'm not sure I got less confused, though...
> 
> > Setting this to true makes sure that cmake will use (universally) the upstream package config,
> 
> What is upstream in this case?

With "upstream" I mean what the official Zstandard library provides. Their cmake configuration provides a different target name than the `FindZSTD.cmake` script from cmake. So to be very precise, these two have calls to `add_library(<targetname>)` for different target names. Upstream uses target name `zstd::libzstd_shared`, and `FindZSTD.cmake` from cmake uses `ZSTD::ZSTD`.

> > This is what the PR fixes, by looking for the alternative target name.
> 
> OK. However, isn't there a similar issue with `hiredis`?

I haven't enabled redis, so maybe that is why I did not encounter problems. However, generally it is really seldom that cmake find scripts differ from the official package configuration from the upstream package. 

> > Ideally cmake should fix the naming to unify ZSTD target names.
> 
> Do you mean that CMake has special naming for libzstd specifically? Or do you refer to some more general issue with target names?

This seems to be very specific to `libzstd`.

> > But until this may happen, this PR is a solution to the same issue.
> 
> I don't like special cases, especially those I don't understand. slightly_smiling_face Is there an open CMake issue or effort to fix this? Or is it something that you just personally think should happen?

I did not report it upstream, and I just personally think should happen.

cristianadam:
Please note that there is no upstream CMake `FindZSTD.cmake` find module, as seen at https://cmake.org/cmake/help/latest/manual/cmake-modules.7.html#find-modules  

CMake doesn't accept any new find modules, the policy is to have projects provide either a find module or better a config module.

ccache has https://github.com/ccache/ccache/blob/master/cmake/Findzstd.cmake which can be changed to be in sync with the upstream zstd convention.

CMakeLists.txt
```diff
@@ -98,6 +98,11 @@ endif()
 
 option(ZSTD_FROM_INTERNET "Download and use libzstd from the Internet" ${ZSTD_FROM_INTERNET_DEFAULT})
 find_package(zstd 1.1.2 REQUIRED)
+if(TARGET zstd::libzstd_shared)
```
cristianadam, line 101: Please don't do this. Modify the `ZSTD::ZSTD` alias library in https://github.com/ccache/ccache/blob/master/cmake/Findzstd.cmake to point to the `zstd::libzstd_shared`.

There is no need to make CMake code uglier than it has to be.

cristianadam: `find_package(zstd 1.1.2 REQUIRED)` needs to change to `find_package(zstd 1.1.2 MODULE REQUIRED)`.

Then in https://github.com/ccache/ccache/blob/master/cmake/Findzstd.cmake you change:

```cmake
...
else()
  find_package(zstd CONFIG)
  if (zstd_FOUND and TARGET zstd::libzstd_shared)
    add_library(ZSTD::ZSTD ALIAS zstd::libzstd_shared)
  else()
    find_library(ZSTD_LIBRARY zstd)
    find_path(ZSTD_INCLUDE_DIR zstd.h)

    include(FindPackageHandleStandardArgs)
    find_package_handle_standard_args(
      zstd "please install libzstd or use -DZSTD_FROM_INTERNET=ON"
      ZSTD_INCLUDE_DIR ZSTD_LIBRARY)
    mark_as_advanced(ZSTD_INCLUDE_DIR ZSTD_LIBRARY)

    add_library(ZSTD::ZSTD UNKNOWN IMPORTED)
    set_target_properties(
      ZSTD::ZSTD
      PROPERTIES
      IMPORTED_LOCATION "${ZSTD_LIBRARY}"
      INTERFACE_INCLUDE_DIRECTORIES "${ZSTD_INCLUDE_DIR}")
  endif()
endif()
```

emmenlau:
> Please note that there is no upstream CMake `FindZSTD.cmake` find module, as seen at https://cmake.org/cmake/help/latest/manual/cmake-modules.7.html#find-modules

Ah, I see! I was happily confused because the cmake sources ship `Utilities/cmcurl/CMake/FindZstd.cmake`.

> Please don't do this. Modify the ZSTD::ZSTD alias library in https://github.com/ccache/ccache/blob/master/cmake/Findzstd.cmake to point to the zstd::libzstd_shared.

Under these circumstances I fully agree that this should be rather fixed in ccache. But then it seems the "issue" is fully self-contained in ccache, correct? May I suggest then just to rename ccache's `ZSTD::ZSTD` to match the upstream target name (whatever target name the Zstandard cmake configuration provides), something like `zstd::libzstd_shared`? Its probably just a mistake that the names do not match.

jrosdahl:
> May I suggest then just to rename ccache's `ZSTD::ZSTD` to match the upstream target name (whatever target name the Zstandard cmake configuration provides), something like `zstd::libzstd_shared`?

That would be OK with me, except maybe that `zstd::libzstd_shared` actually won't be a shared library if [using `-DZSTD_LIBRARY` to specify a static library](https://github.com/ccache/ccache/blob/05ac8dc345e758a3d31859e01b26a0104c0707e6/doc/INSTALL.md#L24). But I'm not very knowledgeable in what (and what not) to do in CMake, so I suggest following @cristianadam's advise in general.

> Its probably just a mistake that the names do not match.

I don't think there was any intention of using the same name as in Zstandard's own CMake rules (since they aren't used), so I wouldn't call it a mistake.

emmenlau:
> > Its probably just a mistake that the names do not match.
> 
> I don't think there was any intention of using the same name as in Zstandard's own CMake rules (since they aren't used), so I wouldn't call it a mistake.

Mistake is certainly not a good formulation then and I humbly take that back. What I meant to say is that it can be beneficial when the name matches the one from upstream (as in Zstandard's own CMake rules), because i.e. the cmake config can be used without any glue code.

cristianadam:
Actually all that's needed to support the `-DCMAKE_FIND_PACKAGE_PREFER_CONFIG=ON`  is this:

```diff
diff --git a/CMakeLists.txt b/CMakeLists.txt
index fcc2134b..e3d3e1c8 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -97,7 +97,7 @@ if(MSVC AND NOT CMAKE_TOOLCHAIN_FILE MATCHES "vcpkg|conan")
 endif()

 option(ZSTD_FROM_INTERNET "Download and use libzstd from the Internet" ${ZSTD_FROM_INTERNET_DEFAULT})
-find_package(zstd 1.1.2 REQUIRED)
+find_package(zstd 1.1.2 MODULE REQUIRED)

 option(REDIS_STORAGE_BACKEND "Enable Redis secondary storage" ON)
 if(REDIS_STORAGE_BACKEND)
```

Because ccache's `Findzlib.cmake` looks after a library called `zstd` and a header `zstd.h` and creates its `ZSTD::ZSTD` imported target.

I tested this with ccache configured with:

```
   -DZSTD_FROM_INTERNET=OFF
   -DCMAKE_PREFIX_PATH=/path/to/my/zstd-install
``` 

I think that's the main purpose of this PR, not being able to build ccache using `DCMAKE_FIND_PACKAGE_PREFER_CONFIG=ON` and not weather ccache is using the same CMake nomenclature as upstream zstd package.

cristianadam:
I've opened up https://github.com/ccache/ccache/pull/1030 with the one liner change.

emmenlau:
> Actually all that's needed to support the `-DCMAKE_FIND_PACKAGE_PREFER_CONFIG=ON` is this:
> 
> ```diff
> diff --git a/CMakeLists.txt b/CMakeLists.txt
> index fcc2134b..e3d3e1c8 100644
> --- a/CMakeLists.txt
> +++ b/CMakeLists.txt
> @@ -97,7 +97,7 @@ if(MSVC AND NOT CMAKE_TOOLCHAIN_FILE MATCHES "vcpkg|conan")
>  endif()
> 
>  option(ZSTD_FROM_INTERNET "Download and use libzstd from the Internet" ${ZSTD_FROM_INTERNET_DEFAULT})
> -find_package(zstd 1.1.2 REQUIRED)
> +find_package(zstd 1.1.2 MODULE REQUIRED)
> 
>  option(REDIS_STORAGE_BACKEND "Enable Redis secondary storage" ON)
>  if(REDIS_STORAGE_BACKEND)
> ```
[...]
> I think that's the main purpose of this PR, not being able to build ccache using `DCMAKE_FIND_PACKAGE_PREFER_CONFIG=ON` and not whether ccache is using the same CMake nomenclature as upstream zstd package.

Well, the reason to prefer `CMAKE_FIND_PACKAGE_PREFER_CONFIG=ON` in the first place was that the cmake package configuration promises advantages over the find-scripts. For example, cmake package configuration keeps transitive linker dependencies, includes, and compiler defines. My reasoning was that it's beneficial to adopt this more powerful standard wherever possible. Even if a specific package does not currently benefit from it (because Zstandard may currently not have dependencies or defines), it may still benefit from it in the future.

But thats really just my two cents, and I really just wanted to contribute to ccache what from my angle looked like a (however tiny) improvement. Feel free to go with either change, or close the PR alltogether, I can work around the actual issue in my side too.

jrosdahl:
Thanks. I think I understand the situation now and I went with #1030. The main reason for that is that I would prefer not to depend on other modules' behavior and naming.

PR #1005, opened 2022-02-05
fix: Support UTF-16LE .rsp files

vvainola:
Visual Studio writes its .rsp files with UTF-16 encoding
which need to be converted to UTF-8 so that ccache
can be used within Visual Studio

### DIFF
--- a/src/Util.cpp
+++ b/src/Util.cpp
@@ -48,7 +48,9 @@ extern "C" {
 
 #include <algorithm>
 #include <climits>
+#include <codecvt>
 #include <fstream>
+#include <locale>
 
 #ifndef HAVE_DIRENT_H
 #  include <filesystem>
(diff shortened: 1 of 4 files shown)


src/Util.cpp
```diff
@@ -1075,6 +1085,14 @@ read_file(const std::string& path, size_t size_hint)
...
 
   result.resize(pos);
+  // Convert to UTF-8 if the contents start with UTF-16 little-endian BOM
+  if (has_utf16_le_bom(result)) {
+    result.erase(0, 2); // Remove BOM
+    std::u16string result_as_u16((result.size() / 2) + 1, '\0');
+    result_as_u16 = reinterpret_cast<const char16_t*>(result.c_str());
+    std::wstring_convert<std::codecvt_utf8_utf16<char16_t>, char16_t> converter;
+    result = converter.to_bytes(result_as_u16);
+  }
```
jrosdahl, line 1095: Sounds like a good idea to support reading UTF-16LE-encoded text files when parsing rsp files. However, I don't think that `Util::read_file` is the right place to do that – it's a function for reading the content of a file, not necessarily a text file. Put differently, there are other contexts where letting `Util::read_file` unconditionally convert to UTF-8 is superfluous at best.

I suggest introducing a `Util::read_text_file` function that calls `Util::read_file` to get the content and then converts if there an UTF-16 BOM. Then use `Util::read_text_file` in places where it makes sense, e.g. in `Args::from_atfile`.

jrosdahl:
Thanks.

PR #1006, opened 2022-02-12
Add documentation that compiler plugins are hashed too

mephinet:
I've been bisecting the Linux kernel for a few days now, and had a hard time figuring out why ccache had a hit rate around 3% even though I've been recompiling the same code over and over again - until I learned about the existence of compiler plugins today. As the kernel's GCC plugins are rebuilt after every `make clean`, this kept triggering the compilercheck on every build, even though the compiler itself never changed (of course). Therefore adding a hint at the compilercheck documentation for future-me that plugins are considered in this check too.


jrosdahl:
Thanks.

PR #1011, opened 2022-02-25
fix: Make conversion to relative paths more reliable on Windows

emzeat:
When testing `ccache` in my projects using MSVC via cmake and ninja I found that the relative path handling triggered via the "base_dir" config option would not work for all compiler invokations generated by CMake.

This is especially critical as some CIs like e.g. drone use unique build directories and as such caching depends on paths to be always relative.

This PR addresses the two issues I found in the code which prevented base paths to be properly generated in my setup.

### DIFF
--- a/src/Util.cpp
+++ b/src/Util.cpp
@@ -233,6 +233,31 @@ base_name(string_view path)
   return n == std::string::npos ? path : path.substr(n + 1);
 }
 
+bool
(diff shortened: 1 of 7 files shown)


emzeat:
Actually I found another issue remaining with #includes also not translated to relative paths. I will look into this so downgrading to WIP again.

emzeat:
Sorry for the back and forth. Good to go now.

jrosdahl:
Thanks.

PR #1015, opened 2022-03-02
Handle end of args compiler option

jacobly0:
Adds handling for clang group compiler option `--`.

### DIFF
--- a/src/argprocessing.cpp
+++ b/src/argprocessing.cpp
@@ -56,6 +56,7 @@ struct ArgumentProcessingState
   ColorDiagnostics color_diagnostics = ColorDiagnostics::automatic;
   bool found_directives_only = false;
   bool found_rewrite_includes = false;
+  bool found_end_of_options = false;
   nonstd::optional<std::string> found_xarch_arch;
 
   std::string explicit_language;    // As specified with -x.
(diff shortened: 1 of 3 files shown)


jrosdahl:
I would appreciate it if you could split PR into one PR per fix.

jacobly0:
> I would appreciate it if you could split PR into one PR per fix.

Split into 4 separate PRs.

jrosdahl:
> Adds handling for clang group compiler option `--`.

@jacobly0: Could you describe what this PR aims to fix? I can't find any documentation of what `--` does when passed to Clang.

src/ccache.cpp
```diff
@@ -965,6 +965,12 @@ to_cache(Context& ctx,
   Util::unsetenv("DEPENDENCIES_OUTPUT");
   Util::unsetenv("SUNPRO_DEPENDENCIES");
 
+#ifndef __APPLE__
```
jrosdahl, line 968: A difference in compiler behavior shouldn't be solved by looking at the platform at compile time. What if an older Clang version is used on Linux or a newer Clang version is used on macOS?

src/ccache.cpp
```diff
@@ -965,6 +965,12 @@ to_cache(Context& ctx,
...
+  // Older Clang versions on macOS didn't support this.
+  if (ctx.config.is_compiler_group_clang()) {
+    args.push_back("--");
```
jrosdahl, line 971: Why always include `--`? Would it be possible to only include it when it's on the original command line?

src/argprocessing.cpp
```diff
@@ -262,743 +263,755 @@ process_arg(const Context& ctx,
 {
   size_t& i = args_index;
 
-  // The user knows best: just swallow the next arg.
-  if (args[i] == "--ccache-skip") {
-    i++;
-    if (i == args.size()) {
-      LOG_RAW("--ccache-skip lacks an argument");
-      return Statistic::bad_compiler_arguments;
+  if (!state.found_end_of_options) {
```
jrosdahl, line 266: I would prefer not adding this extra indention level. I'll refactor the code to make it easier to do the change you propose.

jrosdahl: Done in 62f3217ee9f2b3b147ad9b31fc317b10e9dfc225.

jacobly0:
I added this because some build/compiler/os (probably cmake/clang-cl/windows?) combination was causing ccache to copy a `--` input argument to the wrong place in the final argument list, however after taking another look it seems likely that I assumed completely the wrong meaning for this argument, which would mean that this PR needs to be rewritten.

PR #1019, opened 2022-03-05
msvc: support directory input for -Fo

### LINKED ISSUE #1018: Support for MSVC with Visual Studio project files
With the new announcement that ccache 4.6 now supports windows I tried it out using Visual Studio project files. There are multiple ways to utilize ccache in that scenario. Here are described two ways: https://stackoverflow.com/questions/30761708/how-to-use-clang-compiler-with-msbuild/ and there is a third way by replacing the cl.exe (e.g. rename to cl-real.exe) by ccache.exe and then set CCACHE_COMPILER cl-real.exe and set CCACHE_COMPILERTYPE to msvc.

When using that I ran into multiple issues - the biggest one being that multiple files are passed to cl.exe 

> [2022-03-01T17:43:46.530839 236620] Multiple input files: <nameA>.cpp and <nameB>.cpp
> 
> [2022-03-01T17:43:46.530839 236620] Failed; falling back to running the real compiler
> 
> [2022-03-01T17:43:46.530839 236620] Executing C:\Program Files (x86)\Microsoft Visual Studio\2019\Professional\VC\Tools\MSVC\14.29.30037\bin\HostX64\x64\cl_real.exe C:\Users\myname\AppData\Local\Temp\tmp8b9e1682df6f4cc7be0e622d8809e8e4.rsp
> 
> [2022-03-01T17:43:46.530839 236620] Result: multiple_source_files

But even when compiling a single cpp file within Visual Studio it fails:

> 
> [2022-03-01T17:49:50.170961 448640] Working directory: Z:/_CM/.cache/build/D/3dmodel
> 
> [2022-03-01T17:49:50.170961 448640] Compiler type: msvc
> 
> [2022-03-01T17:49:50.170961 448640] Not a regular file: 3dmodel_Object.dir\Debug\\
> 
> [2022-03-01T17:49:50.171968 448640] Failed; falling back to running the real compiler
> 
> [2022-03-01T17:49:50.171968 448640] Executing C:\Program Files (x86)\Microsoft Visual Studio\2019\Professional\VC\Tools\MSVC\14.29.30037\bin\HostX64\x64\cl_real.exe C:\Users\kmahdi\AppData\Local\Temp\tmp411f254b56f34aaea604542e60e9d4ef.rsp
> 
> [2022-03-01T17:49:50.171968 448640] Result: bad_output_file

Seems like somehow the file path to the files is slightly messed up with those double backslashes `Not a regular file: 3dmodel_Object.dir\Debug\\`.

I just wanted to point out those issues for anyone wondering who the current support for Visual Studio Project files is and maybe there is a possibility to fix them.

orgads:
It is used by msbuild.

Fixes #1018

### DIFF
--- a/src/argprocessing.cpp
+++ b/src/argprocessing.cpp
@@ -1116,8 +1116,20 @@ process_args(Context& ctx)
   }
 
   // Determine output object file.
-  const bool implicit_output_obj = args_info.output_obj.empty();
-  if (implicit_output_obj && !args_info.input_file.empty()) {
+  bool output_obj_by_source = args_info.output_obj.empty();
+  if (!output_obj_by_source && ctx.config.is_compiler_group_msvc()) {
+    if (*args_info.output_obj.rbegin() == '\\') {
+      output_obj_by_source = true;
+    } else {
+      auto st = Stat::stat(args_info.output_obj);
+      if (st && st.is_directory()) {
+        args_info.output_obj.append("\\");
+        output_obj_by_source = true;
+      }
+    }
+  }
+
+  if (output_obj_by_source && !args_info.input_file.empty()) {
     string_view extension;
     if (state.found_S_opt) {
       extension = ".s";
@@ -1126,8 +1138,9 @@ process_args(Context& ctx)
     } else {
       extension = ".obj";
     }
-    args_info.output_obj =
-      Util::change_extension(Util::base_name(args_info.input_file), extension);
+    args_info.output_obj = args_info.output_obj
+                           + Util::change_extension(
+                             Util::base_name(args_info.input_file), extension);
   }
 
   // On argument processing error, return now since we have determined
@@ -1193,7 +1206,7 @@ process_args(Context& ctx)
     args_info.actual_language.find("-header") != std::string::npos
     || Util::is_precompiled_header(args_info.output_obj);
 
-  if (args_info.output_is_precompiled_header && implicit_output_obj) {
+  if (args_info.output_is_precompiled_header && output_obj_by_source) {
     args_info.output_obj = args_info.input_file + ".gch";
   }
 


src/argprocessing.cpp
```diff
@@ -1116,7 +1116,19 @@ process_args(Context& ctx)
   }
 
   // Determine output object file.
-  const bool implicit_output_obj = args_info.output_obj.empty();
+  bool implicit_output_obj = args_info.output_obj.empty();
+  if (!implicit_output_obj && ctx.config.is_compiler_group_msvc()) {
+    if (*args_info.output_obj.rbegin() != '\\') {
+      implicit_output_obj = true;
```
jrosdahl, line 1122: I don't understand this logic. If `args_info.output_obj` is not empty, then by definition it's not an implicit output object since an output object filename was passed explicitly.

Also, doesn't this break `/Fo` with a filename? If the compiler is MSVC and `example.c` and `/Foexample.obj` is on the command line, then `!implicit_output_obj && ctx.config.is_compiler_group_msvc()` will be true. Since `*args_info.output_obj.rbegin() != '\\'` will be true, `implicit_output_obj` will be set to true. On line 1132, `implicit_output_obj && !args_info.input_file.empty()` will be true, so `args_info.output_obj` will be set to `example.objexample.obj`. Am I missing something?

orgads: Oops, this should have been == of course!

I reused `implicit_output_obj` to avoid duplication of the extension translation logic. Do you prefer a separate variable for clearance? It will be equivalent to `implicit_output_obj` though.

jrosdahl: I would prefer a  a more descriptive name of `implicit_output_obj`, or separate variable if that makes the code more readable.

orgads: Done

jrosdahl:
Thanks.

PR #1030, opened 2022-03-19
fix: Prefer CMake find module for zstd package

cristianadam:
ccache is providing a `Findzstd.cmake CMake` find module, which is
ignored when `CMAKE_FIND_PACKAGE_PREFER_CONFIG` is set to `ON`.


jrosdahl:
Thanks.

PR #1033, opened 2022-03-23
fix: normalize abs paths when collecting includes to fix basedir option

vvainola:
On windows with basedir to set e.g. C:\Projects, it will
get normalized to C:/Projects. With a header in C:\Projects\repo\some_header.h,
during include reading it may be read as C:\\\\\\\\Projects\\\\\\\\\repo\\\\\\\\some_header.h
with additional slashes. It does not then match basedir and is not changed to
relative when it should be.

Also the header paths printed to stdout need to be changed to relative so that Ninja will
use the relative paths to headers to detect changes instead of absolute paths. Otherwise
Ninja will check for header changes from the original build dir if an object was restored
from cache to another build dir.

### DIFF
--- a/src/ArgsInfo.hpp
+++ b/src/ArgsInfo.hpp
@@ -26,8 +26,11 @@
 // This class holds meta-information derived from the compiler arguments.
 struct ArgsInfo
 {
-  // The source file.
+  // The source file path.
   std::string input_file;
+
+  // The source file path run through Util::normalize_concrete_absolute_path.
+  std::string normalized_input_file;
 
   // In normal compiler operation an output file is created if there is no
   // compiler error. However certain flags like -fsyntax-only change this
--- a/src/argprocessing.cpp
+++ b/src/argprocessing.cpp
@@ -1016,6 +1016,8 @@ process_arg(const Context& ctx,
 
   // Rewrite to relative to increase hit rate.
   args_info.input_file = Util::make_relative_path(ctx, args[i]);
+  args_info.normalized_input_file =
+    Util::normalize_concrete_absolute_path(args_info.input_file);
 
   return nullopt;
 }
--- a/src/ccache.cpp
+++ b/src/ccache.cpp
@@ -553,23 +553,10 @@ process_preprocessed_file(Context& ctx, Hash& hash, const std::string& path)
       if (!ctx.has_absolute_include_headers) {
         ctx.has_absolute_include_headers = util::is_absolute_path(inc_path);
       }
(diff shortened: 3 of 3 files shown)


jrosdahl:
Looks good to me.

@vvainola: I rebased your branch to latest master since there was a merge conflict. This made a newly written profiling test fail due to the change related to `# 1 "CWD//"`. I'll look into that.

jacobly0:
> This made a newly written profiling test fail due to the change related to `# 1 "CWD//"`. I'll look into that.

Should be fixed by #1056.

jrosdahl:
> Should be fixed by #1056.

Thanks for the suggestion, but that only updates the tests to hide the problem instead of fixing it, I'm afraid.

The real problem that the new profiling test caught (by luck) is that https://github.com/ccache/ccache/blob/35fda14aba3fce18a9cd059241537245ba383b84/src/ccache.cpp#L556 normalizes _all_ paths path found in the preprocessed input so that https://github.com/ccache/ccache/blob/35fda14aba3fce18a9cd059241537245ba383b84/src/ccache.cpp#L287 now potentially fails to match. This leads to the _input file_ being treated as a _header file_, thereby getting remembered in the manifest and getting subject to the "Include file ... too new" check, which it shouldn't.

I have also realized that there is another problem with the current PR: `Util::normalize_absolute_path` is [not guaranteed to generate a path to a file that actually exists](https://github.com/ccache/ccache/blob/35fda14aba3fce18a9cd059241537245ba383b84/src/Util.hpp#L277-L279) (this can happen in edge cases with symlinks in the path in combination with `..` segments), so we can't just call `Util::normalize_absolute_path` like that in `process_preprocessed_file`.

jrosdahl:
I wrote:

> This leads to the _input file_ being treated as a _header file_, thereby getting remembered in the manifest and getting subject to the "Include file ... too new" check, which it shouldn't.

This should be fixed by bd6c64b8d9d0c53811459c882c1ad6b9e81e719d.

> I have also realized that there is another problem with the current PR: `Util::normalize_absolute_path` is [not guaranteed to generate a path to a file that actually exists](https://github.com/ccache/ccache/blob/35fda14aba3fce18a9cd059241537245ba383b84/src/Util.hpp#L277-L279) (this can happen in edge cases with symlinks in the path in combination with `..` segments), so we can't just call `Util::normalize_absolute_path` like that in `process_preprocessed_file`.

This should be fixed by 9a05332915d2808f4713437006b3f7c812d9fd74 plus usage of `Util::normalize_concrete_absolute_path` in 31a5d0bcfdd59aaf678f19de0f1fbd986029546c

PR #1037, opened 2022-03-25
fix: Do not add redundant newlines for stdout

orgads:
Tokenizer is used with `include_empty`, and the output typically ends with `\n`.

For each line, tokenizer returns it without the `\n`, and the function appends it. But if the output ends with `\n`, the tokenizer returns an additional empty string, and a redundant `LF` is written to stdout.

Another issue can be if the last line of the original output doesn't end with `\n` at all, ccache added it anyway.

It was probably unnoticed until now since gcc has no output at all, while cl outputs the source file name.

### DIFF
--- a/src/ccache.cpp
+++ b/src/ccache.cpp
@@ -906,6 +906,9 @@ write_result(Context& ctx,
 static std::string
 rewrite_stdout_from_compiler(const Context& ctx, std::string&& stdout_data)
 {
+  using util::Tokenizer;
+  using Mode = Tokenizer::Mode;
+  using IncludeDelimiter = Tokenizer::IncludeDelimiter;
   // distcc-pump outputs lines like this:
   //
   //   __________Using # distcc servers in pump mode
@@ -913,13 +916,12 @@ rewrite_stdout_from_compiler(const Context& ctx, std::string&& stdout_data)
   // We don't want to cache those.
   if (!stdout_data.empty()) {
     std::string new_stdout_text;
-    for (const auto line : util::Tokenizer(
-           stdout_data, "\n", util::Tokenizer::Mode::include_empty)) {
+    for (const auto line : Tokenizer(
+           stdout_data, "\n", Mode::include_empty, IncludeDelimiter::yes)) {
       if (util::starts_with(line, "__________")) {
         Util::send_to_fd(ctx, std::string(line), STDOUT_FILENO);
       } else {
         new_stdout_text.append(line.data(), line.length());
-        new_stdout_text.append("\n");
       }
     }
     return new_stdout_text;
--- a/src/util/Tokenizer.hpp
+++ b/src/util/Tokenizer.hpp
@@ -37,11 +37,14 @@ public:
     skip_last_empty, // Include empty tokens except the last one.
   };
 
+  enum class IncludeDelimiter { no, yes };
+
   // Split `string` into tokens at any of the characters in `separators` which
   // must neither be the empty string nor a nullptr.
   Tokenizer(nonstd::string_view string,
             const char* delimiters,
-            Mode mode = Mode::skip_empty);
+            Mode mode = Mode::skip_empty,
+            IncludeDelimiter include_delimiter = IncludeDelimiter::no);
 
   class Iterator
   {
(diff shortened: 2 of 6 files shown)


src/ccache.cpp
```diff
@@ -912,6 +912,11 @@ rewrite_stdout_from_compiler(const Context& ctx, std::string&& stdout_data)
   //
   // We don't want to cache those.
   if (!stdout_data.empty()) {
+    // Tokenizer is used with include_empty, so strip last \n to avoid a
```
jrosdahl, line 915: How about changing tokenizer mode to `skip_last_empty` instead?

orgads: This could work, but there is still an issue when the output (or at least its last line) doesn't contain a newline at all. See the updated commit.

jrosdahl: Sure, handling that case as well would be good, even though it likely doesn't occur in practice.

src/util/Tokenizer.hpp
```diff
@@ -32,9 +32,10 @@ class Tokenizer
 {
 public:
   enum class Mode {
-    include_empty,   // Include empty tokens.
-    skip_empty,      // Skip empty tokens.
-    skip_last_empty, // Include empty tokens except the last one.
+    include_empty,     // Include empty tokens.
+    skip_empty,        // Skip empty tokens.
+    skip_last_empty,   // Include empty tokens except the last one.
+    include_delimiter, // Include the delimiter
```
jrosdahl, line 38: I like the idea of being able not to strip the delimiter, but I'm not so sure about adding it to the `Mode` list. `Mode` says which tokens to include, while `include_delimiter` says how to process each token, so it's an odd bird. In theory one might want to retain delimiters but also skip empty tokens.

How about adding a fourth `bool strip_delimiter = true` or `bool include_delimiter = false` parameter to `Tokenizer::Tokenizer` instead? (Or even better: an enum class with two values.) It would be OK to `ASSERT(strip_delimiter || mode != include_empty);` for now if that simplifies the implementation (should be essentially the same as what you have now).

src/ccache.cpp
orgads: Done.

src/util/Tokenizer.hpp
orgads: Ok

orgads:
I created 3 separate commits. 1 is test cleanup, 1 for adding the new flag for tokenizer, and the last one is using it for stdout.

There's another potential use - in `rewrite_stderr_to_absolute_paths`. And maybe also in `rewrite_paths`.

unittest/test_util_Tokenizer.cpp
```diff
@@ -142,4 +146,121 @@ TEST_CASE("util::Tokenizer")
...
+  {
+    const auto split = split_func(Mode::include_empty, IncludeDelimiter::yes);
+    {
+      const auto s = split("", "/");
+      REQUIRE(s.size() == 1);
+      CHECK(s[0] == "");
+    }
+    {
+      const auto s = split("/", "/");
+      REQUIRE(s.size() == 1);
```
jrosdahl, line 160: This feels surprising to me. I thought that `IncludeDelimiter::yes` should just affect each token, not how tokens are split. Put differently: For every pair of input and delimiters strings, I would expect the number of tokens to be equal with `IncludeDelimiter::yes` and `IncludeDelimiter::no` in all split modes. Does this make sense to you?

unittest/test_util_Tokenizer.cpp
```diff
@@ -23,40 +23,51 @@
...
+    };
+  };
 
   SUBCASE("include empty tokens")
   {
+    const auto split = split_func(Mode::include_empty);
     {
-      const auto s = Util::split_into_views("", "/", Mode::include_empty);
+      const auto s = split("", "/");
       REQUIRE(s.size() == 1);
```
orgads, line 42: It's the same as here

unittest/test_util_Tokenizer.cpp
orgads: But it is! Did I miss anything?

jrosdahl: With `Mode::include_empty` and `IncludeDelimiter::no`, splitting `"/"` with delimiters `"/"` returns 2 tokens:

https://github.com/ccache/ccache/blob/5486a6809ae39d15e8e5ec88dce977778358ba14/unittest/test_util_Tokenizer.cpp#L46-L47

With `Mode::include_empty` and `IncludeDelimiter::yes`, splitting `"/"` with delimiters `"/"` returns 1 token:

https://github.com/ccache/ccache/blob/5486a6809ae39d15e8e5ec88dce977778358ba14/unittest/test_util_Tokenizer.cpp#L159-L160

unittest/test_util_Tokenizer.cpp
jrosdahl: Right, but that's not the same input string.

unittest/test_util_Tokenizer.cpp
orgads: Ok, you're right of course. I'll try to debug it.

orgads: Done

jrosdahl:
Nice, thanks.

PR #1041, opened 2022-03-29
fix: Fix static linkage with hiredis on Windows

orgads:


orgads:
ping

jrosdahl:
Thanks.

> ping

I do all ccache work in my spare time, which is quite irregular. I also normally process most uncommented issues/PRs in oldest-first order in my mailbox, so the only thing a ping will lead to is more delay since that will make the conversation newer.

orgads:
I didn't mean to pressure, just thought you might have missed it.

I'll try to keep in mind not to ping again.

Thanks for all your work!

PR #1046, opened 2022-04-06
fix: Replace skip_last_empty with IncludeDelimiter

orgads:
It is more accurate, in case the original output doesn't end with \n.


jrosdahl:
Thanks.

PR #1054, opened 2022-04-14
build: Fix build arguments to clang-cl

jacobly0:
Avoids passing `-include` to `clang-cl` which is not a valid argument.


jrosdahl:
Thanks.

PR #1055, opened 2022-04-14
Handle -imsvc compiler option

jacobly0:
Treats the `clang-cl`-only option `-imsvc` the same way as `-isystem`.


jrosdahl:
Thanks.

PR #1056, opened 2022-04-14
Fix sporatic test.profiling failures

jacobly0:
Fixes two unrelated causes of sporatic failures in profiling tests.
1. Using `cut`, which is always line-based, to remove bytes from a binary file can cause problems if the binary timestamp field happens to contain a newline character, which causes `test.profiling` to fail <1% of the time.  Instead use `tail` which has a character mode that ignores newlines.
~~2. Sometimes ccache.log contains `test.c too new` so copy the format of `SETUP` in `direct.bash` to `profiling*.bash`.~~

### DIFF
--- a/test/suites/profiling.bash
+++ b/test/suites/profiling.bash
@@ -2,7 +2,7 @@
 normalize_gcno_file() {
     local from="$1"
     local to="$2"
-    cut -b 13- "${from}" >"${to}"
+    tail -c +13 "${from}" >"${to}"
 }
 
 


test/suites/profiling.bash
```diff
@@ -20,8 +20,12 @@ SUITE_profiling_PROBE() {
 }
 
 SUITE_profiling_SETUP() {
-    echo 'int main(void) { return 0; }' >test.c
     unset CCACHE_NODIRECT
+
+    cat <<EOF >test.c
+int main(void) { return 0; }
+EOF
+    backdate test.c
```
jrosdahl, line 28: As mentioned in https://github.com/ccache/ccache/pull/1033#issuecomment-1114030692, this hides a real problem instead of fixing it.

jrosdahl (changes requested):

test/suites/profiling.bash
jrosdahl: I'll rebase your branch to remove abad5b0c46b6e31e53d1fef3dcbb7d105bb45a6b since I want to apply f4f6e470ed0e396017b3eba90c9a9fa21d0529cf.

jrosdahl (approved):

jrosdahl:
> 2. Sometimes ccache.log contains test.c too new so copy the format of SETUP in direct.bash to profiling*.bash.

I removed this commit in order not to delay the PR any longer, see https://github.com/ccache/ccache/pull/1056#discussion_r862476676.

jrosdahl:
Thanks.

PR #1059, opened 2022-04-21
Fix "Multiple precompiled headers used" error if /Yu option is used after /Fp

telishev:
When using ccache on Windows with MSVC with precompiled headers sometimes the following error occures: "Multiple precompiled headers used: precompiled_header.pch and precompiled_header.pch".  This happens if you specify /Fp before /Yu. I fixed it in the same manner as /Fp handles being after /Yu but not sure if it's the best approach.


jrosdahl:
LGTM. @orgads: Would you like to review this PR?

PR #1063, opened 2022-04-25
fix: Process /Fp and /Yu flags before unknown args

lo1ol:
Passing /Fp or /Yu flags without space between arg leads to situation
when this flags not processed by appropriate handler and
included_pch_file is not set. To avoid this we have to process this
flags before processing unknown args.

### DIFF
--- a/src/argprocessing.cpp
+++ b/src/argprocessing.cpp
@@ -917,6 +917,19 @@ process_arg(const Context& ctx,
     return nullopt;
   }
 
+  // Detect PCH for options with concatenated path.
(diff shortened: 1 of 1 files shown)


jrosdahl:
Thanks for the bug report! Looks like I messed up the case with -Fp/-Yu and concatenated path in c8fb539523c801bf15591c395c3029b4530a7a2f.

The fix you propose seems partially correct, but it doesn't rewrite the path argument to relative like it should (and like the code before c8fb539523c801bf15591c395c3029b4530a7a2f did). I made a more complete fix in 09dea223466e3bd27c8b84f1cc4acbfa1509e061. Please try it out.

lo1ol:
Your solution is works, thanks!

PR #1064, opened 2022-04-27
Support Redis over unix socket

afbjorklund:
Similar to the regular tcp redis url, but uses a unix socket file.

This was some code that I did earlier, when we were discussing Unix socket support for redis (originally memcached)

We didn't find the syntax, since there was a conflict between database and location, but it was a different _scheme_.

Useful if network access is limited, or for using a local proxy.


doc/MANUAL.adoc
```diff
@@ -1142,6 +1142,7 @@ The default is *subdirs*.
 === Redis storage backend
 
 URL format: `+redis://[[USERNAME:]PASSWORD@]HOST[:PORT][/DBNUMBER]+`
+Alternative: `+redis+unix://SOCKET_PATH[?db=DBNUMBER[&password=PASSWORD]]+`
```
jrosdahl, line 1145: I suggest letting db and password be attributes instead of URL parameters (e.g. `redis+unix:///example/path|db=0|password=example`). That way the attributes can apply to both redis and redis+unix (overriding URL parameters), and it's easier to redact the password since the framework already supports that. Does that sound OK to you?

doc/MANUAL.adoc
```diff
@@ -1160,6 +1161,8 @@ Examples:
 
 * `+redis://localhost+`
 * `+redis://p4ssw0rd@cache.example.com:6379/0|connect-timeout=50+`
+* `+redis+unix:///run/redis.sock+`
```
jrosdahl, line 1164: It would be nice to support `redis+unix:/run/redis.sock` as well, similar to the `file` schema.

doc/MANUAL.adoc
afbjorklund: I think the syntax was borrowed from perl: https://metacpan.org/pod/URI::redis

But they _could_ be attributes ("secret") instead of part of the URL (in user_info)

That would probably apply to all forms of URL, though ?

i.e. also to the regular (non-unix) http and redis, right ?

doc/MANUAL.adoc
afbjorklund: Sounds reasonable, I think it is "no hostname" vs "empty hostname" (or somesuch)

doc/MANUAL.adoc
jrosdahl: > I think the syntax was borrowed from perl: [metacpan.org/pod/URI::redis](https://metacpan.org/pod/URI::redis)

I see. I like the `redis+unix://[:PASSWORD@]SOCKET_PATH[?db=DATABASE]` variant. Perhaps support only that one?

afbjorklund: Ok, that should be more similar to the other backends

afbjorklund:
Rewriting the syntax sort of got forgotten, but I can give it a go with the rebase.

doc/MANUAL.adoc
afbjorklund: When using user info, a host is required.

So syntax is: `redis+unix://[[USERNAME:]PASSWORD@localhost]SOCKET_PATH[?db=DBNUMBER[&password=PASSWORD]]`

afbjorklund: Removed the code for password parameter

jrosdahl:
Thanks.

Issue #1049, opened 2022-04-11
Improve manifest handling for primary/secondary storage


bengtj:
We're investigating an issue with high system load in our build cluster caused by looking for files during preprocessing. Then we tested depend mode and the sysload got even worse. It turns out that this was largely an issue with the way manifests are handled in secondary storage.

In short the problem comes from the fact that a manifest file is never updated from secondary storage when there is a manifest for that source file/compiler/flag combination in the local cache. It follows the same method as other files first looking for the local file and then try to fetch it from secondary storage. But as the manifest files are updated with new combinations of header file hashes this does not work well for manifests.

Also, the secondary storage cache manifest files are continuously being overwritten with manifest files based on the local cache manifest for the latest machine that lacked an entry for a header file combination. So the manifest files in secondary storage normally only contains few entries.

We tried a quick fix in calculate_result_and_manifest_key() updating this part in ccache 4.5.1:
```
    const auto manifest_path =
      ctx.storage.get(*manifest_key, core::CacheEntryType::manifest);

    if (manifest_path) {
      LOG("Looking for result key in {}", *manifest_path);
      MTR_BEGIN("manifest", "manifest_get");
      try {
        const auto manifest = read_manifest(*manifest_path);
        result_key = manifest.look_up_result_digest(ctx);
      } catch (const core::Error& e) {
        LOG("Failed to look up result key in {}: {}", *manifest_path, e.what());
      }
      MTR_END("manifest", "manifest_get");
      if (result_key) {
        LOG_RAW("Got result key from manifest");
      } else {
        LOG_RAW("Did not find result key in manifest");
      }
    }
```

To something like this:
1. Look for manifest in local cache only, try to find result key for the set of header files.
2. If no key found, then fetch manifest from secondary cache only and look for result key.
3. If key still not found, then continue with preprocessed hits and real compilation as normal.
4. When manifest is later updated upload the new version of the manifest. Which is now based on the secondary cache manifest with an added entry.

This works so much better in our setup. Entries are now added to the secondary storage manifest whenever header files are changed and other build machines directly takes advantage of this. Before things were quite confusing because the build time on a certain build machine was so dependendant on the state of the manifests in the local cache. As our preprocessed hits are expensive this affected the build times a lot for a long time after a header update.

[...]

Issue #1010, opened 2022-02-22
Add configuration option to disable primary storage and use secondary storage only


mbamelis:
We are using Ccache to speedup our CI/CD pipelines that build and test our C++ applications. This is working wonderfully with the secondary storage backend (e.g. a Redis server) so the different pipelines get to share cached results.

However, our pipelines actually run an ephemeral container so primary storage is empty at the start and filling it is useless since the storage is not carried over to future builds. Therefore, it would be useful if we could configure Ccache to skip primary storage completely and only use secondary storage.

We've looked into `max-size` and `max-files` but putting those to "0" equals "no limit". Giving them a low value will trigger Ccache's cleanup logic (near) constantly.

**So we are looking for a way to configure Ccache to operate in a kind of "secondary storage only" mode for use in ephemeral environments.**

jrosdahl:
This feature is planned for – see https://github.com/ccache/ccache/issues/856#issuecomment-875839782. The reason why `read_only_primary` can't just be added right away is that the code for secondary storage currently depends on the result landing in primary storage first (since that was faster to implement in order to get basic secondary storage functionality in place). In other words: some refactoring is needed first, and I haven't had time to do that yet.

mbamelis:
I'm happy to see my request is not nonsense and you had actually already planned to implement it!

u625030:
We would also very much be interested in this feature as we have a very similar use case.

jrosdahl:
> When is the release of 4.7 planned?

Ccache releases don't follow any schedule. There is no specific plan for 4.7; it will be released when it's ready.

Issue #856, opened 2021-06-06
Storage backends: API and framework


jrosdahl:
Design and implement an API and framework for simple key-value storage backends. This includes configuration parsing (see #855) and new statistics counters. Only needs to support one secondary backend for now.

afbjorklund:
Something that I found recently was that it was good to add a "rm" function, next to the usual "set" and "get" functions. It made it easier to do testing, that could clean up after themselves... Previously we just restarted memcached or wiped the directory, to clean the cache. But for some backends, one uses the same storage also for testing. And then it can be useful to do a more directed clean, to remove items from the cache by key. Not sure it is needed during normal operations, since it does auto-clean.

jrosdahl:
* [API for secondary storage](https://github.com/ccache/ccache/blob/2b3368cfc2a076e3205e0d7f400b4f97db227dae/src/storage/SecondaryStorage.hpp)
* [Example implementation](https://github.com/ccache/ccache/commit/f8789e2f4156cd3413f72c0ce4ae0dc321f70084)

BrightSeraph:
I did not find any hints on this topic:
Will the secondary storage propagate into the primary and/or will the primary eventually drop its content into the secondary?

jrosdahl:
> Will the secondary storage propagate into the primary and/or will the primary eventually drop its content into the secondary?

My current plan looks something like this:

| Primary storage hit/miss | Secondary storage hit/miss | What happens |
| ------- | --------- | ------------ | 
| miss    | miss      | Compile, write to primary<sup>[1]</sup>, write to secondary<sup>[2]</sup> |
| miss    | hit       | Read from secondary, write to primary<sup>[1]</sup><sup>[3]</sup> |
| hit     | -         | Read from primary, don't write to secondary<sup>[4]</sup> |

- <sup>[1]</sup> Unless primary storage is read-only (future config option: `read_only_primary`/`CCACHE_READONLY_PRIMARY`).
- <sup>[2]</sup> Unless secondary storage is read-only (`|read-only` or `|read-only=true` appended to the storage URL).
- <sup>[3]</sup> Unless secondary storage is marked to not share results. The use case for this is if you have a secondary storage that is so fast that it's of no use to also populate the primary storage. I haven't come up with a good name for this yet, though. The best I got so far is `|share-hits=false` appended to the storage URL. Other suggestions are welcome.
[...]

PR #1065, opened 2022-05-02
feat: Fetch/update secondary storage manifest on primary miss

bengtj:
On a manifest key miss in the primary cache the manifest from secondary
storage is fetched. This will incrementally update secondary storage
cache manifest files instead of overwriting them with modified primary
storage manifests. Which in turn makes direct hits being properly shared
between machines through secondary storage.

Partially resolves issue described in #1049

### DIFF
--- a/src/ccache.cpp
+++ b/src/ccache.cpp
@@ -1791,8 +1791,9 @@ calculate_result_and_manifest_key(Context& ctx,
 
     manifest_key = hash.digest();
 
-    const auto manifest_path =
-      ctx.storage.get(*manifest_key, core::CacheEntryType::manifest);
+    auto manifest_path = ctx.storage.get(*manifest_key,
+                                         core::CacheEntryType::manifest,
+                                         storage::Storage::Mode::primary_only);
 
     if (manifest_path) {
       LOG("Looking for result key in {}", *manifest_path);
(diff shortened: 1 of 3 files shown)


bengtj:
Rebased nonstd -> std...

jrosdahl:
Thanks for the PR! It looks like it indeed would improve the situation when using a single secondary storage, but it still has the same fundamental problem when there's more than one secondary storage. That is probably not very common, but I would like to solve the problem in a more generic way. To do that, the storage API needs to be improved so that it passes data via memory instead of via files, which is something I have planned to do anyway to make it possible to use only secondary storage (see #1010 and https://github.com/ccache/ccache/issues/856#issuecomment-875839782).

Since I plan to work in this area in ccache 4.7, I won't apply this PR now since the code and API it touches needs to be rewritten anyway. If I change plans for 4.7 and don't do this, I'll go with this PR instead. Does that sound OK to you?

bengtj:
In our case I guess it depends on when similar functionality arrives upstream (like manifest merge) so that we can upgrade ccache.

We actually had ccache 3.2.9 with a similar patch for secondary cache on nfs before that we never got enough time to rebase. When the same functionality arrived in upstream ccache we updated but the performance got much worse in a strange way. It took us a long time to pinpoint the root cause why there was so much sysload in certain builds and why sharing over nfs did not work as expected. A big part of the problem is that preprocessing C++ is so expensive with high sysload after manifest miss. In short, this is an extreme improvement for our use case.

Maybe you can consider applying the PR and then work in the area? It is only a few lines being changed and I do not think it makes that big of difference for your bigger refactoring :)

(I do understand why you would want to avoid changes after looking at your links. But it would really help us keeping up with newer ccache versions.)

jrosdahl:
> But it would really help us keeping up with newer ccache versions.

Sure, I can apply the PR and then remove the code again, but I'm not sure why that would help you keep up with newer versions. :slightly_smiling_face: Well, maybe except if you want to base your own ccache on an development version from ccache master instead of a released version, in which case there will be a window where the improvement is present. Is that what you're looking for?

bengtj:
I think that is what we're looking for to be able to keep up better. I guess it is mainly if your refactoring does not get into 4.7 though.

I think there are others needing this functionality as well. There was a request for me to create this PR in my original request for manifest merge.

jrosdahl:
> I think that is what we're looking for to be able to keep up better.

Still not sure I understand, but maybe I don't have to. :slightly_smiling_face: 

> I guess it is mainly if your refactoring does not get into 4.7 though.

(Right. That's what I tried to address with _"If I change plans for 4.7 and don't do this, I'll go with this PR instead."_.)

bengtj:
Thanks!

Keep up the good work, ccache is extremely helpful with our large codebase...

bengtj:
Just so I understand:
If you refactor the storage API then the old secondary storage manifest behaviour will be reintroduced? Or will you refactor it so that it works like after this PR?

(I guess the feature of merging manifests is out scope for refactoring.)

bengtj:
Sounds great!

PR #1067, opened 2022-05-02
Add optional Tkrzw secondary storage backend

### LINKED ISSUE #1066: Tkrzw local DBM backend with remote RPC
A new database backend, with the successor to Kyoto Tycoon:

* ~~https://github.com/ccache/ccache/issues/435~~

It offfers both local DBM and remote RPC options, for the database.

The server can be memory-mapped, similar to Redis-on-Flash etc.

https://dbmx.net/tkrzw/
https://github.com/estraier/tkrzw

https://dbmx.net/tkrzw-rpc/
https://github.com/estraier/tkrzw-rpc

afbjorklund:
Tkrzw is a successor to Kyoto Cabinet and Kyoto Tycoon,
and has a Tkrzw-RPC component for the remote access (gRPC).

The library requires C++-17 and is only available on Linux.
It needs to be separately installed, and is optional to use.

Closes #1066


afbjorklund:
Added some tests for RPC, and cleanup the manual.

jrosdahl:
Thanks for the idea, but I'm not willing to maintain this. I wrote some thoughts about what kind of backends we can bundle in https://github.com/ccache/ccache/issues/414#issuecomment-855351889 and https://github.com/ccache/ccache/issues/414#issuecomment-855361565:

> 5. "Simple" backends can be bundled with ccache, where "simple" here means something like "with minimal external dependencies". For instance, I would like to avoid depending on more external libraries unless they are ubiquitous and small enough.
> [...]
> A problem I see with backends not enabled by default is that they make the documentation messy. I would prefer if all bundled backends are expected to be enabled by default so that the configuration doesn't have to say "if enabled", etc.

I have some ideas on how to improve the secondary storage framework to use long-lived helper processes to support TLS-enabled HTTP and Redis, and as a side-effect of that we could actually support non-bundled backends as well. But I also have lots of other higher priority ideas and, as always, not much time to act on them, unfortunately.

Regarding C++17, it's very deliberate to stay on C++14 for now to make it easier to build ccache with old compilers. That said, I have an entry on my to do list to re-evaluate this some time during the year.

afbjorklund:
The original idea was to try a backend based on gRPC, and this one was one of the easier.

With the library, the protobuf here was more straight-forward than using the bazelbuild API.

https://github.com/bazelbuild/remote-apis/blob/main/build/bazel/remote/execution/v2/remote_execution.proto

https://github.com/estraier/tkrzw-rpc/blob/master/tkrzw_rpc.proto

PR #1068, opened 2022-05-04
fix: Check for short reads in Reader::read_str

gjasny:
The `read_int` function checks for short reads so `read_str` should do as well.

Additionally don't trigger undefined behavior with `operator[]` on empty strings.

### DIFF
--- a/src/core/Reader.hpp
+++ b/src/core/Reader.hpp
@@ -72,7 +72,12 @@ inline std::string
 Reader::read_str(const size_t length)
 {
   std::string value(length, 0);
-  read(&value[0], length);
+  if (length > 0) {
+    const auto bytes_read = read(&value[0], length);
+    if (bytes_read != length) {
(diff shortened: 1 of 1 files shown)


jrosdahl:
> The `read_int` function checks for short reads so `read_str` should do as well.

Sounds good, thanks.

> Additionally don't trigger undefined behavior with `operator[]` on empty strings.

Correct me if I'm wrong, but I see nothing wrong what the code. From <https://en.cppreference.com/w/cpp/string/basic_string/operator_at>:

> If `pos == size()`, a reference to the character with value CharT() (the null character) is returned.

jrosdahl:
That's the C++98 version. The C++11 version reads like this:

> If pos is less or equal to the [string length](https://www.cplusplus.com/basic_string::length), the function never throws exceptions (no-throw guarantee). Otherwise, it causes undefined behavior.

PR #1071, opened 2022-05-10
fix: Fix parsing of MSVC response files

jacobly0:
Inside response files passed to clang-cl or msvc:
1. `\` should not be ignored completely, but rather it should only escape `\` and `"`.
2. `'` should not be treated specially.


jrosdahl:
Thanks.

stilllman:
I have an issue that seems related to this bug, see #1216.

PR #1077, opened 2022-05-19
CMake: find zstd & hiredis with their config file first

SpaceIm:
Recent versions of `zstd` & `hiredis` provide a CMake config file. This PR proposes to check first existence of these config files for discovery/injection of these libraries.

There is also a small fix for out of source build, related to generation of the file `version.cpp`.


jrosdahl:
Thanks for the PR.

> Recent versions of `zstd` & `hiredis` provide a CMake config file. This PR proposes to check first existence of these config files for discovery/injection of these libraries.

Could you describe what issue this solves?

> There is also a small fix for out of source build, related to generation of the file `version.cpp`.

Could you describe what issue this solves? I don't have any problem doing out-of-source builds.

PR #1080, opened 2022-05-21
fix(win32execute): Handle space in paths when using response file

Predelnik:
Putting all arguments in quotes is done in `argv_to_string`, however when ccache uses `cmd_args` response file, `argv[0]` ends up being without quotes which ends up badly if it has spaces in it. In my case it resulted in preprocessor error and subsequent cache misses.

### DIFF
--- a/src/execute.cpp
+++ b/src/execute.cpp
@@ -151,7 +151,7 @@ win32execute(const char* path,
     TemporaryFile tmp_file(FMT("{}/cmd_args", temp_dir));
     args = Win32Util::argv_to_string(argv + 1, sh, true);
     Util::write_fd(*tmp_file.fd, args.data(), args.length());
-    args = FMT("{} @{}", full_path, tmp_file.path);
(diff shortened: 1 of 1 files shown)


src/execute.cpp
```diff
@@ -151,7 +151,7 @@ win32execute(const char* path,
     TemporaryFile tmp_file(FMT("{}/cmd_args", temp_dir));
     args = Win32Util::argv_to_string(argv + 1, sh, true);
     Util::write_fd(*tmp_file.fd, args.data(), args.length());
-    args = FMT("{} @{}", full_path, tmp_file.path);
+    args = FMT(R"("{}" @{})", full_path, tmp_file.path);
```
jrosdahl, line 154: Looks like I didn't follow up _"Also, perhaps the compiler path should be double quoted [...]"_ in https://github.com/ccache/ccache/pull/816#discussion_r593750281 properly.

On that note, how about quoting the `@{}` part as well? The temporary response file can be a path with spaces in it.

jrosdahl:
Thanks.

PR #1089, opened 2022-06-06
xxhash.h: Fix build with gcc-12

Mingli-Yu:
Fixes:
  /buildarea/tmp/work/core2-64-poky-linux/ccache/4.6.1-r0/ccache-4.6.1/src/third_party/xxhash.h:3932:1: error: inlining failed in call to 'always_inline' 'XXH3_accumulate_512_sse2': function not considered for inlining
  3932 | XXH3_accumulate_512_sse2( void* XXH_RESTRICT acc,
      | ^~~~~~~~~~~~~~~~~~~~~~~~
  /buildarea/tmp/work/core2-64-poky-linux/ccache/4.6.1-r0/ccache-4.6.1/src/third_party/xxhash.h:4369:9: note: called from here
  4369 |         f_acc512(acc,
      |         ^~~~~~~~~~~~~
  4370 |                  in,
      |                  ~~~
  4371 |                  secret + n*XXH_SECRET_CONSUME_RATE);

Signed-off-by: Mingli Yu <mingli.yu@windriver.com>

### DIFF
--- a/src/third_party/xxhash.h
+++ b/src/third_party/xxhash.h
@@ -1501,7 +1501,7 @@ static void* XXH_memcpy(void* dest, const void* src, size_t size)
 #  define XXH_NO_INLINE static
 /* enable inlining hints */
 #elif defined(__GNUC__) || defined(__clang__)
-#  define XXH_FORCE_INLINE static __inline__ __attribute__((always_inline, unused))
+#  define XXH_FORCE_INLINE static
 #  define XXH_NO_INLINE static __attribute__((noinline))
 #elif defined(_MSC_VER)  /* Visual Studio */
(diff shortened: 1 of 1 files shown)


jrosdahl:
Thanks for the PR, but you're modifying third party code that is just bundled with ccache. I don't want to keep local changes of it since those changes will disappear when bumping to the next version of xxHash.

Instead, please report an issue in [xxHash](https://github.com/Cyan4973/xxHash) and then we can get the upstream fix from them.

Mingli-Yu:
Thanks!
Use  https://github.com/Cyan4973/xxHash/pull/720 to track the issue.

PR #1093, opened 2022-06-09
build: Fix build with newer MSVC 2022 versions

### LINKED ISSUE #1076: Compile error with latest Visual Studio 2022 update
Once I installed the latest Visual Studio 2022 update, I am getting the following compile error when building. I don't know enough about C++ to know what this means exactly:

```console
[1/103] Building CXX object src\CMakeF...cache_framework.dir\AtomicFile.cpp.obj
FAILED: src/CMakeFiles/ccache_framework.dir/AtomicFile.cpp.obj
C:\PROGRA~1\MICROS~1\2022\COMMUN~1\VC\Tools\MSVC\1432~1.313\bin\Hostx64\x64\cl.exe  /nologo /TP -DHAVE_REDIS_STORAGE_BACKEND -DSTATIC_GETOPT -DWIN32_LEAN_AND_MEAN -DXXH_STATIC_LINKING_ONLY -Dnssv_CONFIG_SELECT_STRING_VIEW=nssv_STRING_VIEW_NONSTD -IC:\Users\rkitover\source\repos\ccache\build-msvc -IC:\Users\rkitover\source\repos\ccache\src -IC:\Users\rkitover\source\repos\ccache\build-msvc\zstd-1.5.2\lib -IC:\Users\rkitover\source\repos\ccache\build-msvc\hiredis-1.0.2\include /DWIN32 /D_WINDOWS  /GR /EHsc /MD /O2 /Ob2 /DNDEBUG /FIC:/Users/rkitover/source/repos/ccache/build-msvc/config.h /Zc:preprocessor /Zc:__cplusplus /D_CRT_SECURE_NO_WARNINGS /WX /W4 /wd5105 /wd4244 /wd4245 /wd4267 /wd4706 /wd4996 -std:c++17 /showIncludes /Fosrc\CMakeFiles\ccache_framework.dir\AtomicFile.cpp.obj /Fdsrc\CMakeFiles\ccache_framework.dir\ccache_framework.pdb /FS -c C:\Users\rkitover\source\repos\ccache\src\AtomicFile.cpp
C:\Users\rkitover\source\repos\ccache\src\core/exceptions.hpp(57): error C2664: 'std::string fmt::v8::format<>(fmt::v8::basic_format_string<char>)': cannot convert argument 1 from 'core::Error' to 'fmt::v8::basic_format_string<char>'
C:\Users\rkitover\source\repos\ccache\src\core/exceptions.hpp(57): note: No user-defined-conversion operator available that can perform this conversion, or the operator cannot be called
C:\Users\rkitover\source\repos\ccache\src\third_party/fmt/core.h(3117): note: see declaration of 'fmt::v8::format'
C:\Users\rkitover\source\repos\ccache\src\AtomicFile.cpp(48): note: see reference to function template instantiation 'core::Error::Error<core::Error&>(core::Error &)' being compiled
C:\Users\rkitover\source\repos\ccache\src\core/exceptions.hpp(57): error C2280: 'core::ErrorBase::ErrorBase(void)': attempting to reference a deleted function
C:\Users\rkitover\source\repos\ccache\src\core/exceptions.hpp(36): note: compiler has generated 'core::ErrorBase::ErrorBase' here
C:\Users\rkitover\source\repos\ccache\src\core/exceptions.hpp(36): note: 'core::ErrorBase::ErrorBase(void)': function was implicitly deleted because a base class 'std::runtime_error' has either no appropriate default constructor or overload resolution was ambiguous
C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Tools\MSVC\14.32.31326\include\stdexcept(101): note: see declaration of 'std::runtime_error'
ninja: build stopped: subcommand failed.
```

With Visual Studio 2019 this error does not occur.

Predelnik:
So this is the simplest fix I came up with, if we specify the first argument to be exactly of type `const char *` then we definitely cannot intersect with special member functions.

The only downside is probably inability to use compile time argument checking with `FMT_STRING` but since it is not done currently I think it should not be a problem.

I understand that it may seem like a workaround but in reality this construct is fragile and could become broken for other compiler as I demonstrate here with adding `catch` (thanks to Jonathan Caves from Microsoft for pointing this out) https://gcc.godbolt.org/z/qG1rjPMsr 

Fixes #1076

### DIFF
--- a/src/core/exceptions.hpp
+++ b/src/core/exceptions.hpp
@@ -43,7 +43,7 @@ public:
   Error(const std::string& message);
 
   // `args` are forwarded to `fmt::format`.
-  template<typename... T> inline Error(T&&... args);
+  template<typename... T> inline Error(const char* format, T&&... args);
 };
 
 inline Error::Error(const std::string& message) : ErrorBase(message)
@@ -51,8 +51,8 @@ inline Error::Error(const std::string& message) : ErrorBase(message)
 }
 
 template<typename... T>
-inline Error::Error(T&&... args)
-  : ErrorBase(fmt::format(std::forward<T>(args)...))
+inline Error::Error(const char* format, T&&... args)
+  : ErrorBase(fmt::format(format, std::forward<T>(args)...))
 {
 }
 
@@ -65,7 +65,7 @@ public:
   Fatal(const std::string& message);
 
   // `args` are forwarded to `fmt::format`.
-  template<typename... T> inline Fatal(T&&... args);
+  template<typename... T> inline Fatal(const char* format, T&&... args);
 };
 
 inline Fatal::Fatal(const std::string& message) : ErrorBase(message)
@@ -73,8 +73,8 @@ inline Fatal::Fatal(const std::string& message) : ErrorBase(message)
 }
 
 template<typename... T>
-inline Fatal::Fatal(T&&... args)
-  : ErrorBase(fmt::format(std::forward<T>(args)...))
+inline Fatal::Fatal(const char* format, T&&... args)
+  : ErrorBase(fmt::format(format, std::forward<T>(args)...))
 {
 }
 


jrosdahl:
> The only downside is probably inability to use compile time argument checking with `FMT_STRING` but since it is not done currently I think it should not be a problem.

I have used `FMT_STRING` [in some other places](https://github.com/ccache/ccache/blob/7c980edc1df8277aab086eec6ccaa4e66c8d2456/src/Logging.hpp#L39) and it would be nice to use it for the exception classes as well, but I'm fine with this fix for now.

jrosdahl:
Thanks.

PR #1100, opened 2022-06-20
Support for intel compiler on windows

DanielRichtmann:
Here's the pull request regarding https://github.com/ccache/ccache/discussions/1092.

What I did is to add icl to the group of windows compilers, i.e., the same that is already done for clang-cl. Without these changes I saw output like 

```
...
Compiler type: other
/nologo is not a regular file, not considering as input file
/EHsc is not a regular file, not considering as input file
/arch:sse2 is not a regular file, not considering as input file
/DWIN64 is not a regular file, not considering as input file
/DWIN32 is not a regular file, not considering as input file
/O2 is not a regular file, not considering as input file
/DNDEBUG is not a regular file, not considering as input file
/Qopt-report:0 is not a regular file, not considering as input file
/Qopt-report-phase:vec is not a regular file, not considering as input file
/Qdiag-disable:cpu-dispatch,11074,11075,13408,3920,2621 is not a regular file, not considering as input file
/Qalign-loops:32 is not a regular file, not considering as input file
/Qfnalign:64 is not a regular file, not considering as input file
/Qaxcore-avx2 is not a regular file, not considering as input file
...
```

which indicates that ccache isn't able to parse the compiler call correctly. Hence I did not see any cache hits. With this PR everything works as expected.

### DIFF
--- a/src/Config.cpp
+++ b/src/Config.cpp
@@ -461,6 +461,7 @@ compiler_type_to_string(CompilerType compiler_type)
 
     CASE(clang);
     CASE(gcc);
+    CASE(icl);
     CASE(msvc);
     CASE(nvcc);
     CASE(other);
--- a/src/Config.hpp
+++ b/src/Config.hpp
@@ -30,7 +30,16 @@
 #include <string>
 #include <unordered_map>
 
-enum class CompilerType { auto_guess, clang, clang_cl, gcc, msvc, nvcc, other };
+enum class CompilerType {
+  auto_guess,
+  clang,
+  clang_cl,
+  gcc,
+  icl,
+  msvc,
+  nvcc,
+  other
+};
 
 std::string compiler_type_to_string(CompilerType compiler_type);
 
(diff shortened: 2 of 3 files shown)


src/Config.cpp
```diff
@@ -461,6 +461,7 @@ compiler_type_to_string(CompilerType compiler_type)
 
     CASE(clang);
     CASE(gcc);
+    CASE(icl);
```
jrosdahl, line 464: Please add `icl` to [Config::parse_compiler_type](https://github.com/ccache/ccache/blob/9536b4fb1f653fb4e10dd84f1569f5d5d732795f/src/Config.cpp#L240) and [the manual](https://github.com/ccache/ccache/blob/9536b4fb1f653fb4e10dd84f1569f5d5d732795f/doc/MANUAL.adoc#L516-L532) as well.

DanielRichtmann: Fixed

PR #1103, opened 2022-06-25
Add https support as storage option

tcShadowWalker:
- Find OpenSsl library, enable https support if found
- Compile httplib with ssl/https support
- Register the 'https' scheme with the http backend as a secondary_storage option

The ccache docs explicitly state "HTTPS is not supported.". That note could be removed with this change.
It strikes me as careless to use unencrypted network-connections for something as sensitive as a build-cache.
I guess insecure http is only intended to be used to connect to a local proxy, and not really to a remote endpoint because of that.
Unless I'm missing something, enabling https seems quick and easy.

---

I needed https support to use Google-Cloud-Storage (GCS) as a backend.
And GCS, of course, does not allow requests using insecure http.
That can be achieved now by configuring:
`
secondary_storage = https://{bucket_name}.storage.googleapis.com|bearer-token={auth_token}
`
Whereas the auth-token can be generated with `gcloud auth print-access-token` (or via Api, of course)
That could be added as an example here: https://github.com/ccache/ccache/wiki/HTTP-storage#google-cloud-storage-gcs

I would have liked to add a test-case, but doing that would require disabling the ssl-certificate check to run a local server.
Regarding the implementation i wasn't sure whether both schemas should point to the same HttpStorage object. I also didn't see an easy way to do that in the global initialization context.

### DIFF
--- a/src/storage/Storage.cpp
+++ b/src/storage/Storage.cpp
@@ -52,6 +52,7 @@ const std::unordered_map<std::string /*scheme*/,
   k_secondary_storage_implementations = {
     {"file", std::make_shared<secondary::FileStorage>()},
     {"http", std::make_shared<secondary::HttpStorage>()},
+    {"https", std::make_shared<secondary::HttpStorage>()},
 #ifdef HAVE_REDIS_STORAGE_BACKEND
     {"redis", std::make_shared<secondary::RedisStorage>()},
 #endif
--- a/src/third_party/CMakeLists.txt
+++ b/src/third_party/CMakeLists.txt
@@ -6,6 +6,13 @@ else()
   target_compile_definitions(third_party PUBLIC -DSTATIC_GETOPT)
 endif()
 
+find_package(OpenSSL)
+if(OPENSSL_FOUND)
+  message(STATUS "Compiling with support for Https")
+  target_compile_definitions(third_party PRIVATE -DCPPHTTPLIB_OPENSSL_SUPPORT=1)
+  target_link_libraries(third_party PRIVATE OpenSSL::SSL OpenSSL::Crypto)
+endif()
+
 if(WIN32)
   target_sources(third_party PRIVATE win32/mktemp.c)
 endif ()


tcShadowWalker:
Damn, only after creating this PR do I now realize there are existing PRs about adding https.
(I scanned only the issue-list first and didn't find any)
Specifically, [894](https://github.com/ccache/ccache/discussions/894).

afbjorklund:
It should be relatively simple, to set up a local nginx as a proxy for the https server.

Maybe it can be added to https://github.com/ccache/ccache/wiki/HTTP-storage ?

afbjorklund:
This one might also work, if using bazel-remote

https://github.com/notnoopci/bazel-remote-proxy

jrosdahl:
Thanks for the PR. I'll close this as a duplicate of #890.

> It strikes me as careless to use unencrypted network-connections for something as sensitive as a build-cache. I guess insecure http is only intended to be used to connect to a local proxy, and not really to a remote endpoint because of that.

Say that you have several build servers on a company-internal trusted network. Do you think that it's careless to use unencrypted HTTP to share a cache between those servers?

No, the HTTP backend is not primarily intended to be used to connect to a local proxy, but you can of course do that if you like.

> Damn, only after creating this PR do I now realize there are existing PRs about adding https. (I scanned only the issue-list first and didn't find any) Specifically, [894](https://github.com/ccache/ccache/discussions/894).
> 
> One discussion point there was the overhead of the ssl(-handshake) overhead. (Although creating multiple requests in parallel should be much more relevant) And while it would be better, of course, to reuse the TCP connections, imo that is not an argument for not supporting https at all.

I agree. That point is more of a discussion of priority, not a blocker. The blocker is the first point (_"Startup cost when not using HTTPS storage"_). See also https://github.com/ccache/ccache/pull/902#issuecomment-993905988 for a similar question regarding encrypted Redis support.

> If we can't use https, we can't use ccache at all. OR have to fiddle around setting up a local proxy.

There is a third option: Since it's free software, you can make your own changes to ccache and use that version instead.

I do have an idea on how to improve the backend framework to be able support https and rediss (and other custom backends as well) without making ccache worse when not using such backends, but I unfortunately cannot make any promise about if and when that will happen.

Duplicate of #890

jrosdahl:
HTTPS support is now available in ccache 4.13!

PR #1108, opened 2022-07-09
feat: Add VS2022 and MSBuild to Windows CI runners

rkitover:
@jrosdahl 

Hi, if you remember I wanted to add VS2022 and MSBuild to the CI runners, since we had issues with both. And now that VS2022 compilation is fixed thanks to @Predelnik I should be able to do that.

By the way, do you know where these variables like `CMAKE_PARAMETERS` and `TEST_CC` are used in the Windows jobs? I could not figure that out, and I've never used GitHub Actions before.


rkitover:
@jrosdahl 

Ok it's working and ready to go, I had to disable tests on the MSBuild runners because they are not compiling, I or someone else should look at that later.

I see that the `ci/build` script is getting run on the Windows jobs, presumably through Git bash or some horrible thing and that is what is using the variables, so never mind.

rkitover:
I'm going to try to figure out why the unit tests don't build with MSBuild here, they do for me locally. Converted this back to draft for now.

rkitover:
I should really fix the Clang-Tidy runner to just do the tidy check and skip the build.

rkitover:
Ok I fixed the problem with compiling unit tests, it turns out that this only affects the VS2019 runner because the vm gh actions uses has an out of date and buggy Windows 10 SDK.

Some header files in said SDK cause compiler errors with `/Zc:preprocessor`, see:

https://developercommunity.visualstudio.com/t/sdk-and-experimentalpreprocessor/343296
.

This only affects `test_Stat.cpp`, so I turned off this flag for this source file, and everything works fine.

The only issue is the warning about overriding `/Zc:preprocessor` with `/Zc:preprocessor-` to turn it off for that one file.

I have no idea why this affects MSBuild but not Ninja, probably because the cmake Ninja generator does not select an SDK but uses a default one which may be newer or older, I don't know.

I think I'm done here, this is ready to go.

jrosdahl:
OK, thanks.

PR #1110, opened 2022-07-09
Allow using dict for compressing and decompressing

### LINKED ISSUE #1109: Dictionary compression for cache
I started some experiments, using the Zstandard dictionary compression.

https://facebook.github.io/zstd/#small-data

In theory, it improves performance and compression ratio for small files.
Remains to be seen, how the dictionary loading time affects in practice...

Since ccache typically only access 1-2 files per invocation, before quitting.
But it _should_ help with files that are "small", as in below 100 KiB or so ?

Made a proof-of-concept, need help with testing.

afbjorklund:
One `M.dict` for manifest, one `R.dict` for result

Cache contents are unreadable without these!

Requires zstd version 1.4.0+, such as 1.5.2

Closes #1109

### DIFF
--- a/doc/MANUAL.adoc
+++ b/doc/MANUAL.adoc
@@ -1289,6 +1289,35 @@ are currently compressed with a different level than the target level will be
 recompressed.
 
 
+== Compression dictionary
+
+Zstandard supports using a dictionary compression, for better performance and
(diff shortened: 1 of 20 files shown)


jrosdahl:
@afbjorklund: Do you plan to work more on this?

PR #1111, opened 2022-07-10
Add download-redis script matching the upload-redis

afbjorklund:
Normally one would have the primary cache around still,
but sometimes it can be useful to download from secondary.

The Redis symantics between `KEYS` and `SCAN` are confusing*,
so add a small helper script to avoid accidents if attempting...

----

\* using KEYS on a redis server downloads _all_ the keys at once,
bringing down the server. using SCAN downloads 10 at a time.

Unfortunately SSDB uses a different syntax, for keys and scan.
So it is currently **not** compatible, before fixing [redis-py](https://pypi.org/project/redis/) to match.

### DIFF
--- a/misc/download-redis
+++ b/misc/download-redis
@@ -1,0 +1,75 @@
+#!/usr/bin/env python3
+
+# This script downloads the contents of the cache from a Redis secondary storage
+
+import redis
+import os
+
+import progress.bar
+import progress.spinner
+import humanize
+
+config = os.getenv("REDIS_CONF", "localhost")
+if ":" in config:
+    host, port = config.rsplit(":", 1)
+    sock = None
+elif config.startswith("/"):
+    host, port, sock = None, None, config
+else:
+    host, port, sock = config, 6379, None
+username = os.getenv("REDIS_USERNAME")
+password = os.getenv("REDIS_PASSWORD")
+context = redis.Redis(host=host, port=port, unix_socket_path=sock, password=password)
+pipe = context.pipeline(transaction=False)
+
+ccache = os.getenv("CCACHE_DIR", os.path.expanduser("~/.cache/ccache"))
+try:
+    count = context.info()["db0"]["keys"]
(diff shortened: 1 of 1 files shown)


jrosdahl:
> Normally one would have the primary cache around still, but sometimes it can be useful to download from secondary.

Sure! I'm OK with bundling contributed scripts like this in the repository. That said, have you considered adding support for uploading and downloading to any secondary storage via ccache?

> * using KEYS on a redis server downloads _all_ the keys at once, bringing down the server. using SCAN downloads 10 at a time.
> 
> Unfortunately SSDB uses a different syntax, for keys and scan. So it is currently **not** compatible, before fixing [redis-py](https://pypi.org/project/redis/) to match.

Perhaps an alternative could be to download all `ccache:00*` keys first, then `ccache:01*` and so on, if that's more compatible with different server implementations?

misc/download-redis
```diff
@@ -0,0 +1,75 @@
...
+files = result = manifest = objects = 0
+size = 0
+columns = os.get_terminal_size()[0]
+width = min(columns - 24, 100)
+bar = progress.bar.Bar(
+    "Downloading...", max=count, fill="=", suffix="%(percent).1f%%", width=width
+)
+if not count:
+    bar = progress.spinner.Spinner("Downloading... ")
+keys = context.scan_iter()
```
jrosdahl, line 39: Perhaps use

```suggestion
keys = context.scan_iter(match="ccache:*")
```

here? Then there will be no need to filter out keys not starting with `ccache:` later.

afbjorklund: It was done that way mostly for the progress bar to match, but also some servers didn't implement "match"

afbjorklund:
> That said, have you considered adding support for uploading and downloading to any secondary storage via ccache?

Possibly considered, but not really implemented anything. The "archive" was as close I got, for the other storage types.

PR #1112, opened 2022-07-10
feat: auto dep FROM_INTERNET when unavailable

rkitover:

This is the auto `FROM_INTERNET` thing I promised to do.

This is almost completely backwards compatible, except the behavior of the `FROM_INTERNET` cmake options is slightly different, they are `ON` by default, and if the user did not enable `STATIC_LINK` and a system library is available, it will be linked shared.

Also a loud warning is printed when a dep is used `FROM_INTERNET`.

Still smoothing out some minor details, so this PR is a draft.

### DIFF
--- a/cmake/Findhiredis.cmake
+++ b/cmake/Findhiredis.cmake
@@ -1,26 +1,57 @@
-if(HIREDIS_FROM_INTERNET)
-  set(hiredis_version "1.0.2")
-  set(hiredis_url https://github.com/redis/hiredis/archive/v${hiredis_version}.tar.gz)
+if(hiredis_FOUND)
+  return()
+endif()
 
-  set(hiredis_dir ${CMAKE_BINARY_DIR}/hiredis-${hiredis_version})
-  set(hiredis_build ${CMAKE_BINARY_DIR}/hiredis-build)
+set(hiredis_FOUND FALSE)
 
-  if(NOT EXISTS "${hiredis_dir}.tar.gz")
-    file(DOWNLOAD "${hiredis_url}" "${hiredis_dir}.tar.gz" STATUS download_status)
-    list(GET download_status 0 error_code)
-    if(error_code)
-      file(REMOVE "${hiredis_dir}.tar.gz")
-      list(GET download_status 1 error_message)
-      message(FATAL "Failed to download hiredis: ${error_message}")
-    endif()
+find_package(PkgConfig)
+if(PKG_CONFIG_FOUND)
+  pkg_check_modules(HIREDIS hiredis>=${hiredis_FIND_VERSION})
+  find_library(HIREDIS_LIBRARY ${HIREDIS_LIBRARIES} HINTS ${HIREDIS_LIBDIR})
+  find_path(HIREDIS_INCLUDE_DIR hiredis/hiredis.h HINTS ${HIREDIS_PREFIX}/include)
+else()
+  find_library(HIREDIS_LIBRARY hiredis)
+  find_path(HIREDIS_INCLUDE_DIR hiredis/hiredis.h)
+endif()
+
+if(HIREDIS_INCLUDE_DIR AND HIREDIS_LIBRARY)
+  mark_as_advanced(HIREDIS_INCLUDE_DIR HIREDIS_LIBRARY)
+
+  add_library(HIREDIS::HIREDIS UNKNOWN IMPORTED)
+  set_target_properties(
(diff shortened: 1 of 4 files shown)


rkitover:
Also here I'm moving the windows `ws2_32` import lib to `Findhiredis.cmake` where I should have put in #1058.

Also, I just noticed that on Linux I have to pass `-DCMAKE_ASM_COMPILER=gcc` or blake3 errors out, I'll open a separate PR to fix that. I tested this on both Gentoo and Ubuntu.

rkitover:
Ok I'm done checking some minor things, this is ready to go.

rkitover:
@afbjorklund Yeah I will add the checksums, no problem.

rkitover:
@afbjorklund Added checksums, also fixed the fatal errors.

rkitover:
@jrosdahl This is ready to go, unless you see any problems.

rkitover:
@jrosdahl Disregard my issue with `CMAKE_ASM_COMPILER` it's a problem with my environment: https://github.com/ccache/ccache/issues/1113#issuecomment-1182615489 .

cmake/Findhiredis.cmake
```diff
@@ -1,4 +1,38 @@
...
+    INTERFACE_COMPILE_OPTIONS "${HIREDIS_CFLAGS_OTHER}"
+    INTERFACE_INCLUDE_DIRECTORIES "${HIREDIS_INCLUDE_DIR}")
+  if(WIN32 AND STATIC_LINK)
+    target_link_libraries(HIREDIS::HIREDIS INTERFACE ws2_32)
+  endif()
+
+  set(hiredis_FOUND TRUE)
+  set(target HIREDIS::HIREDIS)
+elseif(HIREDIS_FROM_INTERNET)
+  message(STATUS "*** WARNING ***: Using hiredis from the internet because it was NOT found and HIREDIS_FROM_INTERNET is TRUE")
```
jrosdahl, line 34: Would it make sense to use `message(WARNING [...])` here (and similar for zstd)?

jrosdahl:
Thanks for the PR.

While on the subject: Since we now require a newer CMake than 3.11 (a078662d9a2d219c4f68c4a298fdca09ebeecbcd), do you think that it would be of interest to start using [FetchContent](https://cmake.org/cmake/help/latest/module/FetchContent.html) instead of `file(DOWNLOAD [...])`, `tar xf` and custom checksumming logic?

cmake/Findhiredis.cmake
rkitover: CMake warnings are very ugly, they are intended more for warnings about cmake code and show the line number etc.. Here I wanted to show a loud warning that also blends into the rest of the output, but I will change this if you insist.

rkitover:
I looked at the FetchContent docs and I don't really understand this stuff right now, it's very complicated, but I got the basic Declare/Populate logic from the docs to work if that's fine with you. It does make things somewhat cleaner.

cmake/Findhiredis.cmake
jrosdahl: OK, then let's keep it as is. :+1:

cmake/Findzstd.cmake
```diff
@@ -2,32 +2,47 @@ if(zstd_FOUND)
...
+  add_library(ZSTD::ZSTD UNKNOWN IMPORTED)
+  set_target_properties(
+    ZSTD::ZSTD
+    PROPERTIES
+    IMPORTED_LOCATION "${ZSTD_LIBRARY}"
+    INTERFACE_INCLUDE_DIRECTORIES "${ZSTD_INCLUDE_DIR}")
+
+  set(zstd_FOUND TRUE)
+elseif(ZSTD_FROM_INTERNET)
+  message(STATUS "*** WARNING ***: Using zstd from the internet because it was NOT found ZSTD_FROM_INTERNET is TRUE")
```
jrosdahl, line 22:
```suggestion
  message(STATUS "*** WARNING ***: Using zstd from the internet because it was NOT found and ZSTD_FROM_INTERNET is TRUE")
```

jrosdahl:
It would be great if you could also update the [installation instructions](https://github.com/ccache/ccache/blob/e007123076cc3a8210d45b8579b16314f906e0f0/doc/INSTALL.md) to match the new behavior.

rkitover:
I rebased/squashed this commit and added a message.

jrosdahl:
Thanks.

PR #1117, opened 2022-07-13
Add script to create archive tarball from cache

afbjorklund:
Small variation on the old `upload-redis` script.

This .tar can be used as a secondary file storage.

### DIFF
--- a/misc/create-archive
+++ b/misc/create-archive
@@ -1,0 +1,65 @@
+#!/usr/bin/env python3
+
+# This script creates an archive of the contents of the cache from primary storage.
+
+import tarfile
+import os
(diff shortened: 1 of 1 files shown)


jrosdahl:
Thanks for the PR.

I don't quite understand the use of this script. Do you want me to merge it? If so, who would use it and when?

afbjorklund:
Like the other scripts, it could be considered a "prototype" of something like `ccache --export`.

How to save the contents of the current cache into a file, that could be used somewhere else.

afbjorklund:
Can keep it in the other issue for now, it has some more background and context

PR #1124, opened 2022-07-21
fix: Move Windows config and cache to LOCALAPPDATA

### LINKED ISSUE #946: Switch to using LOCALAPPDATA as default location on Windows
### Description ###
The current default on Windows of using `getenv("APPDATA")` results in the default location being in `"${USERHOME}/AppData/Roaming"`. In many enterprise environments that is part of a Windows roaming profile, and is synchronized across a potentially slow network connection on login (i.e. during the login screen when no interaction is possible). The usage of this location by ccache has the unfortunate side-effect of synchronizing a large cache of binary data that is [mostly] inherently tied to a local machine (compiler versions, paths, etc.).

Instead, `ccache` should default to using LOCALAPPDATA instead which is intended for use by application to store local-only data (email cache, browser cache, etc.).

I'll try to draft a PR in the next week or so, with the following considerations in mind:
* The new default precedence for the config file should be to use `ccache.conf` in LOCALAPPDATA, then APPDATA. This provides a way to share config while allowing override on a per-machine basis. It also maps well to the existing configuration locations so no migration is required.
* The new default cache directory should be at `LOCALAPPDATA/ccache`
* Issue a warning if a cache directory is detected at `APPDATA/ccache` but is not explicitly configured for that location. The warning should include a recommendation to migrate/delete the cache location.
* Continue using the `APPDATA/ccache` as a default iff it exists but the `LOCALAPPDATA/ccache` version does not

### LINKED ISSUE #1023: Primary config path different in Powershell and Bash
From using ccache 4.6 in a Windows GitHub Actions job:

In Powershell:

```
Primary config:   C:\Users\runneradmin\AppData\Roaming/ccache/ccache.conf
```

In Bash:

```
Primary config:   C:\Users\runneradmin/ccache/ccache.conf
```

rkitover:
Add a variadic template helper function PATH(std::string...) to
construct native paths using std::filesystem.

Throw a fatal error if a cache is detected in APPDATA.

Fix-up default cmake install directories and set system-wide config to
C:\ProgramData\ccache\ccache.conf .

Update manual.

Fix #1023
Fix #946

I think I have the logic we wanted implemented for this, also as a bonus the `~/.ccache` check will work on Windows too now.

I'm sure you'll tell me to do some things differently, so I marked this as a draft.


CMakeLists.txt
```diff
@@ -75,7 +75,14 @@ include(DefaultBuildType)
 #
 # Configuration
 #
+
+if(WIN32 AND (NOT CMAKE_INSTALL_SYSCONFDIR))
```
cristianadam, line 79: This is not needed.  `CMAKE_INSTALL_SYSCONFDIR` is used only when you use the `install` command in CMake and have a `SYSCONF` parameter.

ccache is only installing the executable, no conf file is installed.

doc/MANUAL.adoc
```diff
@@ -309,12 +309,16 @@ this:
 2. Otherwise, if the environment variable `CCACHE_DIR` is set then use
    `$CCACHE_DIR/ccache.conf`.
 3. Otherwise, if <<config_cache_dir,*cache_dir*>> is set in the secondary
-   (system-wide) configuration file then use `<cache_dir>/ccache.conf`.
+   (system-wide) configuration file then use `<cache_dir>/ccache.conf`. The
+   system-wide configuration on Windows is `C:\ProgramData\ccache\ccache.conf`
```
cristianadam, line 313: I don't see in the patch where `C:\ProgramData` is reffered. The part in the CMake file doesn't count.

CMakeLists.txt
rkitover: This is interpolated into `config.h` [here](https://github.com/ccache/ccache/blob/master/cmake/config.h.in#L209).

If you build this branch and run `./ccache -vs` you will see that it takes effect.

doc/MANUAL.adoc
rkitover: See above.

CMakeLists.txt
cristianadam: ok, but then it has a wrong value. `C:\ProgramData` is hardcoded. You can have Windows installed on a different drive, not just `C`. 

Running:

```
$ set | findstr ^ProgramData$
ALLUSERSPROFILE=C:\ProgramData
ProgramData=C:\ProgramData
```

I think `%allusersprofile%` should be used in this case.

doc/MANUAL.adoc
cristianadam: This value is hardcoded, use `%allusersprofile%` instead.

CMakeLists.txt
rkitover: You're right, I forgot about this, I'll adjust the other gnuinstalldirs variables as well.

rkitover: By the way, have you tried working in PowerShell?

I am maintaining a guide for it [here](https://github.com/rkitover/windows-dev-guide).

rkitover: Can you take a look again please?

rkitover:
With the install dir tweaks, which I can tweak further, we should be able to make cmake generate an installer package if we choose to do that.

CMakeLists.txt
rkitover: Actually, now that I think about it again, the variable needs to be expanded at runtime, not compile time.

rkitover: Fixed.

cristianadam:
I think the whole thing with slash to backslash is too much. 

Keep using slashes and just convert from backslash to slash when you get the path from the environment variable. Keeps the code easier to read and this patch smaller.

What's the deal with the `CMAKE_INSTALL_PREFIX`? Unrelated to this change. If somebody builds ccache they can specify the `CMAKE_INSTALL_PREFIX` themselves, or use `cmake --install <build-dir> --prefix <desired-install-prefix>`.

rkitover:
> I think the whole thing with slash to backslash is too much.
>
> Keep using slashes and just convert from backslash to slash when you get the path from the environment variable. Keeps the code easier to read and this patch smaller.

I wanted the `ccache -vs` output to show native paths with backslashes, it's not that bad. Not sure about the placement of the `DIR_SEP` constant though.

> What's the deal with the CMAKE_INSTALL_PREFIX? Unrelated to this change. If somebody builds ccache they can specify the CMAKE_INSTALL_PREFIX themselves, or use cmake --install <build-dir> --prefix <desired-install-prefix>.

The default prefix is `C:\P:rogram Files (x86)` which is wrong for 64 bit builds, which is the type of build most people will have.

I don't know how to check if `CMAKE_INSTALL_PREFIX` contains the default value or was set by the user. However, since I am setting it in the cache it can be changed in the GUI.

rkitover:
Oh I see now, there is `CMAKE_INSTALL_PREFIX_INITIALIZED_TO_DEFAULT` so I'll use that.

rkitover:
Ok, `CMAKE_INSTALL_PREFIX` is fixed, thanks.

rkitover:
I could make some kind of `PATH("foo", "bar", "baz") --> foo\bar\baz` helper to make this nicer instead of passing `DIR_SEP` to `FMT()`.

rkitover:
I made this path helper, does this look nicer?

rkitover:
@Predelnik Your thoughts on this? @cristianadam gave me a few great suggestions.

cmake/InstallDirs.cmake
```diff
@@ -0,0 +1,41 @@
...
+
+    if(NOT program_files)
+      if(NOT CMAKE_C_SIZEOF_DATA_PTR EQUAL 4)
+        set(program_files "\\Program Files")
+      else()
+        set(program_files "\\Program Files (x86)")
+      endif()
+    endif()
+
+    string(REPLACE "\\" "\\\\" program_files "${program_files}")
```
cristianadam, line 18: CMake's native separator format is slash, and the last thing you want to have is mixing slashes and backslashes. It's a nightmare.

Always use slashes and at the input and output convert to and from "native" format. I'm using "native" here because Windows can handle happily slashes.

Use [`file(TO_CMAKE_PATH)` ](https://cmake.org/cmake/help/latest/command/file.html?#to-cmake-path) to convert the `$ENV` paths and have less escaping and in the end simpler code.

doc/MANUAL.adoc
```diff
@@ -314,10 +314,28 @@ this:
...
 
+On Windows, this is the method used:
+
+1. If `CCACHE_CONFIGPATH` is set, use that path.
+2. Otherwise, if the environment variable `CCACHE_DIR` is set then use
+   `$CCACHE_DIR/ccache.conf`.
+3. Otherwise, if <<config_cache_dir,*cache_dir*>> is set in the secondary
+   (system-wide) configuration file then use `<cache_dir>\ccache.conf`. The
+   system-wide configuration on Windows is
+   `$env:ALLUSERSPROFILE\ccache\ccache.conf` by default, `$env:ALLUSERSPROFILE`
```
jrosdahl, line 329: If I understand correctly, the `$env:VAR` syntax is what's used in PowerShell and doesn't work in cmd? Do you think more people will understand that syntax than `%VAR%`? (I didn't until I googled.) And if they're using Bash (Git for Windows) it's yet a different syntax... :thinking: Perhaps we can reformulate the documentation to be agnostic to variable derefencing syntax?

doc/MANUAL.adoc
```diff
@@ -314,10 +314,28 @@ this:
...
+   is usually `C:\ProgramData`.
+4. Otherwise, if there is a legacy `$env:USERPROFILE\.ccache` directory then use
+   `$env:USERPROFILE\.ccache\ccache.conf`.
+6. Otherwise, if `$env:LOCALAPPDATA\ccache\ccache.conf` exists it is used.
+7. Otherwise, if `$env:APPDATA\ccache\ccache.conf` exists it is used.
+
+The default cache dir on Windows is `$env:LOCALAPPDATA\ccache` unless specified
+otherwise by the config. If a cache is detected in `$env:APPDATA\ccache`, which
+was the default previously, a fatal error is thrown.
```
jrosdahl, line 338: I'm not comfortable with potentially breaking people's builds like that. I suggest either just using the legacy location if present.

src/Util.hpp
```diff
@@ -33,6 +34,23 @@
...
+template<typename... T>
+std::string
+PATH(T&... args)
+{
+  std::vector<std::string> arg_list{args...};
+  std::filesystem::path full;
+
+  for (auto&& p_str : arg_list) {
+    std::filesystem::path p(p_str);
+    full = full / p;
```
jrosdahl, line 48: Can this be simplified to

```suggestion
    full = full / p_str;
```

?

jrosdahl (commented):
Thanks for the PR!

cmake/InstallDirs.cmake
rkitover: I just needed this for the `config.h` interpolated variable, and not this one, I will use a separate variable for that.

doc/MANUAL.adoc
rkitover: This is what we agreed to do in #946, the issue is that `APPDATA` can be stored on the network and large caches would involve large file transfers when switching between computers in a domain for example, and since ccache cannot print a warning, a fatal error is the only option. But I will do whatever you suggest.

src/Util.hpp
rkitover: That works, thanks!

cmake/InstallDirs.cmake
rkitover: Fixed, thanks!

doc/MANUAL.adoc
rkitover: Fixed.

doc/MANUAL.adoc
rkitover: But yes, this will break everyone's build if they previously used ccache in powershell or cmd with the defaults, until they manually move their cache to the `LOCALAPPDATA` directory.

rkitover:
The MSBuild runners are already coming in handy, just fixed an error that showed up only in them.

doc/MANUAL.adoc
jrosdahl: > This is what we agreed to do in https://github.com/ccache/ccache/issues/946

Right. Sorry for my short memory and for apparently changing my mind.

> since ccache cannot print a warning, a fatal error is the only option.

For the record: another option is to not even look in `APPDATA` and just leave any cache there. Do you think that would be better?

rkitover: I could remove the check for the cache in `APPDATA`, but still check for the existence of a `ccache.conf` there after `LOCALAPPDATA` in case a user wants a roaming ccache configuration, the default cache location would still be in `LOCALAPPDATA`.

Furthermore, we could add a paragraph somewhere in the appropriate place in the documentation asking users to check for cache directories in `APPDATA` and remove them if found.

cmake/InstallDirs.cmake
```diff
@@ -0,0 +1,52 @@
...
+  if(CMAKE_INSTALL_PREFIX_INITIALIZED_TO_DEFAULT)
+    set(program_files "$ENV{ProgramFiles}")
+
+    # For 32 bit builds.
+    if(CMAKE_C_SIZEOF_DATA_PTR EQUAL 4 AND ENV{ProgramFiles\(x86\)})
+      set(program_files "$ENV{ProgramFiles\(x86\)}")
+    endif()
+
+    if(NOT program_files)
+      if(NOT CMAKE_C_SIZEOF_DATA_PTR EQUAL 4)
```
Predelnik, line 11: I'm not a cmake expert but shouldn't actually `CMAKE_SIZEOF_VOID_P` better be used here? Because it's at least [documented](https://cmake.org/cmake/help/latest/variable/CMAKE_SIZEOF_VOID_P.html).

Actually I found the [issue](https://gitlab.kitware.com/cmake/cmake/-/issues/18312) in cmake tracker discussing if `Program Files` is even correct default anymore but it's 3 years old and no changes have been made it seems.

doc/MANUAL.adoc
```diff
@@ -314,10 +314,28 @@ this:
...
 
+On Windows, this is the method used:
+
+1. If `CCACHE_CONFIGPATH` is set, use that path.
+2. Otherwise, if the environment variable `CCACHE_DIR` is set then use
+   `<CCACHE_DIR>/ccache.conf`.
+3. Otherwise, if <<config_cache_dir,*cache_dir*>> is set in the secondary
+   (system-wide) configuration file then use `<cache_dir>\ccache.conf`. The
+   system-wide configuration on Windows is
+   `<ALLUSERSPROFILE>\ccache\ccache.conf` by default, the `ALLUSERSPROFILE`
```
Predelnik, line 329: Are you sure we shouldn't surround by `%` here and everywhere for environment variables in windows? this notation should be familiar to most windows users I think.

src/Util.hpp
```diff
@@ -33,6 +35,22 @@
 
 class Context;
 
+// Helper variadic template function to construct native paths.
+// Used like: PATH("usr", "local", "bin");
+template<typename... T>
+std::string
+PATH(T&... args)
```
Predelnik, line 42: Is there a reason why the name is uppercase? maybe I'm not familiar with ccache code style but if it's a function then probably it's better to make a name more descriptive like `make_path` for example.

Additionally I think it's better to add `const` before `T` just in case.

src/Util.hpp
```diff
@@ -33,6 +35,22 @@
...
+{
+  std::initializer_list<const std::string> paths{args...};
```
Predelnik, line 44: Since we support C++17 because it's where `<filesystem>` is supported there's actually a way to implement the whole function basically in one line - `return (std::filesystem::path {} / ... / args).string ()`

Predelnik (commented):
I have provided a bit of feedback, hope it's useful.

src/Util.hpp
rkitover: That's amazing. Can you please explain how this works or give me some references to read?

cmake/InstallDirs.cmake
rkitover: The reason I did this was that it was always setting `C:\Program Files (x86)` as the prefix even for 64 bit builds. I think this should be more reasonable and in the future we can generate an `.msi` with CPack or something.

src/Util.hpp
rkitover: How strongly do you feel about this? I could make it something like `Util::make_path(...)` instead, but I thought that since we already have a `FMT(...)` helper this would be in a similar vein.

cmake/InstallDirs.cmake
rkitover: Oh right, that was the issue you linked, never mind.

rkitover:
@Predelnik Thank you very much for taking a look! I've made all but one of the fixes you suggested.

doc/MANUAL.adoc
rkitover: I removed the error and expanded the documentation section to include this paragraph as well as cache directory determining heuristics.

src/Util.hpp
rkitover: @jrosdahl and which would you prefer? `Util::make_path(...)` would probably not be much worse.

Predelnik: I agree that it's for @jrosdahl to decide, depending if there is separation of naming style for macros and functions etc.

src/Util.hpp
Predelnik: Here's a reference:
https://en.cppreference.com/w/cpp/language/fold

cmake/InstallDirs.cmake
```diff
@@ -0,0 +1,52 @@
...
+
+  set(CMAKE_INSTALL_BINDIR     "" CACHE PATH "executables subdirectory" FORCE)
+  set(CMAKE_INSTALL_SBINDIR    "" CACHE PATH "system administration executables subdirectory" FORCE)
+  set(CMAKE_INSTALL_LIBEXECDIR "" CACHE PATH "dependent executables subdirectory" FORCE)
+  set(CMAKE_INSTALL_LIBDIR     "" CACHE PATH "object libraries subdirectory" FORCE)
+endif()
+
+include(GNUInstallDirs)
+
+# Escape SYSCONFDIR for a C macro in config.h .
```
cristianadam, line 43: Why is the escaping needed? You can have the `CMAKE_INSTALL_SYSCONFDIR` which uses slashes and the `std::filesystem::path` will do the right thing. 

In my opinion this adds code that is not needed an complicates things.

rkitover: Just tried this, it actually doesn't:

```cpp
#include <iostream>
#include <filesystem>

int main(int argc, char** argv)
{
    std::cout << (std::filesystem::path{"C:/ProgramData"} / "ccache").string() << std::endl;
    return 0;
}
```

, the result is:

```console
C:/ProgramData\ccache
```

. Also, if it did work it would require always passing the path through something before using it, I'd rather have it be in the best form in the first place.

I could probably define the macro as a `constexpr` function that transforms the slashes, but that just moves the complexity from cmake to C++. But I can do that too.

rkitover: Also tried:

```cpp
std::filesystem::path{"C:/ProgramData", std::filesystem::path::native_format}
```

, but that also doesn't convert the slash.

cristianadam: ok, then `std::filesystem::path` is sillier than I thought.

This whole exercise is purely academic since `ALLUSERSPROFILE` will always exist and this value will never be used.

rkitover: This works though:

```cpp
#include <iostream>
#include <filesystem>

int main(int argc, char** argv)
{
    std::cout << (std::filesystem::path{"C:/ProgramData"} / "ccache")
        .lexically_normal().string() << std::endl;
    return 0;
}
```

, will add that to the wrapper.

Thanks for the idea!

rkitover: I can remove this now if you really want as the macro is only used in one place, and I made it go through the normalizing `PATH(...)` wrapper.

rkitover: I really think the value should be correct at build time. Another thing I could do is hide the gory details by making a `configure_file()` wrapper that does this automatically. @jrosdahl what do you suggest?

rkitover: @cristianadam I removed the unnecessary cache entry and put the two lines into the `GenerateConfigurationFile.cmake` file where the `configure_file()` call is, is this more tolerable?

src/Util.hpp
jrosdahl: `Util::make_path` sounds good to me.

cmake/InstallDirs.cmake
jrosdahl: Do I understand correctly that you want the path printed in logs or by `ccache -sv` to look nice (be in backslash form) in case `%ALLUSERSPROFILE%` isn't set? (I assume that using the unnormalized path works fine.) I agree with @cristianadam that it doesn't sound very important, but I'm fine with keeping the two lines if you like.

jrosdahl: Another option would perhaps be to let the CMake script require that `ALLUSERSPROFILE` is set and exit with an error otherwise?

doc/MANUAL.adoc
```diff
@@ -314,10 +314,47 @@ this:
...
+   environment variable is usually `C:\ProgramData`.
+4. Otherwise, if there is a legacy `%USERPROFILE%\.ccache` directory then use
+   `%USERPROFILE%\.ccache\ccache.conf`.
+6. Otherwise, if `%LOCALAPPDATA%\ccache\ccache.conf` exists it is used.
+7. Otherwise, if `%APPDATA%\ccache\ccache.conf` exists it is used.
+
+The cache directory on Windows is determined as follows:
+
+1. `cache_dir` in the above found configuration if specified.
+2. `%CCACHE_DIR%` if set.
```
jrosdahl, line 347:
```suggestion
1. `%CCACHE_DIR%` if set.
2. `cache_dir` in the above found configuration if specified.
```

doc/MANUAL.adoc
```diff
@@ -314,10 +314,47 @@ this:
...
+3. `%USERPROFILE%\.ccache` if it exists.
+4. Otherwise `%LOCALAPPDATA%\ccache`.
+
+*WARNING:* Previous builds of ccache for Windows defaulted to storing the cache
```
jrosdahl, line 351: The [AsciiDoc syntax for a warning admonition](https://docs.asciidoctor.org/asciidoc/latest/syntax-quick-reference/#admonitions) is to use the plain string `WARNING:`.

```suggestion
WARNING: Previous builds of ccache for Windows defaulted to storing the cache
```

doc/MANUAL.adoc
```diff
@@ -314,10 +314,47 @@ this:
...
    `$HOME/Library/Preferences/ccache/ccache.conf` (macOS) or
    `$HOME/.config/ccache/ccache.conf` (other systems).
 
+The location of the cache directory is determined as follows:
+
+1. `cache_dir` in the above found configuration if specified.
+2. `$CCACHE_DIR` if set.
+3. `$HOME/.ccache` if it exists.
+4. `$XDG_CACHE_HOME/ccache` if `$XDG_CACHE_HOME` is set.
+5. Otherwise `~/.cache/ccache` or `~/Library/Caches/ccache` on Apple systems.
```
jrosdahl, line 327: This is already described for _cache_dir (CCACHE_DIR)_ in the _Configuration options_ section. Would it be enough to refer to that description to avoid having two descriptions?

cmake/InstallDirs.cmake
rkitover: `%ALLUSERSPROFILE%` will always be set with nearly 100% probability. My only issue was that the path written to `config.h` be in the correct form. If you're fine with the two lines I'll keep them, we will need to revisit all of this again anyway, see below.

rkitover:
> Another option would perhaps be to let the CMake script require that ALLUSERSPROFILE is set and exit with an error otherwise?

Actually there are some other issues related to all of this.

The code as it is here will work for fine for  a user `cmake --install` on the machine where it was built, which is a good thing to fix and good enough for now.

However, when you are for example doing cross builds on linux, none of these environment variables like `ALLUSERSPROFILE` or `ProgramFiles` will be available, which is why I check for this in the cmake code and use hopefully reasonable enough defaults without a `SystemRoot`. This doesn't matter very much right now, as we only distribute a binary and the hardcoded `SYSCONFDIR` has practically no chance of being used.

These variables will also sometimes be different on different machines, for example due to `SystemRoot` being different, as @cristianadam pointed out.

I've never generated installers from cmake yet, so I don't know how all of this works, and if we do this I will do more research.

But there is one thing I wanted to address in the near-term, and not in this PR but separately.

I think we should have something like a `-DSTANDARD_BUILD=TRUE` option (specific name TBD) which would use the default for `SYSCONFDIR` instead of an environment variable, and various other things, such as compiler options and assembly targeting the CPUs we support instead of checking the capabilities of the build machine.

src/Util.hpp
rkitover: I've made this change.

doc/MANUAL.adoc
rkitover: Since I made the section "config and cache locations", is it ok if I do the opposite and link the `ccache_dir` option to this section.

jrosdahl: I see `cache_dir`/`CCACHE_DIR` as a configuration option like any other configuration option, so unless you think that it would be significantly better to describe it together with the configuration file location logic I would prefer keeping the description in _Configuration options_.

(In contrast, the configuration file location needs to be described separately since it's not a configuration option.)

cmake/InstallDirs.cmake
jrosdahl: OK.

doc/MANUAL.adoc
rkitover: Ok, I will expand the `cache_dir` description and make it more accurate and link to it from the configuration file section.

rkitover: How's this?

jrosdahl: :+1:

jrosdahl:
Thanks!

PR #1131, opened 2022-08-13
feat: Support multiple -Xarch_* arguments matching -arch

ntadej:
This pull request should bring more flexibility to iOS build caching. I admit that I am by far not a compiler expert so I did a lot of assumptions and I am opening this PR to get feedback.

The current situation is that multiple `-Xarch_*` arguments are not supported. Based on #874 I took a step forward and now match `-Xarch_*` to `-arch` and if they match, treat them as `-arch` arguments so adding them together to the each preprocessor run.

This allows me to efficiently cache multiarch iOS builds but I did not runtime test it yet if the output actually makes sense (the compilation finishes fine).

### DIFF
--- a/src/argprocessing.cpp
+++ b/src/argprocessing.cpp
@@ -66,7 +66,7 @@ struct ArgumentProcessingState
   ColorDiagnostics color_diagnostics = ColorDiagnostics::automatic;
   bool found_directives_only = false;
   bool found_rewrite_includes = false;
-  std::optional<std::string> found_xarch_arch;
+  std::unordered_map<std::string, std::vector<std::string>> xarch_args;
   bool found_mf_opt = false;
   bool found_wp_md_or_mmd_opt = false;
   bool found_md_or_mmd_opt = false;
@@ -399,20 +399,17 @@ process_option_arg(const Context& ctx,
     config.set_direct_mode(false);
   }
 
-  // -Xarch_* options are too hard.
+  // -Xarch_* options need to be handled with care
   if (util::starts_with(arg, "-Xarch_")) {
     if (i == args.size() - 1) {
       LOG("Missing argument to {}", args[i]);
       return Statistic::bad_compiler_arguments;
     }
     const auto arch = arg.substr(7);
-    if (!state.found_xarch_arch) {
-      state.found_xarch_arch = arch;
-    } else if (*state.found_xarch_arch != arch) {
-      LOG_RAW("Multiple different -Xarch_* options not supported");
-      return Statistic::unsupported_compiler_option;
-    }
-    state.common_args.push_back(args[i]);
+    auto [it, inserted] =
+      state.xarch_args.emplace(arch, std::vector<std::string>());
+    it->second.emplace_back(args[i + 1]);
+    ++i;
     return Statistic::none;
   }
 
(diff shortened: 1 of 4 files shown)


birkskyum:
Thanks for looking into this, the ios test is definitely the paint point right now. I like the approach you take, and for reference there are also some [github actions for ccache](https://github.com/marketplace?type=actions&query=ccache+) that we might be able to fall back on.

test/suites/multi_arch.bash
```diff
@@ -36,17 +36,32 @@ SUITE_multi_arch() {
     expect_stat direct_cache_hit 2
     expect_stat cache_miss 3
 
-    # A single -Xarch_* matching -arch is supported.
+    # A multiple -Xarch_* matching -arch are supported.
```
jrosdahl, line 39: While that's true, perhaps the comment no longer makes sense in this context since it only tests one `-Xarch` option? Maybe remove the comment instead?

src/argprocessing.cpp
```diff
@@ -1423,22 +1416,39 @@ process_args(Context& ctx)
     compiler_args.push_back("-dc");
   }
 
-  if (state.found_xarch_arch && !args_info.arch_args.empty()) {
-    if (args_info.arch_args.size() > 1) {
+  if (!state.xarch_args.empty()) {
+    if (args_info.arch_args.size() != state.xarch_args.size()) {
```
jrosdahl, line 1420: If I understand the relationship between `-arch` and `-Xarch` correctly, `-Xarch` options are only relevant for the matching `-arch`, which is why we can transform

```
-arch A -Xarch_A a -arch B -Xarch_B b
```

to two preprocessor calls with

```
-arch A -Xarch_A a
-arch B -Xarch_B b
```

Does that sound correct?

Since the code now already handles the logic of matching `-Xarch` options with `-arch`, perhaps it would be just as easy to support zero or multiple `-Xarch` options? That is, could we transform

```
-arch A -Xarch_A a1 -Xarch_A a2 -arch B -Xarch_B b -arch C
```

into

```
-arch A -Xarch_A a1 -Xarch_A a2
-arch B -Xarch_B b
-arch C
```

?

ntadej: Honestly I'm not sure if the third example should work. I am guessing here a bit the relationship between `-arch A` and `-Xarch_A`. Will try to find some documentation so I feel more confident in my change.

ntadej:
Unfortunately I can not find any reasonable documentation for this. The implementation above with one matching `Xarch` works fine for me for some projects, but not for the others (many cache misses). Not sure how we should proceed here, I'm open to suggestions.

jrosdahl:
> Unfortunately I can not find any reasonable documentation for this. The implementation above with one matching `Xarch` works fine for me for some projects, but not for the others (many cache misses). Not sure how we should proceed here, I'm open to suggestions.

I suggest supporting zero, one or several `-Xarch_A` options (for the same architecture `A`) since it seems a bit limited to only support one. That is, store something like

```c++
std::unordered_map<std::string, std::vector<std::string>> xarch_args;
```

so that `-Xarch_A a1 -Xarch_A a2 -Xarch_B b1` would result in

```c++
xarch_args["A"] == std::vector<std::string>{"a1, "a2"};
xarch_args["B"] == std::vector<std::string>{"b1"};
```

Then when looping over `-arch` values, also pass corresponding `-Xarch_` options to the preprocessor.

As for supporting `-arch A` without any `-Xarch_A`, here's an example that just has `-arch` options: https://github.com/llvm/llvm-project/blob/f116107f2d9342dabf8f1ccecb3f33f9b85d9b48/clang/docs/DriverInternals.rst?plain=1#L199

test/suites/multi_arch.bash
ntadej: Done.

src/argprocessing.cpp
ntadej: All three options above are supported now.

ntadej:
Hi @jrosdahl, I finally had time to work on this. I think I implemented all the changes proposed. Now `-Xarch_*` arguments are stored in a map and mapped with `-arch`.

Tagging also @rblx-kbuck as he provided a conflicting fix in the meantime.

src/argprocessing.cpp
```diff
@@ -384,20 +384,17 @@
...
     }
     const auto arch = arg.substr(7);
-    if (!state.found_xarch_arch) {
-      state.found_xarch_arch = arch;
-    } else if (*state.found_xarch_arch != arch) {
-      LOG_RAW("Multiple different -Xarch_* options not supported");
-      return Statistic::unsupported_compiler_option;
-    }
-    state.common_args.push_back(args[i]);
+    auto [it, inserted] =
```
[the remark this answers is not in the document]
ntadej: Not sure how to fix this nicely.

rblx-kbuck:
I'm not super familiar with the code, but it looks to me like this should work -- Xarch is being handled the same way the compiler would.

Since `ccache` is already forking multiple compiler instances if multiple different `-arch` options are passed, it would technically be safe to skip emitting the `-Xarch_*` when running the compiler, since we know that it will definitely match the `-arch` option passed. i.e.:

```bash
$ clang -arch x86_64 -Xarch_x86_64 -I. -E ...
```

Is equivalent to:

```bash
$ clang -arch x86_64 -I. -E ...
```

But for paranoia's sake, I suppose it's nice to keep it.

src/argprocessing.cpp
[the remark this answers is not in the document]
jrosdahl: I think we can just leave it.

jrosdahl:
Looks good to me.

jrosdahl:
Thanks!

PR #1133, opened 2022-08-14
Make the tests work under Windows

orgads:
Rebased version of #943.

### DIFF
--- a/test/run
+++ b/test/run
@@ -162,6 +162,8 @@ objdump_cmd() {
 objdump_grep_cmd() {
     if $HOST_OS_APPLE; then
         fgrep -q "\"$1\""
+    elif $HOST_OS_WINDOWS || $HOST_OS_CYGWIN; then
+        fgrep -q "$1"
     else
         fgrep -q ": $1"
     fi
@@ -173,11 +175,17 @@ expect_stat() {
     local line
     local value=""
 
+    if $HOST_OS_WINDOWS ; then
+        filter="sed -e s/\r//g"
+    else
+        filter=cat
+    fi
+
     while read -r key value; do
         if [[ $key == $stat ]]; then
             break
         fi
-    done < <($CCACHE --print-stats)
+    done < <($CCACHE --print-stats | $filter )
 
     if [ "$expected_value" != "$value" ]; then
         test_failed_internal "Expected $stat to be $expected_value, actual $value"
@@ -204,7 +212,23 @@ expect_equal_content() {
         test_failed_internal "expect_equal_content: $2 missing"
     fi
     if ! cmp -s "$1" "$2"; then
-        test_failed_internal "$1 and $2 differ: $(echo; diff -u "$1" "$2")"
+        test_failed_internal "$1 and $2 differ"
+    fi
+}
+
+expect_equal_text_content() {
+    if [ ! -e "$1" ]; then
+        test_failed_internal "expect_equal_text_content: $1 missing"
+    fi
+    if [ ! -e "$2" ]; then
+        test_failed_internal "expect_equal_text_content: $2 missing"
+    fi
+    if ! cmp -s "$1" "$2"; then
+        if $HOST_OS_WINDOWS  && diff -u --strip-trailing-cr "$1" "$2" > /dev/null ; then
+            test_failed_internal "$1 and $2 with differ line endings."
+        else
+            test_failed_internal "$1 and $2 differ: $(echo; diff -u "$1" "$2")"
+        fi
     fi
 }
 
@@ -499,6 +523,7 @@ COMPILER_TYPE_GCC=false
 
 COMPILER_USES_LLVM=false
 COMPILER_USES_MINGW=false
+COMPILER_USES_MSVC=false
 
 ABS_ROOT_DIR="$(cd $(dirname "$0"); pwd)"
 readonly HTTP_CLIENT="${ABS_ROOT_DIR}/http-client"
@@ -511,6 +536,7 @@ HOST_OS_WINDOWS=false
 HOST_OS_CYGWIN=false
 
 compiler_version="`$CC --version 2>/dev/null | head -1`"
+
 case $compiler_version in
     *gcc*|*g++*|2.95*)
         COMPILER_TYPE_GCC=true
@@ -525,12 +551,21 @@ case $compiler_version in
         ;;
 esac
 
+case $CC in
+    *MSVC*|*msvc*)
+        COMPILER_USES_MSVC=true
+        ;;
+esac
+
 case $compiler_version in
     *llvm*|*LLVM*)
         COMPILER_USES_LLVM=true
         ;;
     *MINGW*|*mingw*)
         COMPILER_USES_MINGW=true
+        ;;
+    *MSVC*|*msvc*)
+        COMPILER_USES_MSVC=true
         ;;
 esac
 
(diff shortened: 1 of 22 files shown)


orgads:
Woohoo! It passed!

cc @rvogg

test/run
```diff
@@ -622,7 +663,11 @@ echo
 cd $TESTDIR || exit 1
 
 mkdir compiler
+
 COMPILER="$(pwd)/compiler/$(basename "$REAL_COMPILER_BIN")"
+if $HOST_OS_WINDOWS; then 
+    COMPILER="$COMPILER.sh"
```
jrosdahl, line 669: I don't quite understand why shell scripts need to have a `.sh` extension. Is that a requirement in bash in MSYS2? Hmm, or is it because of ccache's strange usage of `Win32::add_exe_prefix`? If it's the latter, would it be possible to fix that instead, i.e., only add `.exe` to the path of the path doesn't already exist?

test/suites/basedir.bash
```diff
@@ -219,7 +225,14 @@ EOF
     fi
 
     # -------------------------------------------------------------------------
-    for option in MF "MF "; do
+    TEST "-MF/-MQ/-MT with absolute paths"
```
jrosdahl, line 228: There's no test here, so I suggest just removing this.

```suggestion
```

test/suites/basedir.bash
```diff
@@ -240,10 +253,17 @@ EOF
...
+        additional_options=
+    else
+        additional_options=(MQ MT)
+    fi
+    for option in "MQ " "MT " $additional_options; do
         TEST "-${option}/absolute/path"
 
         cd dir1
+        echo  CCACHE_BASEDIR="`pwd`" $CCACHE_COMPILE -I`pwd`/include -MD -${option}`pwd`/test.d -c src/test.c
+
```
jrosdahl, line 266: Left-over debug output?

```suggestion
```

test/suites/basedir.bash
```diff
@@ -219,7 +225,14 @@ EOF
     fi
 
     # -------------------------------------------------------------------------
-    for option in MF "MF "; do
+    TEST "-MF/-MQ/-MT with absolute paths"
+
+    if $HOST_OS_WINDOWS; then
+        additional_options=
```
jrosdahl, line 231: Why not test `-MF/path` on Windows?

Also, perhaps don't do this and similar changes since the test suite isn't run on Windows anyway?

test/suites/modules.bash
```diff
@@ -1,9 +1,11 @@
 SUITE_modules_PROBE() {
-    if ! $COMPILER_TYPE_CLANG; then
+    if ! $COMPILER_TYPE_CLANG || $COMPILER_USES_MSVC; then
         echo "-fmodules/-fcxx-modules not supported by compiler"
     else
-        touch test.c
-        $COMPILER -fmodules test.c -S || echo "compiler does not support modules"
+        cat <<EOF >testmodules.cpp
+#include <string>
+EOF
```
jrosdahl, line 7:
```suggestion
        echo '#include <string>' >testmodules.cpp
```

test/suites/pch.bash
```diff
@@ -55,7 +55,9 @@ SUITE_pch() {
 
     pch_suite_common
     if $COMPILER_TYPE_CLANG; then
-        pch_suite_clang
+        if ! [ $COMPILER_BIN = "gcc" ] || $RUN_WIN_XFAIL; then
```
jrosdahl, line 58: Can `$COMPILER_BIN` ever be `gcc` here since `$COMPILER_TYPE_CLANG` is true?

test/run
orgads: Done

test/suites/basedir.bash
orgads: Done.

test/suites/basedir.bash
orgads: Where's the logic for skipping tests? What if these tests *can* run on Windows? How do I enable them?

I don't know why the variants without space were excluded. Maybe @rvogg can comment on that.

test/suites/modules.bash
orgads: Done

test/suites/pch.bash
orgads: Yes. That's the case on MSYS2/Clang{32,64}. You run gcc, but it's in fact clang.

test/suites/basedir.bash
rvogg: As far as I remember it is because the tests run in the MSYS2 shell. The MSYS runtime automatically converts everything that looks like Unix paths to Windows paths. If there is no space between -MF and the path, it will not be converted. The correct variant under Windows would then probably be -MFC:/msys64/...  but that was too much effort for me to do it cleanly. Also I am not sure anymore if this is accepted by the Windows compiler.

test/run
rvogg: As far as I remember, the .sh extension was necessary, otherwise the script was not started in the shell. If the gitbash is installed, *.sh files are assigned to it.

Alternatively you would have to extend ccache to evaluate the shebang and then run the script with the matching executable.

orgads: It didn't work. See [this run](https://github.com/ccache/ccache/runs/7906502881?check_suite_focus=true).

orgads: Reverted to adding sh extension, and removed the commit that modifies `add_exe_suffix`.

test/suites/basedir.bash
jrosdahl: > Where's the logic for skipping tests?

Here:

https://github.com/ccache/ccache/blob/c020954d9da3f885eae559a76e6d9f3587f01c5f/test/suites/basedir.bash#L2-L4

orgads: I prefer to keep these. Looks like they were done to prepare the basedir tests for Windows at a later point.

orgads:
\o/

test/suites/basedir.bash
rvogg: As I mentioned in my original PR, my intention for testing on Windows was to implement the basedir feature for Windows. 
These adaptations come from my side branch where the feature works. 
You can undo the change and re-implement it later if the tests fail, or leave it as is.

orgads:
Rebased, all the dependencies are already merged.

jrosdahl:
Thanks.

jrosdahl:
Looks like some tests are flaky for environment reasons:

https://github.com/ccache/ccache/actions/runs/3237518536/jobs/5304674709
https://github.com/ccache/ccache/actions/runs/3237518536/jobs/5322157485

:-1:

If this continues I'll have to disable them.

orgads:
Yes, I've noticed.

The failure happens while installing packages. Maybe it was a temporary network failure.

Last retry succeeded.

```
  error: failed retrieving file 'mingw-w64-clang-x86_64-crt-git-10.0.0.r109.g0cd93ace0-1-any.pkg.tar.zst.sig' from mirror2.sandyriver.net : Operation too slow. Less than 1 bytes/sec transferred the last 10 seconds
  error: failed retrieving file 'mingw-w64-clang-x86_64-headers-git-10.0.0.r109.g0cd93ace0-1-any.pkg.tar.zst' from mirror.msys2.org : Operation too slow. Less than 1 bytes/sec transferred the last 10 seconds
  warning: failed to retrieve some files
  error: failed to commit transaction (unexpected error)
  Errors occurred, no packages were upgraded.
```

jrosdahl:
> Maybe it was a temporary network failure.

Well, it happened both yesterday and today, so it doesn't feel like a one-time glitch.

orgads:
Ok, I give up :)

Please merge #1179.

PR #1135, opened 2022-08-15
fix: Strip .exe from executable name on version output

orgads:
It currently prints "clang.exe version x.y.z"


jrosdahl:
Thanks.

orgads:
Oops, the commit message should have read ccache.exe, sorry.

PR #1145, opened 2022-08-23
Fix build with musl when using gcc12

kraj:
The build issues are seen with musl + gcc12, the issues are latent with glibc which is indirectly including these headers


jrosdahl:
Thanks.

PR #1156, opened 2022-09-07
feat: Add basic support for yacc/bison files

macdice:
Bison takes as input a .y file and produces a .c file, and optionally also a .h file if "-d" is provided.  The .y file does have CPP #include directives in regions of C code, but those survive unexpanded in the output .c file (for a later build stage to deal with), so "direct mode" seems to be sufficient.

My motivation is to be able to build PostgreSQL with BISON="ccache bison", because grammar file processing is now one of the slowest parts of our build after we ccache'd almost everything else.  This is a minimal proof of concept, enough to work with PostgreSQL, to see if the concept has legs.  I'd be grateful for any feedback.  Thanks!


jrosdahl:
Hi,

Supporting yacc/bison sounds nice and I understand that it could be very useful for some projects.

I would however prefer not to do it by adding more custom logic to the already messy argument handling code (accidental complexity due to supporting several different compilers without proper abstractions; see also #956).

For instance, as the PR is currently written ccache will try to guess the compiler from the yacc/bison binary name and fail, and then fall back to using GCC compiler semantics. Since those GCC semantics don't work for yacc/bison, it is solved by adding special cases for `.y` as a source file and finding `-d` which by luck doesn't have any meaning to other already supported compilers. :slightly_smiling_face: Instead I would like ccache to operate in yacc/bison mode if such a "compiler" is used. Another example of what I feel isn't OK is adding a bogus "yacc" language to `src/language.cpp` – that file contains a list of GCC/Clang languages supported by the `-x` flag.

So in summary, I feel that we need #956 in place before adding yacc/bison support.

macdice:
Thanks.  I've subscribed to that issue, and I'll try again once that infrastructure has started to take shape.  I'll wait for a day or two to see if any other feedback and then close this PR.

macdice:
Closing for now based on feedback.  Will try again when ccache is ready to support funkier kinds of 'compilers'.

PR #1158, opened 2022-09-13
feat: Support auto depend mode for MSVC without /showIncludes

orgads:
If MSVC is executed *without* /showIncludes, and ccache is configured with depend mode, add /showIncludes, and strip the extra output from stdout.

### DIFF
--- a/src/ArgsInfo.hpp
+++ b/src/ArgsInfo.hpp
@@ -75,6 +75,9 @@ struct ArgsInfo
   // Is the compiler being asked to output dependencies?
   bool generating_dependencies = false;
 
+  // Is the compiler being asked to output includes (MSVC -showIncludes)?
+  bool generating_includes = false;
+
   // The dependency target in the dependency file (the object file unless
   // overridden via e.g. -MT or -MQ).
   std::optional<std::string> dependency_target;
--- a/src/Config.cpp
+++ b/src/Config.cpp
@@ -83,6 +83,7 @@ enum class ConfigItem {
   log_file,
   max_files,
   max_size,
+  msvc_dep_prefix,
   namespace_,
   path,
   pch_external_checksum,
@@ -138,6 +139,7 @@ const std::unordered_map<std::string, ConfigKeyTableEntry> k_config_key_table =
     {"log_file", {ConfigItem::log_file}},
     {"max_files", {ConfigItem::max_files}},
     {"max_size", {ConfigItem::max_size}},
+    {"msvc_dep_prefix", {ConfigItem::msvc_dep_prefix}},
     {"namespace", {ConfigItem::namespace_}},
     {"path", {ConfigItem::path}},
     {"pch_external_checksum", {ConfigItem::pch_external_checksum}},
@@ -187,6 +189,7 @@ const std::unordered_map<std::string, std::string> k_env_variable_table = {
   {"LOGFILE", "log_file"},
   {"MAXFILES", "max_files"},
   {"MAXSIZE", "max_size"},
+  {"MSVC_DEP_PREFIX", "msvc_dep_prefix"},
   {"NAMESPACE", "namespace"},
   {"PATH", "path"},
   {"PCH_EXTSUM", "pch_external_checksum"},
@@ -754,6 +757,9 @@ Config::get_string_value(const std::string& key) const
   case ConfigItem::max_size:
     return format_cache_size(m_max_size);
 
+  case ConfigItem::msvc_dep_prefix:
+    return m_msvc_dep_prefix;
+
   case ConfigItem::namespace_:
     return m_namespace;
 
@@ -1000,6 +1006,10 @@ Config::set_item(const std::string& key,
 
   case ConfigItem::max_size:
     m_max_size = Util::parse_size(value);
+    break;
+
+  case ConfigItem::msvc_dep_prefix:
+    m_msvc_dep_prefix = Util::expand_environment_variables(value);
     break;
 
   case ConfigItem::namespace_:
--- a/src/Config.hpp
+++ b/src/Config.hpp
@@ -77,6 +77,7 @@ public:
   const std::string& log_file() const;
   uint64_t max_files() const;
   uint64_t max_size() const;
+  const std::string& msvc_dep_prefix() const;
   const std::string& path() const;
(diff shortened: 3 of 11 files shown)


orgads:
Depends on #992.

justinpryzby:
It seems like the "Including file" output is being elided during the initial compilation, as intended.
But then it's being output when using the cached compilation.

orgads:
Good catch, thanks!

Should be Ok now, not sure if the code is well positioned though. What do you say?

justinpryzby:
On Tue, Sep 13, 2022 at 01:03:51PM -0700, Orgad Shaneh wrote:
> Should be Ok now, not sure if the code is well positioned though. What do you say?

I didn't check the code much, but it seems to do what I wanted.  Thanks.

justinpryzby:
On Thu, Sep 15, 2022 at 08:00:14AM -0700, vvainola wrote:
> I still get "Note: including file:" if using ccache from within VS. On cache hits or when compiling from command line the includes are suppressed.

Could you check if VS is calling CL.EXE with "/showIncludes" ?
If so, that's what's expected.

orgads:
@vvainola How do you use it in VS?

vvainola:
I made small VS solution for reproducing that you can try. In the zip there is a path in Directory.Build.Props file to cl.exe which should be a symlink to ccache.exe. If CCACHE_DEPEND=1, includes are printed on cache miss and if CCACHE_DEPEND is not set they are never printed.
[ccache_demo.zip](https://github.com/ccache/ccache/files/9585222/ccache_demo.zip)

vvainola:
Sorry, it was my local mistake to try to fix this meson issue https://github.com/mesonbuild/meson/pull/10493 that includes from ccache are printed with forward slash and the latest meson release does not accept those. I used util::replace_all(line, "/", "\\") in strip_includes which looks like it took the entire string_view as input instead of just the single line.

orgads:
> I narrowed the issue down to the actual execute call
> 
> https://github.com/ccache/ccache/blob/a39ed5824d8f4343e33b80575a9ba0bf0c73e89f/src/ccache.cpp#L717
> 
> so it seems to have nothing to do with changes in this PR except that with depend mode the showIncludes is added automatically. For some reason the output "leaks" if compiling from Visual Studio but not if compiling with MSBuild from command line. I tried adding std::abort() right before the execute call and then there is no output in Visual Studio but if I add the std::abort() right after then the includes are printed.

I can reproduce. It is really strange. Looks like MSVC has a side-channel to capture the output. I also noticed that although `capture_output` is set, nothing is captured in the temp file.

Any idea how to proceed with it?

justinpryzby:
I don't have an MSVC to test with.
Did you check if this is related to the confusion about stdout vs stderr?

They updated the documentation to match the apparent behavior, but is it
possible that CL writes its includes to a different stream when run from
MSVC ?

On Fri, Sep 30, 2022 at 02:08:17AM -0700, Orgad Shaneh wrote:
> > I narrowed the issue down to the actual execute call
> > 
> > https://github.com/ccache/ccache/blob/a39ed5824d8f4343e33b80575a9ba0bf0c73e89f/src/ccache.cpp#L717
> > 
> > so it seems to have nothing to do with changes in this PR except that with depend mode the showIncludes is added automatically. For some reason the output "leaks" if compiling from Visual Studio but not if compiling with MSBuild from command line. I tried adding std::abort() right before the execute call and then there is no output in Visual Studio but if I add the std::abort() right after then the includes are printed.
> 
> I can reproduce. It is really strange. Looks like MSVC has a side-channel to capture the output. I also noticed that although `capture_output` is set, nothing is captured in the temp file.
> 
> Any idea how to proceed with it?

-- 
Justin Pryzby
System Administrator
Telsasoft
+1-952-707-8581

orgads:
> I don't have an MSVC to test with. Did you check if this is related to the confusion about stdout vs stderr? They updated the documentation to match the apparent behavior, but is it possible that CL writes its includes to a different stream when run from MSVC ?

Nah, both files (stdout and stderr) are empty. The output goes to stdout when running from command-line.

jrosdahl:
Wild guess: Could this be related to https://stackoverflow.com/a/7402354/331771?

vvainola:
> Wild guess: Could this be related to https://stackoverflow.com/a/7402354/331771?

It is indeed. The includes are no longer printed in VS after removing it from the environment given to CreateProcess
```
  LPCH a = GetEnvironmentStrings();
  std::string new_env;
  std::size_t len = strlen(a);
  while (len > 0) {
    std::string env = std::string(a, len);
    if (!util::starts_with(env, "VS_UNICODE_OUTPUT")) {
      new_env += env + '\0';
    }
    a += len + 1;
    len = strlen(a);
  }
  new_env += '\0';

  BOOL ret = CreateProcess(full_path.c_str(),
                           const_cast<char*>(args.c_str()),
                           nullptr,
                           nullptr,
                           1,
                           0,
                           (LPVOID)new_env.c_str(),
                           nullptr,
                           &si,
                           &pi);
```

justinpryzby:
LGTM.  postgresql compiles fine with depend mode, which (in the case of no cache hits) is about 2x faster than without depend mode.
https://cirrus-ci.com/task/4805444794318848

jrosdahl:
I merged this by mistake. Will revert.

I missed that it depended on #992.

orgads:
Why? Do you have rejections on #992?

jrosdahl:
> Why? Do you have rejections on https://github.com/ccache/ccache/pull/992?

Yes, [documentation needs an update and unit tests would be appreciated](https://github.com/ccache/ccache/pull/992#issuecomment-1229532220).

orgads:
Unit tests can be enabled with the other PR that enables them on Windows.

jrosdahl:
> Unit tests can be enabled with the other PR that enables them on Windows.

Sorry, what do you mean?

orgads:
#1133.

jrosdahl:
> #1133.

Those are integration tests, not unit tests. I'm talking about something similar to [unittest/test_Depfile.cpp](https://github.com/ccache/ccache/blob/f8b81c2bf89b15d0eb3fc4fee03d3be81bc06143/unittest/test_Depfile.cpp) which does not depend on a Windows runtime environment.

PR #1161, opened 2022-09-23
fix: small typo in the run script

louiscaron:
Missing space inside bash condition

Without this fix, there is an error returned when executing the run script with the option to keep the test directory `KEEP_TESTDIR`:

    ./run: line 645: [: missing `]'


jrosdahl:
Thanks.

PR #1163, opened 2022-09-23
fix: remove potential windows newline from the stats

louiscaron:
On windows machines, this could happen, the symptom is a failure message like the following:

    FAILED

    Test suite:     basedir (line 22)
    Test case:      Enabled CCACHE_BASEDIR
    Failure reason: Expected direct_cache_hit to be 0, actual 0

### DIFF
--- a/test/run
+++ b/test/run
@@ -179,8 +179,11 @@ expect_stat() {
         fi
     done < <($CCACHE --print-stats)
 
+    # strip any windows newline char from the value
+    value=${value//$'\r'/}
+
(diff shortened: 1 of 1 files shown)


louiscaron:
I have split the windows OS related issues in several Pull requests to make the acceptance process faster.

jrosdahl:
Thanks for the PR.

A change similar to this is already in PR #1133. I think that it would be better to focus on getting that PR ready instead of redoing the changes in another context. Perhaps you can help out with it?

orgads:
This can be closed now.

PR #1164, opened 2022-09-23
fix: fix windows issues in run script

louiscaron:
On windows, the creation of symlinks is not supported so I removed this from the run script.

I also added an extension to the script replacing the compiler because otherwise ccache automatically adds a .exe extension to invoke it.

Signed-off-by: louiscaron <caron_louis@yahoo.fr>

### DIFF
--- a/test/run
+++ b/test/run
@@ -449,9 +449,15 @@ TEST() {
     remove_cache
     rm -rf $ABS_TESTDIR/run $ABS_TESTDIR/run.real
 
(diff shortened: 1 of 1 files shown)


louiscaron:
I have split the windows OS related issues in several Pull requests to make the acceptance process faster.

jrosdahl:
Thanks for the PR.

A change similar to this is already in PR #1133. I think that it would be better to focus on getting that PR ready instead of redoing the changes in another context. Perhaps you can help out with it?

orgads:
This can be closed now.

Issue #1048, opened 2022-04-07
MSVC: support external includes for ccache


venge-vimeo:
### How to reproduce ###

1. Create empty 1.cpp
2. `CCACHE_DEBUG= ccache cl /I"c:\\includes" -external:I"c:\\external_includes" -c 1.cpp`
3. Open 1.obj.ccache-input-text

### Actual behavior ###

/external flag become enabled by default in MSVS 2019 16.10 (https://docs.microsoft.com/en-us/cpp/build/reference/external-external-headers-diagnostics?view=msvc-170)

CMake use this flag for external libraries, i.e. Qt

`-external` include directories are not processed as include directory for ccache, i.e. path is not resolved and `-external` arguments reordered

1) c:\\external_includes
2) c:\\includes

### Expected behavior ###

Includes and external includes should appear in declared order, i.e.

1) c:\\includes
2) c:\\external_includes

### Environment ###

ccache version 4.6

Windows 10, Visual Studio 2022 (msvc 19.31.31105), CMake 3.22

PR #1165, opened 2022-09-26
feat: add support for -external:I for msvc

res0nance:
This should fix the issue as described in https://github.com/ccache/ccache/issues/1048 this has been tested with the repro steps and its found that the include ordering is correct with respect to each other.


jrosdahl:
Thanks.

PR #1166, opened 2022-09-26
fix: spaces between target and colon in dependency files

louiscaron:
Support dependency files that are generated with spaces between the target and the colon sign.  This is the case for the tasking compiler but since this is supported by definition of the target, it could be generated theoretically by any compiler

### DIFF
--- a/src/Depfile.cpp
+++ b/src/Depfile.cpp
@@ -146,6 +146,30 @@ tokenize(std::string_view file_content)
 {
   // A dependency file uses Makefile syntax. This is not perfect parser but
   // should be enough for parsing a regular dependency file.
+  // enhancement:
+  // - space between target and colon
+  // - no space between colon and first pre-requisite
+  // the later is pretty complex because of the windows paths which are
+  // identical to a target-colon-prerequisite without spaces (e.g. cat:/meow vs.
+  // c:/meow) here are the tests on windows gnu make 4.3 how it handles this:
+  //  + cat:/meow -> sees "cat" and "/meow"
+  //  + cat:\meow -> sees "cat" and "\meow"
+  //  + cat:\ meow -> sees "cat" and " meow"
+  //  + cat:c:/meow -> sees "cat" and "c:/meow"
+  //  + cat:c:\meow -> sees "cat" and "c:\meow"
+  //  + cat:c: -> target pattern contains no '%'.  Stop.
+  //  + cat:c:\ -> target pattern contains no '%'.  Stop.
+  //  + cat:c:/ -> sees "cat" and "c:/"
+  //  + cat:c:meow -> target pattern contains no '%'.  Stop.
+  //  + c:c:/meow -> sees "c" and "c:/meow"
+  //  + c:c:\meow -> sees "c" and "c:\meow"
+  //  + c:z:\meow -> sees "c" and "z:\meow"
+  //  + c:cd:\meow -> target pattern contains no '%'.  Stop.
+
+  // the logic for a windows path is:
+  //  - if there is a colon, if the previous token is 1 char long
+  //    and that the following char is a slash (fw or bw), then it is
+  //    a windows path
 
   std::vector<std::string> result;
   const size_t length = file_content.size();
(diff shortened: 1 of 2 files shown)


jrosdahl:
Thanks for the PR. Please consider adding a unit test for this to `unittest/test_Depfile.cpp`.

louiscaron:
@jrosdahl : I added the unit tests as you suggested.  I discovered that there were also two other formatting possibilities that are supported by gnu make and that could be generated in dependency files so I added the support for those as well (quoted filenames and no space between colon and first pre-requisite)

louiscaron:
@jrosdahl : I hope this makes sense and that you will accept this change

louiscaron:
@jrosdahl : finally, the code code format is fixed. thanks

jrosdahl:
> @jrosdahl : I added the unit tests as you suggested.

Thanks.

> I discovered that there were also two other formatting possibilities that are supported by gnu make and that could be generated in dependency files so I added the support for those as well (quoted filenames and no space between colon and first pre-requisite)

Handling any amount of spaces around the colon looks good.

However, where did you discover that `"` should be parsed specially? GNU make doesn't do that, and neither GCC nor Clang quotes `"` in filenames when writing the dependency file.

src/Depfile.cpp
```diff
@@ -199,6 +214,22 @@ tokenize(std::string_view file_content)
...
+        if (next == '"') {
+          // skip the last quote
+          ++p;
+          break;
+        } else {
+          token.push_back(next);
+          ++p;
+        }
+      }
+      continue;
```
jrosdahl, line 232: This doesn't match the behavior of GNU make, at least not version 4.3:

```console
% echo '"foo1 foo2": "foo3"' >test.mk
% make -np -f test.mk | grep foo     
make: *** No rule to make target '"foo3"', needed by '"foo1'.  Stop.
.DEFAULT_GOAL := "foo1
"foo3":
foo2": "foo3"
"foo1: "foo3"
```

Note that `"foo1 foo2"` is parsed as two different tokens (`"foo1` and `foo2"`), and those tokens are interpreted verbatim when evaluating the makefile.

jrosdahl (changes requested):

src/Depfile.cpp
louiscaron: All my problems are coming from the tasking support rebase and enhancement (to support the depend mode which provides an extra 30% improvment).

The tasking compiler generates the dependency file with the following changes compared to a standard gcc compiler:
- extra space between target and colon -> supported by gnu make (on the versions I tested 3.82 and 4.3)
- no space between colon and pre-requisite -> supported by gnu make (on the versions I tested 3.82 and 4.3)
- quotes around the file names -> you highlighted that this is not supported by make so I tried to understand why my build environment which is make-based still works, and I discovered that there is a post processing of the generated dependency files (bouhh!! ugly).  

I tried to enhance as much as possible the core of the ccache tool when doing my tests and I thought that all those changes were meaningful for a broader support even for the quotes because they are a standard way of specifying names without escaping characters.

Of course, this is your tool, you can decide to accept this or not.  Could you just tell me what you decide (make the quoting part of the tasking support changes or part of the core) and I will comply to it?

thanks

jrosdahl: > The tasking compiler generates the dependency file with the following changes compared to a standard gcc compiler:
> 
> * extra space between target and colon -> supported by gnu make (on the versions I tested 3.82 and 4.3)
> * no space between colon and pre-requisite -> supported by gnu make (on the versions I tested 3.82 and 4.3)

That's fine (as mentioned in https://github.com/ccache/ccache/pull/1166#issuecomment-1261313008).

> I thought that all those changes were meaningful for a broader support even for the quotes because they are a standard way of specifying names without escaping characters.

Right, but since that's not how quotes work in makefiles (which a `.d` file is), we can't add such parsing to the common code path since that would introduce bugs for the common GCC/Clang case[^1].

[^1]:  For example, parsing a dependency file generated by  `gcc -MD -c 'exa"mple.c'` will be interpreted incorrectly.

louiscaron: Okay, I have removed that change and moved it to the tasking specific behavior

louiscaron: oops, sorry, please do not merge that because now we have an issue with the C: which is seen as a : delimiter even though there are quotes (I removed the quotes mechanism)

louiscaron:
okay, at this point, it looks better, I added depend mode format parsing tests specific to windows, to make sure that I do not break the behavior on windows.

jrosdahl (approved):

jrosdahl:
> okay, at this point, it looks better, I added depend mode format parsing tests specific to windows, to make sure that I do not break the behavior on windows.

Thanks for the thorough investigation.

Issue #1002, opened 2022-01-27
Ignore -frandom-seed compiler option


ivan-tkatchev:
GCC and Clang have a `-frandom-seed` compiler option. The idea is to improve reproducibility by allowing the build system to specify random seeds for symbols that are randomly generated by the compiler.

ccache should ignore `-frandom-seed` in compiler options when looking for cache hits. Currently two compiler invocation command lines that are strictly identical except for the random seed will cause a cache miss, which is incorrect.

See also https://github.com/NixOS/nixpkgs/issues/109033 for context.

jrosdahl:
Thanks for the bug report. However, this is not a bug since ccache works as designed.

I don't think that it sounds like a good idea to let ccache ignore `-frandom-seed` by default since ccache then not return the same things as a call to the compiler would without ccache. What could be done, though, is to make that behavior opt-in by adding a new [sloppiness value](https://ccache.dev/manual/4.5.1.html#config_sloppiness) called `random_seed`.

ivan-tkatchev:
Adding a sloppiness value seems like a great solution.

PR #1168, opened 2022-09-28
feat: Add sloppiness for -frandom-seed

res0nance:
This allows the use different -frandom-seed
values without causing cache misses.

Should fix https://github.com/ccache/ccache/issues/1002


jrosdahl:
Thanks. Looks reasonable, except that `doc/MANUAL.adoc` needs to be updated as well.

jrosdahl:
Thanks.

PR #1169, opened 2022-09-30
build: Use pkgconfig for zstd search

neheb:
pkgconfig is a more reliable way to search for libzstd, especially when not built with the non default cmake build.

Signed-off-by: Rosen Penev <rosenp@gmail.com>

### DIFF
--- a/cmake/Findzstd.cmake
+++ b/cmake/Findzstd.cmake
@@ -4,8 +4,15 @@ endif()
 
 set(zstd_FOUND FALSE)
 
-find_library(ZSTD_LIBRARY zstd)
-find_path(ZSTD_INCLUDE_DIR zstd.h)
+find_package(PkgConfig)
+if(PKG_CONFIG_FOUND)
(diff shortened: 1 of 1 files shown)


jrosdahl:
Doesn't sound unreasonable, I guess.

Should the same thing be done for `cmake/Findredis.cmake`?

@cristianadam: Do you have any opinion on this?

cristianadam:
It looks good. It adds hints to `find_library` and `find_path` from pkgconfig. The previous functionality should not be affected.

I remember a ticket where `vcpkg` was used to provide the 3rd party dependencies, and AFAIK vcpkg is using cmake to compile their packages.

I think `cmake/Findredis.cmake` can also be updated with the same functionality (I assume it has a pkgconfig configuration on *nix)

neheb:
Redis doesn't seem to use pkgconfig. I can't find such a file.

neheb:
Updated to match Findhiredis.cmake

PR #1170, opened 2022-10-01
fix: Capture MSVC stdout/stderr when running from Visual Studio

orgads:
When VS_UNICODE_OUTPUT is set, CL sends its (unicode) output to VS using a side-channel. This prevents the option to capture the output.

Unset this variable to enable capturing the output.

### DIFF
--- a/src/Util.cpp
+++ b/src/Util.cpp
@@ -1364,6 +1364,8 @@ unsetenv(const std::string& name)
 {
 #ifdef HAVE_UNSETENV
   ::unsetenv(name.c_str());
+#elif defined(_WIN32)
+  SetEnvironmentVariable(name.c_str(), NULL);
 #else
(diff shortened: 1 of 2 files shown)


orgads:
See also https://github.com/ninja-build/ninja/issues/824 and https://cmake.org/pipermail/cmake/2011-April/044064.html

orgads:
@jrosdahl Would you prefer moving it somewhere else, to be done only once, and only for MSVC?

orgads:
> @jrosdahl Would you prefer moving it somewhere else, to be done only once, and only for MSVC?

Done :)

jrosdahl:
> > @jrosdahl Would you prefer moving it somewhere else, to be done only once, and only for MSVC?
> 
> Done :)

:+1:

PR #1171, opened 2022-10-01
fix: Fix implementation of unsetenv on Windows

orgads:
`putenv("FOO")` doesn't really unset it. `putnev("FOO=")` sets an empty value, which might be good for most cases, but for really unsetting we need `SetEnvironmentVariable("FOO", NULL)`.


jrosdahl:
Thanks.

PR #1172, opened 2022-10-02
fix: Remove usage of deprecated codecvt header

orgads:
Based on https://stackoverflow.com/a/69410299/764870.

### DIFF
--- a/src/util/file.cpp
+++ b/src/util/file.cpp
@@ -21,6 +21,7 @@
 #include <Fd.hpp>
 #include <Logging.hpp>
 #include <Stat.hpp>
+#include <Win32Util.hpp>
 #include <fmtmacros.hpp>
 #include <util/Bytes.hpp>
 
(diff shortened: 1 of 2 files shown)


src/util/file.cpp
```diff
@@ -171,11 +170,32 @@ read_file(const std::string& path, size_t size_hint)
...
+      const int size = WideCharToMultiByte(CP_UTF8,
+                                           0,
+                                           result_as_u16.c_str(),
+                                           int(result_as_u16.size()),
+                                           nullptr,
+                                           0,
+                                           nullptr,
+                                           nullptr);
+      if (size <= 0)
+        return {};
```
jrosdahl, line 188: I suggest returning an error instead of silently returning an empty string. Perhaps something like this?

```suggestion
      if (size <= 0) {
        return nonstd::make_unexpected(
          FMT("Failed to convert {} from UTF-16LE to UTF-8: {}",
              path,
              Win32Util::error_message(size)));
      }
```

(Untested.)

A test case for this in `unittest/test_util_file.cpp` would be appreciated.

orgads: Done

jrosdahl:
Thanks.

PR #1173, opened 2022-10-02
fix: Store output without line-ending conversion

orgads:
When the cached data is read, the output to fd is binary (Util::send_to_fd), so in order to maintain the original line endings, the output must be stored as binary too.

### DIFF
--- a/src/ccache.cpp
+++ b/src/ccache.cpp
@@ -732,21 +732,24 @@ do_execute(Context& ctx, Args& args, const bool capture_stdout = true)
 
   std::string stdout_data;
   if (capture_stdout) {
-    auto stdout_data_result = util::read_file<std::string>(tmp_stdout.path);
+    auto stdout_data_result = util::read_file<util::Bytes>(tmp_stdout.path);
     if (!stdout_data_result) {
       // The stdout file was removed - cleanup in progress? Better bail out.
       return nonstd::make_unexpected(Statistic::missing_cache_file);
     }
-    stdout_data = std::move(*stdout_data_result);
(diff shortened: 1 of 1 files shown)


orgads:
With this change, unittest fails on MinGW environments. See https://github.com/ccache/ccache/actions/runs/3167606615/jobs/5158225686

Without this change, Clang output is not recovered correctly. See https://github.com/ccache/ccache/actions/runs/3167309740/jobs/5157710197

@jrosdahl can you please advise?

orgads:
I'm really unsure if this is the right way to convert `util::Bytes` to `std::string`...

jrosdahl:
> Yes, reading `tmp_stdout` and `tmp_stderr` as raw data (not interpreting line endings) sounds correct. :+1: Ideally, the `DoExecuteResult` fields should also store raw uninterpreted data. I'll see what I can do

Done in e9ddf6800eaed6ce45bcb01f976995ef423c6ae2.

orgads:
Thank you. Now that complicates #1158 :)

I used `strip_includes` that returns either the original string, or a new string. How can I do that with `string_view`?

jrosdahl:
> Thank you. Now that complicates #1158 :)
> 
> I used `strip_includes` that returns either the original string, or a new string. How can I do that with `string_view`?

Oops. Looks like you could use the same approach as I did in [rewrite_stdout_from_compiler](https://github.com/ccache/ccache/blob/e9ddf6800eaed6ce45bcb01f976995ef423c6ae2/src/ccache.cpp#L917-L918). Would that work out?

orgads:
A bit ugly, but it should be ok.

PR #1174, opened 2022-10-02
feat: add depend mode tests

louiscaron:
The first test exercices a relative path only identical generation.

The second test exhibits a known limitation when the same test case is done with an absolute path (depend file is wrong).


louiscaron:
I needed to understand whether this was an expected behavior or if this was an issue with my test on the tasking compiler.  I assumed that adding more tests is always a good thing, so instead of losing that work, I suggest that it gets merged.

jrosdahl:
Sure, thanks.

PR #1176, opened 2022-10-05
feat: Support auto depend mode for MSVC without /showIncludes

orgads:
If MSVC is executed *without* /showIncludes, and ccache is configured with depend mode, add /showIncludes, and strip the extra output from stdout.

Reviving #1158 which was reverted.

Depends on #992.


orgads:
Now with unit tests and doc.

orgads:
Rebased

jrosdahl:
Thanks.

PR #1178, opened 2022-10-12
build: Add headers to CMake project files

orgads:
Useful for listing them in the IDE project tree, for IDEs that use CMake file api (like Qt Creator).

### DIFF
--- a/src/storage/CMakeLists.txt
+++ b/src/storage/CMakeLists.txt
@@ -6,4 +6,7 @@ set(
   Storage.cpp
 )
 
+file(GLOB headers *.hpp)
+list(APPEND sources ${headers})
+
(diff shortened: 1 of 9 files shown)


src/storage/CMakeLists.txt
```diff
@@ -4,6 +4,8 @@ add_subdirectory(remote)
 set(
   sources
   Storage.cpp
+  Storage.hpp
+  types.hpp
```
jrosdahl, line 8: Would it make sense to glob these as well?

orgads: Sure.

orgads: Done

jrosdahl:
Thanks.

PR #1180, opened 2022-10-14
build: Remove Ubuntu 18.04 image from ci

orgads:
It is deprecated.

See https://github.blog/changelog/2022-08-09-github-actions-the-ubuntu-18-04-actions-runner-image-is-being-deprecated-and-will-be-removed-by-12-1-22/

Add GCC 12 on Ubuntu 22.04.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -25,17 +25,9 @@ jobs:
       fail-fast: false
       matrix:
         config:
-          - os: ubuntu-18.04
-            compiler: gcc
-            version: "9"
-
           - os: ubuntu-20.04
             compiler: gcc
             version: "10"
-
-          - os: ubuntu-18.04
-            compiler: clang
-            version: "8"
 
           - os: ubuntu-20.04
             compiler: clang
@@ -57,6 +49,10 @@ jobs:
             compiler: gcc
             version: "11"
 
+          - os: ubuntu-22.04
+            compiler: gcc
+            version: "12"
+
           - os: macOS-11
             compiler: xcode
             version: "11.7"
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yaml
```diff
@@ -70,13 +62,17 @@ jobs:
           if [ "${{ runner.os }}" = "Linux" ]; then
             sudo apt-get update
 
-            packages="elfutils libhiredis-dev libzstd-dev ninja-build pkg-config python3 redis-server redis-tools"
             # Install ld.gold (binutils) and ld.lld (lld) on different runs.
-            if [ "${{ matrix.config.os }}" = "ubuntu-18.04" ]; then
-              sudo apt-get install -y $packages binutils
-            else
-              sudo apt-get install -y $packages lld
-            fi
```
jrosdahl, line 79: binutils was selected for ubuntu-18.04 not because lld doesn't exist in Ubuntu 18.04 but to have a job that uses ld.gold instead of ld.lld (see the comment on line 74 above). I suggest keeping this test and making ubuntu-20.04 and ubuntu-22.04 run different linkers.

orgads: Done

orgads:
We can also change Clang >=11 to run on Ubuntu 22.04, and add 13 and 14. What do you say?

Is there a good reason to test *all* clang versions? Why not test only first and last (9 and 14), and maybe one more between?

jrosdahl:
Thanks.

jrosdahl:
> We can also change Clang >=11 to run on Ubuntu 22.04, and add 13 and 14. What do you say?

Sure, sounds good to me.

> Is there a good reason to test _all_ clang versions? Why not test only first and last (9 and 14), and maybe one more between?

I don't have a strong opinion about that. @AlexanderLanin made it that way in #690 and I think that it has some merit, but it's on the other hand likely consuming more resources without much added value.

AlexanderLanin:
The goal of #690 was to test fewer combinations than before. I tend to agree that testing with half a dozen clang versions likely has no benefit. 
If there are any ifdef for supported features however, then I would test with versions where such features were first introduced.

PR #1190, opened 2022-10-19
build: Fix compilation with GCC 8

orgads:
See https://discourse.cmake.org/t/correct-way-to-link-std-filesystem-with-gcc-8/4121


jrosdahl:
Thanks.

PR #1199, opened 2022-10-27
fix: Process the argument following a -Xarch argument

### LINKED ISSUE #1198: Compiler argument following a -Xarch argument is always ignored.
### How to reproduce ###

1. Construct a compiler command-line that uses a `-Xarch_*` parameter prior to a parameter important to `ccache`; e.g. `-Xarch_x86_64 -include/path/to/some.cxx.pch`
2. Execute a compilation prefixed by `ccache`, using this command-line.

### Actual behavior ###
`ccache` ignores the parameter after `-Xarch_*`. (I.e. it treats `-Xarch_*` similarly to `--ccache-skip`.) In some cases, this causes it to read incorrect data from the cache, when instead the compiler should be executed again. (For example, if the argument following was to include a PCH which has been changed.)

### Expected behavior ###
`ccache` should unconditionally parse the parameter after an `-Xarch_*`, because `ccache` already validates that all `-Xarch_*` architectures match each other as well as the architecture specified with `-arch`. Even if the parameter following the `-Xarch_*` parameter would have been ignored by the compiler, it is generally safer that `ccache` interpret it than to ignore it completely.

### Environment ###

ccache version master.c33609aa (current git master)
ccache release versions 4.7.1 and 4.6.3 are also susceptible. I suspect all versions supporting `-Xarch` have this behavior.

Compiler: Apple clang 13.0.0
Operating system: MacOS 12.2.1
Architecture: arm64 (M1) cross-compiling to x86_64
Toolchain: CMake 3.22.2, Ninja 1.10.2

This is most severe on MacOS, especially in combination with CMake and precompiled headers. CMake unconditionally prefixes a `-include/path/to/file.pch` with a `-Xarch` on this platform, which causes `ccache` to ignore any changes to the PCH and return a false cache hit, leading to a later linking failure or corrupted binary. `-Xarch` could also in theory accidentally smuggle in compiler opts that are unsupported by `ccache`.

### Patch ###

I'm hoping to submit a PR to resolve this soon, once I write a unit test. This can be easily patched by removing the following 2 lines in `src/argprocessing.cpp`:

```patch
diff --git a/src/argprocessing.cpp b/src/argprocessing.cpp
index b095b7a7..e26ca590 100644
--- a/src/argprocessing.cpp
+++ b/src/argprocessing.cpp
@@ -385,8 +385,6 @@ process_option_arg(const Context& ctx,
       return Statistic::unsupported_compiler_option;
     }
     state.common_args.push_back(args[i]);
-    state.common_args.push_back(args[i + 1]);
-    ++i;
     return Statistic::none;
   }
```

rblx-kbuck:
Since there are already checks enforcing that all -Xarch arguments match each other and -arch, we can assume that the compiler would also interpret the following argument, so ccache should interpret it too.

Fixes #1198.

### DIFF
--- a/src/argprocessing.cpp
+++ b/src/argprocessing.cpp
@@ -385,8 +385,6 @@ process_option_arg(const Context& ctx,
       return Statistic::unsupported_compiler_option;
     }
     state.common_args.push_back(args[i]);
-    state.common_args.push_back(args[i + 1]);
-    ++i;
     return Statistic::none;
   }
 
--- a/test/suites/multi_arch.bash
+++ b/test/suites/multi_arch.bash
@@ -47,6 +47,10 @@ SUITE_multi_arch() {
     expect_stat direct_cache_hit 3
     expect_stat cache_miss 4
 
+    # The parameter following -Xarch should be processed.
+    $CCACHE_COMPILE -arch x86_64 -Xarch_x86_64 -analyze -I. -c test1.c
+    expect_stat unsupported_compiler_option 1
+
     # -------------------------------------------------------------------------
     TEST "cache hit, preprocessor mode"
 


jrosdahl:
Thanks.

PR #1206, opened 2022-11-02
Add possibility of setting remote subdirs levels

afbjorklund:
Make it possible to use 1024 subdirectories / buckets,
by increasing levels = 1 to levels = 2 to get 256 x 256.
    
Each level after that will increase by another 1024x,
since it will use base32 digits instead of base16 digits.

----

When you have a cache directory, with more than 256k files.
Then you might want to to divide them into 1024 "buckets" :

levels=1
`storage/00/0b8psj1hu4710bvein42a6lp97qn5dc`

levels=2
`storage/00/0b/8psj1hu4710bvein42a6lp97qn5dc`

This is similar to the local cache levels, but uses 2 digits each.
There is probably no real reason to go above subdir level 2 ?

Most of the code "borrowed" from the local cache.

### DIFF
--- a/doc/MANUAL.adoc
+++ b/doc/MANUAL.adoc
@@ -1158,6 +1158,8 @@ Optional attributes:
 --
 +
 The default is *subdirs*.
+* *levels*: The number of levels of subdirectories to use for *subdirs*. The
+  default is 1.
 * *umask*: This attribute (an octal integer) overrides the umask to use for
   files and directories in the cache directory.
 * *update-mtime*: If *true*, update the modification time (mtime) of cache
@@ -1209,6 +1211,8 @@ values.
 --
 +
 The default is *subdirs*.
+* *levels*: The number of levels of subdirectories to use for *subdirs*. The
+  default is 1.
 * *operation-timeout*: Timeout (in ms) for HTTP requests. The default is 10000.
 
 
--- a/misc/convert-levels
+++ b/misc/convert-levels
@@ -1,0 +1,82 @@
+#!/usr/bin/env python3
+
+# This script converts the number of levels in remote storage.
+
+import os
+import sys
+
+import progress.bar
+import humanize
+
+levels = 2
+digits = 2
+
+storage = "."
+filelist = []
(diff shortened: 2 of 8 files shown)


jrosdahl:
Thanks for the PR.

When I implemented support for `layout=subdirs`, I deliberately didn't make the levels configurable. I just chose a layout that I thought would work for all use cases. I'm not aware of any reasonable use case where one would want to use more than 256 subdirectories, so I would like to keep things simple unless there is a compelling use case for making it more configurable.

Note that non-ancient file systems have no problem dealing with directories with hundreds of thousands of files. The old `CCACHE_NLEVELS` setting might have made sense for file systems available 20 years ago, but for modern file systems I don't think that it makes sense to have it configurable – a typical user does not understand the trade-offs and either just relies on the default or chooses a too deep structure "to make things fast". For instance, I know that somebody set `CCACHE_LEVELS=5` and then complained that cache cleanup took a very long time. That is because using 4<sup>5</sup>=1048576 subdirectories means that there will only be a couple of files in each directory even for very large caches and that is simply suboptimal. Using `levels=2` with your suggestion means 256<sup>2</sup>=65536 subdirectories, and that would also be suboptimal unless you have like 10 billion files in your cache, or maybe not even then.

As you probably know, I removed the `CCACHE_NLEVELS` setting in ccache 4.0. I debated with myself whether to just always use 256 subdirectories, but I ended up choosing 2-4 levels dynamically in order to keep supporting users who still use slow legacy file systems, if any. However, since ccache remote storage is a new use case, I think that we can just say that you can't use large remote caches on legacy file systems. Does this make sense to you?

So, does your suggestion originate from a problem that you want to solve? If so, could you describe the problem?

I can also note that if you want to increase the number of buckets for some reason you can use the sharding feature.

afbjorklund:
> So, does your suggestion originate from a problem that you want to solve? If so, could you describe the problem?

Some of the nfs subdirectories had over 10,000 files but if that is not a problem then we will just go with 256 subdirs.

PR #1211, opened 2022-11-04
Add proof-of-concept RPC storage backend

afbjorklund:
This allows calling _another_ storage backend over rpc,
communicating with a persistent "rpc-server" backend.
    
It uses [rpclib](http://rpclib.net/), to send the backend request using [msgpack](https://msgpack.org/).
Compared with grpc, this is simpler and self-contained.


afbjorklund:
I made some experiments to enable this backend by default, but was a bit scared of the size...

```
264K	liblibhiredis_static.a
16M	liblibrpc_static.a
```

So it is currently OFF by default.

PR #1221, opened 2022-11-12
fix: Do not create /run directory on systems that don't have it when …

olegsidorkin:
…run as root

Some systems (e.g. FreeBSD) don't have /run directory. So when ccache is run as root it creates /run/... and uses it as tmp, that is not very good since / is not supposed to be written during the build. Fix this behavior by checking that /run is present before creating any subfolders.

Alternative idea was to have some kind of knob in config to move tmp into the cache dir (e.g. ```temporary_dir=${CCACHE_DIR}/tmp```} but I think having things working by default is a better idea.
This knpd still requires some patching because otherwise ccache will create /run on every call to default_temporary_dir() that is called, for example, from LocalStorage::finalize().

### DIFF
--- a/src/Config.cpp
+++ b/src/Config.cpp
@@ -1115,9 +1115,11 @@ Config::default_temporary_dir() const
 {
   static const std::string run_user_tmp_dir = [] {
 #ifdef HAVE_GETEUID
-    auto dir = FMT("/run/user/{}/ccache-tmp", geteuid());
-    if (Util::create_dir(dir) && access(dir.c_str(), W_OK) == 0) {
-      return dir;
+    if (Stat::stat("/run").is_directory()) {
+      auto dir = FMT("/run/user/{}/ccache-tmp", geteuid());
+      if (Util::create_dir(dir) && access(dir.c_str(), W_OK) == 0) {
+        return dir;
+      }
     }
 #endif
     return std::string();


afbjorklund:
It is a systemd thing, it creates one tmpfs per user. Previously, `/tmp` (or TMPDIR) was used also on linux.

https://www.freedesktop.org/software/systemd/man/user@.service.html

https://www.freedesktop.org/software/systemd/man/pam_systemd.html

It was introduced with the inode cache: 213d9883a0c3944749f8512eeb784d8572ad8d91 (not sure how well that cache translates to FreeBSD?)

olegsidorkin:
> It is a systemd thing, it creates one tmpfs per user. Previously, `/tmp` (or TMPDIR) was used also on linux.
FreeBSD systems use /var/run/user for this purpose.

> It was introduced with the inode cache: [213d988](https://github.com/ccache/ccache/commit/213d9883a0c3944749f8512eeb784d8572ad8d91) (not sure how well that cache translates to FreeBSD?)

Yep, but initially there was a ```stat()``` call, so it didn't create new dirs.

On FreeBSD inode cache successfully finds all necessary structures and is enabled according to the ```config.h``` content. However ufs and zfs that are used on freebsd are not on the list of supported FSs so I'm going to check it.
If there are some already existing tests they will help a lot.

afbjorklund:
I'm not sure why it even bothers creating them, since they are not really "directories" but "mounted file systems"

```
TARGET         SOURCE FSTYPE OPTIONS
/run/user/1000 tmpfs  tmpfs  rw,nosuid,nodev,relatime,size=1609880k,mode=700,uid=1000,gid=1000
```

But I suppose everyone tries to humor systemd, by creating `/run/user` and `/etc/os-release` where it don't exist ?

afbjorklund:
> Alternative idea was to have some kind of knob in config to move tmp into the cache dir (e.g. `temporary_dir=${CCACHE_DIR}/tmp`} but I think having things working by default is a better idea.

Theoretically it should have used $XDG_RUNTIME_DIR instead of hardcoding it, but I'm with you on defaults.
Better to have it working, by falling back to the cache directory instead of using the (missing) user directory...

jrosdahl:
Thanks.

jrosdahl:
@afbjorklund wrote:

> Theoretically it should have used $XDG_RUNTIME_DIR instead of hardcoding it, but I'm with you on defaults. Better to have it working, by falling back to the cache directory instead of using the (missing) user directory...

Using `$XDG_RUNTIME_DIR` sounds good; I was not aware of it when reviewing #577. :+1: Done in 876509ae8b95f76adebdfa1f2380d75a49f9871d.

afbjorklund:
The main problem is that it might not be universally set, so still need the other one as a fallback anyway.

jrosdahl:
> The main problem is that it might not be universally set, so still need the other one as a fallback anyway.

Which systems create `/run/user/<uid>` but don't set `XDG_RUNTIME_DIR`?

afbjorklund:
For ccache it shouldn't make a difference, but not all systems are using systemd (mostly container-based system left, but)

As far as I know, all machines that use systemd also make sure to set XDG_RUNTIME_DIR as part of the login session ?

So maybe you are right, if available it is set (the environment variable) - otherwise fall back to (some kind of) TMPDIR.

This inode cache is a bit of a special case, I think it is assuming that it is stored on tmpfs - but it probably doesn't matter...

jrosdahl:
> For ccache it shouldn't make a difference, but not all systems are using systemd (mostly container-based system left, but)

I know that. But for a fallback to `/run/user/<uid>` to make sense, there must exist non-systemd systems that create `/run/user/<uid>`.

The XDG spec says _"If $XDG_RUNTIME_DIR is not set applications should fall back to a replacement directory with similar capabilities"_, and ccache already falls back to `<cache_dir>/tmp` which has been the default for a long time before #577.

> This inode cache is a bit of a special case, I think it is assuming that it is stored on tmpfs - but it probably doesn't matter...

No, there is no such assumption.

PR #1222, opened 2022-11-14
fix: posix_fallocate can return EINVAL if filesystem doesn't support …

olegsidorkin:
…it. Fallback to emulation in this case.

E.g. ZFS does so on FreeBSD (haven't tested with ZFS on linux).

This fixes Utill::fallocate unit tests on ZFS.


jrosdahl:
Thanks.

PR #1227, opened 2022-11-23
fix: Fix edge case where a non-temporal identifier is misidentified

erijo:
If a non-temporal identifier, that ends with a temporal macro, happens to be at the end of the buffer with the temporal suffix starting on the avx boundary, then it would be incorrectly classified as a temporal macro. This since the helper function lacks the context to see that the data before the match is something that invalidates the match.


jrosdahl:
Thanks.

PR #1229, opened 2022-12-06
fix: Use spinlocks for inodes cache buckets protection

olegsidorkin:
These changes makes InodeCache working on FreeBSD (corresponging FSs were added to the list of supported filesystems).
Inode cache tests passes now on BSD and building a few large projects doesn't produces anything strange (inode cache file was created and updated during build).

### DIFF
--- a/src/InodeCache.cpp
+++ b/src/InodeCache.cpp
@@ -33,6 +33,7 @@
 
 #include <fcntl.h>
 #include <libgen.h>
+#include <sched.h>
 #include <sys/mman.h>
 #include <unistd.h>
 
@@ -67,7 +68,7 @@ namespace {
 // Note: The key is hashed using the main hash algorithm, so the version number
 // does not need to be incremented if said algorithm is changed (except if the
 // digest size changes since that affects the entry format).
-const uint32_t k_version = 1;
+const uint32_t k_version = 2;
 
 // Note: Increment the version number if constants affecting storage size are
 // changed.
@@ -87,6 +88,8 @@ static_assert(
   "Numeric value is part of key, increment version number if changed.");
 
 const void* MMAP_FAILED = reinterpret_cast<void*>(-1); // NOLINT: Must cast here
+
+const auto MAX_LOCK_DURATION = util::Duration(5);
 
 bool
 fd_is_on_known_to_work_file_system(int fd)
@@ -119,7 +122,10 @@ fd_is_on_known_to_work_file_system(int fd)
       // Is a filesystem you know works with the inode cache missing in this
       // list? Please submit an issue or pull request to the ccache project.
       "apfs",
+      "tmpfs",
+      "ufs",
       "xfs",
+      "zfs",
     };
     if (std::find(known_to_work_filesystems.begin(),
                   known_to_work_filesystems.end(),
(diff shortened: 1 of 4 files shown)


liljenzin (approved):

src/InodeCache.cpp
```diff
@@ -140,6 +144,35 @@ fd_is_on_known_to_work_file_system(int fd)
...
+      pid_t pid = 0;
+      if (owner_pid.load(std::memory_order_relaxed) == 0
+          && owner_pid.compare_exchange_weak(
+            pid, self_pid, std::memory_order_acquire)) {
+        return false;
+      }
+      sched_yield();
+    }
+    pid_t lock_owner = owner_pid.load(std::memory_order_relaxed);
+    if (lock_owner != 0 && getpgid(lock_owner) == -1 && errno == ESRCH) {
```
jrosdahl, line 161: Why `getpgid` here when `getpid` is used to set `m_self_pid`?

liljenzin: Because you can't use getpid to to check if a different pid is still alive while getpgid will error if not?

jrosdahl: I see, it's a variation of `kill(pid, 0)` that I haven't seen before.

olegsidorkin: Process may have not enough rights to send a kill, however getpgid is always allowed.

src/InodeCache.cpp
```diff
@@ -162,7 +195,7 @@ struct InodeCache::Entry
 
 struct InodeCache::Bucket
 {
-  pthread_mutex_t mt;
+  std::atomic<pid_t> owner_pid;
```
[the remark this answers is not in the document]
olegsidorkin: Done, thanks

src/InodeCache.cpp
liljenzin: For the record note that also getpgid may fail with EPERM by spec:
https://pubs.opengroup.org/onlinepubs/9699919799/functions/getpgid.html

But regardless of using getpgid or kill, EPERM is most likely indicating the process is still around and then it should be the right decision to not remove the lock since that assumption can't cause false positives even if wrong.

src/InodeCache.cpp
```diff
@@ -258,40 +306,21 @@ InodeCache::with_bucket(const Digest& key_digest,
...
     }
-    err = pthread_mutex_consistent(&bucket->mt);
-    if (err) {
-      LOG(
-        "Can't consolidate stale mutex at index {}: {}", index, strerror(err));
-      LOG_RAW("Consider removing the inode cache file if the problem persists");
-      return false;
-    }
     LOG("Wiping bucket at index {} because of stale mutex", index);
     memset(bucket->entries, 0, sizeof(Bucket::entries));
```
liljenzin, line 315: Clearing the bucket is no longer safe since the old lock owner, if still alive, could be writing new content to the same bucket simultaneously, thus leaving the bucket in undefined state. So instead we need to wipe the entire cache file by unmapping, unlinking and creating a new one. See InodeCache::drop() and InodeCache::initialize().

olegsidorkin: Done

liljenzin: Looks like this conversation is somehow still marked as unresolved, but I can't figure out how to resolve it since it got obsoleted by the force push.

olegsidorkin: Resolved it (I didn't resolve it immediately to check that implementation is right).

jrosdahl:
Thanks.

PR #1233, opened 2022-12-21
args: do not escape backslashes on msvc

### LINKED ISSUE #1216: MSVC: escaped backslash in rsp file is processed incorrectly
We have an issue that is somewhat related to #1072, as far as I can tell: we have a preprocessor definition that contains an escaped backslash, which ends up being passed to `cl.exe`as a non-escaped backslash, making our compilation fail. It works correctly when building with `ninja` but not when building with Visual Studio, I assume it is due to the use of `.rsp` files? I made a minimal repro case similar to the one in #1072.

### How to reproduce ###

`a.cpp`
```
const char c = DIRSEP;
```
`a.rsp`
```
"-DDIRSEP='\\'"
```
Executing
1. `cl /c a.cpp @a.rsp`
2. `ccache cl /c a.cpp @a.rsp`

### Actual behavior ###

1. compiles without error
2. fails compiling with the following error:

> a.cpp(1): error C2001: newline in constant
> a.cpp(2): fatal error C1004: unexpected end-of-file found

because `ccache` passes `-DDIRSEP='\'` to `cl.exe` instead of `-DDIRSEP='\\'`

### Expected behavior ###

Both compile without error

### Environment ###

ccache version 4.6.1, 4.7.3
VS 2019

res0nance:
Fixes #1216 , I tested the repro example provided and it should in fact compile. As such it seems that escaping backslashes is wrong all the other bug reports related to rsp parsing seems to only indicate issues with `"` whose escape behaviour remains.

Original fix #1071 did not indicate as to why backslashes should be escaped. As such I think its appropriate to revert that behaviour.

### DIFF
--- a/src/Args.cpp
+++ b/src/Args.cpp
@@ -77,7 +77,7 @@ Args::from_atfile(const std::string& filename, AtFileFormat format)
         }
         break;
       case AtFileFormat::msvc:
-        if (*pos != '"' && *pos != '\\') {
+        if (*pos != '"') {
           pos--;
         }
         break;
--- a/unittest/test_Args.cpp
+++ b/unittest/test_Args.cpp
@@ -139,12 +139,12 @@ TEST_CASE("Args::from_atfile")
     CHECK(args[6] == "seve\nth");
   }
 
-  SUBCASE("Only escape double quote and backslash in alternate format")
+  SUBCASE("Only escape double quote in alternate format")
   {
-    util::write_file("at_file", "\"\\\"\\a\\ \\\\\\ \\b\\\"\"\\");
+    util::write_file("at_file", "\"\\\"\\a\\ \\b\\\"\"\\");
     args = *Args::from_atfile("at_file", Args::AtFileFormat::msvc);
     CHECK(args.size() == 1);
-    CHECK(args[0] == "\"\\a\\ \\\\ \\b\"\\");
+    CHECK(args[0] == "\"\\a\\ \\b\"\\");
   }
 
   SUBCASE("Ignore single quote in alternate format")
@@ -154,6 +154,14 @@ TEST_CASE("Args::from_atfile")
     CHECK(args.size() == 2);
     CHECK(args[0] == "'a");
     CHECK(args[1] == "b'");
+  }
+
+  SUBCASE("Do not escape backslash in alternate format")
+  {
+    util::write_file("at_file", "\"-DDIRSEP='\\\\'\"");
+    args = *Args::from_atfile("at_file", Args::AtFileFormat::msvc);
+    CHECK(args.size() == 1);
+    CHECK(args[0] == "-DDIRSEP='\\\\'");
   }
 }
 


jrosdahl:
Thanks.

@jacobly0: Please comment if you believe that backslashes _should_ be escaped.

jacobly0:
The issue, as I recall, was that an argument that ends with a backslash, such as a path to a directory, will escape the following space, which causes it to get merged with the following argument, but I do not remember what compiler this happened with so I would need to do some testing.

jrosdahl:
>However, I do not remember which compiler this happened with, so I would need to do some testing.

That would be appreciated.

jacobly0:
I wrote a script to demonstrate the expansion behavior of at files, and it's clear that my original implementation was missing some details.
```sh
#!/bin/sh
echo def > atfile.c
for cl in cl clang-cl; do
    echo
    echo "Using compiler $cl:"
    while read -r def; do
        if [ -z "$def" ]; then
            echo
            continue
        fi
        echo -n "$def => "
        echo "/w /nologo /E /Ddef=$def atfile.c" > atfile.rsp
        $cl @atfile.rsp 2>&1 | grep --invert '^#\|^atfile\.c$'
    done <<'EOF'
\
\\
'\\'
"\\"
'"\\"'
"'\\'"
'''\\'''
''"\\"''
'"'\\'"'
'""\\""'
"''\\''"
"'"\\"'"
""'\\'""
"""\\"""


\'\'
'\'\''
"\'\'"
''\'\'''
'"\'\'"'
"'\'\''"
""\'\'""
'''\'\''''
''"\'\'"''
'"'\'\''"'
'""\'\'""'
"''\'\'''"
"'"\'\'"'"
""'\'\''""
"""\'\'"""

\"\"
'\"\"'
"\"\""
''\"\"''
'"\"\""'
"'\"\"'"
""\"\"""
'''\"\"'''
''"\"\""''
'"'\"\"'"'
'""\"\"""'
"''\"\"''"
"'"\"\""'"
""'\"\"'""
"""\"\""""
EOF
done
```
```
$ ./atfile.sh

Using compiler cl:
\ => \
\\ => \\
'\\' => '\\'
"\\" => \
'"\\"' => '\'
"'\\'" => '\\'
'''\\''' => '''\\'''
''"\\"'' => ''\''
'"'\\'"' => ''\\''
'""\\""' => '\'
"''\\''" => ''\\''
"'"\\"'" => '\'
""'\\'"" => '\\'
"""\\""" => "\"


\'\' => \'\'
'\'\'' => '\'\''
"\'\'" => \'\'
''\'\''' => ''\'\'''
'"\'\'"' => '\'\''
"'\'\''" => '\'\''
""\'\'"" => \'\'
'''\'\'''' => '''\'\''''
''"\'\'"'' => ''\'\'''
'"'\'\''"' => ''\'\'''
'""\'\'""' => '\'\''
"''\'\'''" => ''\'\'''
"'"\'\'"'" => '\'\''
""'\'\''"" => '\'\''
"""\'\'""" => "\'\'"

\"\" => ""
'\"\"' => '""'
"\"\"" => ""
''\"\"'' => ''""''
'"\"\""' => '""'
"'\"\"'" => '""'
""\"\""" => ""
'''\"\"''' => '''""'''
''"\"\""'' => ''""''
'"'\"\"'"' => ''""''
'""\"\"""' => '""'
"''\"\"''" => ''""''
"'"\"\""'" => '""'
""'\"\"'"" => '""'
"""\"\"""" => """"

Using compiler clang-cl:
\ => \
\\ => \\
'\\' => '\\'
"\\" => \
'"\\"' => '\'
"'\\'" => '\\'
'''\\''' => '''\\'''
''"\\"'' => ''\''
'"'\\'"' => ''\\''
'""\\""' => '\'
"''\\''" => ''\\''
"'"\\"'" => '\'
""'\\'"" => '\\'
"""\\""" => "\"


\'\' => \'\'
'\'\'' => '\'\''
"\'\'" => \'\'
''\'\''' => ''\'\'''
'"\'\'"' => '\'\''
"'\'\''" => '\'\''
""\'\'"" => \'\'
'''\'\'''' => '''\'\''''
''"\'\'"'' => ''\'\'''
'"'\'\''"' => ''\'\'''
'""\'\'""' => '\'\''
"''\'\'''" => ''\'\'''
"'"\'\'"'" => '\'\''
""'\'\''"" => '\'\''
"""\'\'""" => "\'\'"

\"\" => ""
'\"\"' => '""'
"\"\"" => ""
''\"\"'' => ''""''
'"\"\""' => '""'
"'\"\"'" => '""'
""\"\""" => ""
'''\"\"''' => '''""'''
''"\"\""'' => ''""''
'"'\"\"'"' => ''""''
'""\"\"""' => '""'
"''\"\"''" => ''""''
"'"\"\""'" => '""'
""'\"\"'"" => '""'
"""\"\"""" => """"

```
It appears that `\'` is never parsed specially, `\\` expands to `\` only when the innermost matching quotes are double quotes, `\"` always expands to `"`, and there is some double quote eliding that I don't fully understand.

jrosdahl:
@jacobly0 wrote:
> I wrote a script to demonstrate msvc's expansion behavior of at files, and it's clear that my original implementation was missing some details.

Thanks! I created #1247 if anybody wants to follow up on this.

PR #1234, opened 2022-12-21
ci: Add missing compilers

res0nance:
CI is missing a few compilers add those newer compilers.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -58,6 +58,10 @@ jobs:
             version: "13"
 
           - os: ubuntu-22.04
+            compiler: clang
+            version: "14"
+
+          - os: ubuntu-22.04
             compiler: gcc
             version: "11"
 
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -68,6 +68,10 @@ jobs:
           - os: macOS-11
             compiler: xcode
             version: "12.5.1"
+
+          - os: macOS-12
+            compiler: xcode
+            version: "13.4.1"
     steps:
       - name: Install dependencies
         run: |
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -72,6 +72,10 @@ jobs:
           - os: macOS-12
             compiler: xcode
             version: "13.4.1"
+
+          - os: macOS-12
+            compiler: xcode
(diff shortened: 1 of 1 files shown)


jrosdahl:
Adding test jobs for Xcode 14 and Clang 14 makes sense. Thanks.

Do you think that testing all available Clang version has much value? And testing both Xcode 13.2 and 13.4? In #1180 we were leaning towards not testing all Clang versions to save some resources.

res0nance:
> Adding test jobs for Xcode 14 and Clang 14 makes sense. Thanks.
> 
> Do you think that testing all available Clang version has much value? And testing both Xcode 13.2 and 13.4? In #1180 we were leaning towards not testing all Clang versions to save some resources.

Great question, I think it depends on what our goals are for support.

Do we support older clang / gcc releases or should we only officially support a subset.

13.2.1 and 13.4.1 was picked because they have different [clang / llvm versions](https://en.wikipedia.org/wiki/Xcode)

Since we currently support every version we can even download. I simply added every version with a different compiler.

jrosdahl:
Older versions are supported. For example, GCC 4.7 is still supported for run-time usage with ccache (but not for building ccache). There is no official goal for which set of compiler versions to support; in practice all support is on a best effort basis.

I think that it at least doesn't make much sense to test different minor versions. Maybe not even all major versions, especially for Clang which releases major versions twice as often as GCC.

It seems like Xcode 13.2.1 uses Clang 13.0.0 and Xcode 13.4.1 uses Clang 13.1.6. I suggest picking 13.4.1 unless there's something specific with 13.2.1 that we want to test.

res0nance:
> Older versions are supported. For example, GCC 4.7 is still supported for run-time usage with ccache (but not for building ccache). There is no official goal for which set of compiler versions to support; in practice all support is on a best effort basis.
> 
> I think that it at least doesn't make much sense to test different minor versions. Maybe not even all major versions, especially for Clang which releases major versions twice as often as GCC.
> 
> It seems like Xcode 13.2.1 uses Clang 13.0.0 and Xcode 13.4.1 uses Clang 13.1.6. I suggest picking 13.4.1 unless there's something specific with 13.2.1 that we want to test.

Makes sense to me, I've dropped 13.2.1

jrosdahl:
Thanks.

Issue #1038, opened 2022-03-26
Support MSVC variables CL and _CL_


jrosdahl:
The [MSVC CL and \_CL_ variables](https://docs.microsoft.com/en-us/cpp/build/reference/cl-environment-variables?view=msvc-170) may contain compiler options that ccache needs to parse. As of a2988ace17d61dd8bb1342ffa8b500471307ebde, ccache considers those variables too hard and falls back to just running the compiler. It would be good to implement support for parsing those variables instead of bailing out.

See also #1022.

PR #1255, opened 2023-02-24
win: add support for CL and _CL_ flags

res0nance:
Should resolve https://github.com/ccache/ccache/issues/1038

### DIFF
--- a/src/ccache.cpp
+++ b/src/ccache.cpp
@@ -2390,6 +2390,18 @@ do_cache_compilation(Context& ctx)
   // be disabled.
   Util::setenv("CCACHE_DISABLE", "1");
 
+  // If on windows check CL and _CL_ env variables and add them to
+  // the argument list
+  if (ctx.config.is_compiler_group_msvc()) {
+    if (getenv("CL")) {
+      ctx.prepend_args = Args::from_string(getenv("CL"));
+    }
+
+    if (getenv("_CL_")) {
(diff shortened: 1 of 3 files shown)


src/ccache.cpp
```diff
@@ -2390,6 +2390,18 @@ do_cache_compilation(Context& ctx)
...
+  // If on windows check CL and _CL_ env variables and add them to
+  // the argument list
+  if (ctx.config.is_compiler_group_msvc()) {
+    if (getenv("CL")) {
+      ctx.prepend_args = Args::from_string(getenv("CL"));
+    }
+
+    if (getenv("_CL_")) {
+      ctx.append_args = Args::from_string(getenv("_CL_"));
+    }
```
jrosdahl, line 2402: `Args::from_string` just splits on whitespace, which won't work well here since the variables are interpreted differently by the compiler. For example, if you do `set CL=/I"foo bar"` then the value needs to be parsed to `{"/Ifoo bar"}`, not `{"/I\"foo", "bar\""}`.

So part of the task is to find out how the compiler interprets the variables and implement corresponding parse functionality. It might be the same syntax as parsed by `Args::from_atfile(..., Args::AtFileFormat::msvc`, but that needs to be investigated.

res0nance: I think I'll need to rethink how to get this to work correctly,

PR #1260, opened 2023-03-16
Make ccache compile with clang-cl on Windows

tru:
This makes ccache build out of the box with clang-cl on Windows. I haven't done any benchmarks yet since I couldn't find any benchmarking tests in the source that executes on Windows so I don't know if it improves things. But usually clang-cl + LTO is faster than MSVC on Windows.


jrosdahl:
Thanks.

PR #1261, opened 2023-03-17
remove CMake header checks

neheb:
C++17 allows moving to the compiler, speeding compilation up.


neheb:
Didn't know ccache can't handle __has_include. Probably should abandon this.

Issue #1239, opened 2023-01-12
Support overriding MSVC /Z* options 


firewave:
Visual Studio applies compiler options from left to right allowing you to override existing one by appending a different option:
```
4>cl : command line warning D9025: overriding '/Zi' with '/Z7' [D:\a\cppcheck\cppcheck\lib\cppcheck.vcxproj]
```

This can also be achieved by via environment variables by specifying `_CL_`: https://learn.microsoft.com/en-us/cpp/build/reference/cl-environment-variables

Unfortunately `ccache` looks at the options as a whole and so you cannot simply override an option:
```
[2023-01-12T18:38:19.220631 4848 ] Compiler type: msvc
[2023-01-12T18:38:19.220631 4848 ] Compiler option -Zi is unsupported
[2023-01-12T18:38:19.220631 4848 ] Failed; falling back to running the real compiler
```

The same might be true for GNU-like compiler which also apply options from left to right but I have no example for that.

jrosdahl:
I don't think there's a way to implement a reliable generic detection for this, so it likely has to be on a case-by-case basis. And I agree that teaching ccache that the last of `/Z7`, `/Zi` or `/ZI` "wins" (if that's indeed the case) would be useful.

> This can also be achieved by via environment variables by specifying `_CL_`: [learn.microsoft.com/en-us/cpp/build/reference/cl-environment-variables](https://learn.microsoft.com/en-us/cpp/build/reference/cl-environment-variables)

Related: #1038

firewave:
I think just limiting it to specific options makes a lot of sense and should (hopefully) also make it easy to implement as well.

It does "win" - hence the compiler warning `D9025`.

PR #1263, opened 2023-03-20
fix: exclude /Zc flags from MSVC debug parameter (/Z) detection

### LINKED ISSUE #1262: MSVC /Zc: parameters are no longer supported since version 4.8
### How to reproduce ###

Using MSVC and compiling a source file with any `/Zc:<flag>` parameter (C++ conformance modes), e.g. `/Zc:throwingNew`

### Actual behavior ###

ccache complains about this compiler option:
`Compiler option /Zc:throwingNew is unsupported`

### Expected behavior ###

In version 4.7.4 the conformance options worked just fine.

### Environment ###

ccache version 4.8

I figured out that this issue goes away when I revert d099c1804f56e7a357c8c8473cacf186da0d5d3e.

A brief look into the patch revealed that the simple check for `-Z` wrongly matches the conformance flags as well.

srohmen:
This fixes #1262

### DIFF
--- a/src/argprocessing.cpp
+++ b/src/argprocessing.cpp
@@ -610,9 +610,13 @@ process_option_arg(const Context& ctx,
 
   if (config.is_compiler_group_msvc() && !config.is_compiler_group_clang()
       && util::starts_with(arg, "-Z")) {
-    state.last_seen_msvc_z_option = args[i];
-    state.common_args.push_back(args[i]);
-    return Statistic::none;
+    // Exclude other options starting with /Z (/Zc), which are not debug flags
+    const char debug_mode = arg[2];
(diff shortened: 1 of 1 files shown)


jrosdahl:
Looking at the MSVC documention, there are lots of other options that begin with `/Z`, so I don't think that the right fix is to make a special case for `/Zc`. Instead I think we should make a special case for `/Zi`, `/ZI` and `/Z/` and leave other `/Z` options alone. This was the intention when implementing #1239, but I trusted the issue title too much without checking the docs.

tru:
I tested the latest version of this patch on our code base and it worked just fine.

jrosdahl:
Thanks.

Issue #1259, opened 2023-03-15
`clang-cl` and /Zi and /Z7


tru:
Hi,

I know that /Zi is complicated and not supported correctly by ccache, as described in #1040. But using `clang-cl` we could allow `/Zi` because it behaves exactly as `/Z7` and doesn't write a `.pdb` file per object file, see: https://github.com/llvm/llvm-project/blob/main/clang/include/clang/Driver/Options.td#L6884

PR #1266, opened 2023-03-22
Don't treat /Zi as unsupported for clang-cl.

tru:
For MSVC /Zi is unsupported since it writes a additional
.pdb file per each .obj file and it creates some messy
interaction with ccache. But for clang-cl /Zi is actually
treated as /Z7 and only embeds the debug info in the .obj
file so it makes sense to allow this flag when compiling
with clang-cl.


tru:
Fixes #1259

jrosdahl:
Thanks.

Issue #1267, opened 2023-03-23
RFE: suppport llc binary


squeed:
Clang includes the [llc binary](https://llvm.org/docs/CommandGuide/llc.html), which works similar to `clang` itself. A typical invocation might look like

```
llc -march=bpf -mcpu=v2 -filetype=obj -o bpf_xdp.o bpf_xdp.ll
```

which compiles `bpf_xdp.ll` in to `bpf_xdp.o`.

Would adding this be difficult? Or is this a good-first-issue sort of thing?

squeed:
AFAICT the only real issue is that `llc` just has an implicit input file, rather than a `-c` flag.

squeed:
So, I have a hacky patch that works for this; it just removes the requirement for `-c` if the binary is `llc`. How would be the best way to solve this? Add a new CompilerType and branch off of that?

jrosdahl:
Supporting llc sounds nice and I understand that it could be useful for some projects.

I would however prefer not to do it by adding more custom logic to the already messy argument handling code (accidental complexity due to supporting several different compilers without proper abstractions; see also #956).

Some notes from a cursory glance at `llc --help`:

- `-I=<string>` sets some kind of include path. Does ccache need to know about it?
- Using the syntax `-o=obj.o` won't work without adaptation. For instance, `ccache llc -c test.ll -o test.o` followed by `ccache llc -c test.ll -o=test.o` will erroneously produce the file `=obj.o`.
- `--filetype=...` changes name of the implicit output file.
- `--split-dwarf-output=` produces a file that needs to be cached.
- Same thing for `--time-trace-file=`.
- Does `--summary-file=` specify an input or output? If input it needs to be included in the hash. If output it needs to be cached.
- Using `-x=` instead of `-x` won't work. Also, input languages `ir` and `mir` should probably be added.
- Options `--polly-dump-after`, `--polly-dump-after-file`, `--polly-dump-before` and `--polly-dump-before-file` seems to specify output files that need to be cached.

squeed:
Thanks for the comments! Looks like this is a bit more subtle than I initially realized, but seems doable for sure.

I agree that the argument handling code is probably not the best place for this kind of logic. Is there an existing model for how to do complier-specific parsing? (I must confess, C++ is not my strong suit).

It seems like a PR supporting "--opt=val"-style arguments would be a good first step.

tru:
[...]

PR #1268, opened 2023-03-24
feat: support llc compiler, .ll files

squeed:
llc works, essentially, identcally to clang, except that the -c option is implicit. Thus, in the llc case, we cannot require it.

llc is commonly used for generating BPF objects, so it makes sense for ccache to support it, IMHO.

Additionally, indicate that .ll files are assembly. They're actually LLVM IR, but the effect is the same.

Fixes: #1267

Thanks!


squeed:
As was observed in #1267, this isn't quite the right approach. Closing for now.

Issue #1040, opened 2022-03-28
Support for MSVC's Zi


janisozaur:
ccache should support `/Zi`, which seems to be the most prolific switch used for debugging information.

Based on discussion held in https://developercommunity.visualstudio.com/t/absolute-paths-for-associated-pdb-files-are-record/573023 it appears the issue with `/Zi` is the fact that full file paths are being embedded in resulting objects, leading to possible caching issues. That same thread discusses a new switch having been added to MSVC 16.4p1, `/d1trimfile` which is capable of canonicalizing the paths into format that should pose no threat to caches.

I'm interested in trying to solve this issue, but not familiar with ccache code yet and somewhat time-constrained. I will try submitting something within a week or two or admitting defeat.

janisozaur:
Note: I'm not a Windows/MSVC developer, so please let me know if there's something I'm missing or don't understand.

I started thinking about it and reading into "documentation" (the discussion linked in top post) my plan would be something along those lines:
1. Verify if `/Zi` is present
2. Verify `/d1trimfile` is present
3. Verify if `/d1trimfile` and `$CWD`+`/Fd` options share common prefix
  - length of the common prefix must be equal to length of path provided to `/d1trimfile`
4. Trim the common prefix from `$CWD`+`/Fd` path and use remainder as part of the key
5. If any of those checks fails, provide info in debug output about expected/actual values
6. Proceed to cache all generated objects

Does that sound like a valid approach? Are there options already implemented that require specific values for two arguments? Should I abuse any of the existing options https://github.com/ccache/ccache/blob/0b3c5f17de5af6b5f0520295d96eb21c6ebe5a60/src/compopt.cpp#L25-L46 or should I create a new one?

janisozaur:
Small update: still working on this, according to the plan laid out in previous post. I started by taking a closer look at the tests (great stuff) and trying to fake from here. Some unexpected personal thing came up which will delay work on this.

misery:
Maybe you want to look into other solutions. MSVC and debugging info was a big problem in the past.

* https://github.com/frerich/clcache/issues/30
* https://github.com/mozilla/sccache (See Readme)
* https://github.com/mbitsnbites/buildcache/issues/104

janisozaur:
Thanks, these look useful. Just had a quick look at the first one, where some relation between pdb, mspdbsrv clears up some confusion.

janisozaur:
I did get around to testing MSVC a bit.

[...]

PR #1274, opened 2023-04-09
win32: automatically kill compiler subprocess

arBmind:
adds missing feature for usage in IDEs on Windows to abort the child compiler when ccache is killed by any means.

A similar behaviour is present on Posix systems by default.

Ccache has a signal handler that adds some cleanup.
This would require to hook some break functions on win32 as well.
I have no experience with that.

This helps to not wait for 10 Minutes until the compiler finishes naturally.

### DIFF
--- a/src/execute.cpp
+++ b/src/execute.cpp
@@ -105,6 +105,75 @@ win32execute(const char* path,
              int fd_stderr,
              const std::string& temp_dir)
 {
+  BOOL isProcessInJob;
+  DWORD dwCreationFlags = 0;
+  {
+    BOOL jobSuccess =
+      IsProcessInJob(GetCurrentProcess(), NULL, &isProcessInJob);
+    if (!jobSuccess) {
+      DWORD error = GetLastError();
+      LOG("failed to IsProcessInJob: {} ({})",
+          Win32Util::error_message(error),
+          error);
+      return 0;
+    }
+    if (isProcessInJob) {
+      JOBOBJECT_EXTENDED_LIMIT_INFORMATION jobInfo = {};
+      BOOL querySuccess =
+        QueryInformationJobObject(NULL,
+                                  JobObjectExtendedLimitInformation,
+                                  &jobInfo,
+                                  sizeof(jobInfo),
+                                  NULL);
+      if (!querySuccess) {
+        DWORD error = GetLastError();
+        LOG("failed to QueryInformationJobObject: {} ({})",
+            Win32Util::error_message(error),
+            error);
+        return 0;
+      }
+
+      bool isKillActive = (jobInfo.BasicLimitInformation.LimitFlags
+                           & JOB_OBJECT_LIMIT_KILL_ON_JOB_CLOSE);
+      bool allowBeakAway = (jobInfo.BasicLimitInformation.LimitFlags
+                            & JOB_OBJECT_LIMIT_BREAKAWAY_OK);
+      if (!isKillActive && allowBeakAway) {
+        isProcessInJob = false;
+        dwCreationFlags = CREATE_BREAKAWAY_FROM_JOB | CREATE_SUSPENDED;
+      }
+    } else {
+      dwCreationFlags = CREATE_SUSPENDED;
+    }
+  }
+  HANDLE hJob = NULL;
+  if (!isProcessInJob) {
+    hJob = CreateJobObject(NULL, NULL);
+    if (hJob == NULL) {
+      DWORD error = GetLastError();
+      LOG("failed to CreateJobObject: {} ({})",
+          Win32Util::error_message(error),
+          error);
+      return -1;
+    }
+
+    {
+      // Set the job object to terminate all child processes when the parent
+      // process is killed
+      JOBOBJECT_EXTENDED_LIMIT_INFORMATION jobInfo = {};
+      jobInfo.BasicLimitInformation.LimitFlags =
+        JOB_OBJECT_LIMIT_KILL_ON_JOB_CLOSE;
+      BOOL jobSuccess = SetInformationJobObject(
+        hJob, JobObjectExtendedLimitInformation, &jobInfo, sizeof(jobInfo));
+      if (!jobSuccess) {
+        DWORD error = GetLastError();
+        LOG("failed to JobObjectExtendedLimitInformation: {} ({})",
+            Win32Util::error_message(error),
+            error);
+        return -1;
+      }
+    }
+  }
+
   PROCESS_INFORMATION pi;
   memset(&pi, 0x00, sizeof(pi));
 
@@ -161,7 +230,7 @@ win32execute(const char* path,
                            nullptr,
                            nullptr,
                            1,
-                           0,
+                           dwCreationFlags,
                            nullptr,
                            nullptr,
                            &si,
@@ -178,12 +247,27 @@ win32execute(const char* path,
         error);
     return -1;
   }
+  if (hJob) {
+    BOOL assignSuccess = AssignProcessToJobObject(hJob, pi.hProcess);
+    if (!assignSuccess) {
+      TerminateProcess(pi.hProcess, 1);
+
+      DWORD error = GetLastError();
+      LOG("failed to assign process to job object {}: {} ({})",
+          full_path,
+          Win32Util::error_message(error),
+          error);
+      return -1;
+    }
+    ResumeThread(pi.hThread);
+  }
   WaitForSingleObject(pi.hProcess, INFINITE);
 
   DWORD exitcode;
   GetExitCodeProcess(pi.hProcess, &exitcode);
   CloseHandle(pi.hProcess);
   CloseHandle(pi.hThread);
+  CloseHandle(hJob);
   if (!doreturn) {
     exit(exitcode);
   }


jrosdahl:
Sounds like a good idea. Have you looked into the failing tests?

@ccache/windows-contributors: Would you like to review this PR?

cristianadam (approved):
It's how one would implement the feature on Windows. See https://stackoverflow.com/questions/53208/how-do-i-automatically-destroy-child-processes-in-windows 

Qt Creator uses the same functionality for its `qtcreator_ctrlc_stub` executable.

jrosdahl:
Still waiting for an answer to ["_Have you looked into the failing tests?_"](https://github.com/ccache/ccache/pull/1274#issuecomment-1517997624) – the test suite consistently fails like this:

```
 3/46 Test  #2: test.base ........................***Failed    2.48 sec
Compiler:         gcc (/mingw64/bin/gcc)
Compiler version: gcc.exe (Rev6, Built by MSYS2 project) 13.1.0

Running test suite base
  Base caseccache: error: execute_noreturn of D:/a/ccache/ccache/build/testdir/1248/compiler/gcc.sh failed: No such file or directory

FAILED

Test suite:     base (line 9)
Test case:      Base case
Failure reason: Expected cache_miss to be 1, actual 0
```

arBmind:
@jrosdahl I have not looked into the failing tests. It seems kind of random that most tests succeed, but some special platforms fail.

jrosdahl:
> @jrosdahl I have not looked into the failing tests. It seems kind of random that most tests succeed, but some special platforms fail.

I tried the commit in question on top of latest master and the same tests still failed. It doesn't seem random to me; there's consistently a failure in all Windows jobs. (Some Windows jobs not marked as failed, but that's because they have `allow_test_failures: true`.)

arBmind:
> > @jrosdahl I have not looked into the failing tests. It seems kind of random that most tests succeed, but some special platforms fail.
> 
> I tried the commit in question on top of latest master and the same tests still failed. It doesn't seem random to me; there's consistently a failure in all Windows jobs. (Some, or actually most, Windows jobs are not marked as failed, but that's because they have `allow_test_failures: true`.)

Ah, ok. I did not know that CI tests results are ignored for most windows combinations.

I took a quick look into the build logs of "VS2022 64-bit Ninja". But I'm still stuck.
```
 4/46 Test  #4: test.cache_levels ................***Failed    2.38 sec
D:/a/ccache/ccache/test/suites/base.bash: line 556: syntax error near unexpected token `&&'
D:/a/ccache/ccache/test/suites/base.bash: line 556: `       && $COMPILER -E test1.c -gz >preprocessed.i 2>/dev/null \'
```
How does the bash syntax fail for changing ccache code? - Seems totally unrelated but all tests fail on this. I have not touched this file at all. Is ctest is using the wrong bash version?

rkitover:
On line 555 of `test/suites/base.bash` there is a missing `\` at the end to continue the line.

For some reason though the tests passed for me locally with this branch, but this is definitely wrong.

Could be a bash version difference or something like that.

rkitover:
Ok I see, the `\` is in master but not in this branch.

@arBmind Can you please rebase on master and force push again, this will re-run the CI.

arBmind:
@rkitover Thanks. Okay, now we got some different errors, I cannot interpret:
```error: execute_noreturn of D:/a/ccache/ccache/build/testdir/1216/compiler/clang.sh failed: No such file or directory```

Seems rather unhelpful. There are some checks that fail, but nothing that seems related to my changes.

**Comment by rkitover** [comment]
I can report that both master and this branch pass tests for me in the MSYS2 UCRT64 environment.

Also this branch is currently up to date with master.

So we have now a CI issue that affects master, I will dig into this a bit more.

**Comment by rkitover** [comment]
Tests fail for me locally with the 64 bit ninja build on master, here is the log:

https://gist.github.com/rkitover/4910dd0bcfe7b1ec6fc3a8d60764c2d0

**Comment by arBmind** [comment]
> https://gist.github.com/rkitover/4910dd0bcfe7b1ec6fc3a8d60764c2d0

It seems that that the failing tests are kind of unrelated.
The Gist has 3 failed tests and complains about Microsoft specific modules not present in clang.

GA runs complain about missing files "clang.sh" - which apparently could not be launched. But it's beyond my understanding why we need a ccache to invoke a shell script to remove environment variables before running the actual `clang` instance.

**Comment by jrosdahl** [comment]
> GA runs complain about missing files "clang.sh" - which apparently could not be launched. But it's beyond my understanding why we need a ccache to invoke a shell script to remove environment variables before running the actual clang instance.

The script that unsets some environment variables before calling the real compiler is created by the test framework, which then calls the script instead of the real compiler. That was done since the real compiler could actually itself be a wrapper script that in turn uses ccache, which messes up test expectations (see #914 and 3c47a8a23d773f5d32174ddbb3d3e9409a3f201e). Then  de5778ac6a7c4722454f6032822c2bb76533ebde tweaked this solution in an effort to make the tests work on Windows as well.

However, I guess it's likely that such a wrapper script isn't ever needed on Windows systems, so you could try something like this:

```patch
--- a/test/run
+++ b/test/run
@@ -664,17 +664,18 @@ cd $TESTDIR || exit 1
 
 mkdir compiler
 
-COMPILER="$(pwd)/compiler/$(basename "$REAL_COMPILER_BIN")"
 if $HOST_OS_WINDOWS; then
-    COMPILER="$COMPILER.sh"
-fi
-cat >"$COMPILER" <<EOF
+    COMPILER="$CC"
+else
+    COMPILER="$(pwd)/compiler/$(basename "$REAL_COMPILER_BIN")"
+    cat >"$COMPILER" <<EOF
 #!/bin/sh
 
 CCACHE_DISABLE=1 CCACHE_COMPILER= CCACHE_PREFIX= \
   exec $REAL_COMPILER_BIN$COMPILER_ARGS "\$@"
 EOF
-chmod +x "$COMPILER"
+    chmod +x "$COMPILER"
+fi
 
 [ -z "${VERBOSE:-}" ] && verbose=false || verbose=true
 [ "$1" = "-v" ] && { verbose=true; shift; }
```

arBmind:
Okay, all tests pass.

It seems GA launches the ccache in a job context that does not allow breaking away. I changed the code to detect that and fall back to the old behaviour in that case.

jrosdahl:
Thanks.

jrosdahl:
Note: This was reverted in a2bf2956d78599b2397081a9e7b253a569c7c932 / 2e0caea3845d3f6f8eb7c5d742d72f271c6c6eeb due to problems described in #1386.

@arBmind, @rkitover, @cristianadam: Any ideas on how to implement this in a way that doesn't break the use case described in #1386 are welcome.

arBmind:
@jrosdahl We use ccache on many systems and never had this issue. So I cannot relate or text fixes to your problem here.

I am also kind of disappointed now, that you broke this nice tool for our use case again.

cristianadam:
Having ccache kill the compiler subprocess would have on Windows the same behaviour as on Linux.

At https://github.com/ccache/ccache/issues/1386 we have ccache crashing causing then the compiler subprocess to be killed, and instead of finding out why ccache is crashing and fixing that the revert of the kill of the compiler subprocess was chosen.

I think having the compiler running when the build process has been stopped is not cool, at least this could have been a feature.

cristianadam:
One could build ccache as `RelWithDebInfo` enable [Core Mini Dumps](https://learn.microsoft.com/en-us/windows/win32/wer/collecting-user-mode-dumps) and find out where and why ccache is crashing resulting in force termination of the compiler subprocesses. 

I personally never saw the `fatal error C1090: PDB API call failed, error code '23': (0x000006BA)` error in practice.

jrosdahl:
@arBmind wrote:
> We use ccache on many systems and never had this issue. So I cannot relate or text fixes to your problem here.

That's perfectly OK! It's often not possible to test or foresee problems given the diversity of environments that ccache operates in.

If you have a good solution I'm all ears (as I already mentioned). One possible alternative would be to make termination of the compiler process opt-in via configuration on Windows, but that would feel very clunky and it would of course be better if things just work. Until we know more and have decided what to do, I will leave the feature out since it makes legitimate build scenarios fail, which is worse than compiler processes living longer than they should.

@cristianadam wrote:
> At https://github.com/ccache/ccache/issues/1386 we have ccache crashing causing then the compiler subprocess to be killed, and instead of finding out why ccache is crashing and fixing that the revert of the kill of the compiler subprocess was chosen.

It sounds like you have misunderstood the issue. It's not ccache that is crashing, it's MSVC that writes the error message and exits with failure. Ccache is just happily executing the compiler. My only speculation is that perhaps MSVC in the `/Zi` case starts some background process that lives longer than the compiler and the process tree is then killed transitively when ccache exits?

> I think having the compiler running when the build process has been stopped is not cool

I don't think anybody thinks that.

> I personally never saw the fatal error C1090: PDB API call failed, error code '23': (0x000006BA) error in practice.

That's of course great. I can reproduce the problem easily, though. Unfortunately I'm not very good at debugging and fixing problems on Windows.

Could you guys perhaps give me some slack and try to part of the solution instead of accusing me of breaking things when I'm unbreaking things?

cristianadam:
> That's of course great. I can reproduce the problem easily, though. Unfortunately I'm not very good at debugging and fixing problems on Windows.

Does `/Zf (Faster PDB generation)` help? From the documentation https://learn.microsoft.com/en-us/cpp/build/reference/zf?view=msvc-170 it looks like it would fix exactly this issue.

Generally I think `/Z7` is the best option for ccache, since `/Zi` is not yet supported, as per https://github.com/ccache/ccache/issues/1040 or?

I assume ccache starts a `cl.exe` which then starts the `mspdbsrv.exe` process that merges all the pdb information and it gets killed because of the process tree.

Should ccache start its own `mspdbsrv.exe`? This should fix the premature termination.

jrosdahl:
> Does `/Zf (Faster PDB generation)` help? From the documentation [learn.microsoft.com/en-us/cpp/build/reference/zf?view=msvc-170](https://learn.microsoft.com/en-us/cpp/build/reference/zf?view=msvc-170) it looks like it would fix exactly this issue.

Thanks for the reference. It could be interesting to try it in order to learn more about the problem space, but I don't think that it would get us closer to fixing the problem.

To clarify, in case it's not obvious: the problem is not that _my_ build fails. I can easily work around that. The problem is that ccache crashes a build that _does_ use `/Zi` for whatever reason, for instance if ccache is set up to be used for all compiler invocations, both for builds with and without debug information. Unsupported compilation flags should just make ccache execute the compiler without interference, regardless of which those compilation flags are.

> I assume ccache starts a `cl.exe` which then starts the `mspdbsrv.exe` process that merges all the pdb information and it gets killed because of the process tree.

Yes, that sounds likely and matches my gut feeling. (This also means that the Windows version of terminating the compiler is not exactly like that of the POSIX version since the latter does not kill the whole process group.)

Is it possible (and wanted) to get a similar behavior with the Windows API, i.e. kill the compiler but not processes it has created?

> Should ccache start its own `mspdbsrv.exe`? This should fix the premature termination.

I don't know.

Here's what I'll do unless somebody figures out something better:

* If `/Zi` or `/ZI` is not used, set up job termination.
* If `/Zi` or `/ZI` is used, don't set up job termination.

This should fix the problem (at the expense of suboptimal behavior when aborting when `/Zi` or `/ZI` is used).

arBmind:
So I guess we have to change the Job logic to use JOB_OBJECT_LIMIT_SILENT_BREAKAWAY_OK. This will allow sub processes of the spawned process to not be part of the job by default. (See: https://learn.microsoft.com/en-us/windows/win32/api/winnt/ns-winnt-jobobject_basic_limit_information and https://learn.microsoft.com/en-us/windows/win32/procthread/job-objects#managing-a-process-tree-that-uses-job-objects)

As I have no way to verify that this works, I would love to see someone to pick this up from here.

jrosdahl:
I'll see if I can find some time next week to try `JOB_OBJECT_LIMIT_SILENT_BREAKAWAY_OK`.

jrosdahl:
With

```patch
--- a/src/execute.cpp
+++ b/src/execute.cpp
@@ -174,7 +174,8 @@ win32execute(const char* path,
       // process is killed.
       JOBOBJECT_EXTENDED_LIMIT_INFORMATION jobInfo = {};
       jobInfo.BasicLimitInformation.LimitFlags =
-        JOB_OBJECT_LIMIT_KILL_ON_JOB_CLOSE;
+        JOB_OBJECT_LIMIT_KILL_ON_JOB_CLOSE
+        | JOB_OBJECT_LIMIT_SILENT_BREAKAWAY_OK;
       BOOL job_success = SetInformationJobObject(
         job, JobObjectExtendedLimitInformation, &jobInfo, sizeof(jobInfo));
       if (!job_success) {
```

I can no longer reproduce a build failure, so going with that.

Issue #1278, opened 2023-04-17
Windows 64-bit MSBuild CI jobs fail


jrosdahl:
The [Windows VS2019 64-bit MSBuild](https://github.com/ccache/ccache/actions/runs/4659417272/jobs/8246283413#logs) and [Windows VS2022 64-bit MSBuild](https://github.com/ccache/ccache/actions/runs/4659417272/jobs/8246283873#logs) GitHub actions jobs have started failing on master due to some changes in the CI environment. @rkitover: FYI.

I'm going to disable them for now.

rkitover:
@ruimaciel Different issue, please open an issue if you are having problems with all the details.

PR #1290, opened 2023-05-22
Fix MSBuild CI jobs + minor cmake fixes

rkitover:
- Fix the MSBuild CI jobs #1278 
- Update blake3 from 1.3.1 to 71a2646
- Fix cmake dev warning when downloading zstd/hiredis
- Add a basic CMakeSettings.json for Visual Studio IDE with Ninja as the generator

### DIFF
--- a/cmake/Findhiredis.cmake
+++ b/cmake/Findhiredis.cmake
@@ -2,6 +2,11 @@ if(hiredis_FOUND)
   return()
 endif()
 
+if(POLICY CMP0135)
+  # Set timestamps on extracted files to time of extraction.
+  cmake_policy(SET CMP0135 NEW)
+endif()
+
 set(hiredis_FOUND FALSE)
 
 if(HIREDIS_FROM_INTERNET AND NOT HIREDIS_FROM_INTERNET STREQUAL "AUTO")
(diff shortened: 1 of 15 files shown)


cmake/Findhiredis.cmake
```diff
@@ -60,6 +60,7 @@ if(do_download)
     URL https://github.com/redis/hiredis/archive/refs/tags/v${hiredis_version}.tar.gz
     SOURCE_DIR ${hiredis_dir}
     BINARY_DIR ${hiredis_build}
+    DOWNLOAD_EXTRACT_TIMESTAMP TRUE
```
jrosdahl, line 63: Looks like this was introduced in CMake 3.24? What happens if the user has an older CMake?

src/third_party/blake3/blake3_avx2_x86-64_windows_gnu.S
```diff
@@ -1784,7 +1784,7 @@ blake3_hash_many_avx2:
         vmovdqu xmmword ptr [rbx+0x10], xmm1
         jmp     4b
 
-.section .rodata
+.section .rdata
```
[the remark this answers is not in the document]
rkitover: Will fix.

cmake/Findhiredis.cmake
rkitover: It does in fact break cmake 3.18 on Debian 11, good catch there.

rkitover: Fixed.

jrosdahl:
Thanks.

PR #1295, opened 2023-06-02
fix: Keep the order of MSVC showIncludes output

huangqinjin:
The bug is introduced in commit 3ec0d4ec961cb0c67ff1603b83a6026d9f3abbb5.


jrosdahl:
Thanks!

Issue #1247, opened 2023-02-01
Improve MSVC RSP file parsing


jrosdahl:
From @jacobly0 in https://github.com/ccache/ccache/issues/1233#issuecomment-1411015937:            

> I wrote a script to demonstrate msvc's expansion behavior of at files, and it's clear that my original implementation was missing some details.
> ```sh
> #!/bin/sh
> echo def > atfile.c
> for cl in cl clang-cl; do
>     echo
>     echo "Using compiler $cl:"
>     while read -r def; do
>         if [ -z "$def" ]; then
>             echo
>             continue
>         fi
>         echo -n "$def => "
>         echo "/w /nologo /E /Ddef=$def atfile.c" > atfile.rsp
>         $cl @atfile.rsp 2>&1 | grep --invert '^#\|^atfile\.c$'
>     done <<'EOF'
> \
> \\
> '\\'
> "\\"
> '"\\"'
> "'\\'"
> '''\\'''
> ''"\\"''
> '"'\\'"'
> '""\\""'
> "''\\''"
> "'"\\"'"
> ""'\\'""
> """\\"""
> 
> 
> \'\'
> '\'\''
> "\'\'"
> ''\'\'''
> '"\'\'"'
> "'\'\''"
> ""\'\'""
> '''\'\''''
> ''"\'\'"''
> '"'\'\''"'
> '""\'\'""'
> "''\'\'''"
> "'"\'\'"'"
> ""'\'\''""
> """\'\'"""
> 
> \"\"
> '\"\"'
> "\"\""
> ''\"\"''
> '"\"\""'
> "'\"\"'"
> ""\"\"""
> '''\"\"'''
> ''"\"\""''
> '"'\"\"'"'
> '""\"\"""'
> "''\"\"''"
> "'"\"\""'"
> ""'\"\"'""
> """\"\""""
> EOF
> done
> ```
>
> ```
> $ ./atfile.sh
> 
> Using compiler cl:
> \ => \
> \\ => \\
> '\\' => '\\'
> "\\" => \
> '"\\"' => '\'
> "'\\'" => '\\'
> '''\\''' => '''\\'''
> ''"\\"'' => ''\''
> '"'\\'"' => ''\\''
> '""\\""' => '\'
> "''\\''" => ''\\''
> "'"\\"'" => '\'
> ""'\\'"" => '\\'
> """\\""" => "\"
> 
> 
> \'\' => \'\'
> '\'\'' => '\'\''
> "\'\'" => \'\'
[...]

PR #1303, opened 2023-06-15
fix: Improve handling of backslash in MSVC .rsp

ptc-jhoerner:
Parse correctly backslash escaped by backslash as in `/Fo"Dir\Dir\\"`.

The flag in the unit test `/Fo"N.dir\Release\\"` is taken from .rsp from VS22 project build. VS22 solution was generated by cmake 3.26.4. I'm not sure what is driving the use of the trailing backslash in the path used in `/Fo` flag, but these cases are definitely occurring in the wild for us.

I wasn't able to find spec for this case, but these rsp files with these escapes are correctly handled by CL. It makes sense that backslash can escape backslash, because otherwise it would not be possible to have backslash before `"` as the trailing backslash would otherwise escape `"`.

Cheers,
Jiri

[image](http://www.ptc.com/)
**Jiri Hörner**
Senior Software Development Engineer, Vuforia

jhoerner@ptc.com

[vuforia.com](http://www.vuforia.com/)

---
Parametric Technology Gesellschaft m.b.H., Ungargasse 37, 1030 Wien, Austria. Firmensitz: Wien, FN: 111171 m, Firmenbuchgericht: Handelsgericht Wien.
 
The information contained in this email transmission is confidential and may be privileged. If you are not the intended recipient, any use, dissemination, distribution, publication, or copying of the information contained in this email is strictly prohibited.  If you have received this email in error, please immediately notify me by calling the above number and delete the email from your system. Thank you for your co-operation.

Copyright © 2020 PTC Inc. and/or all its affiliates or subsidiaries. All rights reserved

### DIFF
--- a/src/Args.cpp
+++ b/src/Args.cpp
@@ -77,7 +77,7 @@ Args::from_atfile(const std::string& filename, AtFileFormat format)
         }
         break;
       case AtFileFormat::msvc:
-        if (*pos != '"') {
+        if (*pos != '"' && *pos != '\\') {
           pos--;
         }
         break;
--- a/src/Args.cpp
+++ b/src/Args.cpp
@@ -60,7 +60,7 @@ Args::from_atfile(const std::string& filename, AtFileFormat format)
   auto pos = argtext->c_str();
   std::string argbuf;
   argbuf.resize(argtext->length() + 1);
-  auto argpos = argbuf.begin();
+  auto argpos = argbuf.data();
 
   // Used to track quoting state; if \0 we are not inside quotes. Otherwise
   // stores the quoting character that started it for matching the end quote.
@@ -69,17 +69,40 @@ Args::from_atfile(const std::string& filename, AtFileFormat format)
   while (true) {
     switch (*pos) {
     case '\\':
-      pos++;
       switch (format) {
       case AtFileFormat::gcc:
+        pos++;
         if (*pos == '\0') {
           continue;
         }
         break;
       case AtFileFormat::msvc:
-        if (*pos != '"' && *pos != '\\') {
-          pos--;
+        size_t count = 0;
+        while (*pos == '\\') {
+          count++;
+          pos++;
         }
+        if (*pos == '"') {
+          if (count == 1) {
+            // simple escape \"
+            break;
+          }
+          if (count % 2 != 0) {
+            // If an odd number of backslashes is followed by a double quotation
+            // mark, one backslash is placed in the argv array for every pair of
+            // backslashes, and the double quotation mark is "escaped" by the
+            // remaining backslash
+            pos--;
+          }
+          std::memset(argpos, '\\', count / 2);
+          argpos += count / 2;
+        } else {
+          // Backslashes are interpreted literally, unless they immediately
+          // precede a double quotation mark.
+          std::memset(argpos, '\\', count);
+          argpos += count;
+        }
+        continue;
         break;
       }
       break;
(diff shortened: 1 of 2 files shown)


ptc-jhoerner:
I'll be out next week, so I might not be able to immediately respond here, but I'll take a look at the review feedback when I'm back. Sorry if this PR will look dead next week.

jrosdahl:
Thanks for the PR!

There have been several attempts (at least #1071 and #1233) to fix the original implementation (from #506), but it's still incorrect as noted in #1247. It would be great if somebody could turn @jacobly0's investigation (https://github.com/ccache/ccache/issues/1247#issue-1566713819) into a test case and fix the implementation accordingly. Would you be interested in doing that?

That said, if this fix is one step towards the end goal that would be OK as well, but I'm a bit afraid that making small tweaks to the algorithm on a case-by-case basis rather than considering the full scope is just continuing a game of whac-a-mole.

ptc-jhoerner:
I can take a look where we stand w.r.t. to the test cases mentioned in https://github.com/ccache/ccache/issues/1247 after my changes. But even if we won't be able to fix all of them easily, I think it is still worth fixing at least this case - the same problem we hit is also mentioned in https://github.com/ccache/ccache/issues/1247#issuecomment-1585727296, so it looks like quite widespread issue.

ptc-jhoerner:
ok, I have took another look at this and fixed all examples from #1247 as well as the examples given in the Microsoft documentation you linked. I have added all of these examples to Unit tests and removed one of the existing tests that was not representing an actual MSVC behaviour.

The logic is a bit tricky, but I was able to fit it into the existing cross-platforms code. For some of the behaviour (quote right after closing quote) neither of the two documentation pieces mention the special behaviour, but I was able to find a StackOverflow answer mentioning this which aligned with behaviour observed in #1247. With that I'm not aware of any cases that wouldn't be handled correctly right now. With the test coverage, I think the implementation is pretty solid now. If you find even more edge cases in the future, I'm afraid one might need to resort to using `CommandLineToArgvW` or split the logic from gcc parsing, because it is hitting already a decent degree of complexity.

jrosdahl:
Awesome, thanks!

> using `CommandLineToArgvW`

I would be perfectly OK with using it if the syntax it parses is the same as in `.rsp` files do. (Not sure why it wasn't used in #506; perhaps it was easier to just do minor tweaks to `Args::from_atfile`?) Do you know of any drawbacks of using it?

> or split the logic from gcc parsing, because it is hitting already a decent degree of complexity.

Agreed, and perhaps we've already hit the limit.

ptc-jhoerner:
`CommandLineToArgvW` seems to do the right thing. It does not support comments as per https://learn.microsoft.com/en-us/visualstudio/msbuild/msbuild-response-files?view=vs-2022, but it not clear whether comments are supported in CL response files or whether this is supported on for msbuild response files. (I'm not sure whether they are necessarily the same thing.) We obviously don't handle these comments either at the moment, but at least with the current implementation could introduce the handling of comments more easily than when using `CommandLineToArgvW`, I believe.

To change the implementation to use `CommandLineToArgvW`, clearly one would have to convert UTF-8 to UTF-16 or load the file in UTF-16 directly and then convert the args to UTF-8. Not sure whether you already have some infrastructure to handle this or at least some patterns you use. One can use MultiByteToWideChar and friends, but it is probably going to cost some performance compared to the current implementation. Not sure how much performance is concern here. In the most common use-case rsp files are going to be most likely with every invocation.

So far in my tests I wasn't able to find a difference between the current implementation and `CommandLineToArgvW` they seem to be equivalent.

jrosdahl:
Thanks. Then let's go with your implementation for now. If somebody wants to work on using `CommandLineToArgvW` instead, feel free to do so.

> [...] one would have to convert UTF-8 to UTF-16 or load the file in UTF-16 directly and then convert the args to UTF-8. Not sure whether you already have some infrastructure to handle this or at least some patterns you use.

The only existing code for file encoding on Windows is [the UTF-16 to UTF-8 conversion to support reading Windows response files](https://github.com/ccache/ccache/blob/2737d79e282de72d1b29128f437173e4892ced5e/src/util/file.cpp#L164-L197) from #1005 + #1172.

PR #1307, opened 2023-07-01
feat: Expand  environment variables in debug_dir configuration entry

cupu:
Config directives such as cache_dir expand environment variables in their values ; debug_dir does not.
This is actually useful for me as I have  a ccache config I distribute to developers , I use their home directory for some ccache directory configs but could not do it for debug_dir.


jrosdahl:
Thanks.

I've had a note saying "support variable expansion of all configuration variables" in my to do list for a long time, so I'll implement a more generic solution after merging this.

PR #1312, opened 2023-07-26
feat: Allow specifying debug level

### LINKED ISSUE #1285: Add possibility to dump *ccache-log files only
### Description ###
Currently debug mode causes ccache to generate 5 files for every object - `.ccache-log` and 4 `.ccache-input*` ones.

It would be useful to have possibility to generate `.ccache-log` only. Maybe add an additional option or add possibility to specify debug level.

Rationale: During our work, we wanted to gather statistics which particular files are taken from cache and which are rebuilt. We used `.ccache-log` files for this but found that in addition to that logs, we get 300+Gb of `.ccache-input*` files which are not required for us but affect the build process (we can simply become out of space).

dsilakov:
... via debug_level option or CCACHE_DEBUGLEVEL env variable.

With debug_level=2 (default), ccache behavior in debug mode is not changed.

With debug_level=1, ccache will only dump *.ccache-log files (and not the *input* ones). This is useful for getting per-file caching results.

Closes #1285

### DIFF
--- a/doc/MANUAL.adoc
+++ b/doc/MANUAL.adoc
@@ -681,6 +681,15 @@ is `/home/user` and the object file is `build/output.o` then the debug log will
 be written to `/example/home/user/build/output.o.ccache-log`. See also
 _<<Cache debugging>>_.
 
+[#config_debug_level]
+*debug_level* (*CCACHE_DEBUGLEVEL*)::
+
+    Specifies amount of information dumped when the <<config_debug,debug mode>>
+    is enabled. If set 1, only `<objectfile>.<timestamp>.ccache-log` files with
+    per-file caching statistics are dumped. Otherwise, all possible debug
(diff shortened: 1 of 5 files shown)


jrosdahl (changes requested):
Looks good to me. The only thing I see missing is documentation of the new configuration option [after `debug_dir`](https://github.com/ccache/ccache/blob/689168c292f1ed26c5f4a3070aeb649dad7facb5/doc/MANUAL.adoc?plain=1#L683) in the manual.

dsilakov:
> Looks good to me. The only thing I see missing is documentation of the new configuration option [after `debug_dir`](https://github.com/ccache/ccache/blob/689168c292f1ed26c5f4a3070aeb649dad7facb5/doc/MANUAL.adoc?plain=1#L683) in the manual.

Added.

jrosdahl (approved):

jrosdahl:
Thanks.

PR #1318, opened 2023-08-02
fix: relative path handling for clang PCH header

jiulongw:
cmake may generate both -include and -include-pch. Example:

c++ ... -Xclang -include-pch -Xclang /abs/path/cmake_pch.hxx.pch
        -Xclang -include -Xclang /abs/path/cmake_pch.hxx ...

Both paths needs to be converted to relative path if CCACHE_BASEDIR is set for better cache hit rate in some CI systems.


jrosdahl:
Thanks.

PR #1323, opened 2023-08-22
fix: Allow wildcards in port url validation

### LINKED ISSUE #1321: CCACHE_REMOTE_STORAGE sharding doesn't accept valid ports
### How to reproduce ###

1. Try to set `CCACHE_REMOTE_STORAGE="redis://${{ IP }}:*|shards=${{ PORT1 }},${{ PORT2 }},${{ PORT3 }}"` e.g. `$env:CCACHE_REMOTE_STORAGE="redis://10.0.0.4:*|shards=6379,6380,6381"`
 which fails with `ccache : error : Cannot parse URL redis://10.0.0.4:*: Port '*' in 'redis://10.0.0.4:*' is invalid`
2. Set `CCACHE_REMOTE_STORAGE="redis://${{ IP }}*|shards=:${{ PORT1 }},:${{ PORT2 }},:${{ PORT3 }}"` e.g. `$env:CCACHE_REMOTE_STORAGE="redis://10.0.0.4*|shards=:6379,:6380,:6381"` and it works with proper sharding.

### Actual behavior ###
You can't use ports as shards, but you can bypass the URL validation with a dirty hack.

### Expected behavior ###
Shards should work with ports, to allow clustering on the same host (e.g. with Docker containers) and shard parameters should be validated.

### Environment ###

ccache version 4.7.4

afbjorklund:
Closes #1321

Separated the http client/server log based on port number, since otherwise you could only start one.

### DIFF
--- a/src/storage/Storage.cpp
+++ b/src/storage/Storage.cpp
@@ -174,12 +174,31 @@ parse_storage_config(const std::string_view entry)
           name = shard;
         }
 
+        const Url shard_url = util::replace_first(url_str, "*", name);
+        try {
+          std::ignore = shard_url.str();
+        } catch (const std::exception& e) {
+          throw core::Error(FMT(
+            "Cannot parse URL {} for shard {}: {}", url_str, name, e.what()));
+        }
         result.shards.push_back({std::string(name), weight});
       }
     }
 
     result.params.attributes.push_back(
       {std::string(key), value, std::string(raw_value)});
+  }
+
+  if (result.shards.empty()) {
+    try {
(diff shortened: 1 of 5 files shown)


afbjorklund:
Fixed the clang-tidy issue.

afbjorklund:
Will abandon this in favor of replacing the template `Url` with a string instead.

The tests could probably be reused, to verify and check for any regressions.

afbjorklund:
Will abandon this in favor of replacing the template `Url` with a string instead.

The tests could probably be reused, to verify and check for any regressions.

PR #1331, opened 2023-09-05
fix: src/InodeCache.cpp: avoid sign-compare warning on FreeBSD

pstorz:
As on FreeBSD 13.2 `statfs.f_bavail` is signed, so InodeCache.cpp cannot be compiled but gets a sign-compare warning:

  src/InodeCache.cpp:409:30: error: comparison of integers of different signs:
    'long' and 'unsigned long' [-Werror,-Wsign-compare]
      if (buf.f_bavail * 512 < k_min_fs_mib_left * 1024 * 1024) {
          ~~~~~~~~~~~~~~~~~~ ^ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~bareos
  1 error generated.

The problem is avoided by using static_cast to unsigned long on FreeBSD. 
To detect FreeBSD, now `sys/param.h` is included unconditionally.

### DIFF
--- a/src/InodeCache.cpp
+++ b/src/InodeCache.cpp
@@ -416,7 +416,8 @@ InodeCache::initialize()
         LOG("fstatfs failed: {}", strerror(errno));
         return false;
       }
-      if (buf.f_bavail * 512 < k_min_fs_mib_left * 1024 * 1024) {
+      if (static_cast<uint64_t>(buf.f_bavail) * 512
+          < k_min_fs_mib_left * 1024 * 1024) {
         LOG("Filesystem has less than {} MiB free space, not using inode cache",
             k_min_fs_mib_left);
         return false;


jrosdahl:
> so InodeCache.cpp cannot be compiled

Just a note: It's just a warning so there should be no problem compiling the code in ["non-dev mode"](https://github.com/ccache/ccache/blob/e3e2b296e03bd55d0eca74b49da802728f84c09d/CMakeLists.txt#L55-L62).

Anyway, sounds OK to cast.

src/InodeCache.cpp
```diff
@@ -416,7 +416,12 @@ InodeCache::initialize()
         LOG("fstatfs failed: {}", strerror(errno));
         return false;
       }
-      if (buf.f_bavail * 512 < k_min_fs_mib_left * 1024 * 1024) {
+#if __FREEBSD__
+      if (static_cast<unsigned long>(buf.f_bavail) * 512
+          < k_min_fs_mib_left * 1024 * 1024) {
+#else
+      if ((buf.f_bavail) * 512 < k_min_fs_mib_left * 1024 * 1024) {
+#endif
```
jrosdahl, line 424: How about always casting to `uint64_t` regardless of platform? That way we don't need to duplicate the code.

pstorz: If that does not create problems on other platforms, it would be a great solution as we could avoid the extra include and the conditional compilation.

pstorz: I have updated the PR to do what you suggested and rebased it on master.

jrosdahl:
Thanks.

PR #1338, opened 2023-10-16
fix: Windows is_absolute_path() crash on drive letter

mablanchard:
Prevents `path::is_absolute_path(path)` from reading characters further than `path` string length.


jrosdahl:
Thanks.

PR #1342, opened 2023-10-21
fix: devmode unused function warnings on new MSVC

rkitover:
MSVC apparently has overrides for the two definitions of operator== in
unittest/test_util_string.cpp in the <utility> header, raising a warning
about them being local unused functions:

https://learn.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-4-c4505?view=msvc-170

Disable this warning type in cmake/DevModeWarnings.cmake for MSVC.

### DIFF
--- a/cmake/DevModeWarnings.cmake
+++ b/cmake/DevModeWarnings.cmake
@@ -121,5 +121,7 @@ elseif(MSVC)
     /wd4706
     # Non-underscore-prefixed POSIX functions:
     /wd4996
+    # Dead local functions overridden by headers:
+    /wd4505
   )
 endif()


rkitover:
Looks like checking only for `_MSC_VER` will not be sufficient, I will have to track down the specific change to a specific version of MSVC.

rkitover:
I'm sorry, I was unaware of dev mode. I disabled this warning class, if you guys don't mind.

rkitover:
It's failing some tests on WSL, is this of interest to anyone to bother fixing?

rkitover:
@jrosdahl I had an idea, I want to set `CMAKE_CXX_COMPILER_LAUNCHER` et al. to `ccache` if it's in the `PATH` when dev mode is active. And now that I learned about dev mode, it seems like it would be a perfect addition.

jrosdahl:
Thanks.

PR #1343, opened 2023-10-21
feat: set both C/CXX launchers if either is set

rkitover:
Add cmake/EnableCcache.cmake and load it from the main file.

Make the standard cmake variables CMAKE_CXX_COMPILER_LAUNCHER and
CMAKE_C_COMPILER_LAUNCHER consistent with each other if either is not
set.


jrosdahl:
Thanks for the PR! I understand why you'd want this.

However, Ccache used to have something similar (not scoped to the developer mode) that was added as part of #759 (ccache 4.2) and then eventually removed in 477da4a63eacc3b19acbc3146bd055c3ef8f7208 (ccache 4.6). File history:
https://github.com/ccache/ccache/commits/477da4a63eacc3b19acbc3146bd055c3ef8f7208/cmake/UseCcache.cmake

As mentioned in 477da4a63eacc3b19acbc3146bd055c3ef8f7208, I think that if the automatic enabling of ccache as the compiler launcher leads to double invocations of ccache then something's wrong. Since we now require CMake ≥ 3.15, I guess we could use [`file(READ_SYMLINK ...)`](https://cmake.org/cmake/help/latest/command/file.html#read-symlink) to detect this, though. That would leave the more philosophical question if it should be up to the user to enable or not enable ccache, or any other alternative compiler launcher for that matter.

rkitover:
I'll leave the philosophical issue untouched for now, and I don't understand the double-launching issue.

I'd like to keep this part however, this will allow me to pass `-DCMAKE_CXX_COMPILER_LAUNCHER=ccache` on my build command line and have it also apply to the C sources.

rkitover:
This macos CI failure is weird.

jrosdahl:
> I don't understand the double-launching issue

Looking closer at this, the double invocation that could happen in in some scenarios was actually fixed later as a kind of side effect of 451d92c801d6463bcc54c090b871a1d77f26d9d4, so then I'm OK with doing it like you originally suggested.

rkitover:
@jrosdahl I'm sorry, I meant to follow up on this, but life has been difficult and I also got distracted with other things. I will follow up in a new PR. Also if there are any other build issues you'd like me to look at let me know.

jrosdahl:
> I'm sorry, I meant to follow up on this, but life has been difficult and I also got distracted with other things. I will follow up in a new PR. Also if there are any other build issues you'd like me to look at let me know.

No problems, and thanks! I too have limited time for ccache (as always).

PR #1345, opened 2023-10-31
(feat)add support for clang --analyze

mbel0:
add support for static analysis option in clang: --analyze
static analysis is performed separately from compilation and usually takes more time,
but its results can be captured the same way as regular compilation results: some
warnings may be printed and result is stored in .plist file instead of .o

this is partial solution for static analysis: -analyze option for cc1 is not supported
as I am macOS using Xcode, so modules are always enabled and I had to use direct mode
--analyzer-output option is also not supported, only default .plist format

I have not added much of the testing as I found no similar tests and not sure what would
 be needed. If you provide some hints or guidance here I probably could fix that

### DIFF
--- a/src/argprocessing.cpp
+++ b/src/argprocessing.cpp
@@ -60,6 +60,7 @@ struct ArgumentProcessingState
   bool found_c_opt = false;
   bool found_dc_opt = false;
   bool found_S_opt = false;
+  bool found_analyze_opt = false;
   bool found_pch = false;
   bool found_fpch_preprocess = false;
   bool found_Yu = false;
@@ -527,6 +528,12 @@ process_option_arg(const Context& ctx,
   if (arg == "-S") {
     state.common_args.push_back(args[i]);
     state.found_S_opt = true;
+    return Statistic::none;
+  }
+
+  // --analyze changes the default extension too
+  if (arg == "--analyze") {
+    state.found_analyze_opt = true;
     return Statistic::none;
   }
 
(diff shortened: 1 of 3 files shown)


src/argprocessing.cpp
```diff
@@ -530,6 +531,12 @@ process_option_arg(const Context& ctx,
     return Statistic::none;
   }
 
+  // --analyze changes the default extension too
+  if (arg == "--analyze") {
+    state.found_analyze_opt = true;
```
jrosdahl, line 536: How about adding

```c++
    state.common_args.push_back(args[i]);
```

here, similar to how it's done for `-S`? Then there would be no need to call `compiler_args.push_back("--analyze");` in `process_arg`.

src/argprocessing.cpp
```diff
@@ -1221,8 +1229,14 @@ process_args(Context& ctx)
   }
 
   if (output_obj_by_source && !args_info.input_file.empty()) {
-    std::string_view extension =
-      state.found_S_opt ? ".s" : get_default_object_file_extension(ctx.config);
+    std::string_view extension;
+    if (state.found_S_opt) {
+      extension = ".s";
+    } else if (state.found_analyze_opt) {
+      extension = ".plist";
```
jrosdahl, line 1236: I just tried how the different variants of `--analyzer-output` affect the default extension, and it seems like it's indeed always `.plist`.

However, I also noticed that for `--analyzer-output html` the result will be written to a `.html` file in a directory, e.g. `example.plist/report-141d7e.html` if compiling `example.c`. This has to be taken care of, for instance by simply not supporting the `--analyzer-output html` case and return `Statistic::unsupported_compiler_option` then.

jrosdahl (changes requested):
Thanks for the PR.

src/argprocessing.cpp
mbel0: fixed

src/argprocessing.cpp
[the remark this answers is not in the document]
mbel0: thanks for advice. fixed

jrosdahl:
Thanks.

Issue #1347, opened 2023-11-03
clang-cl wrongly detected when a symlink to clang


DarkShadow44:
### How to reproduce ###
1. Use a clang-cl binary that has clang-cl symlinked to clang
2. Enable ccache logging
3. Check log for detected compiler

### Actual behavior ###
clang-cl is wrongly detected as clang

### Expected behavior ###
clang-cl should be detected as clang

### Environment ###
ccache version 4.8.3
Also tried https://github.com/ccache/ccache/commit/e51e10edee8caf88256f6fcfcbfa5a06172443c4

### Other ###
I figured out what causes this: https://github.com/ccache/ccache/blob/e51e10edee8caf88256f6fcfcbfa5a06172443c4/src/ccache.cpp#L256-L268
This part will resolve all symlinks, leading to the information that it is clang-cl being lost.

I'll add a PR soon. My solution was to just check the original compiler name for "clang-cl". I don't expect a symlink called clang-cl being supposed to be normal clang, I do think the info in the symlink name is important.

PR #1348, opened 2023-11-03
fix: Detect symlink called clang-cl as clang-cl insted of normal clang

DarkShadow44:
This is my proposed fix for #1347.

clang-cl should still be detected as clang-cl when it's a symlink to clang.


jrosdahl:
Thanks for the PR.

I would like to generalize this to first match the unresolved name and then match the resolved name if the first results in `CompilerType::other`. But I can do that as a follow-up.

Issue #1352, opened 2023-11-09
add support for Texas Instruments c6000 compiler


bence42:
The TI c6000 compiler is currently not officially supported, and also does not work unofficially, even partially.
The main reason is the strange command line:

| gcc | TI             |
|-----|----------------|
| -c  | --compile_only |
| -o  | --output_file  |
| -E  | --preproc_only |
| -I  | --include_path |

https://www.ti.com/lit/ug/sprui04f/sprui04f.pdf?ts=1699474432748

PR #1353, opened 2023-11-09
feat: Handle Texas Instruments c6000 compiler arguments (#1352)

bence42:
This PR adds basic support for TI c6000 compiler.
With this change ccache is able to produce a direct hit, and fall back to preprocessor mode too. "Called for link" is not an issue as the TI linker is a separate executable (lnk6x).


bence42:
@jrosdahl Could you please check this and maybe guide me a little here? This is my first contribution.

jrosdahl:
Thanks for the PR.

I'm unlikely to accept this until #956 is done, sorry.

PR #1359, opened 2023-11-15
Use GHA feature in build.yaml

szepeviktor:
Split "Install dependencies" step into two.

Review without indentation change: https://github.com/ccache/ccache/pull/1359/files?diff=unified&w=1

`apt` is an interactive tool, not guaranteed to run without a terminal.


jrosdahl:
Thanks.

PR #1360, opened 2023-11-15
Remove matrix.config from build.yaml

szepeviktor:
Actually, there is no such thing as `matrix.config`.
https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstrategymatrix

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -20,12 +20,12 @@ jobs:
     env:
       CMAKE_GENERATOR: Ninja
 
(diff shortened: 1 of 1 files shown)


jrosdahl:
> Actually, there is no such thing as `matrix.config`.

Of course there is since the current configuration works, so I guess `matrix.config` has been deprecated some time after we started using it. Good cleanup nevertheless.

szepeviktor:
> Of course there is since the current configuration works

Do not believe your eyes 🙃 facts are ephemeral, standards/schemas prevail.

PR #1380, opened 2024-01-16
feat: implement `--print-stats-json`

kaspar030:
This variant of `--print-stats` outputs the statistics as a one line json object.

I did a fair share of ad-hoc conversion of the regular "--print-stats" output, suffering the startup time of Python, thinking "this would be so easy at the source" ...

### DIFF
--- a/src/core/Statistics.cpp
+++ b/src/core/Statistics.cpp
@@ -548,6 +548,32 @@ Statistics::format_machine_readable(const Config& config,
   return util::join(lines, "");
 }
 
+std::string
+Statistics::format_json(const Config& config,
+                        const util::TimePoint& last_updated) const
+{
+  std::vector<std::string> lines;
+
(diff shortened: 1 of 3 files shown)


kaspar030:
I do have two more changes:

- one adding `--print-log-stats` (`--show-log-stats` but in `--print-stats` format),
- one adding `--print-log-stats-json` (`--show-log-stats` but in JSON format).

Maybe it makes sense to have them all in this PR?

jrosdahl:
Thanks for the MR.

Adding a `--print-log-stats` companion to `--show-log-stats` sounds good. However,  I suggest adding a `--format FORMAT` option (where format can be json or tab, defaulting to tab) instead of duplicating the `--print-*` options. Does that make sense to you?

kaspar030:
> Does that make sense to you?

It does! I'll PR the `--print-log-stats` seperately, then add `--format=...`.

PR #1381, opened 2024-01-18
feat: add `--print-log-stats`

kaspar030:
This adds a "machine readable" variant of `--show-log-stats`, similar to `--print-stats` as a variant of `--show-stats`.

See https://github.com/ccache/ccache/discussions/1379 and https://github.com/ccache/ccache/pull/1380.


jrosdahl:
Thanks.

PR #1382, opened 2024-01-18
feat: add `--print-stats` format selection, JSON format

kaspar030:
This PR adds infrastructure to have selectable output format for `--print-stats`/`--print-log-stats`, and adds JSON format in addition to the previous (now default) tab separated format.

*edit* this is based on #1381

See #1379, #1380, #1381.

### DIFF
--- a/src/core/Statistics.cpp
+++ b/src/core/Statistics.cpp
@@ -529,6 +529,8 @@ Statistics::format_machine_readable(const Config& config,
                                     const StatisticFormat format) const
 {
   switch (format) {
+  case StatisticFormat::Json:
+    return Statistics::format_json(config, last_updated);
   case StatisticFormat::Tab:
     /* fall through */
   default:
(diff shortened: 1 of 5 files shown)


kaspar030:
I'm wondering if the one-line json should end with `\n`.

src/core/Statistics.cpp
```diff
@@ -548,6 +563,32 @@ Statistics::format_machine_readable(const Config& config,
...
+    if (!(field.flags & FLAG_NEVER)) {
+      add_line(field.id, m_counters.get(field.statistic));
+    }
+  }
+
+  add_line("max_cache_size_kibibyte", config.max_size() / 1024);
+  add_line("max_files_in_cache", config.max_files());
+
+  std::sort(lines.begin(), lines.end());
+  std::string result = "{" + util::join(lines, ",") + "}";
```
jrosdahl, line 588: > I'm wondering if the one-line json should end with \n.

Yes, I would prefer ending with a newline. I also wouldn't mind line breaks after statistics lines to make it more human readable.

src/core/mainoptions.cpp
```diff
@@ -168,11 +168,16 @@ Options for scripting or debugging:
                                PATH
         --extract-result PATH  extract file data stored in result file at PATH
                                to the current working directory
+    -f, --format FORMAT        specify format for --print-log-stats and
```
jrosdahl, line 171: Since this is used with a long option anyway, and typically only from other programs (not interactively), I suggest removing the short `-f` form. That would save `-f` for future options that might be more commonly used than specifying the format.

jrosdahl (commented):
Looks good in general!

jrosdahl (commented):
Oh, and documentation of `--format` seems to be missing in `MANUAL.adoc`.

src/core/mainoptions.cpp
kaspar030: Agreed!

src/core/Statistics.cpp
kaspar030: Done! I've also added two spaces of indentation.

Personally I'd actually prefer a one-line "compact" version. Would that justify another `--format` choice, "json-compact" (or "json-pretty" and make "json" compact")? E.g., `jq` has `-c`/`--compact-output` which doesn't pretty-print.

(But I'm also fine with piping through `jq -c` to get my one-liner. :) )

kaspar030:
(this needs squashing now)

src/core/Statistics.cpp
jrosdahl: Thanks. Let's try to live with one alternative. :slightly_smiling_face:

jrosdahl:
Thanks.

kaspar030:
> Thanks.

Thanks for reviewing!

PR #1401, opened 2024-02-10
feat: add support for -fcallgraph-info

### LINKED ISSUE #1400: GCC 10+ `-fcallgraph-info` files are not cached
GCC version 10.1 and above supports the `-fcallgraph-info` argument, which outputs a `.ci` file alongside the object file that contains callgraph information. This is very useful for things like stack usage analysis. See https://gcc.gnu.org/onlinedocs/gcc-10.1.0/gcc/Developer-Options.html. This file is not cached by `ccache` and does not cause it to fall back as an unsupported option.

### How to reproduce ###

1. `echo 'int main() { return 0; }' > x.c`
2. `ccache gcc -c -fcallgraph-info x.c`
3. `mv x.ci x-original.ci`
4. `ccache gcc -c -fcallgraph-info x.c`
5. `diff x.ci x-original.ci `

### Actual behavior ###
The `.ci` file is not restored and the output is `diff: x.ci: No such file or directory`.

### Expected behavior ###
The `.ci` file is restored and `diff` reports no changes. Ideally, this would be fetching from the cache in a similar fashion to `-fstack-usage` and `.su` files.

### Environment ###
ccache version 4.8.0

GCC version 10.1 or above, running on Linux

res0nance:
Fixes #1400


jrosdahl:
Thanks for implementing this! I'll apply a bug fix first (treat `-fcallgraph-info` as too hard) in 4.9.2 and let this feature target 4.10.

src/argprocessing.cpp
```diff
@@ -815,6 +815,12 @@ process_option_arg(const Context& ctx,
     return Statistic::none;
   }
 
+  if (arg == "-fcallgraph-info") {
```
jrosdahl, line 818: It would be nice to support `-fcallgraph-info=...` as well while we're at it.

PR #1404, opened 2024-02-12
WIP: Allow for building with system versions of many third_party libraries

### LINKED ISSUE #1396: Build with system libraries
Somehow I'm just discovering the ccache bundles a lot of libraries.  Fedora packaging guidelines discourage this and strongly suggest that packages build with the system versions of libraries if possible.  At the moment ccache does not allow for this.  Would you be receptive to patches that would make this easier?

opoplawski:
This is a start on allowing ccache to be built with system versions of libraries currently in src/third_party.  Fixes #1396 

I think I also need to address getopt_long.c.  I suspect that there are many non-standard / extra build options that haven't been addressed yet as well.

This imports more of the full fmt distribution in order to use it's cmake build system.  We could go back to a slimmed down version I think if we wanted to explicitly enable header-only mode.


opoplawski:
Are there systems out there that you still support that don't have getopt_long?  Nothing in your CI suite seems to need it.

opoplawski:
Any comments?

jrosdahl:
Certainly, I'm usually answering as soon as I can but not sooner. :grin:

> Are there systems out there that you still support that don't have
> getopt_long? Nothing in your CI suite seems to need it.

We support more than what's tested in CI, for instance [_"other Unix-like platforms"_](https://ccache.dev/platform-compiler-language-support.html). That said, I'm not aware of any reasonably modern system that lacks support. There is probably no other way to find out than to remove it and see if anybody reacts. I'll do that.

Before going further on this PR I think it would be a good idea to discuss the overall strategy. Let's do that in #1396.

jrosdahl:
Superseded by #1429.

Issue #1407, opened 2024-02-25
Implement -fprofile-update=atomic


xTachyon:
Hello. I was looking through the stats and I noticed `Uncacheable calls => Unsupported compiler option`. While looking through the logs I found this line:
`Unknown profiling option: -fprofile-update=atomic`

Compiling without that option seems to not be uncacheable anymore, so this seems to be my problem.

Any plans to support that flag?

PR #1408, opened 2024-02-27
feat: add -fprofile-update support

res0nance:
Fixes https://github.com/ccache/ccache/issues/1407

This option AFAICT should be simple as it only affects how the profile is updated


jrosdahl:
Thanks.

Issue #1415, opened 2024-03-17
ccache lockfiles do not work in msys2


Kreijstal:
### How to reproduce ###

1. git clone master in msys2
2.  compile
3.  run test suite

### Actual behavior ###

as discussed on #1412
lock loops hangs forever.

### Expected behavior ###

This loop should eventually exit, but it hangs forever, or test should pass.

### Environment ###

ccache version 68c7decaaa1c808ac77636714ab02f3343c7d743 (git master)

msys2 enviroment, no mingw
windows 11
msys2 already knows this https://github.com/msys2/MSYS2-packages/issues/4449

jrosdahl:
Thanks for the bug report.

> msys2 already knows this https://github.com/msys2/MSYS2-packages/issues/4449

I followed the trail and see that you wrote this in https://github.com/msys2/MSYS2-packages/issues/4311#issuecomment-1987177326:

> Ok, so what I've discovered is that ccache tests symlinks, and Msys implements them as copying. Which confuses the test suite.

It's not the test suite that gets confused – it points out that the lockfile implementation is broken on msys2 since is relies on working POSIX semantics for symlinks. Since the lockfile implementation does not work on msys2, ccache does not work properly on msys2 even if you avoid running the test suite.

The fix would be to either implement proper symlink support on msys2 or to implement lockfiles differently on msys2. And since there is already a [proper Windows implementation in place](https://github.com/ccache/ccache/blob/68c7decaaa1c808ac77636714ab02f3343c7d743/src/ccache/util/LockFile.cpp#L357) I suspect that the latter would be easy to do.

jrosdahl:
@Kreijstal: Could you try if latest ccache master works better?

Kreijstal:
```
/home/topkek/git/ccache/src/ccache/util/wincompat.hpp:67:12: fatal error: direct.h: No such file or directory
   67 | #  include <direct.h>
      |            ^~~~~~~~~~
compilation terminated.
```
That's okay, I'll need to figure out how to access win32 from cygwin/msys2, but it seems promising.
https://cygwin.com/cygwin-ug-net/using.html#pathnames-win32-api
I think I'll try solving this issue with this knowledge, thank you.

jrosdahl:
@Kreijstal: Try what happens with

```diff
--- a/src/ccache/util/wincompat.hpp
+++ b/src/ccache/util/wincompat.hpp
@@ -18,7 +18,7 @@
 
 #pragma once
 
[...]

PR #1416, opened 2024-03-17
msys2 patch: lock mechanism uses file instead of symlink.

Kreijstal:
MSYS2 will not let you create a symlink to a nonexistent file, only to an already existing file. To overcome this issue we use files, and write content on it. 

For Cygwin/MSYS2, the patch leverages POSIX `open` with the flags `O_WRONLY | O_CREAT | O_EXCL`, effectively creating an exclusive lock file. As suggested by @jrosdahl , thank you so much for the suggestion. 
This still does not pass the test suite, but makes ccache usable.
I look forward for you review.

See issue #1415

### DIFF
--- a/src/ccache/util/LockFile.cpp
+++ b/src/ccache/util/LockFile.cpp
@@ -43,7 +43,9 @@ const uint32_t k_max_sleep_time_ms = 50;
 #ifndef _WIN32
 const util::Duration k_staleness_limit(2);
 #endif
-
+#ifdef __CYGWIN__
+#  include <fcntl.h>
+#endif
 namespace fs = util::filesystem;
 
 using pstr = util::PathString;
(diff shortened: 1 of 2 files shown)


Kreijstal:
Ah, I see some errors in the test suite, I'll try to correct that first.

src/ccache/util/LockFile.cpp
```diff
@@ -236,16 +238,31 @@ LockFile::do_acquire(const bool blocking)
...
       // We got the lock.
       return true;
     }
+#  endif
 
     int saved_errno = errno;
     if (saved_errno == ENOENT) {
       // Directory doesn't exist?
+#  ifdef __CYGWIN__
+      if (!fs::exists(m_lock_file.parent_path())
```
jrosdahl, line 261: For my own understanding: Can `open(..., ... O_CREAT ...)` set `errno` to `ENOENT` in msys2 for other reasons than _"a directory component in pathname does not exist or is a dangling symbolic link"_? If not, I guess the `fs::exists` call is superfluous?

jrosdahl (commented):
Looks reasonable.

src/ccache/util/LockFile.cpp
Kreijstal: The fs::exists was added because it solves the infinite loop problem, apparently in some cases, it returns `true` even when the path exists.

jrosdahl:
Thanks.

PR #1418, opened 2024-03-20
Read file in do_hash_file as Bytes and optimize read_file size_hint on Windows

clemenswasser:
Reading the file in do_hash_file as Bytes skips CRLF to LF conversion and is therefore way faster.
Optimize read_file size_hint on Windows by getting the file size of the existing handle instead of opening a duplicate one.

### DIFF
--- a/src/ccache/hashutil.cpp
+++ b/src/ccache/hashutil.cpp
@@ -194,7 +194,7 @@ do_hash_file(const Context& ctx,
   (void)ctx;
 #endif
 
-  const auto data = util::read_file<std::string>(path, size_hint);
+  const auto data = util::read_file<util::Bytes>(path, size_hint);
   if (!data) {
     LOG("Failed to read {}: {}", path, data.error());
     return HashSourceCodeResult(HashSourceCode::error);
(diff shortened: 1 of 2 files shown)


clemenswasser:
Some performance numbers for a completely cached CMake clean rebuild (includes linking time, so impact is bigger than it seems)
before:
```console
> hyperfine -p 'cmake --build .\build-msvc-debug\ -t clean' 'cmake --build .\build-msvc-debug\'
Benchmark 1: cmake --build .\build-msvc-debug\
  Time (mean ± σ):      7.530 s ±  0.110 s    [User: 14.683 s, System: 10.512 s]
  Range (min … max):    7.304 s …  7.690 s    10 runs
```
after:
```console
> hyperfine -p 'cmake --build .\build-msvc-debug\ -t clean' 'cmake --build .\build-msvc-debug\'
Benchmark 1: cmake --build .\build-msvc-debug\
  Time (mean ± σ):      7.232 s ±  0.099 s    [User: 10.662 s, System: 7.441 s]
  Range (min … max):    7.096 s …  7.390 s    10 runs
```

clemenswasser:
I've used the `.ninja_log` file to look how much this saves without linking, etc., by summing all durations of the obj files.
Before: 61737ms
After: 53393ms
So for a completely cached CMake rebuild, ccache is now ~1.16 times faster on Windows :^)

jrosdahl:
Thanks.

Issue #1421, opened 2024-04-07
Potential ways to improve the .incbin support (or lack thereof)


bartdesmet:
First of all, thanks for the work on `ccache`. We've successfully rolled it out in our development environment, both on local developer machines and in a distributed CI environment (using `bazel-remote`), and it's been a great time-saver and productivity booster thus far.

While onboarding `ccache` in our fairly big code base, we stumbled upon two issues that caused lack of caching, and we've been able to address both.

The first issue was with files generated using `protoc` that get compiled a few milliseconds after being generated. This caused `ccache` to be conversative due to its timestamp checks, so I used the sloppiness options to suppress that behavior.

The second issue is with the use of `.incbin` directives in our source code. Unlike issues reported by other folks, our use of `.incbin` may be a bit more esoteric. We use `.incbin` as part of a header used to embed GDB debugger scripts in binaries to enhance our developers' debugging experience with custom components:

```cpp
#define DEFINE_GDB_PY_SCRIPT(script_name, file_name)                                                                   \
    asm(R"(
.pushsection ".debug_gdb_scripts", "MS", @progbits, 1
.byte 4 /* Python inlined */
.ascii ")" script_name R"(
"
.incbin ")" file_name R"("
.byte 0
.popsection
)");
```

The macro is used in a variety of our libraries, some of which are header-only (for template-based types) and end up getting included in a lot of compilation units. This resulted in `ccache` suppressing caching behavior for many of these files.

Our use of embedded GDB scripts is intentional to streamline the debugging experience without having to package and distribute debugger extension scripts and have developers set up the GDB environment to load these scripts.
 
I worked around this by introducing another build flag that toggles a `-D` flag to `#define NO_GDB_EXTENSIONS`, which is then used by the header containing the macro above to omit the `asm` section altogether. Needless to say, this then hampers the debugger experience because scripts get omitted, in exchange for a better build performance.

This got me thinking about possible enhancements in this space.

First, I was a little surprised there's no `#define` set by `ccache` when running the preprocessor in order to detect that it's processing the code. The reason I'm saying "surprised" is because tools like `clang-tidy` set `__clang_analyzer__` to detect presence of the tool. Quoting their docs:

> When the static analyzer is using Clang to parse source files, it implicitly defines the preprocessor macro __clang_analyzer__. While discouraged, code can use this macro to selectively exclude code the analyzer examines.

[...]

PR #1423, opened 2024-04-09
feat: add sloppiness for incbin handling

res0nance:
Code to address some issues in #1421

### DIFF
--- a/doc/MANUAL.adoc
+++ b/doc/MANUAL.adoc
@@ -1045,6 +1045,11 @@ preprocessing first.
     different directories, with the tradeoff of potentially getting an incorrect
     directory in the `.gcno` file. *gcno_cwd* also disables hashing of the
     current working directory if `-fprofile-abs-path` is used.
+*incbin*::
+    By default, ccache will ignore all files containing an `.incbin` directive.
+    While this is the correct behaviour as ccache does not detect incbin changes,
+    this restriction can make some projects difficult to cache. This sloppiness
+    will pretend the `.incbin` directive doesn't exist and simply allow caching.
 *include_file_ctime*::
     By default, ccache will disable caching if a source code file has a status
     change time (ctime) after the start of the ccache invocation. This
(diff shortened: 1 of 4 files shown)


bartdesmet:
I've just confirmed that the `incbin` sloppiness setting does indeed work for our codebase:

```
Found potential unsupported .incbin directive in source code.But continuing due to enabled sloppy incbin handling
```

One nit is to clean up the message a little bit:

```
Found potential unsupported .incbin directive in source code but continuing due to enabled sloppy incbin handling
```

It may also be useful, though likely a longer "feat", to have warning/error codes that get emitted in traces for all possible error paths. Those could be cryptic (but documented) numeric error codes a la `WRN0016` or `ERR0017`, or be somewhat readable alphanumeric strings a la `error-unsupported-incbin-directive` (whatever feels more natural for the ecosystem in which `ccache` operates, e.g. do what gcc and clang tend to do for error reporting).

This would allow to `grep` on something unique versus a random English string that may be altered when spelling or grammar gets corrected and whatnot. That way, we can automate log scanning steps, and potentially even wire up detected warnings or errors to cause a build failure, so such things don't go unnoticed silently.

In fact, there could even be `ccache` command line arguments to do warning-as-error behavior etc.

doc/MANUAL.adoc
```diff
@@ -1045,6 +1045,11 @@ preprocessing first.
     different directories, with the tradeoff of potentially getting an incorrect
     directory in the `.gcno` file. *gcno_cwd* also disables hashing of the
     current working directory if `-fprofile-abs-path` is used.
+*incbin*::
+    By default, ccache will ignore all files containing an .incbin directive.
+    While this is the correct behaviour as ccache does not detect incbin changes.
+    This restriction can make some projects difficult to cache. This sloppiness
```
jrosdahl, line 1051:
```suggestion
    By default, ccache will ignore all files containing an `.incbin` directive.
    While this is the correct behaviour as ccache does not detect incbin changes,
    this restriction can make some projects difficult to cache. This sloppiness
```

jrosdahl (approved):

PR #1426, opened 2024-04-10
ci: disable building test on macos binaries

res0nance:
We don't run the tests when building macOS universal binaries as its covered elsewhere so we can save some compile time by removing test builds.


jrosdahl:
Thanks.

PR #1427, opened 2024-04-10
feat: use sendfile(2) for file copying

res0nance:
Use sendfile for file copying as it avoids copying the data to user-space and back to kernel space.

### DIFF
--- a/src/ccache/util/file.cpp
+++ b/src/ccache/util/file.cpp
@@ -66,6 +66,10 @@
 #  include <windows.h>
 #endif
 
+#ifdef HAVE_SYS_SENDFILE_H
+#  include <sys/sendfile.h>
+#endif
+
 namespace fs = util::filesystem;
 
 using pstr = util::PathString;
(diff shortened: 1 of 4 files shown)


src/ccache/util/file.cpp
```diff
@@ -66,6 +66,10 @@
 #  include <windows.h>
 #endif
 
+#ifdef __linux__
+#  include <sys/sendfile.h>
+#endif
```
jrosdahl, line 71: There are more systems than Linux that have `sys/sendfile.h`. I suggest adding a check in `cmake/GenerateConfigurationFile.cmake` + `cmake/config.h.in` and then use `#ifdef HAVE_SYS_SENDFILE_H` instead.

src/ccache/util/file.cpp
```diff
@@ -100,9 +104,25 @@ copy_file(const fs::path& src, const fs::path& dest, ViaTmpFile via_tmp_file)
         FMT("Failed to open {} for writing: {}", dest, strerror(errno)));
     }
   }
+
+#if defined(__linux__)
+  DirEntry dir_entry(src, *src_fd);
+  if (!dir_entry) {
+    return tl::unexpected(FMT("Failed to stat: {}", src));
```
jrosdahl, line 111:
```suggestion
    return tl::unexpected(FMT("Failed to stat {}: {}", src, strerror(errno));
```

src/ccache/util/file.cpp
```diff
@@ -100,9 +104,25 @@ copy_file(const fs::path& src, const fs::path& dest, ViaTmpFile via_tmp_file)
...
+  }
+  ssize_t bytes_left = dir_entry.size();
+  while (bytes_left > 0) {
+    ssize_t n = sendfile(*dest_fd, *src_fd, nullptr, bytes_left);
+    if (n < 0) {
+      return tl::unexpected(FMT("Failed to copy: {} to {}", src, dest));
+    }
```
jrosdahl, line 118: The [sendfile man page](https://man7.org/linux/man-pages/man2/sendfile.2.html) says:

> Applications may wish to fall back to read(2) and write(2) in the case where sendfile() fails with EINVAL or ENOSYS.

Do you think that would make sense?

unittest/test_util_file.cpp
```diff
@@ -71,23 +71,26 @@ TEST_CASE("util::likely_size_on_disk")
   CHECK(util::likely_size_on_disk(4097) == 8192);
 }
 
-TEST_CASE("util::read_file and util::write_file, text data")
+TEST_CASE("util::read_filem util::write_file and util::copy_file, text data")
```
jrosdahl, line 74:
```suggestion
TEST_CASE("util::read_file, util::write_file and util::copy_file, text data")
```

PR #1428, opened 2024-04-11
bump: update blake3 to v1.5.1

res0nance:
Update our dependencies


jrosdahl (commented):
Thanks. There is one more thing to do: update version here:

https://github.com/ccache/ccache/blob/8916d30837c5c56de6011ed7a188a9025890ba0e/LICENSE.adoc?plain=1#L53

PR #1431, opened 2024-04-14
perf: use native windows copyfile

res0nance:
Copyfile improvement for windows

### DIFF
--- a/src/ccache/util/file.cpp
+++ b/src/ccache/util/file.cpp
@@ -80,8 +80,34 @@ using pstr = util::PathString;
 
 namespace util {
 
-tl::expected<void, std::string>
-copy_file(const fs::path& src, const fs::path& dest, ViaTmpFile via_tmp_file)
+#ifdef _WIN32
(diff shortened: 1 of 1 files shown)


src/ccache/util/file.cpp
```diff
@@ -138,16 +153,16 @@ copy_file(const fs::path& src, const fs::path& dest, ViaTmpFile via_tmp_file)
...
 #  if defined(HAVE_SYS_SENDFILE_H)
-  }
+}
 #  endif
-#endif
 
-  dest_fd.close();
-  src_fd.close();
+dest_fd.close();
+src_fd.close();
```
jrosdahl, line 164: I don't mind the improvement, but I'm hesitant to accept it in its current form since clang-format apparently gets lost in the maze of twisty little ifdefs. Is it possible to restructure to a less twisty form?

Also, note that `copy_file` is only used in [hard link mode](https://ccache.dev/manual/4.9.1.html#config_hard_link) so it's only relevant in a niche use case.

PR #1451, opened 2024-05-13
perf: Use posix_spawn when hashing

mrpippy:
This should improve performance on all OSes. I thought I would convert hashing first and then `execute()` next, since it's slightly more complicated.

One question I have is whether there should be error handling on the `posix_spawn_file_actions_add*` functions, these seem very unlikely to fail (similar to `close` or `dup2` calls).

### DIFF
--- a/cmake/GenerateConfigurationFile.cmake
+++ b/cmake/GenerateConfigurationFile.cmake
@@ -4,6 +4,7 @@ set(include_files
     dirent.h
     linux/fs.h
     pwd.h
+    spawn.h
     sys/clonefile.h
     sys/file.h
     sys/ioctl.h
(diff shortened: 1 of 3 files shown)


jrosdahl:
Sure, why not.
 
> One question I have is whether there should be error handling on the `posix_spawn_file_actions_add*` functions, these seem very unlikely to fail (similar to `close` or `dup2` calls).

Even if they never fail on common systems, perhaps it's good to check return values anyway to catch any problems on non-standard platforms or libc implementations?

mrpippy:
> Sure, why not.
> 
> > One question I have is whether there should be error handling on the `posix_spawn_file_actions_add*` functions, these seem very unlikely to fail (similar to `close` or `dup2` calls).
> 
> Even if they never fail on common systems, perhaps it's good to check return values anyway to catch any problems on non-standard platforms or libc implementations?

True, it's simple to add, just pushed a new version.

PR #1453, opened 2024-05-17
fix: TAKES_/CONCAT_ARG for msvc options

### LINKED ISSUE #1452: Wrong TAKES_/CONCAT_ARG for msvc options
During my latest patches I noticed that many `msvc` related options have incorrectly defined `TAKES_ARG` and `TAKES_CONCAT_ARG`.

I checked all these options like `TAKES_ARG` can contain space between the option and option value, and `TAKES_CONCAT_ARG` can't contain space between. Then I verified this at msvc [Compiler options listed](https://learn.microsoft.com/en-us/cpp/build/reference/compiler-options-listed-alphabetically?view=msvc-170).

And this is what I noted down:

```
/Fp /Yu (remove TAKES_ARG, space is not allowed)
/AI (OK, works with space even docs doesn't state it and syntax is as: /AIfile)
/FU (OK, docs doesn't specify, but Syntax is described as: /FU file)
/FI /external:I /Yc (OK)
/Tc /Tp (add TAKES_CONCAT_ARG, space is optional)
```

I also tested `/Fp` and `/Yu` options using the cl.exe and the compiler throws an error if there is a space between.

Did I miss something or am I right?

silverqx:
/Fp and /Yu doesn't allow space between option and option value.

Fixes #1452

### DIFF
--- a/src/ccache/compopt.cpp
+++ b/src/ccache/compopt.cpp
@@ -74,7 +74,7 @@ const CompOpt compopts[] = {
   {"-F", AFFECTS_CPP | TAKES_ARG | TAKES_CONCAT_ARG | TAKES_PATH},
   {"-FI", AFFECTS_CPP | TAKES_ARG | TAKES_CONCAT_ARG | TAKES_PATH}, // msvc
   {"-FU", AFFECTS_CPP | TAKES_ARG | TAKES_CONCAT_ARG | TAKES_PATH}, // msvc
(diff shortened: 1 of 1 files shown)


silverqx:
I had to remove `TAKES_CONCAT_ARG` for /Tc and /Tp options because they are also tagged as `TOO_HARD` and because of this, the `TEST_CASE("option_table_should_be_sorted")` with `compopt_verify_sortedness_and_flags()` call fails.

jrosdahl:
Thank!

silverqx:
Thx for review.

PR #1454, opened 2024-05-17
feat: Add print-version option

silverqx:
I have added a new `--print-version` option as ccache will support PCH on msvc and I will need to somehow detect the ccache version to correctly decide if PCH should be disabled (`CMAKE_DISABLE_PRECOMPILE_HEADERS`). 

So if ccache >=4.10 then don't disable PCH and if it's below then set `CMAKE_DISABLE_PRECOMPILE_HEADERS=ON`.


jrosdahl:
Sounds good to make it easier to script things.

Suggestions:

- Other options for scripting are placed under _Options for scripting or debugging:_, so move the option to that section.
- There are no other options named "dump", but `--print-*` is used for options that print machine-parsable data, so I suggest calling it `--print-version`.
- Also print a newline character after the version.

jrosdahl:
Thanks.

PR #1459, opened 2024-05-27
fix: small fixes

silverqx:
Small fixes discovered during reviews.

### DIFF
--- a/src/ccache/util/wincompat.hpp
+++ b/src/ccache/util/wincompat.hpp
@@ -29,8 +29,8 @@
 // From:
 // http://mesos.apache.org/api/latest/c++/3rdparty_2stout_2include_2stout_2windows_8hpp_source.html
 #  ifdef _MSC_VER
-const mode_t S_IRUSR = mode_t{_S_IREAD};
-const mode_t S_IWUSR = mode_t{_S_IWRITE};
+constexpr auto S_IRUSR = mode_t{_S_IREAD};
+constexpr auto S_IWUSR = mode_t{_S_IWRITE};
 #  endif
 
 #  ifndef S_IFIFO
--- a/src/ccache/InodeCache.cpp
+++ b/src/ccache/InodeCache.cpp
@@ -155,7 +155,7 @@ fd_is_on_known_to_work_file_system(int fd)
   }
   return known_to_work;
 #else
-  HANDLE file = (HANDLE)_get_osfhandle(fd);
+  HANDLE file = reinterpret_cast<HANDLE>(_get_osfhandle(fd));
   if (file == INVALID_HANDLE_VALUE) {
     return false;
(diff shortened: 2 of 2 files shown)


src/ccache/util/wincompat.hpp
```diff
@@ -29,8 +29,8 @@
 // From:
 // http://mesos.apache.org/api/latest/c++/3rdparty_2stout_2include_2stout_2windows_8hpp_source.html
 #  ifdef _MSC_VER
-const mode_t S_IRUSR = mode_t{_S_IREAD};
-const mode_t S_IWUSR = mode_t{_S_IWRITE};
+constexpr auto S_IRUSR = mode_t{_S_IREAD};
+constexpr auto S_IWUSR = mode_t{_S_IWRITE};
```
silverqx, line 33: No problem to revert it if you don't like it.

src/ccache/ccache.cpp
```diff
@@ -2492,10 +2492,6 @@ do_cache_compilation(Context& ctx)
 
   ProcessArgsResult processed = process_args(ctx);
 
-  if (processed.error) {
-    return tl::unexpected(*processed.error);
-  }
-
```
silverqx, line 2498: This condition can never be false, it's useless? If you look [here](https://github.com/ccache/ccache/blob/master/src/ccache/argprocessing.cpp#L1747-L1752) and [here](https://github.com/ccache/ccache/blob/master/src/ccache/argprocessing.hpp#L34-L41).

silverqx (commented):
I added comments to the modified code.

src/ccache/ccache.cpp
jrosdahl: You seem to have missed the https://github.com/ccache/ccache/blob/baea34b407f5f4f036fce8ce6f5d7873c572467f/src/ccache/argprocessing.hpp#L33 constructor, for instance invoked here: https://github.com/ccache/ccache/blob/baea34b407f5f4f036fce8ce6f5d7873c572467f/src/ccache/argprocessing.cpp#L1286

jrosdahl: The construction comes from before `tl::expected` was introduced. I'll refactor it to use `tl::expected` instead.

silverqx: I see there are 13 of them 😆 I totally missed it, the method is so long.

silverqx: I have reverted this change.

jrosdahl (approved):

src/ccache/ccache.cpp
jrosdahl: > I'll refactor it to use tl::expected instead.

Done in 181ffcff3186.

PR #1471, opened 2024-06-15
perf: Use posix_spawn to execute the compiler

mrpippy:
Like the change in #1451, this should improve performance on all OSes.

### DIFF
--- a/src/ccache/execute.cpp
+++ b/src/ccache/execute.cpp
@@ -41,6 +41,10 @@
 
 #include <vector>
 
+#ifdef HAVE_SPAWN_H
+#  include <spawn.h>
+#endif
+
 #ifdef HAVE_UNISTD_H
 #  include <unistd.h>
 #endif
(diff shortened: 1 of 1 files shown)


src/ccache/execute.cpp
```diff
@@ -299,30 +303,46 @@ execute(Context& ctx,
...
-    exit(execv(argv[0], const_cast<char* const*>(argv)));
+  pid_t pid;
+  extern char** environ;
+  result = posix_spawn(&pid,
+                       argv[0],
+                       &file_actions,
+                       nullptr,
+                       const_cast<char* const*>(argv),
+                       environ);
+
```
jrosdahl, line 331: If the process receives a signal here (after `posix_spawn` but before `SignalHandlerBlocker` then the child won't be killed by the signal handler. Unless there's a problem with doing so I suggest moving `posix_spawn` to inside the signal handler blocker block, just like the `fork` call was before.

mrpippy: Good point, fixed.

jrosdahl:
Thanks.

PR #1475, opened 2024-06-26
fix: Add the --compile option to be compliant with nvcc

Labouh:
The nvcc (CUDA) compiler has a extra option compliant with -c: --compile. https://docs.nvidia.com/cuda/cuda-compiler-driver-nvcc/#supported-phases

So ccache should see the --compile option as the -c option (especially because the CUDA MVS plugin use it!)


jrosdahl:
Thanks.

PR #1496, opened 2024-08-14
feat: Add ability to send arbitrary http headers to http storage

dsrowell:
This change adds the ability to send arbitrary http headers to the http backend storage. There are two main use cases:
1. An http server may require certain headers in order to process files correctly. For example, a server might require "content-type: application/octet-stream" to handle binary data.
2. It could be useful to collect metadata, such as the git commit id of branch being built, or the merge-base. This could be used to optimize the backend storage and/or collect metrics on usage.

### DIFF
--- a/doc/MANUAL.adoc
+++ b/doc/MANUAL.adoc
@@ -1309,6 +1309,9 @@ values.
 --
 +
 The default is *subdirs*.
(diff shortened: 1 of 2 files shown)


jrosdahl:
Thanks!

PR #1503, opened 2024-08-27
chore: Use ASCII chars

huangqinjin:
MSVC uses locale encoding by default. The UTF-8 chars cause
```
warning C4819: The file contains a character that cannot be represented in the current code page (936).
Save the file in Unicode format to prevent data loss.
```

### DIFF
--- a/src/ccache/storage/remote/redisstorage.cpp
+++ b/src/ccache/storage/remote/redisstorage.cpp
@@ -32,7 +32,7 @@
 #  include <sys/utime.h> // for timeval
 #endif
 
-// Ignore "ISO C++ forbids flexible array member ‘buf’" warning from -Wpedantic.
+// Ignore "ISO C++ forbids flexible array member 'buf'" warning from -Wpedantic.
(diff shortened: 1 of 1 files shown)


jrosdahl:
Thanks for the PR.

I would prefer to instruct MSVC to accept UTF-8 characters instead. If I understand correctly that can be done with [`/utf-8`](https://learn.microsoft.com/en-us/cpp/build/reference/utf-8-set-source-and-executable-character-sets-to-utf-8?view=msvc-170).

jrosdahl (changes requested):
Thanks for the PR.

I would prefer to instruct MSVC to accept UTF-8 characters instead. If I understand correctly that can be done with [/utf-8](https://learn.microsoft.com/en-us/cpp/build/reference/utf-8-set-source-and-executable-character-sets-to-utf-8?view=msvc-170).

huangqinjin:
Agree. `/utf-8` is best solution. Sent #1508 .

PR #1504, opened 2024-09-06
feat: Option absolute_paths_in_stderr also rewrites relative paths in MSVC diagnostics messages

huangqinjin:


jrosdahl:
Thanks.

PR #1509, opened 2024-09-11
feat: Relativize absolute paths within base_dir in MSVC diagnostics messages

huangqinjin:
MSVC /FC option causes the paths in diagnostics messages become absolute.
Convert them to relative to avoid leakage of workspace paths.
User can specify absolute_paths_in_stderr=true to convert them back to absolute.


jrosdahl:
Thanks.

PR #1510, opened 2024-09-11
fix(http-storage): Make sure 'keep-alive' can be overridden as expected

moha-gh:
The changes introduced in dfcdb68 made it impossible to disable HTTP keep-alive
by setting the 'keep-alive' attribute in the backend string accordingly: The
default of 'true' is applied after parsing of the attributes, overriding the
value configured by the user (if any). Rectify this.


jrosdahl:
Thanks.

PR #1522, opened 2024-10-08
feat: Implement atfile_format config option to control rsp file format

MaxEW707:
This PR implements a feature to fix a ccache issue we ran into.
Currently when a response file is seen on the command line the quoting behaviour is determined by the compiler type.
clang supports both windows and posix quoting via the `--rsp-quoting=[windows|posix]` option.

Due to NDA restrictions I cannot say what vendor we are using which is also why I cannot add a new compiler_type to special case this vendor's fork of clang.
We are building using a vendor clang to cross-compile to the vendor platform on windows. 
This vendor clang is effectively hardcoded with `--rsp-quoting=windows`.
The vendor MSBuild tasks emit response files with only windows quoting behaviour.

For everything else including the driver mode this vendor clang is clang, behaves like clang and uses clang options. It does not use the cl driver mode so clang-cl is not an option here to get ccache to assume `msvc` rsp format.

We need a way to tell ccache to treat response files as `msvc` even though we are using clang.

### DIFF
--- a/src/ccache/args.cpp
+++ b/src/ccache/args.cpp
@@ -18,7 +18,9 @@
 
 #include "args.hpp"
 
+#include <ccache/config.hpp>
 #include <ccache/core/exceptions.hpp>
+#include <ccache/util/assertions.hpp>
 #include <ccache/util/file.hpp>
 #include <ccache/util/logging.hpp>
 #include <ccache/util/string.hpp>
@@ -48,6 +50,8 @@ Args::from_string(std::string_view command)
 std::optional<Args>
 Args::from_atfile(const std::string& filename, AtFileFormat format)
 {
+  ASSERT(format != AtFileFormat::guess_from_compiler);
+
   const auto argtext = util::read_file<std::string>(filename);
   if (!argtext) {
     LOG("Failed to read atfile {}: {}", filename, argtext.error());
--- a/src/ccache/args.cpp
+++ b/src/ccache/args.cpp
@@ -72,6 +72,8 @@ Args::from_atfile(const std::string& filename, AtFileFormat format)
     switch (*pos) {
     case '\\':
       switch (format) {
+      default:
+        break;
       case AtFileFormat::gcc:
         pos++;
         if (*pos == '\0') {
--- a/src/ccache/args.cpp
+++ b/src/ccache/args.cpp
@@ -72,7 +72,8 @@ Args::from_atfile(const std::string& filename, AtFileFormat format)
     switch (*pos) {
     case '\\':
       switch (format) {
-      default:
+      case AtFileFormat::auto_guess:
+        ASSERT(false); // Can't happen
         break;
       case AtFileFormat::gcc:
         pos++;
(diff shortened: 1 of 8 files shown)


src/ccache/config.cpp
```diff
@@ -522,6 +537,21 @@ home_directory()
 #endif
 }
 
+std::string
+atfile_format_to_string(AtFileFormat atfile_format)
+{
+  switch (atfile_format) {
+  case AtFileFormat::guess_from_compiler:
+    return "guess_from_compiler";
```
jrosdahl, line 545: This means that e.g. `ccache -p` can print `atfile_format = guess_from_compiler`, but that's not a valid setting according to `parse_atfile_format`. I suggest either using the empty string as the default or (probably better) making the default value token valid as a setting, similar to how `compiler_type` works. I would then also suggest `auto` instead of `guess_from_compiler` for symmetry with `compiler_type`.

src/ccache/args.cpp
```diff
@@ -68,6 +72,8 @@ Args::from_atfile(const std::string& filename, AtFileFormat format)
     switch (*pos) {
     case '\\':
       switch (format) {
+      default:
+        break;
```
jrosdahl, line 76: Suggestion: list all enum values instead of defaulting, e.g.:

```suggestion
      case guess_from_compiler:
        ASSERT(false); // Can't happen
        break;
```

unittest/test_config.cpp
```diff
@@ -308,6 +308,52 @@ TEST_CASE("Config::update_from_environment")
   CHECK(!config.compression());
 }
 
+TEST_CASE("Config:atfile_format")
```
jrosdahl, line 311:
```suggestion
TEST_CASE("Config::atfile_format")
```

jrosdahl (changes requested):
Sounds reasonable to me.

It would be great if you could also document the new setting in `doc/MANUAL.adoc`.

src/ccache/config.cpp
```diff
@@ -71,6 +71,7 @@ namespace {
 
 enum class ConfigItem {
   absolute_paths_in_stderr,
+  atfile_format,
```
jrosdahl, line 74: Since MSVC calls these files "response files" and GCC AFAIK doesn't have a specific name for them, maybe it would be better to call the setting `response_file_format`?

MaxEW707: Let me know if you also want me to change `enum class AtFileFormat` to `enum class ResponseFileFormat`.

Let me know if you want me to change the public options from `gcc` and `msvc` to `posix` and `windows`. I prefer `posix` and `windows` to match the clang `--rsp-quoting=[posix|windows]` option since we actually just care about the quoting behaviour of the response file. We don't care about the compiler default quoting behaviour.

jrosdahl: > Let me know if you also want me to change `enum class AtFileFormat` to `enum class ResponseFileFormat`.

You can leave it for now if you like.

> Let me know if you want me to change the public option values from `gcc` and `msvc` to `posix` and `windows`.

Yes, that sounds good.

jrosdahl:
Thanks.

Issue #1252, opened 2023-02-16
ccache can't handle C+20 Named modules properly now


ChuanqiXu9:
The C++20 Named Modules introduce new dependent information to the project. So the legacy dependency discovering method can't work properly for C++20 Named Modules. And now clang-scan-deps has just landed the support for discovering the depdendency for named modules. We can find the manual at: https://clang.llvm.org/docs/StandardCPlusPlusModules.html#discover-dependencies. It should be pretty cool that ccache can support named modules.

ChuanqiXu9:
> Can I work on this?

I don't know ccache a lot. But you can contact me if you need information from modules : )

bharsaklemukesh975:
Hello @ChuanqiXu9 , Could you please provide some examples of how to use modules? I will start working on it. Looking forward to learning and implementing modules.

ChuanqiXu9:
> Hello @ChuanqiXu9 , Could you please provide some examples of how to use modules? I will start working on it. Looking forward to learning and implementing modules.

Hello. https://clang.llvm.org/docs/StandardCPlusPlusModules.html is the document about how to use C++20 modules in clang. And the blog https://www.kitware.com/import-cmake-c20-modules/ tells how to use cmake to compile C++20 modules. It may be a good start to use cmake to compile the hello world example.

And for the ccache, I think the key point is the p1689 format produced from clang-scan-deps. You can find the document in https://clang.llvm.org/docs/StandardCPlusPlusModules.html#discover-dependencies. And you can find the formal paper in https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2022/p1689r5.html.

For the more complex example, I think https://github.com/alibaba/async_simple/tree/CXX20Modules is the most complex example for modules as far as I know. You can find there a lot of module interfaces there. So I think it would be a great success if the ccache works for that one. Note that for the above example in async_simple, you would better to use the same standard library otherwise there may be problems.

kelteseth:
We now track ccache modules support progress at https://arewemodulesyet.org/tools/ . Feel free to create a PR if the status changes at https://github.com/kelteseth/arewemodulesyet/ 😊

PR #1523, opened 2024-10-11
Implement support for C++20 modules

silvanshade:
This is a PR intended to implement support for C++20 modules. The relevant issue is #1252.

I've submitted this early as a draft PR in case there is important feedback that might influence how this should be designed.

The current plan is roughly the following:

- [x] implement support for parsing [p1689r5](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2022/p1689r5.html) module dependency files
- [ ] add argument parsing for the appropriate module related arguments for `clang`, `gcc`, and `cl`
- [ ] add support for caching module related artifacts for `clang`, `gcc`, and `cl`
- [ ] process p1689 files similar to how depfiles are currently processed for depend mode


jrosdahl:
Thanks for working on this.

koplas:
I know this WIP, but how can this be tested? I compiled this branch with the feature enabled, but touching a file results in a cascade of recompilations. I tried it with GCC 14.2.1 and clang 21.

silvanshade:
@koplas the best way to understand what is currently working is to keep an eye on the [cxx_modules](https://github.com/ccache/ccache/pull/1523/commits/11d614d2665f48b5ead9f32e426e0fe3fad69387#diff-cd89dc4a7c6c8f7753b178e17b537304e41afd4be74d67d1b2ac76b03a824860) test. That test will continue to be updated as the rest of the support is built out.

Currently only GCC 14 is supported with regard to the caching of module artifacts.

Support for MSVC is also almost ready. Most of the pieces are in place for that, I just need to implement some additional processing for the `/sourceDependencies` output.

Clang support will arrive after that. It's not particularly difficult but does require some extra steps since the need to use `clang-scan-deps` requires processing in two phases. (I believe eventually `clang-scan-deps` functionality will be merged into the main frontend making this unnecessary in the long run).

As for how to test with GCC 14, you can take a look at the example here:

https://github.com/ccache/ccache/blob/11d614d2665f48b5ead9f32e426e0fe3fad69387/test/suites/cxx_modules.bash#L91-L93

Direct depend mode needs to be enabled for CCACHE and most of those module-related arguments need to be detected for the compiler invocation by `ccache`.

You can also set the environment variable `CCACHE_DEBUG=1` while compiling and check the logs that `ccache` creates for more detail about what is happening and whether the module artifacts were cached/restored.

koplas:
Thank you for your response! Enabling depend mode solved the issue, and I no longer experience recompilation cascades.

avikivity:
Was the work here abandoned? I was looking forward to using modules (although, without the distcc equivalent, it will still be difficult).

silvanshade:
> Was the work here abandoned?

It hasn't been abandoned but I don't know when it will be completed.

What isn't visible in this PR is additional work I've done on preparing patches for both Clang and GCC which fill in most of the missing information for the p1689 files, in order to avoid having to rely on messy and error-prone heuristics to try and reconstruct or infer that information.

The Clang work in particular is somewhat substantial because it replaces the need for `clang-scan-deps` entirely.

I tried to avoid needing to go this route because obviously getting PRs upstream to facilitate this has its own challenges. But the reality is that the current state of module support and p1689 support is not sufficient for a good implementation of this feature.

I am hoping I will have the chance to revisit this sometime early next year, and get those PRs submitted upstream, but I can't promise anything.

mathstuf:
Do you have anything written up on what needs to change in P1689 support? (My interest is that I authored P1689.)

As for having `clang` do it instead of `clang-scan-deps`, that was my preferred direction as well, but I got lost in the layers of abstraction without much progress before an upstream developer carried that torch. See this issue for upstream awareness of that: https://github.com/llvm/llvm-project/issues/76726

silvanshade:
> Do you have anything written up on what needs to change in P1689 support? (My interest is that I authored P1689.)

I don't have anything concrete written up but that would be helpful.

From what I remember of the last time I worked on this, it wasn't so much that the p1689 format itself was the issue as it was that Clang and GCC just didn't tend to fill in many of the fields that could have been used when constructing the manifests for caching.

I don't remember the specifics, but things like knowing which input source file was compiled, what the corresponding object file was, what the other inputs/headers were, etc., basically all of the same kind of info that is reconstructed from the depfiles.

> As for having `clang` do it instead of `clang-scan-deps`, that was my preferred direction as well, but I got lost in the layers of abstraction without much progress before an upstream developer carried that torch. See this issue for upstream awareness of that: [llvm/llvm-project#76726](https://github.com/llvm/llvm-project/issues/76726)

I see.

I did in fact manage to get a single-pass front-end implementation to work, which emits the p1689 data during compilation (like with Makefile depfiles), but the implementation purposely lacks a lot of the functionality of `clang-scan-deps` which is beyond the scope of the p1689 data, like the stuff related to scanning for `compile_commands.json`.

That, and the fact that `clang-scan-deps` was structured as a separate tool, with much of the logic not implemented through the existing module analysis libraries, meant that I had to re-implement a fair amount of the p1689 analysis in order to make it available to the frontend, rather than being able to re-use what was already there from that tool.

It wasn't clear to me how the project would react to that, although my impression was that `clang-scan-deps` should probably continue to exist as a separate tool, including its more advanced functionality.

Both the PRs for Clang and for GCC were actually in a usable state though, populating all of the fields specified in the p1689 spec (with the option for extra vendor specific data), but they needed some refactoring and commentary and justification before submission.

Since people seem to be interested, I'll see if I can make those implementations available and link them here.

ChuanqiXu9:
Just out of curiosity, I noticed https://github.com/mozilla/sccache/pull/2516  I think sccache are more complex than ccache, but that PR is 10x smaller than this. Would you like to tell the difference? Sorry if you feel uncomfortable, I am just interested.

silvanshade:
> Just out of curiosity, I noticed [mozilla/sccache#2516](https://github.com/mozilla/sccache/pull/2516) I think sccache are more complex than ccache, but that PR is 10x smaller than this. Would you like to tell the difference? Sorry if you feel uncomfortable, I am just interested.

At a glance, that PR is doing far less than what this PR is doing, feature-wise.

That PR appears to only partially support Clang (not GCC nor MSVC) based solely on specific command-line arguments. They don't parse the P1689 JSON files at all, nor the MSVC custom JSON format, nor process the Make depfiles for module artifacts (in GCC's case).

It's a starting point but I wouldn't expect that approach to be a complete solution as far as C++20 module support goes.

EDIT: Another difference, which isn't directly related to the feature support, but does make this PR larger, is that in some cases I've included updated documentation, bug-fixes, refactorings, substantial tests, and introduced some new abstractions (e.g., for the compiler) in order to make the code more maintainable and to help offset the additional complexity of module support.

TroyKomodo:
> > Just out of curiosity, I noticed [mozilla/sccache#2516](https://github.com/mozilla/sccache/pull/2516) I think sccache are more complex than ccache, but that PR is 10x smaller than this. Would you like to tell the difference? Sorry if you feel uncomfortable, I am just interested.
> 
> At a glance, that PR is doing far less than what this PR is doing, feature-wise.
> 
> That PR appears to only partially support Clang (not GCC nor MSVC) based solely on specific command-line arguments. They don't parse the P1689 JSON files at all, nor the MSVC custom JSON format, nor process the Make depfiles for module artifacts (in GCC's case).
> 
> It's a starting point but I wouldn't expect that approach to be a complete solution as far as C++20 module support goes.
> 
> EDIT: Another difference, which isn't directly related to the feature support, but does make this PR larger, is that in some cases I've included updated documentation, bug-fixes, refactorings, substantial tests, and introduced some new abstractions (e.g., for the compiler) in order to make the code more maintainable and to help offset the additional complexity of module support.

I'm the author of the other PR, maybe im missing something but in the context of sccache I dont think we would ever need to parse the P1689 output at all since we are not a build system so we dont need to decide how & what order to compile the modules, we rely on the build system (cmake or any other c++20 modules supported build system) to generate the correct compile commands and then we use those to correctly cache. This approach should work for both MSVC and Clang. The gcc approach is a bit annoying since it requires a bit more work, however cmake seems to use clang's dep scanner and then generates similar compile commands to clang.

Maybe im missing something but why do you need to do all the extra work specifically for ccache?

silvanshade:
> Maybe im missing something but why do you need to do all the extra work specifically for ccache?

It's been a while since I've thought about thought deeply about this so I probably don't remember all of the exact details.

The problem though is that if you don't parse the P1689 files, which I really think should be the closest thing to the source of truth from the perspective of external tooling, then you require on some other tool making all of the exact dependency details explicit through a big collection of command line flags.

Currently this does work if you're compiling using a build system that provides this (like CMake) and everything is fully explicit on the command line. But it doesn't work if you're compiling with GCC where some of this information is implicit or potentially contained in non-standard entries in the Make depfiles (which I either was parsing or ignoring here depending on the exact context).

It's also reasonable to assume I think that at some point the compiler frontends, particularly Clang's, might move more in the direction of providing a very lightweight interface for compiling a project with modules where almost all of the necessary information is implicit. The patches I worked on for the Clang frontend (which merged the clang-scan-deps functionality into the main frontend) worked more like this.

There's also the fact that `ccache` already parses Make depfiles for computing some dependency information when constructing the manifests. This is described to some extent [here](https://ccache.dev/manual/4.12.1.html#_the_depend_mode).

Theoretically, the P1689 provide a much more precise format for communicating the same information and more, so as long as ccache continues to function the way it does, it makes sense to also use the information from there. The problem is that (without the patches I mentioned working on), Clang and GCC leave a lot of the optional fields unpopulated.

From a design perspective, I think it's also better in the long run for maintenance if the caching tools could rely more on the P1689 format, and less on parsing the compiler flags. The latter is a lot messier, because flags are different between compilers and between versions, and also just due to the nature of having to parse and normalize all of that into something usable.

I don't know much about how sccache works so hard to say much beyond that. I would be curious to see how well the approach your considering seems to work in practice. I suspect getting a nice solution for GCC may be a bit involved. I'm not sure that relying on `clang-scan-deps` for that (if that's what CMake is doing) is a great solution but maybe it would be enough for most users at least given how things are right now.

PR #1525, opened 2024-10-14
Make tests pass in 2038

### LINKED ISSUE #1524: Tests fail after 2038
While working on reproducible builds for openSUSE, I found that
our ccache-4.10.2 fails to build in 2038.

### How to reproduce ###

1. run tests with the system clock set to 2038 , e.g. with `qemu -rtc base=2038-02-02T00:00:00`

### Actual behavior ###

```
 21/45 Test #21: test.nocpp2 ......................***Failed    1.41 sec
 Compiler version: gcc (SUSE Linux) 14.2.1 20241007 [revision 4af44f2cf7d281f3e4f3957efce10e8b2ccb2ad3]

 Running test suite nocpp2.......................
 FAILED

 Test suite:     nocpp2 (line 291)
 Test case:      Too new source file
 Failure reason: Expected modified_input_file to be 1, actual 0

 Actual statistics counters
 ==========================
...
 The following tests FAILED: 
          1 - test.base (Failed)
         21 - test.nocpp2 (Failed)
 Errors while running CTest
```

### Expected behavior ###

tests should keep working at least 15y in the future.

### Environment ###

ccache version 4.10.2

OS openSUSE Tumblweed 20241013 x86_64

I tried this patch, but it still failed in the same way:
```diff
--- a/test/suites/base.bash
+++ b/test/suites/base.bash
@@ -285,7 +285,7 @@ fi
     TEST "Too new source file"

     touch new.c
-    touch -t 203801010000 new.c
+    touch -t 303801010000 new.c

     $CCACHE_COMPILE -c new.c
     expect_stat modified_input_file 1
```

bmwiedemann:
Make tests pass in 2038
    
Fixes #1524
    
we use relative future timestamps to keep systems with 32-bit time_t working until 2037 (this should give them 1 year advance warning that they need to fix something).
    
Use the 3rd of January of the following year to ensure that it is in the future, even when starting a second before new year.

### DIFF
--- a/test/suites/base.bash
+++ b/test/suites/base.bash
@@ -284,8 +284,9 @@ fi
     # -------------------------------------------------------------------------
     TEST "Too new source file"
 
+    futuretimestamp=$(expr 1 + $(date -u +%Y))01030000
     touch new.c
-    touch -t 203801010000 new.c
+    touch -t $futuretimestamp new.c
 
     $CCACHE_COMPILE -c new.c
     expect_stat modified_input_file 1
@@ -304,7 +305,7 @@ EOF
     cat <<EOF >new.h
 int test;
 EOF
-    touch -t 203801010000 new.h
+    touch -t $futuretimestamp new.h
 
     $CCACHE_COMPILE -c new.c
     expect_stat modified_input_file 1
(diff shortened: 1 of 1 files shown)


test/suites/base.bash
```diff
@@ -285,7 +285,7 @@ fi
     TEST "Too new source file"
 
     touch new.c
-    touch -t 203801010000 new.c
+    touch -d "1 year" new.c
```
jrosdahl, line 288: As indicated by the failing tests, we can't assume that `touch` is of the GNU kind.

bmwiedemann: Good point. I updated it to use the January 3rd of the following year, so it can be 2-367 days in the future.

PR #1531, opened 2024-11-14
fix: Pass -fprofile-abs-path through to gcc

mkj:
When caching support for -fprofile-abs-path was added, the argument was omitted from the gcc commandline.

Fixes: 195011ad1b1f ("feat: Add support for GCC -fprofile-abs-path option")


mkj:
Not sure why the mingw test is failing now, seems unrelated to this PR

jrosdahl:
> Not sure why the mingw test is failing now, seems unrelated to this PR

Looks like GitHub's "windows-latest" environment has received a new Clang version.

jrosdahl:
Thanks!

PR #1533, opened 2024-12-06
feat: Add support for Intel's LLVM-based compilers

sarnex:
Intel's LLVM-based compilers are `icx` and `icx-cl`. Add support for them [based](https://github.com/ccache/ccache/pull/1100) on the changes that added support for Intel's legacy compiler, `icl`.

I noticed this because I was trying to use `ccache` with `icx` on Windows and it wasn't working. 

I didn't add a new enum for `icx-cl` because it can be treated exactly the same as `icx`.

Manually verified basic functionality.

### DIFF
--- a/src/ccache/config.hpp
+++ b/src/ccache/config.hpp
@@ -104,7 +104,7 @@ public:
   // Return true for Clang and clang-cl.
   bool is_compiler_group_clang() const;
 
-  // Return true for MSVC (cl.exe), clang-cl, icl, and icx.
+  // Return true for MSVC (cl.exe), clang-cl, icl, icx and icx-cl.
   bool is_compiler_group_msvc() const;
 
   util::SizeUnitPrefixType size_unit_prefix_type() const;
--- a/src/ccache/config.hpp
+++ b/src/ccache/config.hpp
@@ -41,6 +41,7 @@ enum class CompilerType {
   gcc,
   icl,
   icx,
+  icx_cl,
   msvc,
   nvcc,
   other
@@ -101,10 +102,10 @@ public:
   const std::filesystem::path& temporary_dir() const;
   std::optional<mode_t> umask() const;
 
-  // Return true for Clang and clang-cl.
+  // Return true for Clang, clang-cl and icx (not on Windows).
   bool is_compiler_group_clang() const;
 
-  // Return true for MSVC (cl.exe), clang-cl, icl, icx and icx-cl.
+  // Return true for MSVC (cl.exe), clang-cl, icl, icx-cl, and icx (on Windows).
   bool is_compiler_group_msvc() const;
 
   util::SizeUnitPrefixType size_unit_prefix_type() const;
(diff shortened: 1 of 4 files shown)


src/ccache/config.hpp
```diff
@@ -294,7 +295,8 @@ Config::is_compiler_group_msvc() const
 {
   return m_compiler_type == CompilerType::msvc
          || m_compiler_type == CompilerType::clang_cl
-         || m_compiler_type == CompilerType::icl;
+         || m_compiler_type == CompilerType::icl
+         || m_compiler_type == CompilerType::icx;
```
jrosdahl, line 299: I don't think that it will work out to unconditionally say that ICX is a MSVC compiler since it's called ICX on non-Windows (POSIX/Linux) platforms as well.

jrosdahl (changes requested):

src/ccache/config.hpp
sarnex: Thanks for the review. I'm embarrassed to say I'm not super familiar with the codebase here, so it removing these enums from the MSVC function enough, or do I need to add them to another function?

jrosdahl: I don't know since I haven't investigated the semantics of non-Linux icx and icx-cl. It seems clear that icx-cl should be in the "MSVC compiler group" but I don't know about icx.

Since you wrote _"I noticed this because I was trying to use ccache with icx on Windows and it wasn't working"_, do I understand correctly that you used icx (not icx-cl) on Windows and something wasn't working? If so, what didn't work?

sarnex: Thanks. I discussed this with an expert in the area internally and `icx` can be treated as an alias to `icx-cl` on Windows. Is the change okay as-is in that case?

I was trying to use CMake on Windows with ccache set as the launcher, as in:
```
cmake  -DCMAKE_C_COMPILER=icx -DCMAKE_CXX_COMPILER=icx -DCMAKE_C_COMPILER_LAUNCHER=ccache  -DCMAKE_CXX_COMPILER_LAUNCHER=ccache ...
```
and I found that no files were being cached, some empty folders were created but that's about it.

I reproduced the same issue above with `icx-cl` on Windows, and confirmed the change here works there too.

jrosdahl: > `icx` can be treated as an alias to `icx-cl` on Windows. Is the change okay as-is in that case? It's different on Linux, so should we guard the change with OS-specific ifdefs so on Linux `icx` is part of the `clang` group?

On Linux icx should be in the clang group, yes.

On Windows, icx should be in the msvc group if icx actually behaves like MSVC on Windows and not like GCC.  For instance, does `icx -MD` on Windows mean [use the multithread-specific and DLL-specific version of the run-time library](https://learn.microsoft.com/en-us/cpp/build/reference/md-mt-ld-use-run-time-library?view=msvc-170) or [enable writing of a dependency file](https://gcc.gnu.org/onlinedocs/gcc/Preprocessor-Options.html#index-MD)?

sarnex: just pushed a commit which should hopefully implement the desired behavior. i implemented it as windows vs non-windows instead of windows vs linux because icx only acts like MSVC on windows. thanks for you help

jrosdahl:
Thanks!

sarnex:
Thanks for the review and advice!

PR #1537, opened 2024-12-13
Add HTTP_STORAGE_BACKEND={ON/OFF} option.

sobomax:
Hi, first of all, thanks for a great project! I use it for 20+ years across multiple projects and products and it always does the job.

Now, this patch basically adds ability to turn the http backend off. I need this option since the library it uses is not stable on some of the platforms (I will open another PR on adding GitHub actions on 30+ platforms/architectures), so with this option I can just disable it on those that don't. In any case, using that backend is probably a very niche application (over 20+ years I never felt a need for it for example). 

You can see it used here for example:

https://github.com/sobomax/ccache/blob/21d94300de80c725dbb44983e2afbc668099e04c/misc/get-arch-buildargs#L20

Also if you are curious, here are some of the latest builds:

https://github.com/sobomax/ccache/actions/runs/12045852367

Some of the resulting containers are uploaded here:

https://github.com/sobomax/ccache/pkgs/container/ccache/versions

Thanks in advance!


jrosdahl:
Thanks!

PR #1538, opened 2024-12-13
Refresh dockerfiles.

sobomax:
This change refreshes dockerfiles, so that they actually work. This includes:

- Fix: E: Package 'gcc-multilib' has no installation candidate in ubuntu*.
- CentOS images are EOLed.
- Add Debian 12.

It also includes a change to allow BASE_IMAGE to be overridden, this is just a convenience hook for cases when someone wants to use some other version of the base image (or add ccache into an existing custom image that is derived from those).

It also add name to a layer, so again it's easy to extend and build upon, since someone who uses this image is probably just going to pull one or few binaries and throw out the rest.

### DIFF
--- a/dockerfiles/alpine-3.11/Dockerfile
+++ b/dockerfiles/alpine-3.11/Dockerfile
@@ -1,4 +1,5 @@
-FROM alpine:3.11
+ARG BASE_IMAGE=alpine:3.11
+FROM ${BASE_IMAGE} AS build
 
 RUN apk add --no-cache \
         bash \
--- a/dockerfiles/alpine-3.15/Dockerfile
+++ b/dockerfiles/alpine-3.15/Dockerfile
@@ -1,4 +1,5 @@
-FROM alpine:3.15
+ARG BASE_IMAGE=alpine:3.15
+FROM ${BASE_IMAGE} AS build
 
 RUN apk add --no-cache \
         bash \
--- a/dockerfiles/debian-11/Dockerfile
+++ b/dockerfiles/debian-11/Dockerfile
@@ -1,5 +1,7 @@
-FROM debian:11
+ARG BASE_IMAGE=debian:11
+FROM ${BASE_IMAGE} AS build
 
+ARG GCC_APT="gcc-multilib"
 RUN apt-get update \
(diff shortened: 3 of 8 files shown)


jrosdahl:
Thanks for the PR.

> Fix: E: Package 'gcc-multilib' has no installation candidate in ubuntu*.

It works for me to install gcc-multilib in ubuntu-20.04, ubuntu-22.04 and ubuntu-24.04, so I'm not sure what problems you see? The reason for installing gcc-multilib is to make it possible to test 32-bit builds. I'd prefer to keep it.

> CentOS images are EOLed.

That's true. However, the ccache binary release is still built in a CentOS 7 container, so I'd prefer to keep the corresponding Dockerfile until I've bumped to something newer.

> Add Debian 12.

:+1:

> It also includes a change to allow BASE_IMAGE to be overridden, this is just a convenience hook for cases when someone wants to use some other version of the base image (or add ccache into an existing custom image that is derived from those).

Sounds OK. (I'm however not sure if anybody will use them that way. I'm not aware of anybody else than me using them locally.)

The main reason why the Dockerfiles exist in the source tree is to build and test via `misc/test-all-systems`. Thus it would be good to update that script as well to match what's available.

sobomax:
Hey @jrosdahl thanks for taking a look. As I mentioned this is pre-requisite of the larger work that I want to submit for your consideration in a separate PR to add automatic CI/CD pipeline for your project, so it pushes multi-arch images to one or more container repositories. Then folks like me will be able to consume that directly from the registry without going through packaging system.

Issue with the having the multiarch is that ist's does not work on non-i386 (i.e. arm, mips, riscv etc). The whole idea of this change is that you won't need to do anything to test all of those if the bigger PR is merged. I do not have a big issue if you want to keep stale Dockerfiles in the tree, I'd just have to ignore them specifically, which I wanted to avoid as much as possible. But I am fine either way.

ScreenShot1395

sobomax:
I've pushed that bigger job, as part of the #1557 just to demo. You can see the 9 new jobs here (one per dockerfile) https://github.com/ccache/ccache/actions/runs/12678615680

Issue #1536, opened 2024-12-13
absolute_paths_in_stderr does not rewrite all paths


gitmodimo:
Current implementation of `rewrite_stderr_to_absolute_paths` matches only "In file included from " string [1](https://github.com/ccache/ccache/blob/852f0740dc746364b5c5bc604e01afd9f7e365d8/src/ccache/core/common.cpp#L96)
However gcc outputs multiple other vartiations depending on context. [2](https://gcc.gnu.org/git/?p=gcc.git;a=blob;f=gcc/diagnostic-format-text.cc;h=856d25e8482cd0bff39bd8076e6e529e184362cc;hb=HEAD#l679) Those other paths do not get rewritten to absolute. 

### How to reproduce ###

Produce error in nested included file or module

### Actual behavior ###

Only lines stsrting with "In file included from "  are rewritten to absolutem. 
Lines that start with "    from"  has original path i.e. Not rewritten to absolute path. 

### Expected behavior ###

All paths should be rewritten

### Environment ###

4.10.1
ccache version X.Y.Z

PR #1539, opened 2024-12-13
Fix issue #1536

gitmodimo:
This PR fixes reported issue #1536

### DIFF
--- a/src/ccache/core/common.cpp
+++ b/src/ccache/core/common.cpp
@@ -93,7 +93,17 @@ make_relative_path(const Context& ctx, const fs::path& path)
 std::string
 rewrite_stderr_to_absolute_paths(std::string_view text)
 {
-  const std::string_view in_file_included_from = "In file included from ";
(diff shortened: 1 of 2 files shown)


jrosdahl:
Looks good to me. Could you add some more cases to the [`core::rewrite_stderr_to_absolute_paths`](https://github.com/ccache/ccache/blob/852f0740dc746364b5c5bc604e01afd9f7e365d8/unittest/test_core_common.cpp#L50) test as well?

gitmodimo:
> Looks good to me. Could you add some more cases to the [`core::rewrite_stderr_to_absolute_paths`](https://github.com/ccache/ccache/blob/852f0740dc746364b5c5bc604e01afd9f7e365d8/unittest/test_core_common.cpp#L50) test as well?

Done. I was not expecting to find a bug that way, but there it was.

jrosdahl:
Thanks.

PR #1557, opened 2025-01-08
Build and push multiarch container images for a number of distros

sobomax:
This patch adds ability to build, test and push containers into GitHub Container registry so that anyone can get the latest version of the code from there instead of going through packaging system. I tried to retain as much of original infrastructure as possible and also enable all platforms supported by a given image with some exceptions (when the platform is broken in some major way).

For the purposes of this PR you can ignore "Refresh dockerfiles." part, which is already provided as part of the #1538 PR.


jrosdahl:
Thanks for the PR.

I have to admit that I don't quite understand the motivation behind building and publishing container images like this. Could you expand a bit on why you think this would be a good idea? In particular, why would it be preferable to use these prebuilt images instead of building them yourself?

My concerns: The set of Dockerfiles in the source tree has been added quite arbitrarily and have been updated whenever I or some other ccache _developers_ want to test something locally. They have never been intended to include all supported OS variants. If we publish these images/builds for consumption by ccache _users_ and add instructions for how to use them then I anticipate that users will request additional image/build variants. I'm not particularly stoked about this additional maintenance burden unless there is a clear benefit. Put differently, I strive to minimize the amount of potential things I have to tend to so that the limited time I have for ccache is freed up to work on other things.

sobomax:
@jrosdahl my motivation for this is quite simple: as I maintain several projects (5-6 and the number keeps growing) that rely on ccache to speed-up multi-arch container builds, I want a simple and consistent way to install latest ccache in all those builds.

The actual trigger for this work was when I was trying to debug low caching ratios building one of my projects (https://github.com/sippy/rtpproxy/) and found out that it was caused by the use of the -include directive that has only been relatively recently added into the ccache and has not propagated into particular distros that I was building against. My immediate impulse was to just build ccache from source, but that would add complexity and source of failures which is not something that I want but also this would have to be repeated over multiple projects. So I started looking into building a container and pushing it to my private ghcr repo just for ubuntu and debian that I care about. It worked pretty well, so I extended it to other distros just to see how it goes, found and fixed few issues.

That all being said, the ghcr part is not critical to me, I am totally fine with pulling from my own ghcr account. I'd still advocate for you to take the build part of the new CI (i.e. the "PR" version), then you would always know if any of the dockerfiles gets stale or broken. This would also take care of automatic testing not only on i386 and more exotic things like s390 or mips, which is not something your CI is doing now. And reduce the burden of maintaining my fork for me.

How does that sound? Let me know if you want me to update the PR along those lines. Thanks!

Arriya2532 (approved):

johnnynunez:
arm builds? aarch64?

sobomax:
> arm builds? aarch64?

arm

> arm builds? aarch64?

Yes and +5 others, most useful riscv, most obscure s390.

-Max

ScreenShot1441

sobomax:
Here is a short summary of supported bulds along with the stats of self-caching mechanism. Interestingly enough, on 80% of platforms ccache build cannot cache itself with default options. I have not looked into specific reasons for that, but if you have any ideas let me know. Another interesting observation is that based on number of files compiled the docker image for alpine is not including some dependencies, causing number of files compiled to be 30% higher.
```
+----------------+---------+--------+----------+----------------+----------+-------------+
| Platform       | Calls   | Hits   | Direct   | Preprocessed   | Misses   | Size (GB)   |
+================+=========+========+==========+================+==========+=============+
| almalinux-8                                                                            |
+----------------+---------+--------+----------+----------------+----------+-------------+
| amd64          | 154     | 154    | 151      | 3              | 0        | 0.0         |
+----------------+---------+--------+----------+----------------+----------+-------------+
| arm64          | 155     | 155    | 152      | 3              | 0        | 0.0         |
+----------------+---------+--------+----------+----------------+----------+-------------+
| ppc64le        | 154     | 154    | 151      | 3              | 0        | 0.0         |
+----------------+---------+--------+----------+----------------+----------+-------------+
| s390x          | 154     | 154    | 151      | 3              | 0        | 0.0         |
+----------------+---------+--------+----------+----------------+----------+-------------+
| almalinux-9                                                                            |
+----------------+---------+--------+----------+----------------+----------+-------------+
| amd64          | 154     | 145    | 133      | 12             | 9        | 0.0         |
+----------------+---------+--------+----------+----------------+----------+-------------+
| arm64          | 155     | 146    | 141      | 5              | 9        | 0.0         |
+----------------+---------+--------+----------+----------------+----------+-------------+
| ppc64le        | 154     | 154    | 151      | 3              | 0        | 0.0         |
+----------------+---------+--------+----------+----------------+----------+-------------+
| s390x          | 154     | 151    | 147      | 4              | 3        | 0.0         |
+----------------+---------+--------+----------+----------------+----------+-------------+
| alpine-3.21                                                                            |
+----------------+---------+--------+----------+----------------+----------+-------------+
| 386            | 199     | 3      | 2        | 1              | 196      | 0.0         |
+----------------+---------+--------+----------+----------------+----------+-------------+
| amd64          | 199     | 3      | 2        | 1              | 196      | 0.0         |
+----------------+---------+--------+----------+----------------+----------+-------------+
| arm32 v6       | 188     | 1      | 1        | 0              | 187      | 0.0         |
+----------------+---------+--------+----------+----------------+----------+-------------+
| arm32 v7       | 188     | 1      | 1        | 0              | 187      | 0.0         |
+----------------+---------+--------+----------+----------------+----------+-------------+
| arm64          | 188     | 1      | 1        | 0              | 187      | 0.0         |
+----------------+---------+--------+----------+----------------+----------+-------------+
| ppc64le        | 188     | 2      | 1        | 1              | 186      | 0.0         |
+----------------+---------+--------+----------+----------------+----------+-------------+
| riscv64        | 188     | 2      | 2        | 0              | 186      | 0.0         |
+----------------+---------+--------+----------+----------------+----------+-------------+
| s390x          | 188     | 2      | 2        | 0              | 186      | 0.0         |
+----------------+---------+--------+----------+----------------+----------+-------------+
| debian-11                                                                              |
+----------------+---------+--------+----------+----------------+----------+-------------+
| 386            | 150     | 34     | 33       | 1              | 116      | 0.0         |
+----------------+---------+--------+----------+----------------+----------+-------------+
| amd64          | 150     | 34     | 33       | 1              | 116      | 0.0         |
+----------------+---------+--------+----------+----------------+----------+-------------+
| arm32 v7       | 139     | 30     | 30       | 0              | 109      | 0.0         |
+----------------+---------+--------+----------+----------------+----------+-------------+
| arm64          | 139     | 30     | 30       | 0              | 109      | 0.0         |
+----------------+---------+--------+----------+----------------+----------+-------------+
| debian-12                                                                              |
+----------------+---------+--------+----------+----------------+----------+-------------+
| 386            | 148     | 2      | 2        | 0              | 146      | 0.0         |
+----------------+---------+--------+----------+----------------+----------+-------------+
| amd64          | 148     | 2      | 2        | 0              | 146      | 0.0         |
+----------------+---------+--------+----------+----------------+----------+-------------+
| arm32 v5       | 139     | 2      | 1        | 1              | 137      | 0.0         |
+----------------+---------+--------+----------+----------------+----------+-------------+
| arm32 v7       | 139     | 2      | 1        | 1              | 137      | 0.0         |
+----------------+---------+--------+----------+----------------+----------+-------------+
| arm64          | 139     | 2      | 1        | 1              | 137      | 0.0         |
+----------------+---------+--------+----------+----------------+----------+-------------+
| mips64le       | 138     | 2      | 2        | 0              | 136      | 0.0         |
+----------------+---------+--------+----------+----------------+----------+-------------+
| ppc64le        | 138     | 2      | 2        | 0              | 136      | 0.0         |
+----------------+---------+--------+----------+----------------+----------+-------------+
| s390x          | 138     | 2      | 2        | 0              | 136      | 0.0         |
+----------------+---------+--------+----------+----------------+----------+-------------+
| fedora-42                                                                              |
+----------------+---------+--------+----------+----------------+----------+-------------+
| amd64          | 154     | 154    | 151      | 3              | 0        | 0.0         |
+----------------+---------+--------+----------+----------------+----------+-------------+
| arm64          | 155     | 155    | 152      | 3              | 0        | 0.0         |
+----------------+---------+--------+----------+----------------+----------+-------------+
| ppc64le        | 154     | 154    | 151      | 3              | 0        | 0.0         |
+----------------+---------+--------+----------+----------------+----------+-------------+
| s390x          | 154     | 154    | 151      | 3              | 0        | 0.0         |
+----------------+---------+--------+----------+----------------+----------+-------------+
| manylinux-2.28                                                                         |
+----------------+---------+--------+----------+----------------+----------+-------------+
| 386            | 158     | 158    | 155      | 3              | 0        | 0.0         |
+----------------+---------+--------+----------+----------------+----------+-------------+
| amd64          | 158     | 158    | 155      | 3              | 0        | 0.0         |
+----------------+---------+--------+----------+----------------+----------+-------------+
| arm64          | 155     | 155    | 152      | 3              | 0        | 0.0         |
+----------------+---------+--------+----------+----------------+----------+-------------+
| ppc64le        | 154     | 154    | 151      | 3              | 0        | 0.0         |
+----------------+---------+--------+----------+----------------+----------+-------------+
| s390x          | 154     | 154    | 151      | 3              | 0        | 0.0         |
+----------------+---------+--------+----------+----------------+----------+-------------+
| ubuntu-20.04                                                                           |
+----------------+---------+--------+----------+----------------+----------+-------------+
| amd64          | 152     | 37     | 36       | 1              | 115      | 0.0         |
+----------------+---------+--------+----------+----------------+----------+-------------+
| arm64          | 144     | 33     | 32       | 1              | 111      | 0.0         |
+----------------+---------+--------+----------+----------------+----------+-------------+
| ppc64le        | 142     | 29     | 26       | 3              | 113      | 0.0         |
+----------------+---------+--------+----------+----------------+----------+-------------+
| riscv64        | 142     | 29     | 24       | 5              | 113      | 0.0         |
+----------------+---------+--------+----------+----------------+----------+-------------+
| s390x          | 142     | 32     | 31       | 1              | 110      | 0.0         |
+----------------+---------+--------+----------+----------------+----------+-------------+
| ubuntu-22.04                                                                           |
+----------------+---------+--------+----------+----------------+----------+-------------+
| amd64          | 149     | 37     | 33       | 4              | 112      | 0.0         |
+----------------+---------+--------+----------+----------------+----------+-------------+
| arm32 v7       | 139     | 30     | 28       | 2              | 109      | 0.0         |
+----------------+---------+--------+----------+----------------+----------+-------------+
| arm64          | 139     | 30     | 28       | 2              | 109      | 0.0         |
+----------------+---------+--------+----------+----------------+----------+-------------+
| ppc64le        | 139     | 29     | 29       | 0              | 110      | 0.0         |
+----------------+---------+--------+----------+----------------+----------+-------------+
| riscv64        | 139     | 26     | 19       | 7              | 113      | 0.0         |
+----------------+---------+--------+----------+----------------+----------+-------------+
| s390x          | 139     | 28     | 28       | 0              | 111      | 0.0         |
+----------------+---------+--------+----------+----------------+----------+-------------+
| ubuntu-24.04                                                                           |
+----------------+---------+--------+----------+----------------+----------+-------------+
| amd64          | 148     | 9      | 9        | 0              | 139      | 0.0         |
+----------------+---------+--------+----------+----------------+----------+-------------+
| arm32 v7       | 138     | 2      | 2        | 0              | 136      | 0.0         |
+----------------+---------+--------+----------+----------------+----------+-------------+
| arm64          | 138     | 2      | 2        | 0              | 136      | 0.0         |
+----------------+---------+--------+----------+----------------+----------+-------------+
| ppc64le        | 138     | 2      | 2        | 0              | 136      | 0.0         |
+----------------+---------+--------+----------+----------------+----------+-------------+
| riscv64        | 138     | 2      | 1        | 1              | 136      | 0.0         |
+----------------+---------+--------+----------+----------------+----------+-------------+
| s390x          | 137     | 2      | 2        | 0              | 135      | 0.0         |
+----------------+---------+--------+----------+----------------+----------+-------------+

**Comment by sobomax** [comment]
P.S. If someone want to looks into the issues with defficiencies here is the script to pull the latest stats:

```
#!/bin/sh

set -e

tdir="`mktemp -d`"
for t in *-*
do
  img="ghcr.io/sobomax/ccache:latest-${t}"
  docker pull -q "${img}"
  PLATFORMS="`docker manifest inspect "${img}" | \
   jq -r '.manifests[] | "\(.platform.os)/\(.platform.architecture)\(if .platform.variant != null then "/\(.platform.variant)" else "" end)"' | \
   grep -v unknown`"
  for p in ${PLATFORMS}
  do
    docker create --platform ${p} --name tmp-container "${img}" /bin/true
    sp="`echo ${p} | sed 's|[^/]*/|| ; s|arm/|arm32 |'`"
    pdir="${tdir}/${t}/${sp}"
    test ! -d "${pdir}" && mkdir -p "${pdir}"
    lsfile="${pdir}/lastbuild.stats"
    docker cp tmp-container:/root/.ccache/lastbuild.stats "${lsfile}"
    docker rm tmp-container
    cat "${lsfile}"
  done
done
echo "Output in: ${tdir}"
```

The branch can be forked and run in a private repo, resulting in a ccache image in a private ghcr space of the fork owner, allowing post-mortem inspection and iterative improvements.

jrosdahl:
I'm very sorry, but I remain unconvinced.

* The images will in practice become official artifacts that need to be maintained and errors fixed when they for some reason no longer work, leading to more work (for me).
* Building and pushing a number of Docker images for each push/merge to master seems like a waste of computing resources. If you want images for building ccache on different systems, wouldn't it be a better idea to do this outside of the ccache repository and making sure that they work with the latest _released_ ccache, not the latest development version of ccache?
* As mentioned earlier, the Dockerfiles in the repo are for developers. PR #1600 is a good example of something I would prefer not to have to deal with (the alma-8 image works to build and test in for me locally, and I prefer the shorter name "alma" – it's just coincidental that the name is equal to the docker image in most cases). I'd rather remove the Dockerfiles from the repo and just keep them locally than making them "official always working".

That being said, I would be happy to mention and link to your images from https://ccache.dev in case others find them helpful.

sobomax:
Well, I am not quite sure how having a separate arm64, risc64 etc could be less wasteful, but ok, you seem far more enthusiastic about that as in #1598 . With a working caching mechanism the additional "waste" could be really minimal, just look at the build logs/times of on the fedora or amla, most of the time is spent in the testing phase as give you not only the test coverage but also some valuable statistics on how the software performs on a variety of kinds of versions of toolchains, libcs and whatnot. Assuming you need/want this coverage, but apparently not so I'll rest my case. ☮️

jrosdahl:
> Well, I am not quite sure how having a separate arm64, risc64 etc could be less wasteful, but ok, you seem far more enthusiastic about that as in #1598 .

I don't quite understand what you mean by this. #1598 is a preparation step for including some more prebuilt binaries for different architectures in ccache releases. That is: one binary for each architecture. This feels a lot different from building a large number of _variants_ of said architectures in the way I understand your proposal.

> With a working caching mechanism the additional "waste" could be really minimal, just look at the build logs/times of on the fedora or amla

With a working caching mechanism I might indeed be more interested. But even with caching of the image builds in place a cache miss could happen at an unfortunate time, for instance in a workflow run for a release or some important PR.

I had a look at the _Build&Push to Registry_ jobs for this PR:

| Job | Duration |
| - | - |
| almalinux:8 |1h 46m 50s|
| almalinux:9 |1h 43m 16s|
| alpine:3.21 | 3h 33m 45s|
| debian:11 | 1h 3m 38s|
| debian:12 |3h 15m 1s|
| fedora:42 |3h 42m 6s|
| quay.io/pypa/manylinux_2_28 | 2h 1m 59s |
| ubuntu:20.04 | 2h 0m 5s |
| ubuntu:22.04 | 2h 31m 19s |
| ubuntu:24.04 | 2h 37m 48s |

Considering this, I'd say it's out of the question to do the builds of any of these images in PR workflows even with caching in place.

Do you have any suggestion on how to improve things?

Only rebuild container images on pushes/merges master but still do ccache build&test on PRs? Obvious downside: Harder to introduce needed changes to the set of platforms.

Only rebuild container images on a schedule, e.g. once a day or week? Same downside as above but could make a caching mechanism unnecessary.

Manually building and pushing images when needed? No.

> most of the time is spent in the testing phase as give you not only the test coverage but also some valuable statistics on how the software performs on a variety of kinds of versions of toolchains, libcs and whatnot. Assuming you need/want this coverage, but apparently not so I'll rest my case. ☮️

I apparently have trouble explaining my gut feelings about this in a good way. I'm not opposed to the "improved test coverage" part per se. It's more about how it is done and about what explicit or implicit promises it would bring.

If we add something similar to what you propose, I would like:

* workflow duration to not become significantly longer even in worst case,
* to be able to have Dockerfiles for some platforms available (for local testing purposes) without necessarily testing them in CI,
* to not have to add and maintain Dockerfiles for lots of different platform variations (which is what I worried could happen when you wrote "[...] push containers into GitHub Container registry so that anyone can get the latest version of the code from there instead of going through packaging system"), and
* freedom to decide what platforms should be tested in CI without external parties depending on which platforms are tested and whether those images are up to date.

PR #1567, opened 2025-02-21
enhance: Support relative paths necessary for react-native builds in different directories

kzlar:
Follow up PR from the discussion here: https://github.com/ccache/ccache/discussions/1566

Essentially, when building React Native under different directories, some flags are being used that are not made relative by ccache, thus caching doesn't work at all.

To resolve this I made the following change:
- Add `make_relative_path` calls for: ivfsoverlay, fmodules-cache-path & fbuild-session-file
- Add unit tests for the flags above as well as a e2e test for fbuild-session-file

Opening this as a draft PR while I'm running some more local tests, would still appreciate any and all feedback though.

### DIFF
--- a/src/ccache/argprocessing.cpp
+++ b/src/ccache/argprocessing.cpp
@@ -1120,6 +1120,11 @@ process_option_arg(const Context& ctx,
     return Statistic::none;
   }
 
+  if (util::starts_with(arg, "-fbuild-session-file")
+      && !(config.sloppiness().contains(core::Sloppy::time_macros))) {
+    args_info.build_session_file = arg.substr(arg.find('=') + 1);
+  }
+
   if (config.sloppiness().contains(core::Sloppy::clang_index_store)
       && arg == "-index-store-path") {
     // Xcode 9 or later calls Clang with this option. The given path includes a
--- a/src/ccache/argsinfo.hpp
+++ b/src/ccache/argsinfo.hpp
@@ -164,4 +164,7 @@ struct ArgsInfo
   // Compilation directory as passed in -ffile-compilation-dir or
   // -fdebug-compilation-dir.
   std::string compilation_dir;
+
+  // Build session file as passed in -fbuild-session-file.
+  std::filesystem::path build_session_file;
 };
--- a/src/ccache/ccache.cpp
+++ b/src/ccache/ccache.cpp
@@ -1688,6 +1688,14 @@ hash_common_info(const Context& ctx, const Args& args, Hash& hash)
     }
   }
 
+  if (!(ctx.args_info.build_session_file.empty())) {
+    // When using -fbuild-session-file, the actual mtime needs to be
+    // added to the hash to prevent false positive cache hits if the
+    // mtime of the file changes.
+    hash.hash_delimiter("-fbuild-session-file mtime");
+    hash.hash(DirEntry(ctx.args_info.build_session_file).mtime().nsec());
+  }
+
   if (!ctx.config.extra_files_to_hash().empty()) {
     for (const auto& path :
          util::split_path_list(ctx.config.extra_files_to_hash())) {
--- a/src/ccache/compopt.cpp
+++ b/src/ccache/compopt.cpp
@@ -124,7 +124,10 @@ const CompOpt compopts[] = {
   {"-emit-pth", AFFECTS_COMP},         // Clang
   {"-external:I",
    AFFECTS_CPP | TAKES_ARG | TAKES_CONCAT_ARG | TAKES_PATH}, // msvc
+  {"-fbuild-session-file=", TAKES_CONCAT_ARG | TAKES_PATH},
   {"-fmodule-header", TOO_HARD},
+  {"-fmodule-map-file=", TAKES_CONCAT_ARG | TAKES_PATH},
+  {"-fmodules-cache-path=", TAKES_CONCAT_ARG | TAKES_PATH},
   {"-fmodules-ts", TOO_HARD},
   {"-fno-working-directory", AFFECTS_CPP},
   {"-fplugin=libcc1plugin", TOO_HARD}, // interaction with GDB
@@ -147,7 +150,7 @@ const CompOpt compopts[] = {
   {"-iquote", AFFECTS_CPP | TAKES_ARG | TAKES_CONCAT_ARG | TAKES_PATH},
   {"-isysroot", AFFECTS_CPP | TAKES_ARG | TAKES_CONCAT_ARG | TAKES_PATH},
   {"-isystem", AFFECTS_CPP | TAKES_ARG | TAKES_CONCAT_ARG | TAKES_PATH},
-  {"-ivfsoverlay", TAKES_ARG},
+  {"-ivfsoverlay", TAKES_ARG | TAKES_PATH},
   {"-ivfsstatcache", TAKES_ARG},
   {"-iwithprefix", AFFECTS_CPP | TAKES_ARG | TAKES_CONCAT_ARG | TAKES_PATH},
   {"-iwithprefixbefore",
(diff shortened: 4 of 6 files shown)


jrosdahl:
Looks good to me for the path rewriting part. Whether it actually works to cache results when using those flags is something I can't tell.

kzlar:
> Looks good to me for the path rewriting part. Whether it actually works to cache results when using those flags is something I can't tell.

Sounds good!

I just added one additional flag that was missing and on the surface it seems like caching is working! I'm seeing cache hits in the stats and build time is reduce from ~6min to ~3min (there's more going on than just c/cpp compilation so I never expected to be much lower than that). I also tested the resulting app and it seems to work as well as without ccache.

Is there anything else you'd like to see tested before you're okay with merging this?

jrosdahl:
> I just added one additional flag that was missing and on the surface it seems like caching is working! I'm seeing cache hits in the stats

That's good. However, the hard part is often to make sure that no false positive cache hits occur because too much "input information" has been thrown away.

> Is there anything else you'd like to see tested before you're okay with merging this?

It would be great if you could evaluate the effects of making `-fmodules-cache-path`, `-fmodule-map-file` and `-fbuild-session-file` arguments relative. For example, do the files they refer to contain absolute paths to the CWD or input/output files? Because if they do there should probably be a sloppiness option to opt in to allowing them at all, similar to `-ivfsoverlay`.

> P.S. I do have one point to double check with you. Currently on correct set-up of `ivfsoverlay`, it seems like the arg is handled later in the function, while my change returns "earlier". Is that okay? Is there anything else happening later that I'd need to consider here?

Right, so `-ivfsoverlay` is handled via the generic option table in `src/compopt.cpp`. If no special handling other than rewriting a path argument to relative is needed it'd be better to add to that table instead. That is, add `TAKES_PATH` for `-ivfsoverlay` and add `TAKES_CONCAT_ARG | TAKES_PATH` entries for the others.

kzlar:
> It would be great if you could evaluate the effects of making `-fmodules-cache-path`, `-fmodule-map-file` and `-fbuild-session-file` arguments relative. For example, do the files they refer to contain absolute paths to the CWD or input/output files? Because if they do there should probably be a sloppiness option to opt in to allowing them at all, similar to `-ivfsoverlay`.
> 
Makes sense. Will do.

> Right, so `-ivfsoverlay` is handled via the generic option table in `src/compopt.cpp`. If no special handling other than rewriting a path argument to relative is needed it'd be better to add to that table instead. That is, add `TAKES_PATH` for `-ivfsoverlay` and add `TAKES_CONCAT_ARG | TAKES_PATH` entries for the others.

Interesting. So `-ivfsoverlay` does work when adding `TAKES_PATH` (see latest patchset). However the rest do not work when I try to add them to compopt.

For example:
I removed the following from `argprocessing.cpp`:
```
  if (util::starts_with(arg, "-fmodules-cache-path=")) {
    auto path = std::string_view(arg).substr(21);
    auto relpath = core::make_relative_path(ctx, path);
    state.common_args.push_back(FMT("-fmodules-cache-path={}", relpath));
    return Statistic::none;
  }
```
and added the following to `compopt.cpp`:
```
  {"-fmodules-cache-path", TAKES_CONCAT_ARG | TAKES_PATH},
```
And I get the following unittest failure (for the new tests I added)
```
/workspaces/ccache/unittest/test_argprocessing.cpp:361:
TEST SUITE: argprocessing
TEST CASE:  fmodules_cache_path_with_separate_argument_should_be_rewritten_if_basedir_is_used

/workspaces/ccache/unittest/test_argprocessing.cpp:381: ERROR: CHECK( result->preprocessor_args[1] == "-fmodules-cache-path=foo/bar" ) is NOT correct!
  values: CHECK( -fmodules-cache-path=/workspaces/ccache/build/testdir/11479/test_18/foo/bar == -fmodules-cache-path=foo/bar )

```

jrosdahl:
> ```
>   {"-fmodules-cache-path", TAKES_CONCAT_ARG | TAKES_PATH},
> ```
 
Try `{"-fmodules-cache-path=", TAKES_CONCAT_ARG | TAKES_PATH},` instead.

kzlar:
> Try `{"-fmodules-cache-path=", TAKES_CONCAT_ARG | TAKES_PATH},` instead.

Ahh that makes sense. Done and tests are passing. Patch-set is updated.

Any thoughts on the analysis of the flags above? From what I can tell the only open question seems to be around `-fmodule-map-file`

jrosdahl:
> 1. `-fbuild-session-file` - according to [this](https://clang.llvm.org/docs/ClangCommandLineReference.html#cmdoption-clang-fbuild-session-file) this only takes the modification time of the file, so using a relative path should not matter at all

Agreed. However, I think that regardless of absolute or relative path ccache should (unless `sloppiness = time_macros`?) include mtime of the file in the input hash since there otherwise will be false positive cache hits if mtime of the file changes.

> 2. `-fmodules-cache-path` - according to [this](https://clang.llvm.org/docs/Modules.html#compilation-model) the module cache file is a binary file with previously compiled modules, it seems like it should be safe to cache this as a relative path.

OK.

> 3. `-fmodule-map-file` - From what I gather it's discouraged to have absolute paths (and indeed none of the react native modules use absolute paths) but it seems technically possible to have absolute paths there. What would it mean to add a sloppiness option for this arg? I'm not really seeing that the sloppiness option for `ivfsoverlay` is used anywhere except for the error, but I'm likely missing something.

If relative paths are used in practice, then perhaps simply don't rewrite the path? That way we don't have to mess with sloppiness options.

kzlar:
> > 3. `-fmodule-map-file` - From what I gather it's discouraged to have absolute paths (and indeed none of the react native modules use absolute paths) but it seems technically possible to have absolute paths there. What would it mean to add a sloppiness option for this arg? I'm not really seeing that the sloppiness option for `ivfsoverlay` is used anywhere except for the error, but I'm likely missing something.
> 
> If relative paths are used in practice, then perhaps simply don't rewrite the path? That way we don't have to mess with sloppiness options.

Sorry I wasn't being clear. It's discouraged to have absolute path in the *contents* of that file.
An example of that file from react-native:
```c++
// Pods/Target Support Files/Yoga/Yoga.modulemap
module yoga {
  umbrella header "Yoga-umbrella.h"

  export *
  module * { export * }
}
```
Technically the umbrella header file path can be absolute but I honestly doubt you'll actually see that happen. Furthermore I think it's even more unlikely that the paths would be build folder dependent, but instead be some global system path. So from my (admittedly limited) understanding, it should be fine for caching with path re-writing, but would appreciate your thoughts here.

jrosdahl:
> Ok, that makes sense, is something like that already done elsewhere so I can imitate the behavior?

Adding information to the input hash is done in [`hash_common_info`](https://github.com/ccache/ccache/blob/fba34380ac416501ed8f80c6983bdfb18e1d5d18/src/ccache/ccache.cpp#L1494), so the path must be known there. I'd do something like this:

1. Add a `build_session_file` entry to `ArgsInfo`.
2. In `process_option_arg` set `args_info.build_session_file` to the path (unless "time_macros sloppiness").
3. In `hash_common_info` hash the mtime using `hash.hash_delimiter("-fbuild-session-file mtime"); hash.hash(DirEntry(path).mtime().nsec())` if the path is set.

>  So from my (admittedly limited) understanding, it should be fine for caching with path re-writing, but would appreciate your thoughts here.

OK, let's go with that.

kzlar:
> Adding information to the input hash is done in [`hash_common_info`](https://github.com/ccache/ccache/blob/fba34380ac416501ed8f80c6983bdfb18e1d5d18/src/ccache/ccache.cpp#L1494), so the path must be known there. I'd do something like this:
> 
> 1. Add a `build_session_file` entry to `ArgsInfo`.
> 2. In `process_option_arg` set `args_info.build_session_file` to the path (unless "time_macros sloppiness").
> 3. In `hash_common_info` hash the mtime using `hash.hash_delimiter("-fbuild-session-file mtime"); hash.hash(DirEntry(path).mtime().nsec())` if the path is set.

Done, let me know if it looks good.

jrosdahl:
Looks good to me.

jrosdahl:
Thanks.

PR #1571, opened 2025-02-27
fix: do not use --offload-compress when preprocessing

amd-isparry:
Clang has a feature where it can compress compiler products such as object files, and this includes when it is just called to preprocess the input file.

Ccache expects to be able to process the output of the preprocessor and is not expecting a compressed binary file. Obvious symptoms are apparently random data in log messages, but ccache appears to produce random results.

The fix is to remove any "--offload-compress" options using the existing facilities when preprocessing.


jrosdahl:
Thanks.

Issue #1488, opened 2024-07-26
Add support for Clang compiling CUDA code


romintomasetti:
### How to reproduce ###

Run the following script somewhere with a decent `clang++` compiler, `ccache 4.10` and the `Cuda` toolchain.

```bash
set -ex

TEST_SOURCE_FILE=test.cpp
TEST_OBJECT_FILE=test.o
TEST_EXECUT_FILE=test.out

# Simple test file with some CUDA code.
rm -r ${TEST_SOURCE_FILE}
cat <<EOF > ${TEST_SOURCE_FILE}
#include "cuda.h"

__device__ double kernel(const double value) {
    return value * 2.;
}

__global__ void entry(const double value) {
    printf("%f transformed to %f.\n", value, kernel(value));
}

int main()
{
    const double value = 7.85;
    entry<<<1,1>>>(value);
    cudaDeviceSynchronize();
    return 0;
}
EOF

# Set ccache debugging.
export CCACHE_DEBUG=true
export CCACHE_DEBUGLEVEL=2
export CCACHE_LOGFILE=$PWD/ccache.log

# Remove previous ccache log file.
rm -f ${CCACHE_LOGFILE}

# Zero ccache statistics.
ccache --zero-stats

# Compile command: use clang with CUDA language enabled.
function compile()
{
    ccache clang++ -x cuda --cuda-gpu-arch=sm_86 -c ${TEST_SOURCE_FILE} -o ${TEST_OBJECT_FILE}
}

# Compile twice.
compile
compile

# Check ccache stats.
ccache -s -x -v

# Remove previous executable.
rm -f ${TEST_EXECUT_FILE}

# Link and run the executable.
clang++ -L /usr/local/cuda/lib64 -lcudart ${TEST_OBJECT_FILE} -o ${TEST_EXECUT_FILE}
./${TEST_EXECUT_FILE}
```

### Actual behavior ###

Caching does not work. Most probably due to failing preprocessor step.

Looking at the logs:

>    [2024-07-26T09:31:10.264304 25321] Running preprocessor
    [2024-07-26T09:31:10.264345 25321] Executing /usr/bin/clang++ --cuda-gpu-arch=sm_86 -x cuda -E -o /workspaces/.../.ccache//tmp/cpp_stdout.tmp.hYklc0.i test.cpp
    [2024-07-26T09:31:10.284040 25321] Preprocessor gave exit status 1

Indeed, running the command myself leads to clang++ complaining that:

[...]

PR #1577, opened 2025-03-14
feat: Add clang cuda support

w169q169:
This pull request introduces support for compiling CUDA code using Clang, addressing [issue #1488](https://github.com/ccache/ccache/issues/1488). With this change, ccache now handles CUDA files when using Clang as the compiler.


src/ccache/argsinfo.hpp
```diff
@@ -141,6 +141,9 @@ struct ArgsInfo
   bool profile_use = false;
   bool profile_generate = false;
 
+  // verbose information
+  bool show_verbose = false;
```
romintomasetti, line 145: I would propose to make this change in a separate PR.

src/ccache/ccache.cpp
```diff
@@ -1324,13 +1325,70 @@ to_cache(Context& ctx,
...
+    FMT(".{}", ctx.config.cpp_extension()));
+  if (!tmp_result) {
+    return tl::unexpected(Statistic::internal_error);
+  }
+
+  const auto& chunk_path = tmp_result->path;
+  tmp_result->fd.close(); // we only need the path, not the open fd
+
+  // 2. Write the chunk contents into the temp file
+  if (auto write_result = util::write_file(chunk_path, chunk); !write_result) {
```
jrosdahl, line 1347:
```suggestion
  if (!util::write_file(chunk_path, chunk)) {
```

test/suites/clang_cu.bash
```diff
@@ -0,0 +1,257 @@
...
+}
+
+clang_cu_PROBE() {
+    if [ -z "$REAL_NVCC" ]; then
+        echo "nvcc is not available"
+    elif ! command -v cuobjdump >/dev/null; then
+        echo "cuobjdump is not available"
+
+    elif ! command -v clang >/dev/null; then
+        echo "clang is not available"
```
jrosdahl, line 17:
```suggestion
        echo "nvcc is not available"
        return
    elif ! command -v cuobjdump >/dev/null; then
        echo "cuobjdump is not available"
        return
    elif ! command -v clang >/dev/null; then
        echo "clang is not available"
        return
```

jrosdahl (commented):
Looks good in general.

PR #1584, opened 2025-03-28
Adding to the manual some context on how to enable ccache when compiling Android

NeutroGe:
docs: adding context for compiling Android using ccache as the linking methods wont work in that specific case


NeutroGe:
PS: it took me literally 4 hours to figure this out when compiling Android, and using ccache when doing it makes a huge difference in the compilation time, so i thought i'd share it here as i expect i'm not the only one to have encounter this issue as Android is a popular mobile OS.

PR #1596, opened 2025-05-30
feat: On a windows build, add the ccache.exe.manifest which enables l…

### LINKED ISSUE #1595: Enable LongPathAware on Windows to support the new Windows 10 longer path
### Description ###

See (Enable long paths in Windows 10, version 1607, and later)[https://learn.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation?tabs=registry#enable-long-paths-in-windows-10-version-1607-and-later] for background.

I was running ccache on a project in a deep folder, and it failed with access errors into the cache folder that I had inside the project.

Downloaded the source and saw that ccache uses fopen and family of function, which on Windows comes from the VCRT. These functions are long path safe, so I turned on long paths by dropping in a SxS manifest.

This made ccache work perfectly. Suggest to either have the manifest be embedded into the .exe, or at least drop it next to the exe so that modern Windows works :)

xoriath:
…ong path awareness in Windows 10+

This fixes #1595

ccache uses a c library for path access that is functional with long paths on Windows 10+. The only thing missing from ccache being able to access longer than MAX_PATH paths, is to set a flag to the loader.

This PR adds the manifest as as Side-by-Side manifest (due to ... msys2 not playing ball with RC/MT.exe), which informs Windows that this executable does not need to see a restricted path space.

Details about this can be seen in (https://learn.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation?tabs=registry#enable-long-paths-in-windows-10-version-1607-and-later)[https://learn.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation?tabs=registry#enable-long-paths-in-windows-10-version-1607-and-later]


jrosdahl:
Making Windows binaries handle long paths sounds very good.

> This PR adds the manifest as as Side-by-Side manifest (due to ... msys2 not playing ball with RC/MT.exe), which informs Windows that this executable does not need to see a restricted path space.

The prebuilt Windows binaries are cross-compiled with MinGW-w64 on Linux, so unless I'm missing something MSYS2 is not part of the equation.

It would indeed be nice to be able to embed the manifest. Could you describe what you mean by _"due to ... msys2 not playing ball with RC/MT.exe"_?

CMakeLists.txt
```diff
@@ -120,6 +120,9 @@ endif()
 # Installation
 #
 install(TARGETS ccache DESTINATION ${CMAKE_INSTALL_BINDIR})
+if(WIN32)
+  install(FILES ${CMAKE_CURRENT_LIST_DIR}/ccache.exe.manifest TYPE BIN DESTINATION ${CMAKE_INSTALL_BINDIR})
```
jrosdahl, line 124: I would prefer to put the manifest in some subdirectory to avoid adding more files to the top directory. Perhaps in `misc`?

xoriath: That makes sense, done :)

xoriath:
> It would indeed be nice to be able to embed the manifest. Could you describe what you mean by _"due to ... msys2 not playing ball with RC/MT.exe"_?

So, at least last time I was trying to do this... it turned into a real hassle.

The flow of embedding this in to a windows binary is,
 - Create the manifest
 - Create a resource file (.rc) describing the binary resources (name, version, strings, embedded xml manifest path)
 - Compile the .rc file with a resource compiler (mt on windows), to create a linkable object with the correct resource sections
 - Link it with the rest of the objects

At least I could never get the mt version in mingw to play ball with the embedded external xml manifest... So there I ended up just having a precompiled lib from a local run of mt that I manually linked... not very ergonomic.

Also, the only real difference between having a side-by-side manifest vs embedding it is w.r.t authenticode signatures... So not really that big of a deal here. The Windows loader caches the SxS manifest on the first load so it should not have any impact on load speed etc ;)

xoriath:
Hmmm the logs from 1a3a6e9 seems like it worked... at least the resource file is compiled and linked in... Is there a way to pull the executable out from the test run before merge, or is it a merge-then-check thing? :) 

In powershell you can pull out parts of the rc information with
```ps
Get-Item .\ccache.exe | Select -Property VersionInfo | Format-List

VersionInfo : File:             C:\Users\Morten\Desktop\ccache\build\Debug\ccache.exe
              InternalName:     ccache
              OriginalFilename: ccache.exe
              FileVersion:      unknown
              FileDescription:  Ccache is a compiler cache. It speeds up recompilation by caching previous compilations
               and detecting when the same compilation is being done again.
              Product:          Ccache
              ProductVersion:   unknown
              Debug:            False
              Patched:          False
              PreRelease:       False
              PrivateBuild:     False
              SpecialBuild:     False
              Language:         English (United States)
```

jrosdahl:
> Also, the only real difference between having a side-by-side manifest vs embedding it is w.r.t authenticode signatures... 

The aspect I'm thinking of is when a user copies ccache.exe somewhere but doesn't know (or forgets) that the manifest file needs to be copied next to ccache.exe as well. So embedding it would be nice! :+1:

> Is there a way to pull the executable out from the test run before merge, or is it a merge-then-check thing? :)

Don't think so. I'm OK with merging it and then checking...

cmake/GenerateVersionFile.cmake
```diff
@@ -2,3 +2,26 @@ configure_file(
   ${CMAKE_SOURCE_DIR}/cmake/version.cpp.in
   ${CMAKE_BINARY_DIR}/src/ccache/version.cpp
   @ONLY)
+
+if(WIN32)
+
+  if(CCACHE_VERSION MATCHES "^([0-9]+)([0-9]+)([0-9]+)$")
```
jrosdahl, line 8: Should this be something like

```suggestion
  if(CCACHE_VERSION MATCHES "^([0-9]+)\.([0-9]+)\.([0-9]+)\$")
```

?

jrosdahl:
Thanks.

Issue #1579, opened 2025-03-18
Add linux arm64/aarch64 binaries to ccache releases


rjaegers:
### Description ###
I am maintaining a [devcontainer](https://github.com/philips-software/amp-devcontainer) that includes ccache and want to keep up to date with GitHub releases as opposed to installing ccache with the package manager. But because I build for both x86_64 and aarch64 target architecture I need to build ccache from source. This takes quite a long time, as the aarch64 build is using emulation.

It would be great if the GitHub Releases contained linux-aarch64 binaries of ccache.

I could not find a GitHub Actions workflow that produces the artifacts, otherwise I would have made a PR.

jrosdahl:
See discussion #1310 and https://github.com/ccache/ccache/discussions/1486#discussioncomment-12035688.

PR #1598, opened 2025-06-11
Add CI jobs for Linux and Windows on ARM

mstorsjo:
These days, Github actions have got free runners with both Linux and Windows on ARM. This adds (or changes existing) CI jobs for both building/testing on those platforms, but also CI jobs for building binaries (for distribution) on these platforms. This would be a step towards https://github.com/ccache/ccache/issues/1579, plus also allows providing binaries for Windows on ARM64. (This would make https://github.com/hendrikmuhs/ccache-action/pull/318 more useful; currently it runs Windows x86_64 ccache binaries emulated.)

The PR contains a couple of CI simplifications, split in individual commits - I'm very open to splitting such things out to a separate preparatory PR if that is preferred.

As for building release binaries for Windows, I'm not sure which method you prefer (cross compilation from Linux, building natively, with mingw tools in msys2 or separately, or with MSVC); based on the existing binaries I presume they're built with mingw tools in one fashion or another. This adds building via the `clangarm64` environment in msys2; it's also possible to cross compile such from Linux with packages from https://github.com/mstorsjo/llvm-mingw/releases. (Those tool envionments are pretty much functionally similar, while one is hooked into the whole msys2 ecosystem, and the llvm-mingw toolchains are entirely standalone.)


jrosdahl:
Thanks!

I currently [build the official Linux x86_64 binary in an AlmaLinux 8 container](https://github.com/ccache/ccache/discussions/1310#discussioncomment-7025908) for compatibility with RHEL8-based distributions (Ubuntu 22.04 has too new libc), and I link it statically with libstdc++. But I can make these adjustments in a later step, so you can keep this as is.

I don't know what lower bar would be appropriate for Linux aarch64 binaries, so I'm fine with using Ubuntu 22.04 until somebody thinks otherwise.

Official Windows binaries are cross-compiled in Linux using Mingw-w64 and [these toolchain files](https://github.com/ccache/ccache/blob/99aa28bd6bb53dbc9d5e4e797d1cdf31899ddb5e/toolchains). Unless you think it would be better to build on Windows I would prefer building them on Linux in CI. @mstorsjo: Is this something you could change in this PR?

Since this PR alone does not make ARM binaries part of upcoming releases I'll unmark it as closing #1579.

(I plan to move most parts of the release process to a GitHub workflow eventually.)

mstorsjo:
> I currently [build the official Linux x86_64 binary in an AlmaLinux 8 container](https://github.com/ccache/ccache/discussions/1310#discussioncomment-7025908) for compatibility with RHEL8-based distributions (Ubuntu 22.04 has too new libc), and I link it statically with libstdc++. But I can make these adjustments in a later step, so you can keep this as is.

Thanks, that sounds like a great plan!

> I don't know what lower bar would be appropriate for Linux aarch64 binaries, so I'm fine with using Ubuntu 22.04 until somebody thinks otherwise.

Using a lower baseline generally is appreciated - I have access to a couple of systems that still are stuck on e.g. Ubuntu 20.04. (But the purposes of this PR mainly is to provide CI for building binaries for use in github actions CI itself, and for that, whatever they have as lowest baseline usually is fine.)

> Official Windows binaries are cross-compiled in Linux using Mingw-w64 and [these toolchain files](https://github.com/ccache/ccache/blob/99aa28bd6bb53dbc9d5e4e797d1cdf31899ddb5e/toolchains). Unless you think it would be better to build on Windows I would prefer building them on Linux in CI. @mstorsjo: Is this something you could change in this PR?

That sounds like a very reasonable way of building release binaries. I'll update this PR to do things that way.

BTW, for building release binaries cross compiled like that - it may be good to have a separate CI step on native Windows to actually verify that they do work (e.g. so they don't depend on some DLL that isn't included etc) - I'll see if I can add a sample CI setup for that too.

> Since this PR alone does not make ARM binaries part of upcoming releases I'll unmark it as closing #1579.

Ok, that's true enough.

> (I plan to move most parts of the release process to a GitHub workflow eventually.)

That sounds like a very good plan!

mstorsjo:
> I currently [build the official Linux x86_64 binary in an AlmaLinux 8 container](https://github.com/ccache/ccache/discussions/1310#discussioncomment-7025908) for compatibility with RHEL8-based distributions (Ubuntu 22.04 has too new libc), and I link it statically with libstdc++. But I can make these adjustments in a later step, so you can keep this as is.
> 
> I don't know what lower bar would be appropriate for Linux aarch64 binaries, so I'm fine with using Ubuntu 22.04 until somebody thinks otherwise.

FWIW it would probably be nice to use a docker container with some older distro as baseline, for the aarch64 binaries as well - so if you adapt it to use docker for the Linux binaries, it's probably best to do the same for both architectures.



> Official Windows binaries are cross-compiled in Linux using Mingw-w64 and [these toolchain files](https://github.com/ccache/ccache/blob/99aa28bd6bb53dbc9d5e4e797d1cdf31899ddb5e/toolchains). Unless you think it would be better to build on Windows I would prefer building them on Linux in CI. @mstorsjo: Is this something you could change in this PR?

I updated the PR to build the Windows binaries on Linux now, using the CMake toolchain files. I added a new toolchain file for aarch64-w64-mingw32, which I made location independent (not hardcoding `/usr/${TOOLCHAIN_PREFIX}` for `CMAKE_FIND_ROOT_PATH`).

(I didn't yet try adding a second step to try to verify running the release binaries with some sort of smoke tests.)

The PR contains a handful of minor CI fixups as preceding commits - let me know if you'd like to merge them separately from the rest, and I can break out a separate PR with just those minor commits.

jrosdahl:
Thanks.

PR #1600, opened 2025-06-17
Unify almalinux support, fix build issue and add almalinux-9 dockerfile

sobomax:
Hi, this is continuation of my effort to provide a multi-architecture pre-built ccache images for number of architectures and linux variants. After syncing my code with the latest master I found that almalinux has been added, however the name used diverges from the official image tag, thereby causing auto-discovery process to fail. 

After fixing that I found that the dependency list is incomplete, causing build to fail.

Last but not least, recently we've started to publish several pre-built pypi packages, which by convention is based on manylinux images, which in turn is based upon almalinux-9 for the latest revision of the manylinux "standard". The said images are successfully used to reduce build times for at least one such project now.

### DIFF
--- a/dockerfiles/almalinux-8/Dockerfile
+++ b/dockerfiles/almalinux-8/Dockerfile
@@ -17,6 +17,8 @@ RUN dnf install -y dnf-plugins-core \
         fmt-devel \
         gcc \
         gcc-c++ \
+        libstdc++-devel \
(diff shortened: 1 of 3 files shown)


jrosdahl:
I have added an alma-9 image and added libstdc++-static to alma-8 and alma-9. I don't want to rename them, sorry.

PR #1601, opened 2025-06-17
Link libatomic statically when STATIC_LINK is set

sobomax:
This patch is needed in order to link libatomic statically on debian arm/v5, where libatomic.a is required but it's hidden under subdirectory and is not visible to the normal linking process even with -static. This causes "static" ccache to depend on libatomic.so.

This option allows me to pass down location of the libatomic.a such as here: https://github.com/sobomax/ccache/blob/d12d99883661b766dc89c16766bdfac1b437295a/misc/get-arch-buildargs#L36

With this I have successfully implemented self-caching mechanism where ccache from a previous image build along with the previous cache set is used to speed-up build of the next revision. What is interesting, however is that effectiveness of this mechanism depends wildly between supported distros even for the same toolchain flavor/version between 0 and 100%.

### DIFF
--- a/cmake/StdAtomic.cmake
+++ b/cmake/StdAtomic.cmake
@@ -1,5 +1,7 @@
 # Check if std::atomic needs -latomic
 
+set(LIBATOMIC_STATIC_PATH "" CACHE PATH "Directory containing static libatomic.a")
+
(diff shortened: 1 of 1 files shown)


jrosdahl:
Thanks.

Issue #1613, opened 2025-07-16
Implementation Proposal: New Storage backend model and Message Format Specification


majitsima:
## 1. Background

#1214 describes a desired implementation for the remote storage backend of ccache. This involves executing a separate process which takes care of the connection with the remote storage. Before committing into a complex implementation of the proposal, a simple [TLV-protocol](https://en.wikipedia.org/wiki/Type%E2%80%93length%E2%80%93value) for this communication is defined with more detail in this document, as well as the file structure and files needed for the implementation. 

## 2. Files

```
src/ccache/utils/socket_interface.hpp
src/ccache/utils/tlv_protocol.hpp
src/ccache/storage/remote/socketbackend/launcher.hpp
src/ccache/storage/remote/socket_storage.hpp
src/ccache/storage/remote/socket_storage.cpp
```

## 3. Implementation Description

This initial implementation introduces two new files in the subdirectory `src/ccache/storage/remote/`, which simply implement the abstract `RemoteStorage` class similar to the http and Redi implementations. So, no structural changes induced.

The `socket_interface.hpp` file is a simple header file that contains a wrapper for the Unix socket (a class named `UnixSocket`) and allows for simple write and read calls and abstracts away the underlying system (Linux, Windows, etc..)

The `launcher.hpp` file contains a single method that takes care of launching the backend process if it is not already running.

The `tlv_protocol.hpp` file contains the implementation of the protocol in this proposal and defines how the client (ccache) communicates with the backend process responsible for remote storage.

All multi-byte values are encoded in network byte order (big-endian).

No maximum packet length is specified. Maximum Length per value: 2^32 bytes (32 bits for length encoding).

#### Data structures

`tlv_protocol.hpp`
- BufferPool: pre-allocated, reusable pool of fixed-size buffers
- TLVParser: parses the tlv-protocol messages
- TLVSerializer: serializes the tlv-protocol messages

`socket_interface.hpp`
- Stream: provides read and write operations for socket stream
- StreamReader: helper for buffering messages
- UnixSocket: complete wrapper for the Unix socket

### a. Packet structure

```yaml
┌────────────────────────────────────────────────────────┐
│                    Message Header                      │
├────────────────────────────────────────────────────────┤
│                    TLV Field 1                         │
├────────────────────────────────────────────────────────┤
│                    TLV Field 2                         │
├────────────────────────────────────────────────────────┤
│                       ...                              │
├────────────────────────────────────────────────────────┤
│                    TLV Field N                         │
[...]

Issue #1214, opened 2022-11-07
New storage backend model


jrosdahl:
## Background

Ccache currently has file, HTTP and Redis remote storage backends. The file and HTTP backends do not depend on any external libraries. The Redis backend depends on the small and ubiquitous Hiredis library. The remote storage backends are part of the ccache source tree and are compiled and linked statically with the ccache executable.

It would be very nice to support more protocols, like HTTPS (#890, #894), Redis over TLS (#902), Azure Blob Storage (#1152), AWS S3 (#1201), Google Cloud Storage (in case the HTTP/HTTPS backend does not suffice) and other cloud services and custom backends.

My approach has been to start out with only bundling backends that have no external dependencies, or only external dependencies that are ubiquitous and small enough. One reason for this is that I want to keep the startup of the ccache executable fast. For instance, linking with libcurl makes the [startup a factor 4 slower on my system](https://github.com/ccache/ccache/pull/676#issuecomment-855353145). Another aspect is that I would prefer not to have to maintain code that I can't easily test myself, such as backends for various cloud services. It would be much better if the people who are interested in a backend are the ones who maintain the code. (It's currently only me who is maintaining ccache and my spare time is not exactly abundant.) And a third aspect is one of distribution: I would like to be able to distribute a ccache package (for instance as part of a Linux distribution) that does not depend on libraries that are not needed for the basic use case (i.e., not using remote storage) and then have support for different remote storage backends in optional add-on packages. This is partly why I have been reluctant to add optional (at compile time) HTTPS support since I want a solution that does not depend on compile-time choices.

Another problem with the current backend framework is that ccache can't keep connections alive and thus not reuse sessions that are costly to set up.

## Proposal

As mentioned in <https://github.com/ccache/ccache/discussions/894#discussioncomment-997373>, I propose that we make ccache automatically start a long-lived protocol-specific helper process (if not already started) and communicate with it over a Unix socket.

Here is a rough design sketch of how it could work, taking HTTPS as an example protocol:

1. Say that ccache has been configured with `remote_storage = https://user:password@example.com/path|param=value`.
2. Ccache connects to a Unix socket named something like `${CACHE_TEMPDIR}/backend-<name>.sock` where `<name>` is a unique hash of the URL and applicable parameters.
    * This makes it easy to handle configuration changes: a new helper process will be started for the new configuration and the old helper process will terminate itself after a while.
4. If the connection is refused or the socket doesn't exist:
[...]

PR #1616, opened 2025-08-05
New storage backend model

majitsima:
This is an implementation of the proposal submitted in #1613 to enable ccache for communicating with a backend helper process, as originally suggested in #1214. This is currently working in combination with a Golang implementation of the external remote backend process found [here](https://github.com/worldline/CcacheAnywhere). 

This is currently a work-in-progress, and any comments and improvements are more than welcome.

### DIFF
--- a/src/ccache/storage/remote/socketstorage.cpp
+++ b/src/ccache/storage/remote/socketstorage.cpp
@@ -4,6 +4,7 @@
 #include "ccache/storage/remote/remotestorage.hpp"
 #include "ccache/storage/remote/socketbackend/tlv_codec.hpp"
 #include "ccache/storage/remote/socketbackend/tlv_constants.hpp"
+#include "ccache/storage/types.hpp"
 #include "ccache/util/bytes.hpp"
 #include "ccache/util/logging.hpp"
 #include "ccache/util/socketinterface.hpp"
@@ -44,7 +45,7 @@ public:
   tl::expected<bool, RemoteStorage::Backend::Failure>
   put(const Hash::Digest& key,
       nonstd::span<const uint8_t> value,
-      bool only_if_missing = false) override;
+      Overwrite overwrite) override;
 
   tl::expected<bool, RemoteStorage::Backend::Failure>
   remove(const Hash::Digest& key) override;
@@ -206,13 +207,13 @@ BackendNode::get(const Hash::Digest& key)
 tl::expected<bool, RemoteStorage::Backend::Failure>
 BackendNode::put(const Hash::Digest& key,
                  nonstd::span<const uint8_t> value,
-                 bool only_if_missing)
+                 Overwrite overwrite)
 {
   auto res = dispatch(tlv::MSG_TYPE_PUT_REQUEST,
                       tlv::FIELD_TYPE_KEY,
                       key,
                       tlv::FIELD_TYPE_FLAGS,
-                      only_if_missing ? uint8_t(0x0) : tlv::OVERWRITE_FLAG,
+                      uint8_t(overwrite == Overwrite::yes),
                       tlv::FIELD_TYPE_VALUE,
                       value);
 


jrosdahl:
I'm not sure what kind of review (if any) you expect at this stage and I'm also not quite sure what your goals are, but here are some quick high level comments (where "server" means the storage backend helper process):

* A document describing how to implement a server would be valuable from early on. It should describe how a server is started and how the wire protocol works. (But if your goal is more about creating proof-of-concept code than production ready code then such a spec document can wait, of course.)
* Regarding how to start the server, my current proposal is to search for `ccache-${scheme}-storage(.exe)` (a) in the directory where the ccache binary is located, (b) in `$PATH`, and (c) in ccache's libexec path. If not found, fall back to built-in backends (http, redis). In a future release we can then deprecate the built-in backends and after a while remove them.
* There should be tests of the client parts.
* For testing the client there needs to be a simple server for usage by tests. Perhaps in Python? See also how `test/http-server` is used by tests.
* It would be nice with a client (perhaps in Python?) for testing servers.
* Why is there a buffer size passed to the server and what does it mean?
* I don't see a need to do socket operations in a separate thread since the protocol is synchronous.
* As mentioned in https://github.com/ccache/ccache/issues/1613#issuecomment-3092166100, the server should send a list of capabilities (currently only "in-band" or another appropriate name) to the client and the client should verify that the server has the needed capability.
* SIMD instructions can't be used without checking at runtime if the CPU supports them (see e.g. `util::cpu_supports_avx2`). Having said that, I see no need to use SIMD instructions for parsing TLV stuff; it's just a small amount of small message headers where any similar optimization will be completely dwarfed by CPU time handling the large payloads.

I plan to make a ccache release relatively soon. After that I'd like to join in on working on the new storage backend model.

My goal is to build production-ready support piece by piece. What can I do to align with your agenda?

One thing that would be valuable to know is if you have done any performance experiments and if so what you learned from them.

majitsima:
Thanks for your comments! I'll address the server executable location issues and remove the SIMD code next. I also plan to add tests similar to the current HTTP tests. Additionally, I've updated the constant socket listening based on your feedback — you're absolutely right that it's unnecessary on the client side.

> I'm not sure what kind of review (if any) you expect at this stage and I'm also not quite sure what your goals are

No worries — I should have been clearer! I was looking for high-level feedback and general recommendations, which is exactly what you provided. I'll focus on addressing the critical issues first, then submit a proper PR.

> A document describing how to implement a server would be valuable from early on. 

Absolutely agree! I've started working on a developer guide for this.

> (But if your goal is more about creating proof-of-concept code than production ready code then such a spec document can wait, of course.)

The intent is making this production-ready, it is still work in progress though.

> Why is there a buffer size passed to the server and what does it mean?

Initially, I had it as a way to synchronize buffer sizes allocated whilst receiving. But I changed the way the server/client listens such that it became less relevant. I can drop it, but I thought maybe it is a nice to have information in case it may help with performance.

> I plan to make a ccache release relatively soon. After that I'd like to join in on working on the new storage backend model.

Awesome!

> My goal is to build production-ready support piece by piece. What can I do to align with your agenda?

I can guarantee that work will continue on making this production-ready until the end of September (where it would hopefully be ready). During this period, please share as much implementation ideas or concerns about the current approach as you can =D While I can't guarantee availability beyond September, I'm definitely open to continued collaboration and discussions.

> One thing that would be valuable to know is if you have done any performance experiments and if so what you learned from them.

I can report on the backends I have. Just in a brief matter, nothing too detailed. The testing was done on my machine and I just took an average run I had sampled. I implemented a Golang backend for both GCP and HTTP. Parallel to this I also installed the [google-cloud-cpp](https://github.com/googleapis/google-cloud-cpp) library and successfully integrated it into ccache. I just created a new `RemoteStorage::Backend` and implemented the get/put/remove methods. Note that during testing neither ccache nor the backend used zero-copy communication stacks.

I compiled mold with the ccache wrapper and an HTTP backend. The results were:

- Default ccache backend: with cached files 6.779s; no cached files 769.686s
- Separate backend process: with cached files 16.291s; no cached files 872.596s

For the GCP-based backend using google-cloud-cpp, the observed times were:

- Integrated within ccache:   retrieved X in 3591.50ms; retrieved Y in 1201.44ms
- Separate backend process: retrieved X in 827.41ms; retrieved Y in 99.88ms

These measurements show that new approach with a separate process improved performance for GCP, which likely means that the overhead of repeatedly establishing a GCP connection is way bigger than the overhead of passing the data over the socket to our helper. With http, another comparison has to follow once both sides (ccache and the helper) have an optimized communication interface. Broadly, this is the expected benefit: for more complex protocols — especially during setup — the new backend model should reduce overhead substantially.

jrosdahl:
OK, sounds good.

I'd say that performance evaluations are very valuable input at this stage. If the separate helper process model is significantly slower for plain HTTP than the current HTTP storage backend then it's important to figure out what the bottlenecks are.

majitsima:
An update from my side. As it stands, the implementation of the protocol from #1613 is as good as done. However, I am achieving quite poor results when retrieving cached binaries from a local HTTP server. On the other hand, when the remote is empty, the performance is quite similar to the default HTTP backend implemented. I am finding it difficult to pinpoint the exact reasons why Get is significantly slower. My suspicion is that the Golang client is slow at handling the response from HTTP (serializer needs optimizing), but I cannot say for sure. The profiler shows it is allocating quite a large amount of memory.

I am still benchmarking using `mold`; I just pulled the latest version today. Here are the latest sampling results averaged over six different runs:

- With cached binaries: default 8s, new 1m19s
- No cached binaries: default 18m25s, new 18m36s

Moreover, I introduced some tests to check the performance of different components. I will introduce a more complicated test next with a Python server and finish writing the guide for creating remote clients.

jrosdahl:
Thanks for the update.

That sounds extremely slow indeed. There should be no need for the proxy to allocate memory other than a fixed-size buffer (per connection) for the bulk data: receive data from server into buffer, send data in buffer to Unix socket, repeat. In Go with `io.CopyN`.

majitsima:
Well, after a deep dive into profiling and optimising every nook and cranny, it turns out the culprit was much simpler than I thought! 😅

I was convinced there was a complex bottleneck in the Go proxy or the underlying protocols, especially with those daunting `real` times. Right [here](https://github.com/worldline/ccache/blob/dev/new-storage-model/src/ccache/storage/remote/socketstorage.cpp#L82), there was a `sleep(1)` call before establishing each connection in the caller! I cannot believe it! I've removed it, and the results are finally looking like they're worth the effort.

Benchmarking with `mold` (entries are cached) and building with `time cmake --build build -j=8`:

**Default implementation:**
```
real    0m4.836s
user    0m8.523s
sys     0m2.516s
```

**New model (after removing sleep):**
```
real    0m2.896s
user    0m6.761s
sys     0m2.320s
```

**Golang proxy performance: (30 second inactivity timeout)**
```
real    0m33.836s
user    0m0.460s
sys     0m0.412s
```

It’s quite satisfying to cross this off the to-do list! I have some simple modification for the protocol that I'd like to discuss, which I'll add to the corresponding open discussion.

jrosdahl:
Since nothing has happened for a month and your Worldline mail address now bounces, I assume that you have moved on?

jrosdahl:
Thanks for prototyping/working on this issue, but I'll have to close this PR due to inactivity.

PR #1618, opened 2025-08-11
ci: Enable tests in the msys2 mingw64+clang configuration

mstorsjo:
Instead of disabling tests altogether, pass the -Wno-deprecated option, working around https://github.com/doctest/doctest/issues/900 in a different fashion.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -190,7 +190,7 @@ jobs:
 
           - sys: mingw64
             compiler: clang
-            extra_cmake_params: -DENABLE_TESTING=OFF # clang currently triggers warnings with doctest
+            extra_cmake_params: -DCMAKE_CXX_FLAGS=-Wno-deprecated # doctest issue #900
             os: windows-2025
 
           - sys: clangarm64


jrosdahl:
Looks good to me.

Regarding the failed job I've seen the _Windows MSYS2 clangarm64 clang_ job fail several times before due to segfaults in what seems to be the CTest runner? I've seen it for different test suites (test.base and now in test.direct). Do you have any idea? 

It's too often to be acceptable unfortunately, so I'm inclined to disable the test run if it keeps happening.

mstorsjo:
> Regarding the failed job I've seen the _Windows MSYS2 clangarm64 clang_ job fail several times before due to segfaults in what seems to be the CTest runner? I've seen it for different test suites (test.base and now in test.direct). Do you have any idea?
> 
> It's too often to be acceptable unfortunately, so I'm inclined to disable the test run if it keeps happening.

Yeah, I noticed that. Sounds very reasonable to disable it, because it's indeed only adding noise in this form - I'm sorry about this!

It did seem to run stable when I iterated a fair number of rounds of this when I prepared #1598, so I wonder if it is some environmental change that is causing it (potentially msys2 shipping some broken package, or something in an updated github actions runner image which is breaking things). I'll try to look into it, if I manage to reproduce the issue and figure it out. But in the meantime, yes, it sounds best to disable it to avoid unnecessary noise from it.

PR #1623, opened 2025-08-21
Use std::array instead of const char * []

autoantwort:


autoantwort:
I don't know why the build failed. It seems that the Job hangs.

PR #1624, opened 2025-08-24
build(blake3): Don't use NEON on aarch64_be

neuschaefer:
This fixes the build on aarch64 big-endian, by correctly falling back to the generic implementation.

Note: The upstream CMakeLists.txt looks completely different, which is why I didn't send the patch upstream first.

### DIFF
--- a/src/third_party/blake3/CMakeLists.txt
+++ b/src/third_party/blake3/CMakeLists.txt
@@ -115,13 +115,15 @@ _add_blake3_source_if_enabled(sse41 "" "-msse4.1" "_mm_test_all_ones(_mm_set1_ep
 _add_blake3_source_if_enabled(avx2 "/arch:AVX2" "-mavx2" "_mm256_abs_epi8(_mm256_set1_epi32(42))")
 _add_blake3_source_if_enabled(avx512 "/arch:AVX512" "-mavx512f -mavx512vl" "_mm256_abs_epi64(_mm256_set1_epi32(42))")
 
-# Neon is always available on AArch64
+# Neon is always available on AArch64, but blake3_neon.c only supports little-endian
 if(CMAKE_SIZEOF_VOID_P EQUAL 8)
   # https://developer.arm.com/architectures/instruction-sets/simd-isas/neon/intrinsics
   check_c_source_compiles(
     [=[
       #include <arm_neon.h>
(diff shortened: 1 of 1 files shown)


src/third_party/blake3/CMakeLists.txt
```diff
@@ -115,13 +115,15 @@ _add_blake3_source_if_enabled(sse41 "" "-msse4.1" "_mm_test_all_ones(_mm_set1_ep
 _add_blake3_source_if_enabled(avx2 "/arch:AVX2" "-mavx2" "_mm256_abs_epi8(_mm256_set1_epi32(42))")
 _add_blake3_source_if_enabled(avx512 "/arch:AVX512" "-mavx512f -mavx512vl" "_mm256_abs_epi64(_mm256_set1_epi32(42))")
 
-# Neon is always available on AArch64
+# Neon is always available on AArch64, but blake3_neon.c only supports little-endian
 if(CMAKE_SIZEOF_VOID_P EQUAL 8)
```
neuschaefer, line 119: Possible alternative approach: `AND CMAKE_C_BYTE_ORDER EQUAL LITTLE_ENDIAN`

src/third_party/blake3/CMakeLists.txt
```diff
@@ -115,13 +115,15 @@ _add_blake3_source_if_enabled(sse41 "" "-msse4.1" "_mm_test_all_ones(_mm_set1_ep
...
   # https://developer.arm.com/architectures/instruction-sets/simd-isas/neon/intrinsics
   check_c_source_compiles(
     [=[
       #include <arm_neon.h>
+      #if __BYTE_ORDER__ == __ORDER_LITTLE_ENDIAN__
       int main() { vdupq_n_s32(42); return 0; }
+      #endif
```
jrosdahl, line 126: Maybe I'm missing something, but doesn't this always compile correctly if `__BYTE_ORDER__ != __ORDER_LITTLE_ENDIAN__` so that `HAVE_NEON` will become true?

neuschaefer: I have tested this approach on aarch64_be, it fails the HAVE_NEON test as intended

jrosdahl:
Something's up with the GitHub workflow, failing jobs. I'll look into it later.

jrosdahl:
Thanks.

PR #1626, opened 2025-08-26
tests windows: Handle GetVolumePathName errors, prevent access violation

autoantwort:
For example GetVolumePathName returns `F:/git_projects` on my computer which does not fit in a char[4]. As a result GetVolumePathName returns an error and a std::string is initialized from uninitialized memory which may not contain a '\0' resulting in out of bounds memory reads.

### DIFF
--- a/unittest/test_argprocessing.cpp
+++ b/unittest/test_argprocessing.cpp
@@ -48,10 +48,11 @@ get_root()
 #ifndef _WIN32
   return "/";
 #else
-  char volume[4]; // "C:\"
-  GetVolumePathName(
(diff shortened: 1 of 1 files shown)


jrosdahl:
Oh, you found a small corner of the code where `std::filesystem`-ification hasn't reached yet.
What do you think about letting the function look something like this instead?

```c++
fs::path
get_root()
{
  auto cwd = fs::current_path();
  if (!cwd) {
    FAIL("get_root failed: ", cwd.error());
  }
  return cwd->root_path();
}
```

PR #1627, opened 2025-08-26
read_file windows: Fix access violation

autoantwort:
When interpreting a char* string that is null terminated as wchar* string it is not guaranteed that the wchar* string is also null terminated. A char[3] string/memeory region with the content ['a', '\0', '\0', 100] interpreted as wchar* has no ending null character resulting in out of bounds memory reads when trying to construct a std::wstring from it.


jrosdahl:
Thanks.

PR #1628, opened 2025-08-27
HttpStorageBackend::get: Logging: Fix read access violation

autoantwort:
If the request fails and `result.error() != Error::Success` `result->` can return
a `nullptr` resulting in a null pointer dereference.


jrosdahl:
Thanks.

PR #1639, opened 2025-10-05
docs: Make it easier to learn about --zero-stats for easier to read stats

ottok:
Minor improvement to docs so that people who read about `--show-stats` easily discover that `--zero-stats` exists and what is the benefit of it.

### DIFF
--- a/doc/manual.adoc
+++ b/doc/manual.adoc
@@ -183,7 +183,8 @@ documentation.
 *-s*, *--show-stats*::
 
     Print a summary of configuration and statistics counters in human-readable
-    format. Use `-v`/`--verbose` once or twice for more details.
+    format. Use `-v`/`--verbose` once or twice for more details. See
+    <<_cache_statistics,Cache statistics>> for more information.
 
 *-v*, *--verbose*::
 
@@ -1439,6 +1440,18 @@ cleanups (number of performed cleanups, either implicitly due to a cache size
 limit being reached or due to explicit `ccache -c` calls), overall hit rate, hit
 rate for <<The direct mode,direct>>/<<The preprocessor mode,preprocessed>> modes
 and hit rate for local and <<config_remote_storage,remote storage>>.
+
+The statistics counters are not used by ccache itself during builds. This means
+that you can safely reset them at any time using `ccache --zero-stats` without
+affecting the build process. For example, you might reset them before a build so
+that `ccache --show-stats` will only summarize the results from that specific
+build. Alternatively, you can set <<config_stats_log,*stats_log*> before
+starting the build and then run `ccache --show-log-stats` afterward to view
+build-specific statistics. This approach allows the statistics counters to
+continue tracking the entire lifetime of the cache while still giving you
+detailed information for individual builds. Another advantage of *stats_log* is
+that it collects statistics without interference from other concurrent builds
+that access the same cache.
 
 The summary also includes counters called "`Errors`" and "`Uncacheable`", which
 are sums of more detailed counters. To see those detailed counters, use the


jrosdahl:
Thanks for the PR.

However, I don't agree that `--zero-stats` should be used in CI since that throws away statistics that should be of interest. Instead I would recommend [`stats_log`/`CCACHE_STATSLOG`](https://ccache.dev/manual/4.12.1.html#config_stats_log) which was added for that use case:

1. `export CCACHE_STATSLOG=ccache.stats`
2. Perform build.
3. `ccache --show-log-stats`

What do you think? And how should this have been documented so that you would have found it?

ottok:
I have been using ccache for maybe 10 years and always zeroed stats before
a run so that I can see if the cache hit rate is close to 100% or not. If
it is not close to 100%, then I know caching wasn't working. That number
would never be close to 100% or close to 0% if the stats are kept from
previous runs.

If you don't recommend zeroing stats on every run, when would you recommend
to zero the stats? When every 10 runs, when the previous runs stats will
skew the current run by over 90%? Or once a year? What is the benefit of
having counters from multiple old runs, what will they help to tell about
the current run?

jrosdahl:
> I have been using ccache for maybe 10 years

For reference, `--show-log-stats` was added 2021 (ccache 4.4).

> If it is not close to 100%, then I know caching wasn't working. That number would never be close to 100% or close to 0% if the stats are kept from previous runs.

Right, and they will be the same if you use `--show-log-stats`. I'm just saying there's a new feature that you can use if you want, and I would like to recommend that instead of the big hammer for a CI build scenario.

> If you don't recommend zeroing stats on every run, when would you recommend to zero the stats? When every 10 runs, when the previous runs stats will skew the current run by over 90%? Or once a year?

You are free to zero your statistics whenever you like. I never zero mine, except perhaps after clearing the whole cache.

> What is the benefit of having counters from multiple old runs, what will they help to tell about the current run?

- `--show-log-stats`: Current build.
- `--show-stats`: Long term statistics. For instance, do certain uncacheable calls pop up over time, and if so what's the percentage? Perhaps something needs to be done if an issue pops up often, or perhaps it can be ignored if it's rare? Does remote storage requests fail often?

Another advantage of `--show-log-stats` is that it can collect statistics without interference from other concurrent builds that access the same cache.

ottok:
Thanks for the examples.

For me when using ccache in GitLab CI where the cache is downloaded before
build and uploaded after the build none of those long term stat benefits
matter. The main question is if ccache hit rate is high or low as a sign if
the cache in general is working or not.

I read about the new feature and while it may be good to some, it seems
overkill for the basic scenario of just trying to save build time when
building the same sources over and over in a CI.

ottok:
Can we at least have some wording that it is not wrong to zero the
stats before every build if ones goal is to have stats for each build,
and it does not have harmful effects?

jrosdahl:
Sure, it would be good to improve documentation. I'm all for that. I think that it would be best not to mention CI and what's typical since not all CI systems are the same. (I have never set up ccache in that way in a CI system.)

What do you think about this suggestion?

```diff
--- a/doc/manual.adoc
+++ b/doc/manual.adoc
@@ -183,7 +183,8 @@ documentation.
 *-s*, *--show-stats*::
 
     Print a summary of configuration and statistics counters in human-readable
-    format. Use `-v`/`--verbose` once or twice for more details.
+    format. Use `-v`/`--verbose` once or twice for more details. See
+    <<_cache_statistics,Cache statistics>> for more information.
 
 *-v*, *--verbose*::
 
@@ -1440,6 +1441,18 @@ limit being reached or due to explicit `ccache -c` calls), overall hit rate, hit
 rate for <<The direct mode,direct>>/<<The preprocessor mode,preprocessed>> modes
 and hit rate for local and <<config_remote_storage,remote storage>>.
 
+The statistics counters are not used by ccache itself during builds. This means
+that you can safely reset them at any time using `ccache --zero-stats` without
+affecting the build process. For example, you might reset them before a build so
+that `ccache --show-stats` will only summarize the results from that specific
+build. Alternatively, you can set <<config_stats_log,*stats_log*> before
+starting the build and then run `ccache --show-log-stats` afterward to view
+build-specific statistics. This approach allows the statistics counters to
+continue tracking the entire lifetime of the cache while still giving you
+detailed information for individual builds. Another advantage of *stats_log* is
+that it collects statistics without interference from other concurrent builds
+that access the same cache.
+
 The summary also includes counters called "`Errors`" and "`Uncacheable`", which
 are sums of more detailed counters. To see those detailed counters, use the
 `-v`/`--verbose` flag. The verbose mode can show the following counters:
```

ottok:
I updated the MR to have the text above but now it does not really have any of my original text, so I should maybe just close this MR?

jrosdahl:
Before going ahead, do you think my suggestion captures most of what you wanted to see?

If you're fine with it I guess you can close the MR if you like and I can then make the commit with you as the coauthor (contributing with the main documentation improvement idea).

ottok:
Yes, I think this is enough to hint that `--zero-stats` exists and can be safely used in a simple use case that is not using the stats log.

jrosdahl:
OK. Thanks for raising this and helping improve the documentation!

PR #1640, opened 2025-10-06
Reconcile max-size suffixes in documentation

Andrej730:
`-max-size` option was using explicit suffixes and in some other places shorter versions were used. Replaced them all with explicit versions to make it more intuitive (no need to remember default suffix) and consistent.

image

Also found an issue with `max_size` description stating that default suffix is `G` (meaning `GB`) and default value is `5G`. While actual default suffix is `Gi` and default value is `5GiB`.


jrosdahl:
Thanks, looks like I missed to update the `max_size` description in ac98dee75104f31c95ebe57909b43ff88e336c45.

Issue #1507, opened 2024-09-10
Add support to -fprofile-prefix-path=path


jdavidpeter:
TLDR:
Current CCache (4.10.2) doesn't support shared caching for coverage builds in different build workspaces using gcc.
By adding support for -fprofile-prefix-path=path gcc option and making sure that "apparent real path" of the object file is not hashed when profile-prefix-path is set, shared caching for coverage could be enabled.

Useful gcc links:
https://gcc.gnu.org/onlinedocs/gcc/Instrumentation-Options.html
https://gcc.gnu.org/onlinedocs/gcc/Overall-Options.html

Context:

In CI environments ccache can drastically increase build space for PR validation builds.
As many CI environments operate on bare metal and are placing workspaces to different directories, caching is only possible if references to build absolute paths in object files are removed.
Using gcc on Linux systems, this is trivial for release builds, easy to achieve for debug (using -ffile-prefix-map) but is seemingly impossible for coverage builds, which produce gcno files build time and gcda files during execution.

For coverage, the compiler generates a gcno file on top of the object file, which is cached and supported by ccache. However, gcc also bakes in the runtime gcda file path in the object file, where the runtime coverage information will be recorded.
By default, this is an absolute path pointing to the same directory as the result object file, so cross-workspace caching is not possible.

There is a build time option, -fprofile-dir=path that can instruct gcc to bake in a different gcda file path in the object files compile time. This new path can be relative, therefore independent of build workspace paths. This option is also supported by ccache.
However, when using this option with realitve paths, the resulting gcda file names will contain a mangled absolute path to the build workspace to avoid potential filename collisions when all gcda files are put in the same folder during execution.
This again, prevents reuse of cache across build workspaces for coverage and I believe was the main reason for a bugfix in version 4.10 which started to hash the "apparent real path" of the object file in the cache key.

However, gcc offers yet another option, -fprofile-prefix-path=path.
When -fprofile-dir is set to a relative path, -fprofile-prefix-path can be set to the root of the build workspace and the absolute path of the build workspace will be cut from the gcda file name baked in the object file.
This makes the resulting gcda file name independent of the build workspace absolute path and thus would enable cache reuse of coverage builds across different build workspaces, e.g. in CI environments.

Unfortunately, -fprofile-prefix-path is not supported by ccache and it just bails when detecting it.

Would it be possible to add support for it?
If so, what parts of the code would need change? I could create a PR if maintainers think this fits to the vision and give me some pointers to start with.

[...]

PR #1644, opened 2025-10-09
feat: Add support for -fcoverage-compilation-dir and -fcoverage-prefix-map clang options

eisaev:
This is similar to https://github.com/ccache/ccache/issues/1507 and https://github.com/ccache/ccache/pull/1515, but for Clang.
 
The current Ccache (4.12.1) doesn't support shared caching for coverage builds in different build workspaces using Clang.
By adding support for the `-fcoverage-compilation-dir=path` and `-fcoverage-prefix-map=old_path=new_path` Clang options and making sure that "apparent real path" of the object file is not hashed when one of these options is set, it would be possible to enable shared caching for coverage.
 
- https://clang.llvm.org/docs/ClangCommandLineReference.html#cmdoption-clang-fcoverage-compilation-dir
- https://clang.llvm.org/docs/ClangCommandLineReference.html#cmdoption-clang-fcoverage-prefix-map


eisaev:
@jrosdahl JFYI: This has only been tested in a limited environment. Within the next week (maybe a couple of weeks) I would like to test this on the actual project infrastructure available to me. So I'm leaving this as a draft for now.

eisaev:
Testing on the infrastructure of the real project didn't reveal any issues.

src/ccache/argprocessing.cpp
```diff
@@ -756,6 +762,15 @@ process_option_arg(const Context& ctx,
     return Statistic::none;
   }
 
+  if (util::starts_with(arg, "-fcoverage-compilation-dir")) {
+    const auto eq_pos = arg.find('=');
+    if (eq_pos != std::string_view::npos) {
+      args_info.coverage_compilation_dir = arg.substr(eq_pos + 1);
+    }
```
jrosdahl: Could this be

```suggestion
  if (util::starts_with(arg, "-fcoverage-compilation-dir=")) {
    const auto eq_pos = arg.find('=');
    args_info.coverage_compilation_dir = arg.substr(eq_pos + 1);
```

?

jrosdahl: Sure.

src/ccache/argsinfo.hpp
```diff
@@ -163,10 +163,16 @@ struct ArgsInfo
   // Relocating debuginfo in the format old=new.
   std::vector<std::string> debug_prefix_maps;
 
+  // Relocating coverage info in the format old=new.
+  std::string coverage_prefix_map;
```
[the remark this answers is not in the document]
eisaev: Reimplemented in [31a1e88](https://github.com/ccache/ccache/pull/1644/commits/31a1e8834fa7f50f5a32dfbaedcc6770a359b3a6) in the same way as `-fdebug-prefix-map`.

src/ccache/argprocessing.cpp
eisaev: OK. Changed in [31a1e88](https://github.com/ccache/ccache/pull/1644/commits/31a1e8834fa7f50f5a32dfbaedcc6770a359b3a6).

src/ccache/ccache.cpp
```diff
@@ -2284,10 +2289,29 @@ hash_profiling_related_data(const Context& ctx, Hash& hash)
...
-        ? ctx.args_info.profile_path
-        : ctx.apparent_cwd / ctx.args_info.profile_path;
+    fs::path profile_path = ctx.args_info.profile_path.is_absolute()
+                              ? ctx.args_info.profile_path
+                              : ctx.apparent_cwd / ctx.args_info.profile_path;
+
+    if (!ctx.args_info.coverage_compilation_dir.empty()) {
+      profile_path = ctx.args_info.coverage_compilation_dir;
+    } else if (!ctx.args_info.coverage_prefix_map.empty()) {
+      for (const auto& map : ctx.args_info.coverage_prefix_map) {
```
jrosdahl: Should be possible to extract this loop to a common helper function ("apply_prefix_remapping") shared with the `-fdebug-prefix-map` case (logging can be done after the loop). I can do that later unless you want to do it as part of this PR.

src/ccache/argsinfo.hpp
```diff
@@ -163,10 +163,16 @@ struct ArgsInfo
   // Relocating debuginfo in the format old=new.
   std::vector<std::string> debug_prefix_maps;
 
+  // Relocating coverage info in the format old=new.
+  std::vector<std::string> coverage_prefix_map;
```
jrosdahl: `coverage_prefix_map` → `coverage_prefix_maps`, similar to `debug_prefix_maps`.

src/ccache/ccache.cpp
jrosdahl: In practice I guess it makes no difference performance-wise since there will be at most a handful entries, but it sounds good to iterate backwards and stop after a match.

jrosdahl:
Thanks.

PR #1646, opened 2025-10-13
Allow static linking of the final ccache binary

berndpetrovitsch:
added the configuration variable STATIC_LINKING (defaulting to false) to enable static linking of the final ccache binary. This allows compilation on recent Linux distributions (which have a new enough cmake et.al.) and use it on old(er) distributions.

Compiles (both dynamic and static linked) on Debian/Sid (Oct 2025) and Fedora-42. The static binary built on Debian/Sid runs on Ubuntu/Disco.

This was shamelessly copied from
https://stackoverflow.com/questions/15521958/cmake-static-and-dynamic-linking-based-on-build-type

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -21,6 +21,13 @@ set(CMAKE_EXPORT_COMPILE_COMMANDS TRUE)
 
 list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake")
 
+# determine, whether we want a static binary
+SET(STATIC_LINKING FALSE CACHE BOOL "Build a static binary?")
+
+# build a static executable for old distros
+IF(STATIC_LINKING)
+    SET(CMAKE_FIND_LIBRARY_SUFFIXES ".a")
+ENDIF(STATIC_LINKING)
 #
 # Minimum compiler requirements (fail gracefully instead of producing cryptic
 # C++ error messages)
@@ -120,6 +127,11 @@ if(WIN32)
   endif()
   target_link_libraries(ccache PRIVATE ccache_win32_manifest)
 endif()
+# build a static executable for old distros
+IF(STATIC_LINKING)
+    SET_TARGET_PROPERTIES(ccache PROPERTIES LINK_SEARCH_END_STATIC 1)
+    SET(CMAKE_EXE_LINKER_FLAGS "-static")
+ENDIF(STATIC_LINKING)
 
 #
 # Documentation


jrosdahl:
Thanks for the PR.

There is already a `STATIC_LINK` option, as mentioned here: 

https://github.com/ccache/ccache/blob/1c6e6d6b5e30524d8b8a490be9169f051c713f31/doc/install.md?plain=1#L64-L66

Here is the implementation: https://github.com/ccache/ccache/blob/1c6e6d6b5e30524d8b8a490be9169f051c713f31/cmake/StaticLinkSupport.cmake

It's intentional that `STATIC_LINK=ON` doesn't use `-static` since linking glibc statically is ["highly discouraged"](https://stackoverflow.com/questions/57476533/why-is-statically-linking-glibc-discouraged). Because of that I'm hesitant to add an official setting that uses `-static`.

For reference, ccache's "mostly static binary release" is built on an old enough distribution (currently Alma Linux 8): <https://github.com/ccache/ccache/blob/1c6e6d6b5e30524d8b8a490be9169f051c713f31/.github/workflows/build.yaml#L314-L344>

This makes the binary work on any reasonably modern Linux, including Ubuntu 19.04.

But if you do want to use `-static` anyway you can do it like this:

```
cmake -D STATIC_LINK=ON -D CMAKE_EXE_LINKER_FLAGS_INIT=-static
```

berndpetrovitsch:
thx for the fast response.
And sorry that I overlooked the "STATIC_LINK" parameter  in the documentation (just near the build description:-().

While I understand the reasoning for not static linking glibc, I get (with "D STATIC_LINK=ON") and do not liking it either. FWIW that runs on an ordinary desktop PC so space is not really an issue.

After compiling with `-D STATIC_LINK=ON` on Debian/sid, running the produced binyra on Ubuntu/disco looks like this:
```
{71708}file ccache
ccache: ELF 64-bit LSB pie executable, x86-64, version 1 (GNU/Linux), dynamically linked, interpreter /lib64/ld-linux-x86-64.so.2, BuildID[sha1]=30272af971fcc461d3b64723a721e73732f98bf9, for GNU/Linux 3.2.0, not stripped
{71709}ldd ccache
./ccache: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.38' not found (required by ./ccache)
./ccache: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.35' not found (required by ./ccache)
./ccache: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.32' not found (required by ./ccache)
./ccache: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.33' not found (required by ./ccache)
./ccache: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.36' not found (required by ./ccache)
./ccache: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.34' not found (required by ./ccache)
	linux-vdso.so.1 (0x00007ffcb0580000)
	libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007f38a2207000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f38a201c000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f38a26e1000)
	libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007f38a2016000)
{71710}./ccache
./ccache: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.38' not found (required by ./ccache)
./ccache: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.35' not found (required by ./ccache)
./ccache: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.32' not found (required by ./ccache)
./ccache: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.33' not found (required by ./ccache)
./ccache: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.36' not found (required by ./ccache)
./ccache: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.34' not found (required by ./ccache)
```
That's the reason for the simple "-static" (and I didn't find any useful docs on how to solve running a binary compiled against a current glibc on an pretty old glibc system).

Debian/Disco has cmake-3.13.4 and ccache apparently needs cmake-3.18. Simply compiling current `cmake` on Debian/Disco needs more new built packages just for the the build of `cmake` (opening cans of worms- tried/started that ...) ....

Of course I like the solution with "-D CMAKE_EXE_LINKER_FLAGS_INIT=-static" better than with my patch (as
my patch doesn't save anything). And I'm not using `cmake` myself (except from copy-pasting command lines from docs) so I'm basically a `cmake` noob ....

jrosdahl:
I understand why you reach for `-static` – it's an easy solution, but not guaranteed to work with glibc. Taking up more space is not the reason why `STATIC_LINK` doesn't use `-static`.

The harder but safer solution is to build on a system with a glibc version older than (or equal to) glibc versions on your other target systems. If you build on Ubuntu 22.04 (an actually supported version :wink:) it will work on Debian unstable.

Currently ccache does have code paths (when using remote HTTP storage) that trigger loading of NSS `.so` files even for a statically linked binary. Based on this I don't want to add an official way to add `-static`. But if `-D CMAKE_EXE_LINKER_FLAGS_INIT=-static` works for you then I suggest using that.

PR #1647, opened 2025-10-14
Handle absolute path in `-fsanitize-blacklist=`

d-alonso:
I was compiling https://github.com/llvm/llvm-project in a couple of different directories and realized ccache stopped working for me, when it previously had worked correctly by setting CCACHE_BASEDIR. I followed the debugging steps on the ccache docs and realized that the hashed invocation was different due to the different absolute path for the `-fsanitize-blacklist=` option, which LLVM [uses](https://github.com/llvm/llvm-project/blob/d49aa40fc703d061f48e520b9c1b63e6646c6907/llvm/cmake/modules/HandleLLVMOptions.cmake#L1118) when UBSAN is enabled.

### DIFF
--- a/src/ccache/argprocessing.cpp
+++ b/src/ccache/argprocessing.cpp
@@ -964,8 +964,10 @@ process_option_arg(const Context& ctx,
   }
 
   if (util::starts_with(arg, "-fsanitize-blacklist=")) {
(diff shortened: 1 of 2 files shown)


jrosdahl:
Thanks!

PR #1648, opened 2025-10-19
chore: Update C++ standard in `.clang-format`

eisaev:
Currently, the project uses the C++17 standard, but `clang-format` is still configured for C++11.


jrosdahl:
Thanks.

PR #1650, opened 2025-10-25
Add `msvc_utf8` option to avoid adding `/utf-8`  to msvc command line for non-utf8 source code

MaxEW707:
Fixes a bug introduced by https://github.com/ccache/ccache/commit/441da4b4408e5e21a15e5f3c6cc6e50a5db92842 in ccache 4.12.x.

`/utf-8` when added to msvc also implicitly enables `/validate-charset` which validates that the source code file, now treated as utf-8, has no invalid utf-8 byte sequences.
We have some super old source code that uses characters in the extended Windows code page range, [128, 255], in code comments. This causes an invalid utf-8 sequence in some files which causes `/validate-charset` to error and thus preprocessing the file to fail.

We can fix our code but unfortunately we also have some headers from third party proprietary libraries that also include extended Windows codepage characters in their comments. Due to licensing we are not allowed to modify code in their headers so effectively we cannot use ccace 4.12.x until we have an option to disable passing `/utf-8` to the preprocessor.

None of our filenames have non-portable characters in them so while I agree `/utf-8` is the best path forward, why I left the default boolean value as true,  the merit for us at the moment is very low. Plus we can't modify vendor headers for the third party libs so we are kinda stuck anyways.

### DIFF
--- a/doc/manual.adoc
+++ b/doc/manual.adoc
@@ -907,6 +907,16 @@ file in `/etc/rsyslog.d`:
     The default prefix is "`Note: including file:`". If you use a localized
     compiler, this should be set accordingly.
 
+[#config_msvc_utf8]
+*msvc_utf8* (*CCACHE_MSVC_UTF8*)::
+
+    This option adds `/utf-8` to the msvc command line when executing the preprocessor to
+    ensure that filenames are not garbled for non-ascii characters.
+    This implicitly enables `/validate-charset` and treats the source code as utf-8 which
+    may cause compilation errors if comments in your code have characters in the [128, 255]
+    range for a given Windows system codepage which results in an invalid utf-8 sequence.
+    The default is true.
+
 [#config_namespace]
 *namespace* (*CCACHE_NAMESPACE*)::
 
@@ -1587,7 +1597,7 @@ An environment variable not supported by ccache was set.
 
 | Unsupported source encoding |
 Source file (or an included header) has unsupported encoding. ccache currently
-requires UTF-8-encoded source code for MSVC.
+requires UTF-8-encoded source code for MSVC when `msvc_utf8` is true.
 
 | Unsupported source language |
 A source language e.g. specified with `-x` was unsupported by ccache.
--- a/src/ccache/ccache.cpp
+++ b/src/ccache/ccache.cpp
@@ -1412,7 +1412,9 @@ get_result_key_from_cpp(Context& ctx, util::Args& args, Hash& hash)
     // compilers that don't exit with a proper status on write error to stdout.
     // See also <https://github.com/llvm/llvm-project/issues/56499>.
     if (ctx.config.is_compiler_group_msvc()) {
-      args.push_back("-utf-8"); // Avoid garbling filenames in output
+      if (ctx.config.msvc_utf8()) {
+        args.push_back("-utf-8"); // Avoid garbling filenames in output
+      }
       args.push_back("-P");
       args.push_back(FMT("-Fi{}", preprocessed_path));
     } else {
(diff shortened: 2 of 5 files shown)


jrosdahl:
Thanks for the PR.

Just to be clear, is the problem you're referring to that ccache logs "Non-UTF-8 source code unsupported in preprocessor mode" and falls back to running the compiler, or do you get another warning/error (C4819)? Can you provide a small reproducer?

I agree that it would be nice to continue supporting non-UTF8 code bases. Ideally in a way that doesn't require adding yet another config knob, so I'd like to understand more about the issue before just making it configurable.

MaxEW707:
I will use the following example code where `é` is `0xE9` in the Windows 1252 code page
```
// testé

void testé() {}
```
`/utf-8` implicitly enables `/validate-charset` which will emit the `C4828` warning, "The file contains a character starting at offset 0xvalue that is illegal in the current source character set (codepage number).".
Since we build with warnings as errors, /WX, this causes a preprocessor error and thus ccache falls back to running the compiler without caching.

We can globally add `/wd4828` to our build system to ignore the warning.
This will blow out our cache on a ccache upgrade due to changing the command line. Inconvenience until the cache gets repopulated.

Disabling the warning also may allow mangled preprocessed output to get thru ccache as it will no longer detect the warning in the output. In my local testing I have seen some characters just removed from the preprocessed output but if the only places where you use these characters are in comments then you will not hit this use case unless you have `keep_comments_cpp` enabled for some reason with msvc.

However depending on the extended codepage and character used the msvc preprocessor will just hard error with `C2018`, "unknown character '0xnn'". I haven't fully figured out when msvc hard errors here when translating from the source character set to the base source character set during preprocessing but there is no workaround here besides changing user code.

In the example above you can build without `/utf-8` however building or preprocessing with `/utf-8` will error with `C2018`.

So far for the cases I hit in our code base it appears I can get away with doing `/wd4828` and live with the cache getting blown away unfortunately.
However I felt it was probably worthwhile to add a ccache option so current code can continue using the codepages it always has without worry while still defaulting `utf-8` to true.

jrosdahl:
Thanks for providing more details! Just wanted to check whether the compilation failed with ccache or just wasn't cached. So this means that ccache works as intended (as of 441da4b4408e5e21a15e5f3c6cc6e50a5db92842), which is good. While this is intentional behavior and not a bug per se, it's of course still unfortunate for users who can't update legacy code bases.

So unless we can come up with a better way, I guess we'll have to live with a new config option. I'm fine with that.

jrosdahl:
Thanks.

jrosdahl:
> So unless we can come up with a better way, I guess we'll have to live with a new config option.

I've now implemented this in a better way: instead of relying on source encoding of the preprocessor output, I use `/sourceDependencies` to extract include file paths. This removes the need for a configuration option. See 496585daf92f4bb7dee22725c9f3161bdde184f2 for details. @MaxEW707: Please give it a try.

MaxEW707:
> > So unless we can come up with a better way, I guess we'll have to live with a new config option.
> 
> I've now implemented this in a better way: instead of relying on source encoding of the preprocessor output, I use `/sourceDependencies` to extract include file paths. This removes the need for a configuration option. See [496585d](https://github.com/ccache/ccache/commit/496585daf92f4bb7dee22725c9f3161bdde184f2) for details. @MaxEW707: Please give it a try.

Hey @jrosdahl I gave your `/sourceDependencies` implementation a try and it works with my test files. This solution works for our codebase and removing my `msvc_utf8` config option in favour of your implementation is good with me.
Do note I only tested with VS2022 since that is what we currently use ccache with. I did not test with VS2017, VS2019 or VS2026.

jrosdahl:
I wrote:

> I've now implemented this in a better way: instead of relying on source encoding of the preprocessor output, I use `/sourceDependencies` to extract include file paths. This removes the need for a configuration option. See [496585d](https://github.com/ccache/ccache/commit/496585daf92f4bb7dee22725c9f3161bdde184f2) for details. @MaxEW707: Please give it a try.

Now reverted since it didn't work out, so the `msvc_utf8` option is back on master.

Issue #1651, opened 2025-10-26
Ccache 12.x Neon detection fails


awawa-dev:
### How to reproduce ###
Compiling the project on arm64 using cmake versions older than 3.20 causes the build process to fail.
The issue was introduce by this PR: https://github.com/ccache/ccache/pull/1624
But this setting in available only in cmake 3.20 and higher: https://cmake.org/cmake/help/latest/variable/CMAKE_LANG_BYTE_ORDER.html
Currently, Ccache requires cmake version 3.18+, so we should be able to use it without errors like this, especially since it's not a critical functionality of the newer cmake.

So in the end CMAKE_LANG_BYTE_ORDER is empty here for cmake 3.18 and new code breaks neon support despite it's present and has correct byte order. Hardcoded `-DCMAKE_C_BYTE_ORDER=LITTLE_ENDIAN` workaround causes the project to be build correctly.

### Actual behavior ###
```
[ 97%] Building CXX object unittest/CMakeFiles/unittest.dir/testutil.cpp.o 47.93
[ 98%] Building CXX object unittest/CMakeFiles/unittest.dir/test_inodecache.cpp.o 49.62
[100%] Linking CXX executable unittest 49.74 /usr/bin/ld: ../src/third_party/blake3/libdep_blake3.a(blake3_dispatch.c.o): in function `blake3_hash_many': 49.74 blake3_dispatch.c:(.text+0xa0): undefined reference to `blake3_hash_many_neon'
49.76 collect2: error: ld returned 1 exit status 49.76 make[2]: *** [unittest/CMakeFiles/unittest.dir/build.make:680: unittest/unittest] Error 1
49.76 make[1]: *** [CMakeFiles/Makefile2:795: unittest/CMakeFiles/unittest.dir/all] Error 2
49.76 make: *** [Makefile:182: all] Error 2 49.76 Make failed ------ 
```

### Expected behavior ###
Compile the project on Debian Bullseye arm64 (cmake  3.18.4)
- Provide different byte order detection.
- or if CMAKE_LANG_BYTE_ORDER is empty assume it's little endian. Just like it happened in Ccache 11.x and older.

### Environment ###
Ccache 12.1, 12.0
Debian Bullseye arm64/armhf

jrosdahl:
@neuschaefer: Would you be interested in fixing this?

neuschaefer:
Thanks for the ping (and excellent report, @awawa-dev), and sorry for the breakage! I'll look into it and open a PR.

PR #1652, opened 2025-10-27
build(blake3): Assume aarch64 to be little-endian on CMake 3.18/3.19

neuschaefer:
The recently introduced detection of big-endian AArch64 uses the CMAKE_C_BYTE_ORDER variable, which was introduced in 3.20, but ccache currently only requires CMake 3.18.

To avoid using an undefined variable, simply assume AArch64 to be little-endian on CMake versions that don't define the aforementioned variable. This is okay because big-endian AArch64 has few users and the subset of them that use CMake 3.18/3.19 is likely negligible.

Reported-by: @awawa-dev
Fixes: https://github.com/ccache/ccache/issues/1651

---

Based on testing with Debian bullseye, which ships CMake 3.18, I was able to reproduce both a build failure and that this commit appears to fix it.

@awawa-dev, could you please test it in your setup as well?


awawa-dev:
Hi
Thank you for the quick fix. Building the version from the PR using the GitHub Action on different platforms, including Debian Bullseye arm64/armhf, confirms that it resolves the issue: https://github.com/awawa-dev/HyperDockerBuilder/actions/runs/18852250754

jrosdahl:
Thanks.

PR #1659, opened 2025-12-08
feat: Add Unix socket support for HTTP storage backend

andrewjcg:
Add http+unix:// URL scheme for connecting to HTTP cache servers via Unix domain sockets, similar to the existing redis+unix:// support.

URL format: http+unix:/path/to/socket?path=/optional/http/path


jrosdahl:
Thanks for the PR.

Could you explain your use case for using a local HTTP server over a UNIX domain socket?

andrewjcg:
Ahh yeah, so we have some containerized builds where we turn off network access, so this would be to allow bind-mounting in a unix domain socket to allow access a remote cache.

jrosdahl:
I see, thanks.

The reason I'm asking is that the next ccache release will include experimental support for #1214. If that works out well, my plan is to _remove_ the built-in HTTP support altogether, so I'm hesitant to add built-in support for http+unix only to remove it again later.

Once the implementation of #1214 is in place, it will be possible to run an HTTP storage helper on the container host and bind-mount its Unix domain socket into the container for similar effect to http+unix.

jrosdahl:
> Once the implementation of https://github.com/ccache/ccache/issues/1214 is in place, it will be possible to run an HTTP storage helper on the container host and bind-mount its Unix domain socket into the container for similar effect to http+unix.

Now available on master, see _Starting a helper explicitly_ on <https://ccache.dev/storage-helpers.html>.

From my perspective, this removes the need to support HTTP over a Unix socket. Do you agree?

PR #1660, opened 2025-12-08
feat: Cache -E preprocessing commands in direct/depend mode

andrewjcg:
Previously, -E commands would bail with called_for_preprocessing. Now they are cached when direct mode or depend mode is enabled, which is the default configuration.


jrosdahl:
Thanks for the PR.

Could you provide some background on when it would be of interest to cache preprocessor calls?

PR #1661, opened 2025-12-13
fix: Do not rewrite -isysroot as relative path

### LINKED ISSUE #1322: base_dir does not play well with Clang relocatable PCH
### How to reproduce ###

1. `touch pch.h`
2. `clang-16 -isysroot ${PWD} --relocatable-pch -x c++-header pch.h -o pch1.pch`
3. `CCACHE_BASEDIR=${PWD} ccache clang-16 -isysroot ${PWD} --relocatable-pch -x c++-header pch.h -o pch2.pch`

### Actual behavior ###

pch1.pch is relocatable relative to current directory, i.e. it contains relative path `pch.h`. pch2.pch is not relocatable because ccache modified `isysroot` argument to `.` which Clang then failed to match header path against.

```
$ grep -r "$(realpath pch.h)" .
Binary file ./pch2.pch matches
```

I am not 100% sure that this is ccache's fault instead of clang's. I think there might be situations where you may want PCH to be relocated against relative path, will clarify it with LLVM devs.

### Expected behavior ###

Produced PCH files are binary identical (which they are if you do not pass CCACHE_BASEDIR).

### Environment ###

ccache version 4.8

lbonn:
Here is an attempt to solve #1322 with one of the proposed approaches from the thread.

To reduce risk, I thought about restricting this behavior to invocations with `--relocatable-pch` but it would probably lead to bigger changes as `hash_argument` is collecting everything in one pass.

Fixes #1322

### DIFF
--- a/src/ccache/argprocessing.cpp
+++ b/src/ccache/argprocessing.cpp
@@ -1253,12 +1253,18 @@ process_option_arg(const Context& ctx,
     // Potentially rewrite path argument to relative path to get better hit
     // rate. A secondary effect is that paths in the standard error output
     // produced by the compiler will be normalized.
-    fs::path relpath = core::make_relative_path(ctx, args[i + next]);
+    fs::path rewritten_path;
+    if (compopt_takes_path_abs(arg)) {
+      // some paths should not be made relative (eg -isysroot)
+      rewritten_path = args[i + next];
+    } else {
+      rewritten_path = core::make_relative_path(ctx, args[i + next]);
+    }
     state.add_common_arg(args[i]);
     if (next == 2) {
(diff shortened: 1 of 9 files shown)


lbonn:
Ah well removing TAKES_PATH probably broke the WIN32 path conversion, it was indeed suspicious from the start. Maybe adding a new compopt flag would be necessary, I'll have a look.

lbonn:
I've noticed that the pch tests actually did not run at all with clang due to some default extension mix-up (maybe it was updated in a more recent clang version).

Additionally, the 2 parts of the change: do not rewrite -isysroot and cache it as relative with --relocatable-pch; have also been split.

PR #1672, opened 2026-01-18
Add remote storage helper support

### LINKED ISSUE #1214: New storage backend model

jrosdahl:
This PR introduces support for external remote storage helpers, enabling ccache to communicate with various storage backends through long-lived helper processes. This architecture keeps protocol-specific code outside ccache's core, allowing independent development and release of storage backend implementations.

Two storage helper implementations are currently available:

- <https://github.com/ccache/ccache-storage-http-cpp>: HTTP(S) helper written in C++
- <https://github.com/ccache/ccache-storage-http-go>: HTTP(S) helper written in Go

Overview of storage helpers: https://ccache.dev/storage-helpers.html

Closes: #1214

## Overview

Storage helpers are executables named `ccache-storage-<scheme>` (e.g. `ccache-storage-https`) that ccache spawns on demand. They:

- Listen on Unix domain sockets (Unix/POSIX) or named pipes (Windows).
- Communicate via a custom IPC protocol, documented in doc/remote_storage_helper_spec.md.
- Keep connections alive to amortize session setup costs.
- Auto-terminate after a configurable idle timeout.

### Storage helper support

- Ccache starts the helper on demand.
- Alternatively, it's possible to connect directly to a manually spawned storage helper using a `crsh:` URL.

### Configuration improvements

- Split `remote_storage` attributes into "properties" (ccache-known) and "custom attributes" (helper-specific). Property syntax: `key=value`, custom attribute syntax: `@key=value`. They are now whitespace-separated instead of pipe-separated values (backward compatible).
- New properties: `data-timeout`, `request-timeout`, `idle-timeout`, `helper`.
- New `libexec_dirs` config option for helper discovery.

### Key encoding standardization

- Reverted to pure base16 encoding for all hash digest keys (was mixed base16/base32hex). This ensures consistency between local storage, remote storage and logs.
- Removed R/M suffixes from filenames for consistency with remote storage.
- The local storage remains backward compatible during a transition period.

### Testing

- Added a new integration test suite for storage helpers: `remote_helper.bash`. These tests use a simplistic storage helper implementation with in-memory storage.
- Added a CLI test client for testing storage helpers.

### Other

- The built-in HTTP and Redis support is retained but will likely be removed in a future release.
- A new `--stop-storage-helpers` command line option is available to gracefully shut down helpers.


tmewett:
Hi Joel - am I right in saying that the change back to base16 keys in 7d64f3b makes it so that file:// remote cache entries created after this change (4.13+) can't be found by older versions?

jrosdahl:
> Hi Joel - am I right in saying that the change back to base16 keys in [7d64f3b](https://github.com/ccache/ccache/commit/7d64f3baf1e99ebbaf185aeef5426f04de0638de) makes it so that file:// remote cache entries created after this change (4.13+) can't be found by older versions?

Yes, that's correct. The backward compatibility lookup that was added in that commit applies to local storage, not `file://` or `http://` remote storage.

As a related point, cache sharing between versions can also be lost when there is a change of what goes into the hashed material (the input hash). In that case the underlying cache key changes rather than just its textual encoding, so the versions naturally end up looking for different entries. This doesn't happen often, but will actually happen in ccache 4.14 (soon to be released).

PR #1043, opened 2022-04-01
fix: Add ASM and ASM_MASM to project() languages

rkitover:
This fixes building and linking blake3 with MSVC for both Ninja and
msbuild.

See #1009.

Signed-off-by: Rafael Kitover <rkitover@gmail.com>

Issue #1009, opened 2022-02-21
Endless loop on startup when built with -static (MSYS2/Windows 10)


vsplesk:
[Logfile_filtered.csv](https://github.com/ccache/ccache/files/8111193/Logfile_filtered.csv)
### How to reproduce ###

1. Install latest MSYS2, with gcc and base-devel and ninja (which brings gcc 11.x)
2. Configure with -DCMAKE_BUILD_TYPE=MinSizeRel -DCMAKE_CXX_FLAGS=-static -DZSTD_FROM_INTERNET=ON -G Ninja -DREDIS_STORAGE_BACKEND=OFF
3. Build with "cmake --build ."
4. Run ./ccache.exe --help (which works)
5. Configure a real project (cmake, with CMACE_CXX_LAUNCHER option pointing to the new ccache.exe)

### Actual behavior ###
ccache gets stuck, showing 100% CPU load on every core

### Expected behavior ###
ccache shall work. It did work the last time I tested the code from PR #942 (Tricore support), however it was compiled with MSVC last time (which did work last year but now MSVC build seems to be totally broken, hence switching to MSYS2 toolchain).

Also see attachment for the monitored system activities - it seems to keep repeating the last ~25-30 operations again and again and again.

### Environment ###
Windows 10 x64, build 1809. Calling the user build process from a git-bash.
See above for other tools involved.

ccache version 4.x
(from current master, https://github.com/ccache/ccache/commit/22a87e468e721649a622ee894e692963d5d4114c )

jrosdahl:
That system activity log unfortunately doesn't tell me anything.

@rkitover, @rvogg, @cristianadam, @llunak: Any ideas here?

rkitover:
I will look, I don't think you're supposed to use `-static` on Windows, just link to static libs, at least I've never done it that way, I don't know.

I will look at fixing MSVC support too, I did that last time, unfortunately everything bitrots.

vsplesk:
Regular MSVC build did not work for me for obscure reasons - failed to link, references to some blake3* code were not resolvable. Although the generated msbuild setup looked okay at the first glance.

Meanwhile I repeated the action on a totally fresh Windows installation. Got MSYS2 and mingw-w64 toolchain (howto from msys2.org) plus cmake and libzstd-devel. Result: same mess, nothing works as soon as a file is specified. Tried to observe with gdb but the issue is elusive, the interrupt often stops "somewhere" and the stack does not tell me much.

[...]

PR #1678, opened 2026-01-26
fix(cmake): Enable assembler compiler conditionally

deckstose:
CMake will call `enable_language` for all languages passed to `project`. This deduplicates the work and only enables one assembler depending on the platform.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,6 +1,6 @@
 cmake_minimum_required(VERSION 3.18)
 
-project(ccache LANGUAGES C CXX ASM ASM_MASM)
+project(ccache LANGUAGES C CXX)
 if(MSVC)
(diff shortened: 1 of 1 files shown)


jrosdahl:
Thanks for the PR.

However, this contradicts 47eceb4d4f0646d6785c6ddfdb28343976549ed3 (see https://github.com/ccache/ccache/issues/1009#issuecomment-1086299665 and #1043), so I suspect this would bring back some problem, but I'm not sure what the actual problem was. @rkitover: Any recollection?

rkitover:
I don't remember exactly, but this will probably be OK. If not, I will fix it correctly.

jrosdahl:
OK, since CI is happy with this and I'm not aware of anybody actually using MSVC to build ccache, I'll go ahead.

PR #1680, opened 2026-01-27
enhance: Use copy_file_range to copy files if available

olegsidorkin:
Check if copy_file_range function is present and use it.

`sys/sendfile.h` header is linux-specific and calling it with file descriptor destination is linux-specific too.
copy_file_range is adopted a little bit wider, however requires newer kernels (5.3+) .

I placed it before sendfile block for tests to pick it before sendfile, but IMO finally it should be committed after sendfile block not to cause degradation on linux with older kernels.

### DIFF
--- a/src/ccache/util/file.cpp
+++ b/src/ccache/util/file.cpp
@@ -161,17 +161,23 @@ copy_file_impl(const fs::path& src,
       FMT("Failed to copy {} to {}: {}", src, dest, strerror(errno)));
   }
   return {};
-#  elif defined(HAVE_SYS_SENDFILE_H)
+#  elif defined(HAVE_COPY_FILE_RANGE) || defined(HAVE_SYS_SENDFILE_H)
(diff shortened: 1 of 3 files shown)


src/ccache/util/file.cpp
```diff
@@ -161,6 +161,28 @@ copy_file_impl(const fs::path& src,
       FMT("Failed to copy {} to {}: {}", src, dest, strerror(errno)));
   }
   return {};
+#  elif defined(HAVE_COPY_FILE_RANGE)
```
jrosdahl: To avoid code duplication, I suggest making this

```
#  elif defined(HAVE_COPY_FILE_RANGE) || defined(HAVE_SYS_SENDFILE_H)
```

and `#ifdef`-ing around the `copy_file_range`/`sendfile` call instead.

Yes, preferring `sendfile` if available is probably the safest bet.

jrosdahl:
Thanks.

PR #1681, opened 2026-01-27
fix(cmake): atomic library build error fixed for Apple Silicon macOS

haaks1998:
A fix has been added for the following build error on Apple Silicon macOS:  
`CMake Error at cmake/StdAtomic.cmake:27 (message): Toolchain doesn't support std::atomic with or without -latomic`

Description:
On macOS, std::atomic is part of the standard library and doesn't require linking against -latomic like on Linux systems. 
To handle this, a check using a CMake flag for Apple devices has been added in the **StdAtomic.cmake** file, setting the `std_atomic_without_libatomic` flag to **_TRUE_** for macOS to avoid linking with -latomic.

### DIFF
--- a/cmake/StdAtomic.cmake
+++ b/cmake/StdAtomic.cmake
@@ -4,20 +4,26 @@ set(LIBATOMIC_STATIC_PATH "" CACHE PATH "Directory containing static libatomic.a
 
 include(CheckCXXSourceCompiles)
 
-set(
-  check_std_atomic_source_code
-  [=[
-    #include <atomic>
-    int main()
-    {
-      std::atomic<long long> x;
-      ++x;
-      (void)x.load();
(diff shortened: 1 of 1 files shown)


jrosdahl:
Thanks for the PR.

The intention of `StdAtomic.cmake` is to figure out if `-latomic` is needed without having to add conditionals on platform, compiler, etc. I guess this means that

https://github.com/ccache/ccache/blob/3927379d201777130dbc1db7096097f427b49009/cmake/StdAtomic.cmake#L10-L17

doesn't compile on Apple Silicon macOS (neither with nor without `-latomic`).

 I would prefer to not add `if(APPLE)` and instead fix the feature check. What error message do you get if you compile that test program manually?

PR #1683, opened 2026-02-09
Feat thinlto

zcfh:
# Background
**ThinLTO** is a linker-time optimization in clang that supports parallel and incremental compilation compared to LTO. However, in normal usage, incremental and parallel compilation are performed internally by the linker. Clang provides an additional [distributed ThinLTO](https://blog.llvm.org/2016/06/thinlto-scalable-and-incremental-lto.html) usage, similar to regular compilation, but with additional input. It allows LTO to be distributed and cached like regular object files, requiring only some adaptation work from the build system.

Currently, I have implemented the above adaptation on my own build system. Additionally, we use ccache as a caching tool, and there have been some changes to ccache; I hope to merge these changes into the upstream implementation.

# Change
The specific difference regarding distributed ThinLTO is the addition of the option `-fthinlto-index=lib1.o.thinlto.bc`. The `thinlto.bc` file already contains the hashes of the other object files it depends on, so simply adding `.thinlto.bc` to the cache's hash calculation is sufficient.


zcfh:
It seems the check process failed because of downloading some dependencies. Should I retry?

src/ccache/argprocessing.cpp
```diff
@@ -735,6 +736,16 @@ process_option_arg(const Context& ctx,
     return Statistic::none;
   }
 
+  if (arg.starts_with("-fthinlto-index=")) {
+    std::string thinlto_index = arg.substr(arg.find('=') + 1);
+    args_info.thinlto_index_path = thinlto_index;
+    // Thinlto backend phase, the extension of input file is .o but the file is
+    // IR. So treat it as assembler file.
+    args_info.actual_language = "assembler";
```
jrosdahl: It doesn't feel good to let the code intentionally use an incorrect language here.

test/suites/distributed_thinlto_clang.bash
```diff
@@ -0,0 +1,80 @@
...
+    $COMPILER -flto=thin -O2 -c -o lib1.o lib1.c
+    $COMPILER -flto=thin -O2 -c -o lib2.o lib2.c
+    # Use compiler to generate thinlto.index.bc.
+    $COMPILER -fuse-ld=lld -Wl,--thinlto-index-only=test.rst test.o lib1.o lib2.o
+    backdate *.c *.o *.bc
+    md5sum *.bc >before.log
+    find -name "*.bc"
+    pwd
+    unset CCACHE_NODIRECT
+    set +x
```
jrosdahl: Looks like left-over debugging stuff?


```suggestion
```

src/ccache/argprocessing.cpp
```diff
@@ -89,6 +89,7 @@ class ArgumentProcessingState
   bool found_wp_md_or_mmd_opt = false;
   bool found_md_or_mmd_opt = false;
   bool found_Wa_a_opt = false;
+  bool found_thinlto_index_opt = false;
```
jrosdahl: Looks unused?

jrosdahl (commented):
Thanks for the PR and thanks for writing tests. Looks good in general.

zcfh:
Thank you for reviewing the code. ISorry, I forgot to clean up some unused code; it's been fixed now.

Also, regarding the change to `args_info.actual_language = "assembler"`,I'd love to see a more elegant approach If it's convenient, please @ me in your new pull request.

Finally, as a side note, when using `thinlto`, you can optionally add the parameter `-x ir`, and the compiler implicitly adds this option by default. This information might be helpful.

src/ccache/argprocessing.cpp
zcfh: I'd love to see more elegant methods. As a side note, thinlto allows to optionally add `-x ir`.

jrosdahl:
> Finally, as a side note, when using thinlto, you can optionally add the parameter -x ir, and the compiler implicitly adds this option by default. This information might be helpful.

That's good to know, thanks.

PR #1684, opened 2026-02-11
refactor: Generalize "already preprocessed" compiler language slightly

jrosdahl:
- Rename `ArgsInfo::direct_i_file` to `preprocess_input_file` (inverting the logic), since that's what it now means.
- Add "ir" as a language and let `-fthinlto-index=` use that language implicitly instead of piggybacking on "assembler".


jrosdahl:
@zcfh: Review welcome.

PR #1689, opened 2026-03-05
chore: use CMake knowledge about helper locations

gjasny:
In our build system we set [EXECUTABLE_OUTPUT_PATH](https://cmake.org/cmake/help/latest/variable/EXECUTABLE_OUTPUT_PATH.html) which breaks the `remote_helper` test. Pass the binary location from CMake to the test runner.

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -8,7 +8,7 @@ function(addtest name)
   set_tests_properties(
     "test.${name}"
     PROPERTIES
-    ENVIRONMENT "CCACHE=$<TARGET_FILE:ccache>;EXIT_IF_SKIPPED=true"
+    ENVIRONMENT "CCACHE=$<TARGET_FILE:ccache>;STORAGE_TEST_HELPER=$<TARGET_FILE:ccache-storage-test>;STORAGE_TEST_CLIENT=$<TARGET_FILE:storage-test-client>;EXIT_IF_SKIPPED=true"
     SKIP_RETURN_CODE 125)
 endfunction()
 
--- a/test/run
+++ b/test/run
@@ -555,6 +555,16 @@ if [ -z "$CCACHE" ]; then
     CCACHE=`pwd`/ccache
 fi
 
+if [ -z "$STORAGE_TEST_HELPER" ]; then
+    STORAGE_TEST_HELPER=$(pwd)/test/storage/helper/ccache-storage-test
+fi
+
+if [ -z "$STORAGE_TEST_CLIENT" ]; then
+    STORAGE_TEST_CLIENT=$(pwd)/test/storage/client/storage-test-client
+fi
+
+readonly CCACHE STORAGE_TEST_HELPER STORAGE_TEST_CLIENT
+
 COMPILER_TYPE_CLANG=false
 COMPILER_TYPE_GCC=false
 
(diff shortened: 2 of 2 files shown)


jrosdahl:
That's a CMake setting I didn't know about.

However, this means that running the suite directly, i.e. not via ctest, won't work. How about generating a settings script with the locations and then let `test/run` source that script?

gjasny:
That's more complicated than expected. For multi-config CMake generators the output paths depend on the configuration. The `CTestTestfile.cmake` takes care of that:

```cmake
if(CTEST_CONFIGURATION_TYPE MATCHES "^([Dd][Ee][Bb][Uu][Gg])$")
  add_test(test.base "bash" "/Users/gregorj/src/ccache/test/run" "base")
  set_tests_properties(test.base PROPERTIES  ENVIRONMENT "CCACHE=/Users/gregorj/src/ccache/_build/Debug/ccache;STORAGE_TEST_HELPER=/Users/gregorj/src/ccache/_build/test/storage/helper/Debug/ccache-storage-test;STORAGE_TEST_CLIENT=/Users/gregorj/src/ccache/_build/test/storage/client/Debug/storage-test-client;EXIT_IF_SKIPPED=true" SKIP_RETURN_CODE "125" WORKING_DIRECTORY "/Users/gregorj/src/ccache/_build" _BACKTRACE_TRIPLES "/Users/gregorj/src/ccache/test/CMakeLists.txt;3;add_test;/Users/gregorj/src/ccache/test/CMakeLists.txt;24;addtest;/Users/gregorj/src/ccache/test/CMakeLists.txt;0;")
elseif(CTEST_CONFIGURATION_TYPE MATCHES "^([Rr][Ee][Ll][Ee][Aa][Ss][Ee])$")
  add_test(test.base "bash" "/Users/gregorj/src/ccache/test/run" "base")
  set_tests_properties(test.base PROPERTIES  ENVIRONMENT "CCACHE=/Users/gregorj/src/ccache/_build/Release/ccache;STORAGE_TEST_HELPER=/Users/gregorj/src/ccache/_build/test/storage/helper/Release/ccache-storage-test;STORAGE_TEST_CLIENT=/Users/gregorj/src/ccache/_build/test/storage/client/Release/storage-test-client;EXIT_IF_SKIPPED=true" SKIP_RETURN_CODE "125" WORKING_DIRECTORY "/Users/gregorj/src/ccache/_build" _BACKTRACE_TRIPLES "/Users/gregorj/src/ccache/test/CMakeLists.txt;3;add_test;/Users/gregorj/src/ccache/test/CMakeLists.txt;24;addtest;/Users/gregorj/src/ccache/test/CMakeLists.txt;0;")
```

I now added a run-wrapper for the individual configurations. You could invoke it with:
```
_build/test/run.Debug.sh remote_helper
```

PS: If you only use single-config generators, the first commit should've continued to work with your use case beause `run` still contained the previously used default.

jrosdahl:
I see, thanks for looking into this. My multi-config-ignorant gut reaction was that it would be good if `../test/run` would still work, but I now realize that this isn't the case in a multi-config scenario even before (cf. #878). With that in mind I'll go with just your first commit, hope that's OK with you.

jrosdahl:
Thanks.

PR #1693, opened 2026-03-09
fix: Do not look for `-fstack-usage` and `-fcallgraph-info` output files when LTO is used

Lauszus:
When `-flto` is used the .su and .ci output files are not generated during compilation, instead they are generated during link time

This means that ccache can simply skip those files when `-fstack-usage` and `-fcallgraph-info` are used in combination with the `-flto` flag

See: https://gcc.gnu.org/onlinedocs/gcc/Developer-Options.html#index-fcallgraph-info

### DIFF
--- a/test/suites/direct.bash
+++ b/test/suites/direct.bash
@@ -640,7 +640,7 @@ EOF
         expect_missing code.su
 
         # clang does not produce .su files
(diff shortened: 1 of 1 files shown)


jrosdahl:
Sounds reasonable, thanks.

Lauszus:
> Sounds reasonable, thanks.

You are welcome :)

PR #1694, opened 2026-03-09
chore: remove obsolete condition

gjasny:
The minimum CMake version is now 3.18. The condition could be removed.


jrosdahl:
Thanks.

PR #1701, opened 2026-03-12
feat: Add support for QCC (QNX) compiler

### LINKED ISSUE #1725: Support QNX Compiler (qcc/q++)
QNX's [`qcc`](https://www.qnx.com/developers/docs/8.0/com.qnx.doc.neutrino.utilities/topic/q/qcc.html) compiler is basically a wrapper over `gcc`.

Supporting it from a ccache perspective should be very similar to supporting gcc. We'd want to add some plumbing to detect it automatically, and might need to switch around some checks to be `if(gcc || qcc)` instead of just `if(gcc)`. We'd also need a couple of other envvar compiler inputs that should always be hashed alongside the usual inputs.

Similar to how `GCC_COLORS` is currently hashed, we would need to hash `QNX_HOST`, `QNX_TARGET`, and `QCC_CONF_PATH`. That last one is optional, the first two are mandatory to invoke the compiler toolchain.

Related to #1609 (though this issue is not unique to Windows)

joybwang:
## Add support for QCC (QNX) compiler

This PR adds support for the QCC compiler, which is the QNX Neutrino RTOS compiler toolchain. QCC is a GCC-based compiler driver, so it shares many behaviors with GCC.

### Changes

**Compiler detection:**
- Auto-detect `qcc` and `q++` executables as `CompilerType::qcc`
- Added `compiler_type = qcc` configuration option

**GCC compatibility:**
- Introduced `is_compiler_group_gcc()` to group GCC and QCC together, since QCC follows GCC conventions for:
  - Diagnostics color (`-fdiagnostics-color`)
  - Dependency target handling with `-Wp,-M[M]D`
  - `GCC_COLORS` environment variable

**QNX-specific handling:**
- Hash QNX environment variables (`QNX_HOST`, `QNX_TARGET`, `QCC_CONF_PATH`) that affect compilation output
- Mark `-set-default` option as `TOO_HARD` (QCC-specific option that changes global state)

**Documentation:**
- Updated manual with `qcc` compiler type
- Added news entry

**Tests:**
- Added unit tests for QCC/q++ compiler detection

### DIFF
--- a/doc/manual.adoc
+++ b/doc/manual.adoc
@@ -683,6 +683,8 @@ _<<Using ccache with other compiler wrappers>>_.
     Microsoft Visual C++ (MSVC).
 *nvcc*::
     NVCC (CUDA) compiler.
+*qcc*::
+    QCC (QNX) compiler.
 *other*::
     Any compiler other than the known types.
 --
--- a/doc/news.adoc
+++ b/doc/news.adoc
@@ -92,6 +92,8 @@ Release date: 2026-03-05
 - Added support for caching distributed ThinLTO for Clang. +
   [small]#_[contributed by GitHub user zcfh]_#
 
+- Added support for QCC (QNX) compiler.
+
 - Added support for caching MSVC `/sourceDependencies` file. +
   [small]#_[contributed by Joel Rosdahl]_#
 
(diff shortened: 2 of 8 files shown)


Shadows-of-Fire (approved):

jrosdahl:
Thanks!

PR #1702, opened 2026-03-12
feat: Add support for ISPC compiler

### LINKED ISSUE #847: Support ISPC language
ISPC language (https://ispc.github.io/) is used in a number of rendering and game dev projects. Some of the projects are fairly large and would benefit from `ccache` support for ISPC. We have specific request from Unreal Engine 4 user to have such support (see here: https://github.com/ispc/ispc/discussions/1944#discussioncomment-709253). I'm sure other project will benefit from that as well.

ISPC command line switches are not a derivative from clang or any other compiler. But it's a relatively short list of switches.

Would you accept a patch that supports ISPC? I'm willing to try implementing it, but would like to check first if there's a chance for merging such patch.

PKizzle:
# Add support for ISPC compiler

Closes #847.

## Summary

This PR adds support for caching [Intel ISPC](https://ispc.github.io/) compilations. ISPC is a compiler for high-performance SIMD programming, widely used in rendering, game engines (e.g., Unreal Engine), and scientific computing. Unlike GCC/Clang, ISPC has its own CLI and can produce multiple output files in a single invocation when targeting multiple ISAs.

## Motivation

ISPC compilations can be slow, especially in multi-target mode where the compiler generates separate object files for each ISA (e.g., SSE4, AVX2, AVX-512). Large projects like Open3D, Embree, and Unreal Engine would benefit from caching these compilations.

## Noteworthy changes

The result serializer now enforces the actual binary format limit of `uint8_t` file counts (255 entries) instead of relying on an older hard-coded deserializer limit of 10 entries. This avoids rejecting valid multi-target ISPC results while still preventing overflow of the serialized `n_files` field. ISPC might easily surpass the old 10 entry limit.


jrosdahl:
Thanks for working on ISPC support. I have meant to comment on this PR for a long time, but as usual I have had too little time to follow up more complex issues in a way that they deserve.

I'm sorry, but this looks too intrusive and complex to add to ccache at this point since ISPC apparently is not very GCC-like (or MSVC-like). As noted in #956, I'm reluctant to add support for a new compiler if that results in too much added complexity. If and when ccache gains an internal compiler abstraction layer, I will reconsider.

PR #1706, opened 2026-03-17
feat: add support for sarif output files.

### LINKED ISSUE #1466: Support caching of additional compiler output like diagnostics
### Actual behavior ###
When I run gcc to emit [Sarif](https://groups.oasis-open.org/communities/tc-community-home2?CommunityKey=c64ae352-bebf-446d-8ebf-018dc7d3eeb0) output with the following command, a _test.o_ file and an addtional _test.cpp.sarif_ file will be written to disk.

`ccache gcc -c ../test.cpp -fdiagnostics-format=sarif-file`

When I delete the outputs and run the same command again ccache will only recreate the _test.o_, but the _test.cpp.sarif_ is missing.

### Expected behavior ###

I would be great if ccache could also cache the sarif output. 
ccache already supports caching of dependency files, which is a quite similar functionality.
Maybe its best to have a generic/configurable functionality for this, since every compiler may has different "byproducts".

### Environment ###
ccache version 4.9.1
gcc 13.2

### References ###
https://gcc.gnu.org/onlinedocs/gcc/Diagnostic-Message-Formatting-Options.html#index-fdiagnostics-format

### LINKED ISSUE #1614: Add support for msvc's `/experimental:log`
MSVC has the flag [`/experimental:log`](https://learn.microsoft.com/en-us/cpp/build/reference/experimental-log) to output warnings in the sarif format to a file. We want to use it because with msbuild and ccache we get interleaved console output. 
The problem is now that when there is a cache hit, the additional file caused by `/experimental:log` is not restored.  
So ccache must store the content of the file generated by  `/experimental:log` and restore it later on.

kfessel:
this:
- rebases #1617 
- adds support for fdiagnostics-add/set-output=sarif

@autoantwort fell free to take these changes and your original PR 

I took this from the original PR by autoantwort 

> Only msvc was tested, worked for simply example.
> Fixes https://github.com/ccache/ccache/issues/1614 and https://github.com/ccache/ccache/issues/1466.
> Open questions that I have not looked at:
>    How to handle absolute paths?
    
I @kfessel  did not and probably will not test with msvc but i also try to touch that code as little as possible if it ever blocks this PR i would remove it

most recent workflow run (since the link is invisible as soon a a push is made and a new run can not be triggerd by me) (i hope i can keep this updated)  https://github.com/ccache/ccache/actions/runs/29922981118

I addressed the clang tidy / format issues  

what went wrong here ?
https://github.com/ccache/ccache/actions/runs/29922981118/job/89202376323
https://github.com/ccache/ccache/actions/runs/30084293456/job/89474416397?pr=1706

now i know -- default file placement changed  https://gcc.gnu.org/gcc-16/changes.html

> GCC's SARIF output now respects the [dump directory](https://gcc.gnu.org/onlinedocs/gcc-16.1.0/gcc/Overall-Options.html#index-dumpdir). For example, given

      gcc \
        -o build-dir/foo.o \
        -fdiagnostics-add-output=sarif
        foo.c
    

> GCC 15 would write the SARIF to foo.c.sarif, whereas GCC 16 now writes it to build-dir/foo.c.sarif.


src/ccache/argprocessing.cpp
```diff
@@ -1548,6 +1622,21 @@ process_args(Context& ctx)
...
+  if (args_info.generating_sarif) {
+    if (ctx.config.is_compiler_group_msvc()) {
+      if (args_info.output_sarif.native().back() == '\\') {
+        args_info.output_sarif /=
+          args_info.input_file.stem().generic_string() + ".sarif";
+      }
+    } else {
+      if (args_info.output_sarif.empty()) {
+        args_info.output_sarif = args_info.input_file;
+        args_info.output_sarif.concat(".sarif");
```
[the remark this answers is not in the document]
kfessel: seems write to `<compiler working dir>/out.c.sarif` for c files 
and to `<compiler working dir>/in.cpp.sarif` for cpp files 

### PRE
```
~/ws_ccache/test-gcc/build$ 
15.2.0
~/ws_ccache/test-gcc$ find 
.
./test
./test/build
./test/in.c
./build
```

### TEST 1 in parent dir to build/
```
~/ws_ccache/test-gcc$ find ~/ws_ccache/test-gcc/  -name '*.sarif' -or -name '*.o' 
~/ws_ccache/test-gcc$ 
~/ws_ccache/test-gcc$ gcc -Wall -fdiagnostics-format=sarif-file -c test/in.c -o build/out.o
~/ws_ccache/test-gcc$ find ~/ws_ccache/test-gcc/  -name '*.sarif' -or -name '*.o' 
./build/out.o
./out.c.sarif
~/ws_ccache/test-gcc$ find .  \( -name '*.sarif' -or -name '*.o' \) -delete 
```

### TEST 2 in source dir to sub dir
```
~/ws_ccache/test-gcc/test$ find ..  -name '*.sarif' -or -name '*.o' 
~/ws_ccache/test-gcc/test$ gcc -Wall -fdiagnostics-format=sarif-file -c in.c -o build/out.o
~/ws_ccache/test-gcc/test$ find ..  -name '*.sarif' -or -name '*.o' 
../test/build/out.o
../test/out.c.sarif
~/ws_ccache/test-gcc/test$ find ..  \( -name '*.sarif' -or -name '*.o' \) -delete
```

### TEST 3 in build dir from src
```
~/ws_ccache/test-gcc/test/build$ find ../..  -name '*.sarif' -or -name '*.o' 
~/ws_ccache/test-gcc/test/build$ gcc -Wall -fdiagnostics-format=sarif-file -c ../in.c -o out.o
~/ws_ccache/test-gcc/test/build$ find ../..  -name '*.sarif' -or -name '*.o' 
../../test/build/out.c.sarif
../../test/build/out.o
~/ws_ccache/test-gcc/test/build$ find ../..  \( -name '*.sarif' -or -name '*.o' \) -delete
```

### TEST 4 in parent build dir
```
~/ws_ccache/test-gcc/build$ find ..  -name '*.sarif' -or -name '*.o' 
~/ws_ccache/test-gcc/build$ gcc -Wall -fdiagnostics-format=sarif-file -c ../test/in.c -o out.o
~/ws_ccache/test-gcc/build$ find ..  -name '*.sarif' -or -name '*.o' 
../build/out.c.sarif
../build/out.o
```

### PRE CPP
```
~/ws_ccache/test-g++$ gcc -dumpversion  
15.2.0
~/ws_ccache/test-g++$ find 
.
./test
./test/build
./test/in.cpp
./build
```

### TEST 1 in parent dir to build/
```
~/ws_ccache/test-g++$ find .  -name '*.sarif' -or -name '*.o'
~/ws_ccache/test-g++$ gcc -Wall -fdiagnostics-format=sarif-file test/in.cpp -o build/out.o
~/ws_ccache/test-g++$ find .  -name '*.sarif' -or -name '*.o'
./build/out.o
./in.cpp.sarif
~/ws_ccache/test-g++$ find .  \( -name '*.sarif' -or -name '*.o' \) -delete
```

### TEST 2 in source dir to sub dir
```
~/ws_ccache/test-g++/test$ find ..  -name '*.sarif' -or -name '*.o'
~/ws_ccache/test-g++/test$ gcc -Wall -fdiagnostics-format=sarif-file in.cpp -o build/out.o
~/ws_ccache/test-g++/test$ find ..  -name '*.sarif' -or -name '*.o'
../test/build/out.o
../test/in.cpp.sarif
~/ws_ccache/test-g++/test$ find ..  \( -name '*.sarif' -or -name '*.o' \) -delete
```

### TEST 3 in build dir from src
```
~/ws_ccache/test-g++/test/build$ find ../..  -name '*.sarif' -or -name '*.o'
~/ws_ccache/test-g++/test/build$ gcc -Wall -fdiagnostics-format=sarif-file ../in.cpp -o out.o
~/ws_ccache/test-g++/test/build$ find ../..  -name '*.sarif' -or -name '*.o'
../../test/build/in.cpp.sarif
../../test/build/out.o
~/ws_ccache/test-g++/test$ find ..  \( -name '*.sarif' -or -name '*.o' \) -delete
```

### TEST 4 in parent build dir
```
~/ws_ccache/test-g++/build$ find ..  -name '*.sarif' -or -name '*.o'
~/ws_ccache/test-g++/build$ gcc -Wall -fdiagnostics-format=sarif-file ../test/in.cpp -o out.o
~/ws_ccache/test-g++/build$ find ..  -name '*.sarif' -or -name '*.o'
../build/in.cpp.sarif
../build/out.o
~/ws_ccache/test-g++/build$ find ..  \( -name '*.sarif' -or -name '*.o' \) -delete
```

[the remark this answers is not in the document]
kfessel: for gfortran it seems to behave like for c

### TEST gfortran 
```
~/ws_ccache/test-gfortran$ gfortran -dumpversion  
15.2.0
~/ws_ccache/test-gfortran$ find .  \( -name '*.sarif' -or -name '*.o' \) -delete
~/ws_ccache/test-gfortran$ find .  -name '*.sarif' -or -name '*.o'
~/ws_ccache/test-gfortran$ find .                                 
.
./src
./src/main.f90
./build
~/ws_ccache/test-gfortran$ gfortran -Wall -fdiagnostics-format=sarif-file src/main.f90 -c -o build/out.o  
~/ws_ccache/test-gfortran$ find .  -name '*.sarif' -or -name '*.o'
./build/out.o
./out.f90.sarif
~/ws_ccache/test-gfortran$ 
```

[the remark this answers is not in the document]
kfessel: The autor of that Sarif feature has done many things around c++ -> i d guess that the intended behaviour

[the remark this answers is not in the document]
kfessel: i just realised my cpp test was wrong ( i missed the -c  -> the -o wasn't the object file but the linker result -> the name was taken from the src ( or maybe a guessed object name)

kfessel:
i rebased again to handle a merge conflict in `test/CMakeLists.txt`

kfessel:
if you want this to be cleaned / suashed or conventional commit titled (per commit ( the PR is ) please tell me.

if you want @autoantwort  to fix the original PR please tell.

if you want me to keep calm ...

src/ccache/argprocessing.cpp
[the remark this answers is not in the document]
kfessel: i think i got it 

```c 
	
      if (args_info.output_sarif.empty()) {
        state.hash_actual_cwd = true;
        args_info.output_sarif = ctx.apparent_cwd;
        args_info.output_sarif /= args_info.output_obj.stem();
        args_info.output_sarif += args_info.input_file.extension();
        args_info.output_sarif += ".sarif";
      }
      
```

kfessel:
The original PR just got closed (by ccache-bot) -- not sure how to continue this one 

- the msvc part stayed mostly the same and was tested by @autoantwort 
- the gcc part is tested by me and used in our CI

src/ccache/argprocessing.cpp
```diff
@@ -1181,6 +1181,80 @@ process_option_arg(const Context& ctx,
     return Statistic::none;
   }
 
+  if (arg.starts_with("-fdiagnostics-format=")) {
+    LOG("-fdiagnostics-format=");
+    using namespace std::string_view_literals;
+    auto param = std::string_view(arg).substr("-fdiagnostics-format="sv.size());
+    if (param == "sarif-file") {
+      args_info.generating_sarif = true;
```
jrosdahl: Looks like `gcc -fdiagnostics-format=sarif-file -fdiagnostics-format=text -c example.c` will enable `generating_sarif`? Only the latest format should apply unless I'm missing something. A test would be good.

src/ccache/argprocessing.cpp
```diff
@@ -1181,6 +1181,80 @@ process_option_arg(const Context& ctx,
...
+        auto file_end = file_param.find(',');
+        if (file_end != std::string_view::npos) {
+          args_info.output_sarif =
+            core::make_relative_path(ctx, file_param.substr(0, file_end));
+        } else {
+          args_info.output_sarif = core::make_relative_path(ctx, file_param);
+        }
+      }
+    }
+    return Statistic::none;
```
jrosdahl: If e.g. `-fdiagnostics-add-output=text` or `-fdiagnostics-set-output=text` is used then the argument is swallowed (not sent to the compiler). It should be added to the compiler argument list unconditionally. It would be great with a test that catches this!

src/ccache/argprocessing.cpp
```diff
@@ -1181,6 +1181,80 @@ process_option_arg(const Context& ctx,
...
+      args_info.generating_sarif = true;
+      if (file_pos != std::string_view::npos) {
+        auto file_param =
+          param.substr(file_pos + std::string_view("file=").size());
+        auto file_end = file_param.find(',');
+        if (file_end != std::string_view::npos) {
+          args_info.output_sarif =
+            core::make_relative_path(ctx, file_param.substr(0, file_end));
+        } else {
+          args_info.output_sarif = core::make_relative_path(ctx, file_param);
```
jrosdahl: If I understand correctly, several diagnostics output sinks can be specified, e.g. with `-fdiagnostics-add-output=sarif:file=file1.sarif -fdiagnostics-add-output=sarif:file=file2.sarif -fdiagnostics-add-output=text`. The code needs to handle this gracefully. Perhaps easiest to just fall back to the compiler if several `-fdiagnostics-add-output` or `-fdiagnostics-set-output` have been seen?

src/ccache/argsinfo.hpp
```diff
@@ -112,6 +115,9 @@ struct ArgsInfo
   // -flto specified.
   bool using_lto = false;
 
+  // Us the compiler being asked to generate sarif output files
```
jrosdahl:
```suggestion
  // Is the compiler being asked to generate sarif output files?
```

jrosdahl:
> The original PR just got closed (by ccache-bot) -- not sure how to continue this one

Let's continue here.

src/ccache/argprocessing.cpp
kfessel: testing  (gcc13.3) 
```
>ls -la
total 4
   8 Jul 21 11:56 .
 192 Jul 21 11:55 ..
 111 Mär 16 21:56 in.c
> gcc -fdiagnostics-format=sarif-file -fdiagnostics-format=text -c in.c
> ls -la
total 12
   36 Jul 21 11:56 .
  192 Jul 21 11:55 ..
  111 Mär 16 21:56 in.c
  470 Jul 21 11:56 in.c.sarif
 1504 Jul 21 11:56 in.o
```
^^ this seems like a bug to me would have expected it to behave like you described  -> tested with gcc15.2.0 behaves the same 


switch them around see if first wins 
```
> ls -la
total 4
   8 Jul 21 11:59 .
 192 Jul 21 11:55 ..
 111 Mär 16 21:56 in.c
> gcc -fdiagnostics-format=text -fdiagnostics-format=sarif-file -c in.c
> ls -la                                                               
total 12
   36 Jul 21 12:00 .
  192 Jul 21 11:55 ..
  111 Mär 16 21:56 in.c
  470 Jul 21 12:00 in.c.sarif
 1504 Jul 21 12:00 in.o
```
no -- sarif wins over text 

but 

sarif-stderr and sarif-file the last one wins -> some one seems to role dice 

or -fdiagnostics-format=text is just ignore if any other 

(`json` output in gcc13.3 behaves the same (also the `json` vs `json-file`)) 

sarif vs json 

last wins 

so `text` is ignored if any other 
for any other last wins 

> // I assume there isn't really a parser for `text` its just a neutral value that does nothing for people/ build chains that like to set every option (and also does not throw an unknown option error)

src/ccache/argprocessing.cpp
kfessel: set and add both add to the list of outputs 

set..:sarif disables the text output 

set..:text enables the text output ( does not disable a previously specified sarif output)

mutiple files may be add or set gcc 15 does not care  in any case they are added

kfessel: tried to handle the mutiple sarif file case ->   
```
args_info_t { std::vector<std::optional<std::filesystem::path>> output_sarif;}
```

ended up in  result retriver seemingly supporting exactly one file per filetype

```
ccache/core/resultretriever.cpp:217:30: error: could not convert ‘((const core::ResultRetriever*)this)->core::ResultRetriever::m_ctx.Context::args_info.ArgsInfo::output_sarif’ from ‘const std::vector<std::optional<std::filesystem::__cxx11::path> >’ to ‘util::filesystem::path’ {aka ‘std::filesystem::__cxx11::path’}
```

dropping that attempt https://github.com/kfessel/ccache/tree/p-sarif-multifile-v4.13.2 

and go the fallback to compiler if mutiple sarif files are requested

kfessel: l1189-l1195 and l1212-l1217 handle that

src/ccache/argprocessing.cpp
kfessel: moved  state.add_compiler_only_arg to l1202

kfessel:
fpush for argprocessing.cpp merge conflict

src/ccache/argprocessing.cpp
jrosdahl: OK. Since the behavior is surprising (potentially unintended, potentially different for different compiler versions), how simply about not supporting multiple `-fdiagnostics-format=`? That seems safer.

src/ccache/argprocessing.cpp
jrosdahl: It feels safer to only support one `-fdiagnostics-add-output=` or `-fdiagnostics-set-output=`. Currently, `-fdiagnostics-add-output=sarif:file=first.sarif -fdiagnostics-set-output=text` produces no SARIF file, but ccache will try to store `first.sarif`.

kfessel: actually it (gcc15) will produce a `first.sarif` -> ccache is able to find that 

`-fdiagnostics-set-output=text` "enables text output"
`-fdiagnostics-add-output=text` "enables text output"
`-fdiagnostics-set-output=sarif` "disables text output, enables sarif output and **adds** another output file (either default location or by file=""-parmeter)
`-fdiagnostics-add-output=sarif` "enables sarif output and adds another output file (either default location or by file=""-parameter)

what i did not write does not happen ( eg -fdiagnostics-set-output=text does not disable sarif output even a previous one) 
your example: `-fdiagnostics-add-output=sarif:file=first.sarif -fdiagnostics-set-output=text`
parameters are processed in order
`state: text output = 1,  sarif output = 0`
`-fdiagnostics-add-output=sarif:file=first.sarif`
`state: text output = 1,  sarif output = 1 filenames{first.sarif}`
`-fdiagnostics-set-output=text`
`state: text output = 1,  sarif output = 1 filenames{first.sarif}`


your other example: `-fdiagnostics-add-output=sarif:file=file1.sarif -fdiagnostics-add-output=sarif:file=file2.sarif -fdiagnostics-add-output=text`
`state: text output = 1,  sarif output = 0`
`-fdiagnostics-add-output=sarif:file=file1.sarif`
`state: text output = 1,  sarif output = 1 filenames{file1.sarif}`
`-fdiagnostics-add-output=sarif:file=file2.sarif`
`state: text output = 1,  sarif output = 1 filenames{file1.sarif,file2.sarif}`
`-fdiagnostics-add-output=text`
`state: text output = 1,  sarif output = 1 filenames{file1.sarif,file2.sarif}`

with set insteadof add: `-fdiagnostics-set-output=sarif:file=file1.sarif -fdiagnostics-set-output=sarif:file=file2.sarif -fdiagnostics-set-output=text`
`state: text output = 1,  sarif output = 0`
`-fdiagnostics-set-output=sarif:file=file1.sarif`
`state: text output = 0,  sarif output = 1 filenames{file1.sarif}`
`-fdiagnostics-set-output=sarif:file=file2.sarif`
`state: text output = 0,  sarif output = 1 filenames{file1.sarif,file2.sarif}`
`-fdiagnostics-set-output=text`
`state: text output = 1,  sarif output = 1 filenames{file1.sarif,file2.sarif}`

another example: `-fdiagnostics-set-output=sarif:file=first.sarif -fdiagnostics-add-output=text -fdiagnostics-set-output=sarif:file=snd.sarif -fdiagnostics-set-output=text`

`state: text output = 1,  sarif output = 0`
`-fdiagnostics-set-output=sarif:file=first.sarif`
`state: text output = 0,  sarif output = 1 filenames{first.sarif}`
`-fdiagnostics-add-output=text`
`state: text output = 1,  sarif output = 1 filename{first.sarif}`
`-fdiagnostics-set-output=sarif:file=snd.sarif`
`state: text output = 0,  sarif output = 2 filename{first.sarif,snd.sarif}`
`-fdiagnostics-set-output=text`
`state: text output = 1,  sarif output = 2 filename{first.sarif,snd.sarif}`

I thought i had it in the comment of that logic but i forgot it in my failed attempt of handling mutiple sarif files

https://github.com/kfessel/ccache/blob/4c53bb510d511b054e56036ea8158cf29557b58e/src/ccache/argprocessing.cpp#L1198-L1202

src/ccache/argprocessing.cpp
```diff
@@ -1181,6 +1181,91 @@ process_option_arg(const Context& ctx,
...
+    auto param =
+      arg_sv.substr(std::string_view("-fdiagnostics-add-output=").size());
+    auto sarif_pos = param.find("sarif");
+    auto file_pos = param.find("file=");
+    if (sarif_pos != std::string_view::npos) {
+      if (!args_info.generating_sarif)
+        args_info.generating_sarif = true;
+      else {
+        LOG("no support for multiple sarif files");
+        return Statistic::unsupported_compiler_option;
```
kfessel: @jrosdahl 
is this return the right way to tell ccache that a set of options (not a specific single one) shall cause a fallback to compiler

```suggestion
        return Statistic::unsupported_compiler_option; //<<this 
```

src/ccache/argprocessing.cpp
kfessel: this behaviour seems to at least have stuck from 13.3 to 15.3 

the very similar behaviour for `-fdiagnostics-set/add-output` is not in 13.3 bacause the option was introduced later

sidenote 
https://gcc.gnu.org/onlinedocs/gcc/Diagnostic-Message-Formatting-Options.html#index-fdiagnostics-set-output
is wrong for gcc 15.2 -fdiagnostics-set-output is **not** replacing all previous fdiagnostics-format=-fdiagnostics-set/add-output=

kfessel:
workflow run before code-formating https://github.com/ccache/ccache/actions/runs/29922981118

src/ccache/argprocessing.cpp
jrosdahl: Yep!

src/ccache/argprocessing.cpp
jrosdahl: I see. This is very messy...

I tested GCC 15.3 myself and observe this behavior:

| Options                                                         | Result       |
|-----------------------------------------------------------------|--------------|
| `-fdiagnostics-format=text`                                     | text         |
| `-fdiagnostics-format=text -fdiagnostics-format=sarif-file`     | file         |
| `-fdiagnostics-format=text -fdiagnostics-add-output=sarif`      | text + file  |
| `-fdiagnostics-format=text -fdiagnostics-set-output=sarif`      | file         |
| `-fdiagnostics-format=sarif-file`                               | file         |
| `-fdiagnostics-format=sarif-file -fdiagnostics-format=text`     | file         |
| `-fdiagnostics-format=sarif-file -fdiagnostics-add-output=text` | text + file  |
| `-fdiagnostics-format=sarif-file -fdiagnostics-set-output=text` | text + file  |
| `-fdiagnostics-add-output=text`                                 | text2        |
| `-fdiagnostics-add-output=text -fdiagnostics-format=sarif-file` | file         |
| `-fdiagnostics-add-output=text -fdiagnostics-add-output=sarif`  | text2 + file |
| `-fdiagnostics-add-output=text -fdiagnostics-set-output=sarif`  | file         |
| `-fdiagnostics-add-output=sarif`                                | text + file  |
| `-fdiagnostics-add-output=sarif -fdiagnostics-format=text`      | text + file  |
| `-fdiagnostics-add-output=sarif -fdiagnostics-add-output=text`  | text2 + file |
| `-fdiagnostics-add-output=sarif -fdiagnostics-set-output=text`  | text + file  |
| `-fdiagnostics-set-output=text`                                 | text         |
| `-fdiagnostics-set-output=text -fdiagnostics-format=sarif-file` | file         |
| `-fdiagnostics-set-output=text -fdiagnostics-add-output=sarif`  | text + file  |
| `-fdiagnostics-set-output=text -fdiagnostics-set-output=sarif`  | file         |
| `-fdiagnostics-set-output=sarif`                                | file         |
| `-fdiagnostics-set-output=sarif -fdiagnostics-format=text`      | file         |
| `-fdiagnostics-set-output=sarif -fdiagnostics-add-output=text`  | text + file  |
| `-fdiagnostics-set-output=sarif -fdiagnostics-set-output=text`  | text + file  |
- `text`: diagnostics to stderr
- `text2`: diagnostics to stderr but duplicated
- `file`: diagnostics to sarif file

So the interaction between `-fdiagnostics-format` and `-fdiagnostics-{add,set}-output` is not at all obvious to me. It would be nice not to have more or less complex logic for this in ccache, especially since it contradicts GCC documentation.

How about this, at least for a start: support (at most) _one_ of `-fdiagnostics-format`, `-fdiagnostics-add-output` and `-fdiagnostics-set-output`. Do you think this would be enough for real usage, or are multiple options actually used in practice?

kfessel:
https://github.com/ccache/ccache/actions/runs/30084293456/job/89474416397?pr=1706

I know what went wrong here 

the default sarif file place is moved from `cwd` to ... `drumroll`  -o directory  < its now version dependent :.(

https://gcc.gnu.org/gcc-16/changes.html#diagnostics

```
GCC's SARIF output now respects the [dump directory](https://gcc.gnu.org/onlinedocs/gcc-16.1.0/gcc/Overall-Options.html#index-dumpdir). For example, given


      gcc \
        -o build-dir/foo.o \
        -fdiagnostics-add-output=sarif
        foo.c
    

GCC 15 would write the SARIF to foo.c.sarif, whereas GCC 16 now writes it to build-dir/foo.c.sarif.
```

src/ccache/argprocessing.cpp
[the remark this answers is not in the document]
kfessel: https://gcc.gnu.org/gcc-16/changes.html

> GCC's SARIF output now respects the [dump directory](https://gcc.gnu.org/onlinedocs/gcc-16.1.0/gcc/Overall-Options.html#index-dumpdir). For example, given


      gcc \
        -o build-dir/foo.o \
        -fdiagnostics-add-output=sarif
        foo.c
    

> GCC 15 would write the SARIF to foo.c.sarif, whereas GCC 16 now writes it to build-dir/foo.c.sarif.

src/ccache/argprocessing.cpp
kfessel: > Do you think this would be enough for real usage, or are multiple options actually used in practice?

for me/us one is enough right now 

-fdiagnostics-add-output=sarif:file=...
to have sarif and text 
-fdiagnostics-set-output=sarif:file=...
for just sarif 
and nothing or 
-fdiagnostics-set-output=text to just have text 

it also seems like sarif will become a catch all and one would append a converter afterwards-- but i am not so sure about this (there is more and more diagnostics and representation) 

eg gcc 16 experimentaly adds html  -> one might like to have all of them text, sarif and html at least with add there would be a documented approach to do that 

I think
multiple add (one per type) 
xor 
one set
with no use of default location ( canged from gcc15 to gcc16)
would be safely supportable 

and good enough for every real use case 
( if one need copys of the same file they can call cp instead of having the compiler write that content twice)
( i would not add output html since i am not interested and it would extend this PR to much )

src/ccache/argprocessing.cpp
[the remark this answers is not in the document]
kfessel: i think that default location support has to go (i could not identify version dependent code paths)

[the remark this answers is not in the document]
jrosdahl: Sounds OK to me to only support explicit locations.

src/ccache/argprocessing.cpp
jrosdahl: OK.

test/suites/sarif.bash
```diff
@@ -11,19 +11,78 @@ SUITE_sarif_SETUP() {
...
+    expect_stat preprocessed_cache_hit 0
+    expect_stat cache_miss 0
+    expect_stat files_in_cache 0
+    # check the two default locations
+    if [ -e output.x.b.sarif ]
+        expect_exist output.x.b.sarif
+    else
+        expect_exist obj/output.x.b.sarif
+    fi
+    SUITE_sarif_cleanup
```
jrosdahl: I suggest starting a new `TEST` instead; each test runs in fresh directory.

src/ccache/argprocessing.cpp
```diff
@@ -1181,6 +1181,154 @@ process_option_arg(const Context& ctx,
...
+        args_info.diagnostics_output += "set:";
+        args_info.diagnostics_output += param;
+        args_info.diagnostics_output += ",";
+      } else {
+        LOG(
+          "no support for combination of -diagnostics-format and another "
+          "-diagnostics-add/set-output or -diagnostics-format");
+        LOG("diagnostics output: {}", args_info.diagnostics_output);
+        return Statistic::unsupported_compiler_option;
+      }
```
jrosdahl: To reduce complexity, I think it would be helpful to do it like this:

- In `process_option_arg`, collect knowledge about which options are used. That is, you could add state variables to `ArgumentProcessingState` like `bool found_fdiagnostics_format` and `bool found_fdiagnostics_set`. If `state.found_fdiagnostics_format || state.found_diagnostics_set || !args_info.output_sarif.empty()` and seeing another `-fdiagnostics-format` or `-fdiagnostics-{add,set}`, then we can fall back to the compiler immediately.
- In `process_args`, act on collected state: fall back to the compiler if `state.found_fdiagnostics_format && !args_info.output_sarif.empty()`, etc.

test/suites/sarif.bash
```diff
@@ -0,0 +1,90 @@
...
+    TEST "Sarif diagnostics"
+
+    # -fdiagnostics-format=sarif-file is not supported due to version dependent default file location
+    $CCACHE_COMPILE -x c -fdiagnostics-format=sarif-file -c src/input.a.b -o obj/output.x.y
+    expect_stat unsupported_compiler_option 1
+    expect_stat preprocessed_cache_hit 0
+    expect_stat cache_miss 0
+    expect_stat files_in_cache 0
+    # check the two default locations
+    if [ -e output.x.b.sarif ]
```
jrosdahl:
```
$ ../test/run sarif   
/path/to/ccache/test/suites/sarif.bash: line 37: syntax error near unexpected token `else'
/path/to/ccache/test/suites/sarif.bash: line 37: `    else'
```

src/ccache/compopt.cpp
```diff
@@ -128,6 +128,8 @@ const CompOpt compopts[] = {
   {"-external:I",
    AFFECTS_CPP | TAKES_ARG | TAKES_CONCAT_ARG | TAKES_PATH                           }, // msvc
   {"-fbuild-session-file=",   TAKES_CONCAT_ARG | TAKES_PATH                          },
+  {"-fdiagnostics-add-output=", TAKES_CONCAT_ARG | TAKES_PATH                        },
+  {"-fdiagnostics-set-output=", TAKES_CONCAT_ARG | TAKES_PATH                        },
```
jrosdahl: These are unnecessary since you handle the options explicitly in argprocessing.cpp. This table is for options that are not handled explicitly.

test/suites/sarif.bash
kfessel:
```suggestion
    if [ -e output.x.b.sarif ] then
```

then is missing 

strange that it succeeds even though it has "syntax error"

test/suites/sarif.bash
kfessel: this would add 5 extra tests (seems much to me for just one feature add)

jrosdahl: I'd say that custom cleanup logic costs more than splitting a large test into several. The test framework is intended to be used that way.

test/suites/sarif.bash
jrosdahl: Yes, that's unfortunate. I've improved it in 9b74826eef4dd6ec8622efa6d1fc6aac479e67b3 now.

kfessel: I ll rebase to not fall for this trap

kfessel: i tested https://github.com/ccache/ccache/commit/9b74826eef4dd6ec8622efa6d1fc6aac479e67b3

```
    Start 47: test.sarif
1/6 Test #47: test.sarif .......................***Failed    0.08 sec
/home/karl.fessel@ml-pa.loc/ws_ccache/ccache-pr/test/suites/sarif1.bash: line 27: syntax error near unexpected token `else'
/home/karl.fessel@ml-pa.loc/ws_ccache/ccache-pr/test/suites/sarif1.bash: line 27: `    else'
Error: Failed to load /home/karl.fessel@ml-pa.loc/ws_ccache/ccache-pr/test/suites/sarif1.bash

    Start 48: test.sarif1
2/6 Test #48: test.sarif1 ......................***Failed    0.08 sec
/home/karl.fessel@ml-pa.loc/ws_ccache/ccache-pr/test/suites/sarif1.bash: line 27: syntax error near unexpected token `else'
/home/karl.fessel@ml-pa.loc/ws_ccache/ccache-pr/test/suites/sarif1.bash: line 27: `    else'
Error: Failed to load /home/karl.fessel@ml-pa.loc/ws_ccache/ccache-pr/test/suites/sarif1.bash

    Start 49: test.sarif2
3/6 Test #49: test.sarif2 ......................***Failed    0.08 sec
/home/karl.fessel@ml-pa.loc/ws_ccache/ccache-pr/test/suites/sarif1.bash: line 27: syntax error near unexpected token `else'
/home/karl.fessel@ml-pa.loc/ws_ccache/ccache-pr/test/suites/sarif1.bash: line 27: `    else'
Error: Failed to load /home/karl.fessel@ml-pa.loc/ws_ccache/ccache-pr/test/suites/sarif1.bash
...

```

first syntax error now fails all tests

if there are two only the first gets a message

kfessel:
```suggestion
    if [ -e output.x.b.sarif ]; then
```

test/suites/sarif1.bash
```diff
@@ -0,0 +1,30 @@
+SUITE_sarif1_PROBE() {
```
jrosdahl: Looks like I failed to explain what I meant clearly enough: I wasn't suggesting to split the sarif suite into many sarif suites, I suggested splitting the long sarif test into several tests in the (single) sarif test suite.

That is:

```
SUITE_sarif() {
    TEST "Sarif diagnostics 1"
    # ...

    TEST "Sarif diagnostics 2"
    # ...
}
```

etc. Each `TEST` starts with a new clean environment, so no need to do custom cleanup between tests.

kfessel: did not see that, thought Test is just a fancy printf, ll rejoin them

kfessel: gcc < 15 do not support some of the options  -> gcc fails the test failed :disappointed:

kfessel:
# gcc16
```
47/53 Testing: test.sarif
47/53 Test: test.sarif
Command: "/nix/store/l6l2q7c6xvlb44rxbkpajssz6cma2hwd-bash-interactive-5.2-p21/bin/bash" "<HOME>/ws_ccache/ccache-pr/test/run" "sarif"
Directory: /mnt/scratch/ccache_build
"test.sarif" start time: Jul 30 12:05 CEST
Output:
----------------------------------------------------------
Compiler:         gcc (/nix/store/1p1wjrxmzj51gn2cy8kycspmqqmp6dkw-gcc-wrapper-16.1.0/bin/gcc)
Compiler version: gcc (GCC) 16.1.0

Running test suite sarif.......
PASSED
<end of output>
Test time =   0.69 sec
----------------------------------------------------------
Test Passed.
"test.sarif" end time: Jul 30 12:05 CEST
"test.sarif" time elapsed: 00:00:00
----------------------------------------------------------
```

# gcc14
```
47/53 Testing: test.sarif
47/53 Test: test.sarif
Command: "/nix/store/l6l2q7c6xvlb44rxbkpajssz6cma2hwd-bash-interactive-5.2-p21/bin/bash" "<HOME>/ws_ccache/ccache-pr/test/run" "sarif"
Directory: /mnt/scratch/ccache_build
"test.sarif" start time: Jul 30 12:12 CEST
Output:
----------------------------------------------------------
Compiler:         gcc (/nix/store/5wxmkw7hp35x8k2vlcs5hdqzz17610fy-gcc-wrapper-14.4.0/bin/gcc)
Compiler version: gcc (GCC) 14.4.0

Running test suite sarif..
INFO: compiler does not support -fdiagnostics-set/add-output; skipping tests

PASSED
<end of output>
Test time =   0.23 sec
----------------------------------------------------------
Test Passed.
"test.sarif" end time: Jul 30 12:12 CEST
"test.sarif" time elapsed: 00:00:00
----------------------------------------------------------
```

test/suites/sarif1.bash
kfessel: workaround:
 run just the first test for compiler that do not support `-fdiagnostics-set/add-output`

src/ccache/argprocessing.cpp
```diff
@@ -1191,6 +1193,143 @@ process_option_arg(const Context& ctx,
...
+  if (arg.starts_with("-fdiagnostics-format=")) {
+    LOG("-fdiagnostics-format=");
+    state.add_compiler_only_arg(args[i]);
+    if (state.found_fdiagnostics_set_output
+        || state.found_fdiagnostics_add_output) {
+      LOG(
+        "no support for combination of -diagnostics-format and another "
+        "-diagnostics-add/set-output or -diagnostics-format");
+      // set this  even in case of early return
+      state.found_fdiagnostics_set_output = true;
```
jrosdahl, line 1264:
```suggestion
```

src/ccache/argprocessing.cpp
```diff
@@ -1191,6 +1193,143 @@ process_option_arg(const Context& ctx,
...
   }
 
+  if (arg.starts_with("-fdiagnostics-format=")) {
+    LOG("-fdiagnostics-format=");
+    state.add_compiler_only_arg(args[i]);
+    if (state.found_fdiagnostics_set_output
+        || state.found_fdiagnostics_add_output) {
+      LOG(
+        "no support for combination of -diagnostics-format and another "
+        "-diagnostics-add/set-output or -diagnostics-format");
```
jrosdahl:
```suggestion
        "No support for combination of -fdiagnostics-format and another"
        " -fdiagnostics-add/set-output or -fdiagnostics-format");
```

src/ccache/argprocessing.cpp
```diff
@@ -90,6 +90,8 @@ class ArgumentProcessingState
   bool found_md_or_mmd_opt = false;
   bool found_Wa_a_opt = false;
   bool rewrite_FI_args = false;
+  bool found_fdiagnostics_set_output = false;
+  bool found_fdiagnostics_add_output = false;
```
jrosdahl, line 94: Since these are only used inside `process_option_arg` they don't belong to `ArgumentProcessingState` since they can be local variables in that function instead. [My earlier suggestion to put them in `state`](https://github.com/ccache/ccache/pull/1706#discussion_r3665399780) was under the assumption that you'd like to read them later in `process_args` for finding unsupported combinations after all options have been parsed.

Also, perhaps it would make sense to combine them into one state variable, perhaps like this?

```c++
enum class DiagnosticOutputMode {
  none,
  add_only, // -fdiagnostics-add-output
  exclusive, // -fdiagnostics-set-output or -fdiagnostics-format
};
```

src/ccache/argprocessing.cpp
```diff
@@ -1191,6 +1193,143 @@ process_option_arg(const Context& ctx,
...
+    arg_is_diagnostics_set_add_output = true;
+  }
+
+  if (arg_is_diagnostics_set_add_output) {
+    auto arg_sv = std::string_view(arg);
+    // "set" and "add" an the same length
+    auto param =
+      arg_sv.substr(std::string_view("-fdiagnostics-set-output=").size());
+    // we care for sarif only
+    auto sarif_pos = param.find("sarif");
```
jrosdahl: This looks fragile. For instance, it will find sarif here: `-fdiagnostics-set-output=somefuturescheme:file=not-sarif.html`. I suggest introducing a simple helper function, e.g. `parse_diagnostics_output_spec`, that parses the spec more strictly.

src/ccache/argprocessing.cpp
```diff
@@ -1191,6 +1193,143 @@ process_option_arg(const Context& ctx,
...
+    if (sarif_pos == std::string_view::npos) {
+      return Statistic::none;
+    }
+    // we need a sarif file parameter
+    auto file_pos = param.find("file=");
```
jrosdahl: Similarly, I think this should be parsed more strictly.

src/ccache/argsinfo.hpp
```diff
@@ -65,6 +65,13 @@ struct ArgsInfo
   // Source dependencies output file (MSVC). Contains pathname if not empty.
   std::filesystem::path output_sd;
 
+  // Sarif diagnostic generation. Contains path if not empty.
+  // after argprocessing this has filename or it is ignored.
+  std::filesystem::path output_sarif;
+
+  // diagnostic output command line option if set else empty.
+  std::string diagnostics_output;
```
jrosdahl: Since `diagnostics_output` is only used to detect whether the option was already seen, it can be replaced by `bool found_msvc_experimental_log` locally in `process_option_arg`.

src/ccache/argprocessing.cpp
```diff
@@ -1191,6 +1193,143 @@ process_option_arg(const Context& ctx,
...
+    }
+    // "-fdiagnostics-format=" replaces the default -> behaves like set
+    state.found_fdiagnostics_set_output = true;
+    // since handling of multiple option of this kind is unclear
+    // fall back to compiler
+    auto arg_sv = std::string_view(arg);
+    auto param =
+      arg_sv.substr(std::string_view("-fdiagnostics-format=").size());
+    if (param == "sarif-file") {
+      args_info.output_sarif = std::filesystem::path(".") / "";
```
jrosdahl: Since the function returns immediately for unsupported sarif file location there should be no need to store a sentinel like this.

```suggestion
```

src/ccache/argprocessing.cpp
```diff
@@ -1191,6 +1193,143 @@ process_option_arg(const Context& ctx,
...
+      if (file_end != std::string_view::npos) {
+        args_info.output_sarif =
+          core::make_relative_path(ctx, file_param.substr(0, file_end));
+      } else {
+        args_info.output_sarif = core::make_relative_path(ctx, file_param);
+      }
+    } else {
+      // this append the '.' directory so output sarif is not empty but also has
+      // no filename
+      args_info.output_sarif = std::filesystem::path(".") / "";
```
jrosdahl: Likewise, I don't see a need to store a sentinel that will never be read anyway.

```suggestion
```

src/ccache/core/resultretriever.cpp
```diff
@@ -211,6 +211,12 @@ ResultRetriever::get_dest_path(FileType file_type) const
 
   case FileType::source_dependencies:
     return m_ctx.args_info.output_sd;
+
+  case FileType::sarif:
+    if (m_ctx.args_info.output_sarif.has_filename()) {
+      return m_ctx.args_info.output_sarif;
+    }
+    break;
```
jrosdahl: I think this could be

```suggestion
    return m_ctx.args_info.output_sarif;
```

instead. The caller will handle an unexpectedly empty path gracefully.

src/ccache/ccache.cpp
```diff
@@ -1070,6 +1070,12 @@ write_result(Context& ctx,
     LOG("Source dependencies file {} missing", ctx.args_info.output_sd);
     return false;
   }
+  if (ctx.args_info.output_sarif.has_filename()
```
jrosdahl: Could this be

```suggestion
  if (!ctx.args_info.output_sarif.empty())
```

instead?

src/ccache/ccache.cpp
```diff
@@ -3024,6 +3030,9 @@ do_cache_compilation(Context& ctx)
   if (!ctx.args_info.output_sd.empty()) {
     LOG("Source dependencies file: {}", ctx.args_info.output_sd);
   }
+  if (ctx.args_info.output_sarif.has_filename()) {
```
jrosdahl:
```suggestion
  if (!ctx.args_info.output_sarif.empty()) {
```

?

.gitignore
```diff
@@ -1,6 +1,9 @@
 # Typical build directories
 /build*/
 
+# Typical cache directory
+/.cache/
```
jrosdahl: Who/what creates files here?

kfessel: i thought i was the ccache build chain ill remove an see if it is recreated

src/ccache/argprocessing.cpp
kfessel: but this information must be available across multiple options beeing parsed -> multiple calls to the 
    process_arg() -> process_option_arg();

i can introduce a local static variable but that seems to not fit with ccaches style of keeping track of state

src/ccache/core/resultretriever.cpp
kfessel: but not a directory name which in gcc case is used to mark `some sarif with default name`

src/ccache/argsinfo.hpp
kfessel: ok lets make it less state  
i ll remove it 
since its already indicated by the non emptyness of output_sarif

.gitignore
kfessel: clangd

kate + clang for lsp

src/ccache/ccache.cpp
kfessel: No, as only a filename is a valid response at this point 
if its a directory (like './' it is not )  valid 

i moved that logic to argparsing

src/ccache/argprocessing.cpp
kfessel: it will be if there are multiple sarif args

src/ccache/argprocessing.cpp
kfessel: but this information must be available across multiple arguments beeing parsed 
-- multiple calls to process_arg() -> process_option_arg();

I can introduce a local static variable but that seems to not fit with ccaches style of keeping track of state

I think declaring an enum for something like this is a bit overengineered

src/ccache/argprocessing.cpp
kfessel: went with a starts_with  to match the sarif  after = 

and with tonkenization for the keys

src/ccache/core/resultretriever.cpp
kfessel: i changed the parse_args to emptythe path  if directory

src/ccache/argprocessing.cpp
kfessel: that should be strict enough

src/ccache/core/resultretriever.cpp
kfessel: moved all the got no filename logic to parse args  and  empty it if there is no filename

src/ccache/argprocessing.cpp
jrosdahl: > but this information must be available across multiple arguments beeing parsed

Oops, you're right that the state still needs to survive between calls to `process_option_arg` even if you don't use it in `process_args` later, my bad.

src/ccache/argprocessing.cpp
jrosdahl: Since `Statistic::unsupported_compiler_option` is returned from `process_option_arg` on the next line, `process_arg` will return the error to `process_args` which sets `argument_error`, leading to error exit later. At least it's the intention that returning an error from `process_option_arg` should be enough.

The general idea is this:

- `process_option_arg`: collect information/state (but an error can be returned immediately for stateless things like missing argument, failure to read a file, etc.)
- `process_args`: act on state collected by `process_option_arg`/`process_arg`

.gitignore
jrosdahl: OK. Perhaps mention clangd in the comment?

src/ccache/argprocessing.cpp
jrosdahl: Even if multiple sarif args are given, ccache will fall back to running the compiler since you return `Statistic::unsupported_compiler_option` immediately after this, so `args_info.output_sarif` won't be read.

src/ccache/argprocessing.cpp
```diff
@@ -1191,6 +1193,145 @@ process_option_arg(const Context& ctx,
...
+      LOG("No support for multiple sarif files");
+      return Statistic::unsupported_compiler_option;
+    }
+
+    auto par_keys = param.substr(std::string_view("sarif:").size());
+
+    for (auto key : util::split_into_views(par_keys, ",")) {
+      if (key.starts_with("file=")) {
+        auto file = key.substr(std::string_view("file=").size());
+        args_info.output_sarif = core::make_relative_path(ctx, file);
```
jrosdahl: To have any effect, `core::make_relative_path` should be applied to the argument sent to the compiler (`state.add_compiler_only_arg(args[i])`). `args_info.output_sarif` is just ccache's internal note about where to find the file and whether that's absolute or relative doesn't matter.

src/ccache/argprocessing.cpp
kfessel: for some reason i think that i like to have that state complete even if already in an error return 

-> i d like to see there if there was a set as well as an add even if the add triggers an error

src/ccache/argprocessing.cpp
kfessel: that' d mean i need to take the argument apart and reassemble 

and if not the desired sarif with file i need to put the original argument

kfessel: went with default just add 

when desired form try to reassemble 
in case of success drop and add the reassembled

src/ccache/argprocessing.cpp
kfessel: like with the found_fdiagnostics_set/add .. i like to get a complete picture

but i added a if that empties the info if it is a directory like this `./` -> other function only need to care for empty or they can be sure its a filename.

src/ccache/argprocessing.cpp
kfessel: i went tonkenization for the keys

src/ccache/argprocessing.cpp
```diff
@@ -1191,6 +1200,163 @@ process_option_arg(const Context& ctx,
...
+    arg_is_diagnostics_set_add_output = true;
+  }
+
+  if (arg_is_diagnostics_set_add_output) {
+    auto reassembled_arg = std::string("");
+    if (arg.starts_with("-fdiagnostics-set-output=")) {
+      reassembled_arg += "-fdiagnostics-set-output=";
+    } else {
+      reassembled_arg += "-fdiagnostics-add-output=";
+    }
```
jrosdahl: Thanks for continuing to work on this!

I think this is becoming unacceptably complex. My suggestions are still:

1. `process_option_arg`:
    - parse and add arguments (using `core::make_relative_path` for paths) - this can be done immediately and should not require `state.drop_compiler_only_arg()` or similar
    - add state to act on later when we have the full picture
        - I think the only state that might be needed is `args_info.output_sarif`, but for better error reporting in logs it might be acceptable with more granular state
    - may bail early but does not need to maintain accurate state in error conditions
    - extract parsing code to helper functions to avoid making `process_option_arg` much longer than it already is
2. `process_args`:
    - act on state collected by `process_option_arg`
    - e.g.: for MSVC, add filename to `output_sarif` if it's a directory
    - could detect unsupported combinations of options here, e.g. both `-fdiagnostics-format` and `-fdiagnostics-*-output`

I realize that it might be difficult to understand exactly what I mean. I don't quite have bandwidth to write very detailed guidance, but I want to fight complexity since ccache already has more than enough of accidental complexity that has accumulated for many years. I would much rather start by only accepting a more easy and narrow use case, so please hear me out and help me to not add unnecessary complexity.

kfessel:
```
D:/a/ccache/ccache/src/ccache/argprocessing.cpp:1331:25: error: no match for 'operator+=' (operand types are 'std::__cxx11::basic_string<char>' and 'std::filesystem::__cxx11::path')
 1331 |         reassembled_arg += args_info.output_sarif;
      |         ~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~
  • there are 5 candidates
```
the c++ used by mingw seems to miss  a solution to covert from path to something  that adds to string -> FMT to mitigate

src/ccache/argprocessing.cpp
kfessel: Actually i also do not like complexity either but i try to have a complete picture at the places where the picture  is drawn (setup  as much of arginfo as possible) even in case of bailing (without introducing new structures, types, enums).
I see you like to fight complexity by stopping to draw the picture as soon as it cannot be finished (return error). 

**If you think you got a better idea how to implement this: Please fell free to take from this PR** what you think is usable (as I did with the original)  (none is of cause also OK), I don't want this PR to block that and I wont feel hurt. When you got a PR  and like an outside review you can ping me, I would like to have this feature upstream (`-fdiagnostics-*-output=sarif:file=FILENAME`), so we can use upstream ccache in our CI and not a patched version.

> * `process_option_arg`:
>   
>   * parse and add arguments (using `core::make_relative_path` for paths) - this can be done immediately and should not require `state.drop_compiler_only_arg()` or similar

there was an immediate addition ( without even looking at the arg (the if is one of the ones i care for aside)), but thats exactly the thing not working they need to be parsed first but i only want to deal with a subset (sarif).

>   * add state to act on later when we have the full picture
this drives the complexity up ( my POV) adding new structure / enums / types
     
>    * I think the only state that might be needed is `args_info.output_sarif`, but for better error reporting in logs it might be acceptable with more granular state

All state that would kind of show what happen during parsing was redacted by your request to delete them as they are not need - you are right but we also lost that information 

>   * may bail early but does not need to maintain accurate state in error conditions

This makes a later view on what happend more difficult  see last point (but also lowers the local complexity)

I give that a try and put the second part of fdiagnostic-add/set-output  handling into a separate funtion.

>   * extract parsing code to helper functions to avoid making `process_option_arg` much longer than it already is

Ahh i see you think that function is too long and like to avoid growing it even more (I kindof treated it as a given that this function is just long i did not treat long as a measure of complexity for this function anymore and just avoided reading most of it) 

> 
> * `process_args`:
>   ...
>   * could detect unsupported combinations of options here, e.g. both `-fdiagnostics-format` and `-fdiagnostics-*-output` 

i think this is easier while processing the options (needs less state recoded)

( the state is the number of set, add and diagnostics-format option and one migth already have gotten something else wrong if not checcking in process_option_arg)

jrosdahl: > **If you think you got a better idea how to implement this: Please fell free to take from this PR** what you think is usable (as I did with the original) (none is of cause also OK), I don't want this PR to block that and I wont feel hurt. When you got a PR and like an outside review you can ping me, I would like to have this feature upstream (`-fdiagnostics-*-output=sarif:file=FILENAME`), so we can use upstream ccache in our CI and not a patched version.

Thank you for being patient and constructive about this. I've now had a look and incorporated your and 
@autoantwort's work into a new PR, marking you as co-authors: #1788. A review and real-life testing would be appreciated!

PR #1718, opened 2026-04-03
ci: Add builds and release on riscv64

luhenry:
This adds building, testing, and releasing of ccache on Linux/riscv64. The runners are made available through the [RISE RISC-V Runners GitHub app](https://github.com/apps/rise-risc-v-runners), and with using `runs-on: ubuntu-24.04-riscv`.

This will unblock projects using ccache on GitHub Actions to target riscv64, since the actions/setup-ccache only advertises architectures  that ccache releases for. For example, I ran into that issue with Llama.cpp when trying to use the same mechanism across all platforms.

Please find more information about RISE at https://riseproject.dev/ and check out the [RISE RISC-V Runners documentation](https://riseproject-dev.github.io/riscv-runner/).


luhenry:
To run the riscv64 jobs running on `ubuntu-24.04-riscv`, you would need to install https://github.com/apps/rise-risc-v-runners on the https://github.com/ccache organization. The list of permissions is documented at https://riseproject-dev.github.io/riscv-runner/docs/getting-started/install.html#for-organizations (it’s basically only “self-hosted runners read+write” as it needs to register runners when a workflow needs to run).

luhenry:
And the workflow that passed on my fork where the app is installed: https://github.com/luhenry/ccache/actions/runs/23955765959

luhenry:
@jrosdahl :wave: thanks for all the amazing work on ccache!

For me to make your life easier, would it help if I broke out this PR into 2 pieces:
- the `build_linux_riscv64_binary_glibc` piece to enable cross-compilation to `linux-riscv64`, and
- the `build_and_test` and `build_linux_musl_static_binary` which do requires native runners?

`ggml-org/llama.cpp` are currently hurting from the lack of readily-available releases on `riscv64`, so trying to figure out how to unblock them.

Happy to submit different PR or rework the change however to make your life easier. Thank you 🙏

jrosdahl:
Thanks for the PR!

I would prefer not having to add a CI dependency to external runners. The cross-compilation for glibc part is uncontroversial, so yes, please split out the native build parts for now.

luhenry:
> I would prefer not having to add a CI dependency to external runners. The cross-compilation for glibc part is uncontroversial, so yes, please split out the native build parts for now.

I fully understand, and will split it. Would you be open to have a musl-static build on QEMU? It wouldn’t require external runners, but it would be quite slow.

luhenry:
Using QEMU to build musl-statis would look like this: https://github.com/luhenry/ccache/compare/master...luhenry:ccache:musl-static. And a CI run: https://github.com/luhenry/ccache/actions/runs/24285089891/job/70913019233#logs

jrosdahl:
> Would you be open to have a musl-static build on QEMU? It wouldn’t require external runners, but it would be quite slow.

Such a slow job would be unfortunate. I guess it _might_ be acceptable to not have it enabled by default but just run it on tags and release branches if the risk that it doesn't work is low enough.

Just a sanity check: so you do need a riscv64 static binary (the glibc "mostly static" isn't enough)?

jrosdahl:
> musl-static

Would it be possible to create and use a custom container image with a musl cross toolchain (hosting it on GHCR)? Or is there already such a container that could be used?

luhenry:
> > Would you be open to have a musl-static build on QEMU? It wouldn’t require external runners, but it would be quite slow.
> 
> Such a slow job would be unfortunate. I guess it _might_ be acceptable to not have it enabled by default but just run it on tags and release branches if the risk that it doesn't work is low enough.

It takes 20min on QEMU on GitHub-hosted runners. I’ll see to enable it on tags and releaS branches only.

> Just a sanity check: so you do need a riscv64 static binary (the glibc "mostly static" isn't enough)?

It’s all about convergence and “working the same”. For example, `ccache-action` which checks if ccache is installed, download if not install locally, and download/upload the cache for a build on GitHub Actions works with the musl-static build. It’s clearly not impossible to adapt it to use the glibc builds on riscv64. But it’s one place where it’s not a hard requirement, and where it would be amazing if ccache was available OOTB for the same platforms as others (amd64 and arm64 here).

jrosdahl:
OK, let's go with the QEMU job for now.

Issue #1727, opened 2026-04-13
Implemented Redis storage helpers and added Redis servers


afbjorklund:
Adding as an issue, so that it is not "lost" in the discussion only...

* https://github.com/ccache/ccache/issues/1214#issuecomment-3765408794

https://ccache.dev/storage-helpers.html

https://ccache.dev/howto/redis-storage.html -->

https://github.com/ccache/ccache/wiki/Redis-storage

1)

I started on "porting" the new storage helpers, from http to redis:

* https://github.com/ccache/ccache/discussions/1716#discussioncomment-16535681

https://github.com/ccache/ccache-storage-http-cpp -->
https://github.com/ccache/ccache-storage-http-go -->

https://github.com/ccache/ccache-storage-redis-cpp
https://github.com/ccache/ccache-storage-redis-go

This is so that the builtin redis storage driver can be deprecated.

The Redis-TLS is not fully implemented yet, only hiredis_ssl linked:

* https://github.com/ccache/ccache/issues/1554

And Redis Cluster is NOT implemented at all in hiredis, only proxy:

* https://github.com/RedisLabs/redis-cluster-proxy

2)

I updated the list of Redis-compatible servers, demoted aerospike:

* https://github.com/ccache/ccache/discussions/1716#discussioncomment-16480207

Besides Redis/Valkey, the recommended server is now Apache Kvrocks:

* https://kvrocks.apache.org/

It has a simple setup (unlike some others), and is supported by Apache:

https://kvrocks.apache.org/blog/kvrocks-graduated-as-tlp (in 2023)

This is when you don't have enough RAM to run Redis (in-memory).

There is also an interesting alternative "Redka" server that uses SQL.

PR #1728, opened 2026-04-13
fix: allow tests to run remote helpers

afbjorklund:
Changed the tests, so that I can run without the builtin drivers.

```
HTTP_STORAGE_BACKEND:BOOL=OFF
REDIS_STORAGE_BACKEND:BOOL=OFF
```

This is for testing the external storage helpers, using "CRSH".

* https://github.com/ccache/ccache/issues/1214#issuecomment-3765408794

* #1727

### DIFF
--- a/test/suites/remote_http.bash
+++ b/test/suites/remote_http.bash
@@ -35,6 +35,8 @@ SUITE_remote_http_PROBE() {
 SUITE_remote_http_SETUP() {
     unset CCACHE_NODIRECT
 
+    export CRSH_LOGFILE="ccache-storage-http.log"
+
     generate_code 1 test.c
 }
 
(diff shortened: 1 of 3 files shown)


test/suites/remote_http.bash
```diff
@@ -43,7 +45,7 @@ SUITE_remote_http() {
     TEST "Subdirs layout"
 
     start_http_server 12780 remote
-    export CCACHE_REMOTE_STORAGE="http://localhost:12780 helper=_builtin_"
```
jrosdahl, line 46: I think that the test suite should continue to test the builtin client code. By removing `helper=_builtin_` the tests now might or might not use the builtin client code, which I don't think is good – `helper=_builtin_`  was added for a reason.

I would be OK with having a way to explicitly opt in to use storage helpers instead, though.

afbjorklund: It should probably be two different tests, I was just sharing how I was testing at the moment

jrosdahl:
Time to close or update this one?

PR #1729, opened 2026-04-14
ci: Enable linux-riscv musl-static builds using QEMU

luhenry:
This is a follow-up to https://github.com/ccache/ccache/pull/1718 where the musl-static binaries are built in QEMU on riscv64. cc @jrosdahl

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -355,7 +355,7 @@ jobs:
           retention-days: 3
 
   build_linux_musl_static_binary:
-    timeout-minutes: 30
+    timeout-minutes: ${{ matrix.qemu && 60 || 30 }}
     name: Linux ${{ matrix.arch }} binary (musl static)
     runs-on: ${{ matrix.runner }}
     strategy:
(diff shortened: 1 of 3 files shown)


.github/workflows/build.yaml
```diff
@@ -366,12 +366,19 @@ jobs:
...
+          - arch: riscv64
+            runner: ubuntu-22.04
+            qemu: true
     steps:
       - name: Get source
         uses: actions/checkout@v4
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v4
       - name: Build in Alpine container
+        if: ${{ !matrix.qemu || github.ref_type == 'tag' || github.event.inputs.test_tag != '' || (github.ref_type == 'branch' && github.ref_name == github.event.repository.default_branch) }}
```
jrosdahl: I'd like it enabled on branches named `*-maint` as well.

luhenry:
And the ci runs for different branches and tags:
* master: https://github.com/luhenry/ccache/actions/runs/24492656962
* not-default-branch: https://github.com/luhenry/ccache/actions/runs/24505468185
* 9.9-maint: https://github.com/luhenry/ccache/actions/runs/24505488715
* [tag] v9.9: https://github.com/luhenry/ccache/actions/runs/24505558267

luhenry:
It triggered on “not-default-branch”, the issue is `[[ false ]]` is interpreted as `-n false` which is true (false here is a non-empty string..)

I’m AFK but will fix as soon as I’m back to my computer.

jrosdahl:
Thanks.

PR #1733, opened 2026-04-19
feat: introduce ECS cache to optimize kernel compilation

cctv18:
Introduce Effective Configuration State (ECS) caching to resolve the long-standing issue in kernel compilation where altering configuration options modifies `autoconf.h`, forcing ccache to fall back to preprocessing mode and significantly reducing compilation speed.

Mechanism:
- When the environment variable `CCACHE_IS_KERNEL_COMPILING="true"` is detected, ccache automatically generates ECS caches for kernel sources.
- During subsequent calls, it directly verifies the hash and the ECS cache validation value of the source code and its referenced headers.
- If these values remain unchanged, ccache achieves a direct hit, bypassing the costly preprocessing step.

### DIFF
--- a/src/ccache/ccache.cpp
+++ b/src/ccache/ccache.cpp
@@ -206,6 +206,29 @@ path_key(const fs::path& path)
   return util::format_base16(hash.digest());
 }
 
+std::string
+portable_path_key(const fs::path& path)
+{
+  fs::path normalized_path = util::lexically_normal(path);
+  std::string normalized = util::pstr(normalized_path).str();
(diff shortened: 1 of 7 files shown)


Numbersf (approved):

jrosdahl:
Sorry, but this feature is not something that will be accepted in ccache.

Among other things, the feature (at least as implemented):

- is very project-specific and intimately tied to behavior of that project
- has several layering violations 
- adds project-specific knowledge to generic code paths
- invents a totally separate storage model with no data eviction
- has no documentation
- has no tests

PR #1735, opened 2026-04-19
Expose base16 source digest in debug input text

chenrui333:
This adds the raw base16 source digest to `ccache-input-text` without changing the legacy digest bytes used for the direct-mode hash.

That makes the debug artifact directly comparable with `ccache --hash-file`.

- https://github.com/Homebrew/homebrew-core/pull/278330
- https://github.com/Homebrew/homebrew-core/pull/271486


jrosdahl:
Thanks for the PR!

However, ccache-input-text now (since fbfa39e322a6) contain base16 source digests, so I'll close this.

PR #1736, opened 2026-04-20
feat: Enable reflinking .dwo files

pzfl:
.dwo files should be reflinked like regular .o files. E.g. on my RelWithDebInfo LLVM build with split dwarf, object files take 2.2GB, whereas .dwo files take up 2.6GB. These files should be trivially shareable across build directories.


jrosdahl:
Thanks!

PR #1739, opened 2026-04-24
Support Ascend C language

wk0911:
This PR adds support for compilation caching of Ascend C. Ascend C is a dedicated programming language tailor-made by Huawei for the Ascend computing platform.


jrosdahl:
Neither GCC nor Clang supports an `asc` language. After a small investigation, this seems to be about a compiler called ascendc (or is it bisheng)? Is that correct? So ascendc accepts `-x asc`, and `-E` emits preprocessed source code similar to that of GCC?

PR #1741, opened 2026-05-04
feat: Rewrite -fprofile-use= path family relative to base_dir

markyang92:
Without this, the absolute path of the profile data file leaks into the
hash, so two workspaces sharing a cache (CI, parallel checkouts) miss
each other on otherwise identical compilations. Content changes still
produce a miss because hash_profile_data_file() hashes the file content.

this change rewrites only the args path while the per-file content hash 
continues to disambiguate, so workspaces with different profile content 
remain distinct.

Tested in profiling_clang.bash: a base_dir cross-workspace case
exercising -fprofile-instr-use=, -fprofile-use=, -fprofile-sample-use=
in both directions (path-only difference hits, content difference misses).

### DIFF
--- a/src/ccache/argprocessing.cpp
+++ b/src/ccache/argprocessing.cpp
@@ -1010,7 +1010,17 @@ process_option_arg(const Context& ctx,
       // The failure is logged by process_profiling_option.
       return Statistic::unsupported_compiler_option;
     }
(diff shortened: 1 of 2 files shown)


jrosdahl:
Thanks!

PR #1744, opened 2026-06-06
chore: add ubuntu-26.04 dockerfile

sobomax:
This adds ubuntu-26.04 dockerfile. Tested with my container multiarch build - works just fine.


jrosdahl:
Thanks.

PR #1745, opened 2026-06-06
ci: update "uses:" to latest versions

sobomax:
This change updated all `uses:` entries in the GH pipeline to the latest versions. This is to avoid warnings about node 20.x deprecated by the gha.


jrosdahl:
Thanks.

PR #1747, opened 2026-06-06
fix: Logic error with read_only storage gating

### LINKED ISSUE #1746: read_only storage gating misbehaving when read_only==false
### Problem description

Problematic Behavior: Using redis remote storage, ccache 4.13.6 was behaving as read-only==true when read-only was explicitly set to false. This behavior is different from the version I was upgrading from (4.10).

https://github.com/tenstorrent/tt-metal/commit/e7e901c802f787cf2c49c341040524d4033e4948

Include read-only=false in the CCACHE_REMOTE_STORAGE config

### Actual behavior

CCache was not writing to redis when read-only=false

### Expected behavior

CCache should write to redis when read-only=false

### Ccache version

4.13.6

### Ccache configuration

_No response_

### Operating system

Linux

### Environment

Clang 20, Ubuntu 22/24

### Logs

```shell

```

ThisIsFineTM:
RemoteStorageEntry::read_only is an std::optional<bool>. Storage::get_backend was treating read_only as a regular bool (likely a residue left over from a refactor?). Operator bool on std::optional is just a has_value check, so the backend storage gating was behaving as read_only even if read_only was set to false.

Closes #1746

### DIFF
--- a/src/ccache/storage/storage.cpp
+++ b/src/ccache/storage/storage.cpp
@@ -582,7 +582,7 @@ Storage::get_backend(RemoteStorageEntry& entry,
                      const std::string_view operation_description,
                      const bool for_writing)
 {
-  if (for_writing && entry.config.read_only) {
+  if (for_writing && entry.config.read_only.value_or(false)) {
     LOG("Not {} {} storage since it is read-only",
         operation_description,
         entry.config.shards.front().url.scheme());


jrosdahl:
Yes, that looks like a regression in #1672.

jrosdahl:
Thanks.

blozano-tt:
Thanks for the quick turnaround on review!

PR #1750, opened 2026-06-21
feat: Add auto depend mode for NVCC on Windows

risingwen:
## Summary

Add NVCC auto depend mode on Windows by having ccache add `-Xcompiler /showIncludes` when `depend_mode=true` and the command line does not already request dependency/include output.

The MSVC-style include lines are then used to compute the depend-mode result key. Since those include lines were injected by ccache, they are stripped from user-visible stdout on both cache miss and cache hit, matching the existing MSVC auto-depend behavior.

If the user explicitly passes `/showIncludes` or `-Xcompiler /showIncludes`, ccache preserves the include output for transparency.

## Tests

- Added argument-processing coverage for explicit `nvcc -Xcompiler /showIncludes`.
- Added argument-processing coverage for NVCC auto depend mode adding `-Xcompiler /showIncludes`.
- Added stdout-stripping coverage for NVCC auto depend mode and explicit showIncludes transparency.
- Verified locally:
  - `cmake --build build-debug --target check_format`
  - `./unittest/unittest --test-case="nvcc_show_includes_via_host_compiler_option"`
  - `./unittest/unittest --test-case="nvcc_auto_depend_mode"`
  - `./unittest/unittest --test-case="strip_includes_from_msvc_show_includes"`
- Verified on Windows with a minimal `.cu` file:
  - auto depend mode strips ccache-injected include output and gets a direct cache hit on the second run
  - explicit `-Xcompiler /showIncludes` preserves include output on both miss and hit and gets a direct cache hit on the second run

### DIFF
--- a/src/ccache/argprocessing.cpp
+++ b/src/ccache/argprocessing.cpp
@@ -376,6 +376,15 @@ is_msvc_z_debug_option(std::string_view arg)
          != std::end(debug_options);
 }
 
+bool
+is_msvc_show_includes_option(std::string_view arg)
+{
+  return arg == "-showIncludes"
+         || arg == "/showIncludes"
+         // clang-cl:
+         || arg == "-showIncludes:user" || arg == "/showIncludes:user";
+}
+
 // Returns std::nullopt if the option wasn't recognized, otherwise the error
 // code (with Statistic::none for "no error").
 std::optional<Statistic>
@@ -931,11 +940,19 @@ process_option_arg(const Context& ctx,
     return Statistic::none;
   }
 
-  if (arg == "-showIncludes"
-      // clang-cl:
-      || arg == "-showIncludes:user") {
+  if (is_msvc_show_includes_option(arg)) {
     args_info.generating_includes = true;
     state.add_compiler_only_arg(args[i]);
+    return Statistic::none;
+  }
+
+  if (ctx.config.compiler_type() == CompilerType::nvcc
+      && (arg == "-Xcompiler" || arg == "--compiler-options")
+      && i < args.size() - 1 && is_msvc_show_includes_option(args[i + 1])) {
+    args_info.generating_includes = true;
+    state.add_compiler_only_arg(args[i]);
+    state.add_compiler_only_arg(args[i + 1]);
+    ++i;
     return Statistic::none;
   }
 
@@ -1912,11 +1929,20 @@ process_args(Context& ctx)
     state.add_compiler_only_arg_no_hash(*diagnostics_color_arg);
   }
 
-  if (ctx.config.depend_mode() && !args_info.generating_includes
-      && ctx.config.compiler_type() == CompilerType::msvc) {
-    ctx.auto_depend_mode = true;
-    args_info.generating_includes = true;
-    state.add_compiler_only_arg_no_hash("/showIncludes");
+  if (ctx.config.depend_mode() && !args_info.generating_includes) {
+    if (ctx.config.compiler_type() == CompilerType::msvc) {
+      ctx.auto_depend_mode = true;
+      args_info.generating_includes = true;
+      state.add_compiler_only_arg_no_hash("/showIncludes");
+    }
+#ifdef _WIN32
+    else if (ctx.config.compiler_type() == CompilerType::nvcc) {
+      ctx.auto_depend_mode = true;
+      args_info.generating_includes = true;
+      state.add_compiler_only_arg_no_hash("-Xcompiler");
+      state.add_compiler_only_arg_no_hash("/showIncludes");
+    }
+#endif
   }
 
   if (state.found_c_opt) {
--- a/src/ccache/compiler/msvc.cpp
+++ b/src/ccache/compiler/msvc.cpp
@@ -57,8 +57,12 @@ strip_includes_from_msvc_show_includes(const Context& ctx,
   using Mode = Tokenizer::Mode;
   using IncludeDelimiter = Tokenizer::IncludeDelimiter;
 
-  if (stdout_data.empty() || !ctx.auto_depend_mode
-      || ctx.config.compiler_type() != CompilerType::msvc) {
+  const bool strip_auto_includes =
+    ctx.auto_depend_mode
+    && (ctx.config.compiler_type() == CompilerType::msvc
+        || ctx.config.compiler_type() == CompilerType::nvcc);
+
+  if (stdout_data.empty() || !strip_auto_includes) {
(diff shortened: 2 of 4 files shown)


risingwen:
Could a maintainer approve the CI run for this fork PR when convenient?

This is a small fix for stripping MSVC-style showIncludes output in nvcc depend mode. I added unit coverage for the updated condition and verified it locally with:

```text
./unittest/unittest --test-case="strip_includes_from_msvc_show_includes"
```

risingwen:
Just to add more context: this PR is intended for nvcc-compatible compilers on Windows that emit MSVC-style showIncludes output while using depend mode.

The existing code already strips showIncludes output for MSVC auto depend mode. This change keeps that behavior unchanged and only extends the same stdout stripping to this narrower condition:

- `compiler_type=nvcc`
- `depend_mode=true`
- `generating_includes=true`

The motivation is that ccache already consumes those include lines for dependency/result-key calculation, so replaying them from cached stdout on a cache hit is not useful and can confuse build systems that parse dependency output.

If preferred, I can add an integration-style test with a fake compiler that emits `Note: including file:` lines to demonstrate the cache-hit stdout behavior.

jrosdahl:
Thanks for the PR.

> Could a maintainer approve the CI run for this fork PR when convenient?

Unfortunately not since no such button appears. I don't know why and I can't find a way to trigger a (re)run. I suggest rebasing to latest master.

> this PR is intended for nvcc-compatible compilers on Windows that emit MSVC-style showIncludes output while using depend mode

Could you supply an example of an actual NVCC command line that triggers includes in stdout where we want to filter it out?

src/ccache/compiler/msvc.cpp
```diff
@@ -57,8 +57,14 @@ strip_includes_from_msvc_show_includes(const Context& ctx,
   using Mode = Tokenizer::Mode;
   using IncludeDelimiter = Tokenizer::IncludeDelimiter;
 
-  if (stdout_data.empty() || !ctx.auto_depend_mode
-      || ctx.config.compiler_type() != CompilerType::msvc) {
+  const bool strip_msvc_auto_includes =
+    ctx.auto_depend_mode && ctx.config.compiler_type() == CompilerType::msvc;
+  const bool strip_nvcc_depend_includes =
+    ctx.config.compiler_type() == CompilerType::nvcc && ctx.config.depend_mode()
+    && ctx.args_info.generating_includes;
```
jrosdahl: I don't understand this part. If `ctx.args_info.generating_includes` is true then `/showIncludes[:user]` has been passed to the compiler, so the compiler should emit includes to stdout. But if ccache strips that output then ccache will not behave transparently (not return what the compiler would return if run without ccache). What am I missing?

risingwen:
Good point, thanks. You are right that using `generating_includes` as the stripping condition was too broad.

What I wanted to handle is the case where ccache itself enables MSVC-style include output for depend mode. In that case the include lines are an implementation detail used to compute the result key, so stripping them from the user-visible stdout is consistent with the existing MSVC auto-depend behavior.

However, if the user explicitly passes `/showIncludes` or `-Xcompiler /showIncludes`, ccache should preserve that stdout for transparency. I have updated the patch to only strip include lines when `ctx.auto_depend_mode` is true, and to make nvcc auto-depend mode add `-Xcompiler /showIncludes` itself.

I verified the adjusted behavior with these cases:

- auto depend mode: `ccache compiler_type=nvcc depend_mode=true nvcc -c test.cu -o test.obj`
  - ccache invokes `nvcc -Xcompiler /showIncludes ...`
  - stdout include lines are stripped
  - the second run is a direct cache hit
- explicit showIncludes: `ccache compiler_type=nvcc depend_mode=true nvcc -c test.cu -o test.obj -Xcompiler /showIncludes`
  - stdout include lines are preserved on both miss and hit
  - the second run is a direct cache hit

jrosdahl:
OK, good. If I understand correctly, this changes the PR to something like "feat: Add auto depend mode for NVCC on Windows", right?

src/ccache/argprocessing.cpp
```diff
@@ -1902,11 +1919,17 @@ process_args(Context& ctx)
...
+  if (ctx.config.depend_mode() && !args_info.generating_includes) {
+    if (ctx.config.compiler_type() == CompilerType::msvc) {
+      ctx.auto_depend_mode = true;
+      args_info.generating_includes = true;
+      state.add_compiler_only_arg_no_hash("/showIncludes");
+    } else if (ctx.config.compiler_type() == CompilerType::nvcc) {
+      ctx.auto_depend_mode = true;
+      args_info.generating_includes = true;
+      state.add_compiler_only_arg_no_hash("-Xcompiler");
+      state.add_compiler_only_arg_no_hash("/showIncludes");
```
jrosdahl: Does NVCC support `/showIncludes` on all platforms?

risingwen:
Yes, that is a more accurate description now. I updated the PR title, description and commit title to `feat: Add auto depend mode for NVCC on Windows`.

src/ccache/argprocessing.cpp
jrosdahl: Tested on my Linux box now:

```console
$ nvcc -Xcompiler /showIncludes -c test.c
gcc-12: fatal error: cannot specify ‘-o’ with ‘-c’, ‘-S’ or ‘-E’ with multiple files
compilation terminated.
$ nvcc -Xcompiler -showIncludes -c test.c
gcc-12: error: unrecognized command-line option ‘-showIncludes’
```

risingwen:
You are right; this is Windows/MSVC-host specific. I should not have applied the NVCC auto-depend behavior on non-Windows platforms.

I updated the patch to guard the NVCC `-Xcompiler /showIncludes` auto-depend path with `_WIN32`. On non-Windows platforms, ccache will no longer add MSVC-style `/showIncludes` for NVCC.

I also updated the argument-processing test so that:

- on Windows, NVCC auto depend mode adds `-Xcompiler /showIncludes`
- on non-Windows, NVCC auto depend mode does not add it

Verified locally with:

```text
cmake --build build-debug --target check_format
./unittest/unittest --test-case="nvcc_show_includes_via_host_compiler_option"
./unittest/unittest --test-case="nvcc_auto_depend_mode"
./unittest/unittest --test-case="strip_includes_from_msvc_show_includes"
```

I also rebuilt and verified the Windows behavior again: ccache still adds `-Xcompiler /showIncludes` on Windows, strips the ccache-injected include output, and gets a direct cache hit on the second run.

PR #1751, opened 2026-06-23
fix: Suppress #line directives when preprocessing with MSVC

### LINKED ISSUE #1731: Suppress #line directives emitted by MSVC preprocessor
In preprocessor mode, ccache uses [`-P` flag](https://github.com/ccache/ccache/blob/5d0c538eed09a9d5831fac6e69e40909f6fe6581/src/ccache/ccache.cpp#L1440-L1445) for MSVC, this instructs the preprocessor to add `#line` directives to the output. As far as I can tell, these emitted paths aren't rewritten to honor `base_dir`.

It is possible to suppress these `#line` directives by specifying *both* `-P` and `-EP`: https://learn.microsoft.com/en-us/cpp/build/reference/p-preprocess-to-a-file

K-ballo:
When MSVC preprocesses with /P, it emits #line directives that embed absolute source paths into the preprocessed output. These directives are not needed for compilation and cause cache misses when the same source is built from different directories, since the preprocessed content varies by path. GCC does not have this issue.

Passing -EP together with -P suppresses the #line directives.

MSVC reference https://learn.microsoft.com/en-us/cpp/build/reference/p-preprocess-to-a-file
Fixes https://github.com/ccache/ccache/issues/1731

### DIFF
--- a/src/ccache/ccache.cpp
+++ b/src/ccache/ccache.cpp
@@ -1453,6 +1453,7 @@ get_result_key_from_cpp(Context& ctx, util::Args& args, Hash& hash)
         args.push_back("-utf-8"); // Avoid garbling filenames in output
       }
       args.push_back("-P");
+      args.push_back("-EP"); // Suppress #line directives in preprocessed output
       args.push_back(FMT("-Fi{}", preprocessed_path));
     } else {
       args.push_back("-E");


jrosdahl:
As mentioned in https://github.com/ccache/ccache/issues/1731#issuecomment-5153081787, the suggested approach doesn't work.

PR #1757, opened 2026-07-02
fix: Disable direct mode for C23 #embed directives

### LINKED ISSUE #1540: ccache doesn't take into account C23 #embed
### How to reproduce ###

```
docker run --rm -it fedora:rawhide
dnf install -y ccache clang
echo "first" > test.bin
printf 'const char embeded[] = {\n#embed "test.bin"\n};' > test.c
ccache clang -std=c23 -c test.c
strings test.o | grep first
echo "second" > test.bin
ccache clang -std=c23 -c test.c
strings test.o | grep second || (echo "did not get second" && /bin/false)
```

### Actual behavior ###

"did not get second" emitted

### Expected behavior ###

it should not been emitted

### Environment ###

ccache-4.10.2-1.fc41.x86_64

ajanicij:
## Summary

Fixes #1540: ccache could return stale cache hits when a source file uses C23 `#embed` and the embedded file changes, because direct mode hashes the `.c` file but does not know about separately embedded files.

This PR implements the **simpler approach** [jrosdahl suggested in review of #1664](https://github.com/ccache/ccache/pull/1664#issuecomment-3909123456): detect `#embed` while hashing source files in `hash_source_code_file` (reusing the buffer already read for temporal-macro checks) and **disable direct mode**, so ccache falls back to preprocessor mode where Clang expands `#embed` correctly. The same scan also detects `.incbin` (respecting existing `incbin` sloppiness), consistent with preprocessor-mode handling.

Thanks to [@haarisr](https://github.com/haarisr) for the earlier work and discussion in #1664 — this PR takes a narrower path focused on the safe fallback rather than tracking embedded files in direct mode.

**Changes:**
- Add `check_for_source_directives()` and `hash_source_code_disables_direct_mode()` in `hashutil`
- Disable direct mode when `#embed` or `.incbin` (without `incbin` sloppiness) is found
- Add integration test `test/suites/embed.bash`
- Add unit tests in `test_hashutil.cpp`
- New inode-cache content type for scanned source files

**Not in this PR (possible follow-ups):**
- Direct-mode support with explicit embedded-file tracking (step 2 from #1664 discussion)
- Combined single-pass scan with temporal-macro detection (performance)
- `embed` sloppiness flag (analogous to `incbin`)

## Test plan

- [x] `./unittest/unittest --test-suite=hashutil`
- [x] `CC=clang-19 CCACHE=$PWD/ccache ../test/run -v embed` (requires Clang with C23 `#embed` support)
- [ ] CI (once opened)


jrosdahl:
Thanks for the PR.

As discussed in #1664, I made a more refined alternative in #1765.

ajanicij:
Great! I am closing this PR, it is not needed any more.

PR #1762, opened 2026-07-11
fix: Hash relative paths in preprocessor output when `base_dir` is set

### LINKED ISSUE #1761: Documentation on `base_dir` inconsistent with actual behavior
### Problem description

The [`base_dir`](https://ccache.dev/manual/4.13.6.html#config_base_dir) documentation says:
> When set, ccache converts absolute paths to relative paths before hashing, ...

This seems to be untrue, according to this line:
https://github.com/ccache/ccache/blob/ddd9437c23337aba77216d9bafe06e495478ac6f/src/ccache/ccache.cpp#L667

`inc_fs_path` holds the absolute path, not the relative path returned from `core::make_path_relative`.

I have a [repository](https://github.com/trustytrojan/ccache-1761) containing everything needed to reproduce the problem on both Windows and Linux.

You may also read the steps below which describe the manual process in a Windows environment:

1. Construct a file tree like below:
    ```
    C:/project
    │   ccache.conf
    │   source.c
    │   
    ├───build1
    │       header1.h
    │       header2.h
    │       
    └───build2
            header1.h
            header2.h
    ```
2. Write the following to `ccache.conf`:
    ```
    hash_dir = false
    base_dir = C:/project
    debug = true
    ```
3. In both `build1` and `build2`, have `header1.h` include `header2.h` using an *absolute path*. For example, `build1/header1.h` would contain `#include "C:/project/build1/header2.h"`. Then have `source.c` contain:
    ```c
    #include <header1.h>
    int main() {}
    ```
4. Clear the cache with `ccache -C`.
5. Populate the cache by running `ccache gcc -I . -c ../source.c -o source.o` inside `build1` (`cd build1` before compiling). I used MinGW GCC on Windows; run something equivalent if using MSVC.
6. Zero out the statistics with `ccache -z`, then repeat step 5 but inside `build2` instead of `build1`.
7. Check the statistics with `ccache -s` and you will see a miss.

### Actual behavior

I expected ccache to only hash the *relative* version of that path. But since `build2` missed the binary from `build1`, and the preprocessor output in the `.ccache-input-text` files differed at the absolute include path we wrote in step 3, it did not hash the relative path.

### Expected behavior

When `base_dir` is set, ccache should only hash the relative version of absolute paths that it finds in preprocessor output. (As a bonus, I would like if this applied to *every* path that is possibly hashed. This puts less of a burden build systems.)

### Another note about the documentation

The documentation on `base_dir` only states the term "relative path", but we never know *what* path is being used as the base for relativization. The documentation makes it sound like the base path of relativization is the first matched directory in the `base_dir` list. However, I had to dig into the source to find out that the base path of relativization is actually ccache's CWD. This should be more clear in the documentation.

### Ccache version

4.13.6

### Operating system

Windows/Linux

trustytrojan:
- Closes #1761 
- When `base_dir` is set and a relative path is computed, hashes the relative path (`inc_path`) instead of the absolute path (`inc_fs_path`).
- In the scenario described in #1761, with this patch applied, the diff in `.ccache-text-input` is gone, and the ccache statistics after `build2` builds `source.c` has 1 (preprocessor) hit and 0 misses.

### DIFF
--- a/src/ccache/ccache.cpp
+++ b/src/ccache/ccache.cpp
@@ -664,7 +664,7 @@ process_preprocessed_file(Context& ctx, Hash& hash, const fs::path& path)
       }
 
       if (inc_fs_path != ctx.apparent_cwd || ctx.config.hash_dir()) {
-        hash.hash(inc_fs_path);
+        hash.hash(inc_path);
       }
 
       TRY(remember_include_file(ctx, inc_fs_path, hash, system, nullptr));


jrosdahl:
Thanks for the PR.

Superseded by 45dbe4d922e56e3e8a62ca1da577c75078d947a5.

PR #1764, opened 2026-07-12
fix: Handle case-insensitive filesystems when ignoring headers

### LINKED ISSUE #1763: Handle case-insensitive filesystems when ignoring headers
After inserting some log statements into `file_path_matches_dir_prefix_or_file()`, I see this when building the `Core` target of [qtbase](https://github.com/qt/qtbase) on Windows:

```
[2026-07-11T18:25:48.472549 10148] dir_prefix_or_file: include/QtCore/6.9.0/QtCore/private/qconfig_p.h
[2026-07-11T18:25:48.472566 10148] file_path:          include\\qtcore\\6.9.0\\QtCore/private/qconfig_p.h
[2026-07-11T18:25:48.472569 10148] result:             false
[2026-07-11T18:25:48.472573 10148] dir_prefix_or_file: include/qtcore/6.9.0/qtcore/private/qconfig_p.h
[2026-07-11T18:25:48.472575 10148] file_path:          include\\qtcore\\6.9.0\\QtCore/private/qconfig_p.h
[2026-07-11T18:25:48.472576 10148] result:             false
```

Unfortunately, it seems some preprocessors make parts of include paths lowercase on Windows (`file_path` in the above log). I cannot anticipate this.

To fix this problem we would need to make both `dir_prefix_or_file` and `file_path` lowercase on Windows before calling `std::mismatch` on them. The `ignore_headers_in_manifest` paths can be transformed to lowercase proactively in `Context::initialize`, and the input file paths can be transformed to lowercase immediately after they are canonicalized here:

https://github.com/ccache/ccache/blob/ddd9437c23337aba77216d9bafe06e495478ac6f/src/ccache/ccache.cpp#L428-L430

trustytrojan:
- Fixes #1763
- Converts both input include file paths and all paths in the `ignore_headers_in_manifest` configuration option to lowercase on Windows before `file_path_matches_dir_prefix_or_file` is called.

### DIFF
--- a/src/ccache/ccache.cpp
+++ b/src/ccache/ccache.cpp
@@ -426,8 +426,12 @@ remember_include_file(Context& ctx,
 
   if (!ctx.ignore_header_paths.empty()) {
     // Canonicalize path for comparison; Clang uses ./header.h.
-    const std::string& canonical_path_str =
+    auto canonical_path_str =
       path_str.str().starts_with("./") ? path_str.str().substr(2) : path_str;
+#ifdef _WIN32
+    // Handle case-insensitive paths by converting to lowercase.
+    canonical_path_str = util::to_lowercase(canonical_path_str);
+#endif
     for (const auto& ignore_header_path : ctx.ignore_header_paths) {
       if (file_path_matches_dir_prefix_or_file(ignore_header_path,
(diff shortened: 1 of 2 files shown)


jrosdahl:
Thanks!

PR #1769, opened 2026-07-27
feat: Implement path mapping

### LINKED ISSUE #1607: Support for stable cache hits despite varying include paths (Conan 2 integration)
### Description 

Greetings,
I’m part of the Conan team and we’re working to provide first-class support for using ccache effectively alongside Conan 2.

ccache already works well in many scenarios when building Conan packages, even when builds happen in isolated folders, as long as we configure:

```sh
hash_dir = false
base_dir = $HOME  # or CI-specific HOME
```

This ensures that the directory structure differences between builds do not prevent cache reuse.

However, there’s one remaining issue we cannot work around: varying include paths. Conan stores each compiled package (e.g. zlib) in a uniquely hashed folder like:
 
```sh
~/.conan2/p/b/zlib916d372395c74/p/include/zlib.h
~/.conan2/p/b/zlibe73b9c09c6910/p/include/zlib.h
```

The content of the headers can be exactly the same, but the path changes each time the package is built or re-exported. This could happen when building in a distributed environment (or in a CI), or locally when testing different configurations.

This leads to:
```
cache_miss
direct_cache_miss
local_storage_miss
local_storage_read_miss
local_storage_read_miss
local_storage_write
local_storage_write
preprocessed_cache_miss
```

I've tried enabling:
```
sloppiness = include_file_ctime,include_file_mtime,file_macro,time_macros
```
But the results are the same. The compilation units that effectively include a header from a conan cache, if include directory (absolute path) changes, despite the header content being identical, it always results in a cache miss. 

### What we would like to achieve

It would be immensely helpful if ccache could optionally treat included headers by content rather than by absolute path, even in direct mode. Perhaps this could be enabled with a new sloppiness option, or some opt-in mode to instruct ccache to:
- Only hash the contents (and maybe size/inode) of included headers.
- Ignore the full include path from the manifest if the content hasn’t changed.

I have read the docs and did not find such an option. I know some company had this issue in the past and could manage to solve it. 

If that is the case, from conan team, we will be pleased to share the exact configuration with our users and allow them to benefit from ccache toolset. 

### Why this matters

Conan uses isolated build directories, which means include paths are not stable across machines or builds, but the actual content of the files is (when the build options and profiles are the same, of course).

Without a way to tell ccache "this file has the same content, even if the path changed", we cannot benefit from cache reuse in a very common scenario: rebuilding a consumer after a dependency is rebuilt into a new Conan folder.

This is one of the last remaining blockers to seamless ccache use in Conan 2 pipelines.

### Environment

- ccache version: 4.11.3 (latest)
- Conan version: 2.x
- Compiler: any
- OS: any

We’d be happy to collaborate on testing, design discussions, or even contribute to a patch or experiment if this kind of behavior could be explored. Thanks for building such a powerful and fast caching tool, we’d love to make it even more compatible with modern dependency managers like conan!

Thank you in advance!

### LINKED ISSUE #1761: Documentation on `base_dir` inconsistent with actual behavior

### LINKED ISSUE #1772: Hashing `VCToolsInstallDir` does not account for different Visual Studio product names
When two developers need to share a remote cache of MSVC binaries, hashing the `VCToolsInstallDir` environment variables does not account for the fact that, even though both developers have the **same MSVC toolset version** (and therefore identical compilers, STL, etc.), one may have a Visual Studio "Build Tools" installation and the other a "Professional" installation.

This is a concern because VS Build Tools has an installation path of:
```
C:\Program Files (x86)\Microsoft Visual Studio\2022\BuildTools
```
While VS Professional has:
```
C:\Program Files\Microsoft Visual Studio\2022\Professional
```

For some reason, Build Tools is installed under `C:\Program Files (x86)`, while Professional (and the other IDE installations) are in `C:\Program Files`. This difference is the main concern. The VS release year in the paths above is important, but is superseded by the environment variable `VSCMD_VER`, which contains the full Visual Studio version (e.g. `17.14.37`).

This same concern exists for the hashing of the environment variables `INCLUDE` and `EXTERNAL_INCLUDE`, since they also contain absolute paths to the local VS installation. These will have to be handled separately.

One solution could be a feature like path mapping (#1769) that could map paths within these variables such that the VS installation path can be abstracted away.

### LINKED ISSUE #1806: `base_dir` does not account for path depth differences
`base_dir` forces your source tree, as well as any other directories required by the build (inside MSVC environment variables, for example), to be at the same depth level relative to the build area (where ccache typically runs from, at least for CMake projects). It may not be common, but a depth difference *can* occur. And if the build area is outside the source tree, the *name* of the source folder also must be the same across two machines. For example, a source area of `C:\src` and a build area of `C:\build` would cause ccache to relativize paths inside `C:\build` to something like `..\..\..\src`. This requires that the name of the source folder is always `src`.

trustytrojan:
### Overview
In GCC/Clang, you can pass `-ffile-prefix-map=<from>=<to>` to have both macros and debug symbols modified to replace absolute, machine-specific paths with abstract paths like `/src` and `/build` that can increase binary reproducibility. ccache can autodetect this compiler flag and use it for mapping certain paths like the CWD and profiler/coverage-related paths.

This PR brings the ability of path mapping to as many hashable paths as possible. Path mapping occurs immediately before hashing in several places:

- `inputfile`,
- any compiler option that takes a path,
- the `path exists` check,
- the path to a precompiled header (which I may consider undoing, since PCH and ccache don't seem to go well together),
- environment variables that can include paths,
- and every include path in preprocessor output.

These are all the places I have needed to transform paths (and where absolute paths can show up), but there may be many more. If any reviewers/onlookers have suggestions, feel free to drop a comment on this PR.

### Related issues
- Closes #1761
- Closes #1772
- Potentially closes #1607
- Closes #1806

### DIFF
--- a/src/ccache/.clang-tidy
+++ b/src/ccache/.clang-tidy
@@ -57,6 +57,7 @@ Checks: '
   -modernize-return-braced-init-list,
   -modernize-use-auto,
   -modernize-use-default-member-init,
+  -modernize-use-designated-initializers,
   -modernize-use-nodiscard,
   -modernize-use-trailing-return-type,
   performance-*,
--- a/src/ccache/ccache.cpp
+++ b/src/ccache/ccache.cpp
@@ -2160,9 +2160,9 @@ hash_argument(const Context& ctx,
   }
 
   // Optionally map and hash include paths.
-  // There may be more than these three include flags to look out for.
+  // There may be more than these include flags to look out for.
   // If needed, a separate PR can add more.
-  for (const auto prefix : {"-I", "-imsvc", "-isystem"}) {
+  for (const auto prefix : {"-I", "-imsvc", "-isystem", "-external:I"}) {
     if (!args[i].starts_with(prefix)) {
       continue;
     }
--- a/src/ccache/ccache.cpp
+++ b/src/ccache/ccache.cpp
@@ -499,7 +499,7 @@ check_included_pch_file(Context& ctx, Hash& hash)
       && !ctx.args_info.generating_pch) {
     fs::path pch_path =
       core::make_relative_path(ctx, ctx.args_info.included_pch_file);
-    hash.hash(pch_path);
+    hash.hash(util::perform_path_mapping(pch_path, ctx.config.path_mapping()));
     TRY(remember_include_file(ctx, pch_path, hash, false, nullptr));
   }
   return {};
@@ -2144,41 +2144,35 @@ hash_argument(const Context& ctx,
   // directories and similar input paths exist. Note that this is not 100%
   // waterproof since it only detects newly appearing directories and not newly
   // appearing header files.
-  if (direct_mode) {
-    std::optional<std::string_view> path;
-    if (compopt_takes_path(args[i]) && i + 1 < args.size()) {
-      path = args[i + 1];
-    } else {
-      path = compopt_prefix_takes_path(args[i]);
-    }
-    if (path) {
-      const auto fsp =
-        util::perform_path_mapping(*path, ctx.config.path_mapping());
-      hash.hash_delimiter("path exists");
-      hash.hash(FMT("{} {}", fsp, fs::exists(fsp) ? "1" : "0"));
-    }
-  }
-
-  // Optionally map and hash include paths.
-  // There may be more than these include flags to look out for.
-  // If needed, a separate PR can add more.
-  for (const auto prefix : {"-I", "-imsvc", "-isystem", "-external:I"}) {
-    if (!args[i].starts_with(prefix)) {
-      continue;
-    }
+  {
     std::optional<std::string_view> path;
     bool space_in_between = false;
+    std::string compopt;
     if (compopt_takes_path(args[i]) && i + 1 < args.size()) {
-      path = args[i + 1];
+      compopt = args[i];
+      path = args[++i]; // Consume both prefix & path
       space_in_between = true;
     } else {
       path = compopt_prefix_takes_path(args[i]);
+      if (path) {
+        compopt = args[i].substr(0, args[i].length() - path->length());
+      }
     }
     if (path) {
-      const auto fsp =
+      const auto mapped_path =
         util::perform_path_mapping(*path, ctx.config.path_mapping());
-      hash.hash_delimiter("arg");
-      hash.hash(FMT("{}{}{}", prefix, space_in_between ? " " : "", fsp));
+      hash.hash_delimiter("path exists");
+      hash.hash(FMT("{} {}", mapped_path, fs::exists(*path) ? "1" : "0"));
+      if (space_in_between) {
+        // Emulate the behavior at the end of this function
+        hash.hash_delimiter("arg");
+        hash.hash(compopt);
+        hash.hash_delimiter("arg");
+        hash.hash(mapped_path);
+      } else {
+        hash.hash_delimiter("arg");
+        hash.hash(FMT("{}{}", compopt, mapped_path));
+      }
       return {};
     }
   }
--- a/src/ccache/ccache.cpp
+++ b/src/ccache/ccache.cpp
@@ -658,7 +658,8 @@ do_process_preprocessed_data(Context& ctx, Hash& hash, util::Bytes&& data)
       inc_path_str.clear(); // inc_path is used from now on
 
       if (inc_path != ctx.apparent_cwd || ctx.config.hash_dir()) {
-        hash.hash(inc_path);
+        hash.hash(
+          util::perform_path_mapping(inc_path, ctx.config.path_mapping()));
       }
 
       TRY(remember_include_file(ctx, inc_path, hash, system, nullptr));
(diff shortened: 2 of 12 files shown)


maksim-petukhou:
This might also help with #1607 if I understand correctly

trustytrojan:
I've come to realize that ccache already has a function to perform path mapping:

https://github.com/ccache/ccache/blob/45dbe4d922e56e3e8a62ca1da577c75078d947a5/src/ccache/ccache.cpp#L1586-L1610

But these mappings are only populated when detecting `-fdebug-prefix-map`, `-ffile-prefix-map`, or `-fcoverage-prefix-map`:

https://github.com/ccache/ccache/blob/45dbe4d922e56e3e8a62ca1da577c75078d947a5/src/ccache/argprocessing.cpp#L744-L757

And they are only used to rewrite the CWD before hashing, or for paths in profiling information.

Now that I know this function exists, it might be worth discussing whether I should reuse it, or keep the new system I have now where `path_mapping` is parsed into a `std::vector<std::pair<std::string, std::string>>` before it is used in `util::perform_path_mapping`. We can instead have the CWD and profiling information hash sites use the new system, and stop processing compiler arguments to populate very specific mappings.

trustytrojan:
I encountered an interesting bug when building larger codebases with `path_mapping` enabled. I forgot to normalize the path before the loop begins in `util::perform_path_mapping`. Although not necessary anymore, below is an argument for why paths coming from preprocessor output should be normalized immediately in `do_process_preprocessed_data` of `ccache.cpp`.

When the preprocessor sees a relative include path inside quotes, it first checks to see if prepending the CWD results in a valid file path. If it does, that path is written to preprocessor output. 

Consider this scenario:

```
src = { header.h }, build = { dir1 = { src.c } }
```

If `src.c` has `#include "../../src/header.h"`, the preprocessor outputs `/path/to/build/dir1/../../src/header.h`. When path mapping occurs, it calls `util::path_starts_with` which normalizes paths before comparison. For `path_mapping = /path/to/src=/src1:/path/to/build=/build1`, and input path being `/path/to/build/dir1/../../src/header.h`, the input path gets normalized to `/path/to/src/header.h` inside `util::path_starts_with`, which returns true when comparing with the first key of `path_mapping`: `/path/to/src`. `util::perform_path_mapping` then decides that the prefix should be mapped, removes 3 leading components from the input path (which has NOT been normalized), and prepends the value of the mapping, returning `/src1/dir1/../../src/header.h`.

This is why (in my opinion) include paths from the preprocessor should be normalized immediately to prevent these kinds of problems in the future.

jrosdahl:
Thanks for the proposal.

I might not fully understand which classes of problems you're aiming to solve, but let me start by clarifying what `base_dir` is and isn't (in case that's unclear):

`base_dir` is only about rewriting paths in the compiler's input to make it appear _as if_ the user had used relative paths when executing the compiler. The idea is that by sending relative paths to the compiler, the result is valid when using the same relative paths in another location since the compiler has never seen any absolute paths. Any absolute paths that the compiler uses anyway (even if it gets relative paths as input) is by design not affected. That's why the manual hints about using `-fdebug-prefix-map` to normalize absolute paths in debug info, and for instance to set `sloppiness = gcno_cwd` to ignore absolute paths produced in code coverage data files. Since `base_dir` rewrites paths that are sent to the compiler, they simply _must_ be relative to the CWD, otherwise it won't work.

Do I understand correctly that `path_mapping` is primarily meant to rewrite paths before hashing them? That could be useful for some types of hashed information (like for #1772). However, you wrote _"The largest benefit of this feature is not having to use the base_dir option"_, which sounds like it would be a full replacement of `base_dir`, which (unless I'm missing something) it isn't.  One example: say that CWD is `/a/b` and we run `gcc -c /a/b/example.c` with `/a/b` rewritten to `/src` before hashing. Then we change directory to `/x/y` and run `gcc -c /x/y/example.c` with `/x/y` rewritten to `/src`. That will indeed be a cache hit, but a `__FILE__` expansion in `example.o` will still be `/a/b/example.c` even though we compiled in `/x/y`. Similarly, a warning message will contain incorrect paths. Of course, that could be an acceptable trade-off for some people.

The aspect of remapping paths to and from a normalized form in the manifest feels reasonable, however. I mentioned something similar in https://github.com/ccache/ccache/issues/1607#issuecomment-3083393079.

trustytrojan:
Hi again, thanks for the response.

In your example, yes, `__FILE__` will point to the original path, but the (quite simple) solution would be to pass to your compiler `-ffile-prefix-map` using the exact same mapping that you gave to ccache. This flag acts on both macros and debug symbols. For MSVC there is `/experimental:deterministic /pathmap:<from>=<to>` which is detailed [here](https://developercommunity.visualstudio.com/t/changing-pdb-source-file-paths-with-experimentalde/502528). I don't understand the warning message part; maybe I've never seen a compiler warning that has paths in it that affected my workflow.

As for the classes of problems this tries to solve, there's the issues I've already linked, but also path mapping provides a more stable way to work around build systems that force the usage of absolute paths. `base_dir` forces your source tree (as well as any potential external directories required by the build) to be at the same depth level relative to the build area (where ccache typically runs from, at least for CMake projects). It may not be common, but a depth difference *can* occur. And if the build area is outside the source tree, the *name* of the source folder also must be the same across two machines. For example, a source area of `C:\src` and a build area of `C:\build` would cause ccache to relativize paths inside `C:\build` to something like `..\..\..\src`. This requires that the name of the source folder is always `src`.

There's also the issue of toolchain-related environment variables containing absolute paths. #1772 and #1797 both go into detail there. Yes, `base_dir` can solve this problem too, but it comes with the same depth difference concern as before.

The "largest benefit" part I wrote is unnecessary; I only said that because I found it a lot easier to debug cache misses when I can see paths without an arbitrary amount of `..`s in them. I've removed that section from the PR description. I don't intend to replace `base_dir`, in fact I haven't touched a single piece of its code in this PR. The features can coexist, and people can choose one or the other.

I hope that clears things up for you.

P.S. I realize this conversation is better suited to be in an issue, so I've made one: #1806

PR #1773, opened 2026-07-31
Avoid hashing the Visual Studio installation path

### LINKED ISSUE #1772: Hashing `VCToolsInstallDir` does not account for different Visual Studio product names

trustytrojan:
- Closes #1772

### DIFF
--- a/src/ccache/ccache.cpp
+++ b/src/ccache/ccache.cpp
@@ -1754,7 +1754,6 @@ hash_common_info(const Context& ctx, const util::Args& args, Hash& hash)
   if (ctx.config.is_compiler_group_msvc()) {
     const char* msvc_env_vars[] = {
       "VCToolsVersion",
-      "VCToolsInstallDir",
     };
     for (const char* name : msvc_env_vars) {
       const char* value = getenv(name);


trustytrojan:
Note to self: this PR can be closed if #1769 also performs path mapping on `VCToolsInstallDir` before it is hashed.

trustytrojan:
[`7e9ef89`](https://github.com/ccache/ccache/commit/7e9ef895b4622f46d50ea16f5e953fcb6a99af69) fixes the problem more elegantly.

PR #1774, opened 2026-08-03
Hash contents of explicit module files (-fmodule-file=)

RedNicStone:
The pcm content referenced by -fmodule-file=<name>=<path> is not part of the preprocessed output of a consumer TU, so ccache's key (command line + preprocessed source) does not change when the module changes. A change to e.g. a template function body in the module therefore produces a false cache hit: the consumer keeps its old object file and silently runs stale code, even across full rebuilds.

Fix: collect the paths from -fmodule-file= arguments and hash the file contents into the common hash, mirroring the existing sanitize-ignorelist handling. Regenerating the pcm from unchanged source yields identical bytes, so unchanged modules still get cache hits.

Adds a test suite (fmodule_file) that verifies a consumer compile is cached while the module is unchanged and that a template body change invalidates the consumer's cache entry.

This fixes an issue that I discovered three years ago and accidentally reported to llvm. I encountered it again today in a new project so I decided to suggest a change.
Fixes https://github.com/llvm/llvm-project/issues/142298

### DIFF
--- a/src/ccache/argprocessing.cpp
+++ b/src/ccache/argprocessing.cpp
@@ -1051,6 +1051,19 @@ process_option_arg(const Context& ctx,
     return Statistic::none;
   }
 
+  if (arg.starts_with("-fmodule-file=")) {
+    // -fmodule-file=<name>=<path> or -fmodule-file=<path> for explicit C++
+    // module imports.
+    constexpr std::string_view module_file_flag = "-fmodule-file=";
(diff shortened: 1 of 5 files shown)


jrosdahl:
Thanks! And a gold star for the added test suite.

RedNicStone:
Thank you for the quick merge! This has been bothering me for a few years now, glad its finally resolved.

PR #1777, opened 2026-08-07
Search for longer compopt prefix matches after `bsearch()`'s returned pointer

### LINKED ISSUE #1776: Binary search in `find_prefix` misses longer compopts
CMake generates certain MSVC (or clang-cl) compilations with [`-FI`](https://learn.microsoft.com/en-us/cpp/build/reference/fi-name-forced-include-file?view=msvc-170) and its path argument concatenated (e.g. `-FIpath/to/file`). Every time [`find_prefix`](https://github.com/ccache/ccache/blob/9f09485511a5c6e273e81c3b3a2ab48adc392f72/src/ccache/compopt.cpp#L211-L221) is called with an option of this form, it always returns a `CompOpt*` to `-F` instead of `-FI`. This causes [`compopt_prefix_takes_path`](https://github.com/ccache/ccache/blob/9f09485511a5c6e273e81c3b3a2ab48adc392f72/src/ccache/compopt.cpp#L318-L328) to return `Ipath/to/file` instead of `path/to/file`, causing unnecessary cache misses since `base_dir`, unless you also insert `I` at the beginning of your `base_dir` paths, cannot match that path (especially if it is an absolute path on Windows).

trustytrojan:
- Fixes #1776
- The newly added assertion in the `prefix_takes_path` unit test verifies that the problem existed and was solved by this change.

### DIFF
--- a/src/ccache/compopt.cpp
+++ b/src/ccache/compopt.cpp
@@ -217,7 +217,18 @@ find_prefix(std::string_view option)
                          std::size(compopts),
                          sizeof(compopts[0]),
                          compare_prefix_compopts);
-  return static_cast<CompOpt*>(result);
+  if (!result) {
+    return {};
+  }
+  // It's possible that option starts with "-FI" and result is "-F".
+  // Search forward for a longer match until prefixes stop matching.
+  auto max_co = static_cast<const CompOpt*>(result);
+  for (auto co = max_co + 1; co < std::end(compopts) && option.starts_with(co->name); ++co) {
+    if (co->name.length() > max_co->name.length()) {
+      max_co = co;
+    }
+  }
+  return max_co;
 }
 
 // Used by unittest/test_compopt.cpp.
--- a/src/ccache/compopt.cpp
+++ b/src/ccache/compopt.cpp
@@ -223,7 +223,9 @@ find_prefix(std::string_view option)
   // It's possible that option starts with "-FI" and result is "-F".
   // Search forward for a longer match until prefixes stop matching.
   auto max_co = static_cast<const CompOpt*>(result);
-  for (auto co = max_co + 1; co < std::end(compopts) && option.starts_with(co->name); ++co) {
+  for (auto co = max_co + 1;
+       co < std::end(compopts) && option.starts_with(co->name);
+       ++co) {
     if (co->name.length() > max_co->name.length()) {
       max_co = co;
     }
--- a/unittest/test_compopt.cpp
+++ b/unittest/test_compopt.cpp
@@ -98,6 +98,7 @@ TEST_CASE("prefix_takes_path")
 {
   CHECK(compopt_prefix_takes_path("-Dfoo") == std::nullopt);
   CHECK(*compopt_prefix_takes_path("-Ifoo") == "foo");
+  CHECK(*compopt_prefix_takes_path("-FIfoo") == "foo");
 }
 
 TEST_SUITE_END();


jrosdahl:
Thanks for noticing the problem and for the PR.

> `bsearch()` did not offer any actual advantage here. In fact, it added a slight overhead by calling the comparator function.

That's odd. When I benchmark this (for a release build on Linux), the linear search version is ≈6 times slower when searching for a handful of keys in different table positions. Are you on Windows/MSVC? Have you benchmarked for different keys, not only `-FI` which only visits 1/5 of the table when searched linearly?

Incidentally, I'm actually working on replacing the option parsing to be (more or less) completely table-driven and already thought about the ambiguous prefixes issue, but I failed to realize that we have such a case even today in the current table/code.

The method I'm using in my still unfinished rewrite is to search with `std::lower_bound`, and (if it's not an exact match) then scan backwards to find the match (if any).

trustytrojan:
> the linear search version is ≈6 times slower 

Sorry, I meant to say that `bsearch` wasn't useful *only in theory* because we know the size of the table is constant. I know that in practice it does save time.

> The method I'm using in my still unfinished rewrite is to search with `std::lower_bound`, and (if it's not an exact match) then scan backwards to find the match (if any).

I did in fact try to solve the problem while still using `bsearch`, but even though [the cppreference page](https://en.cppreference.com/c/algorithm/bsearch) on it says that the `bsearch_s` variant takes the custom data pointer as the 3rd argument, the Windows SDK (10.0.26100.0) `corecrt_search.h` header which declares `bsearch_s` also declares the following as the comparator function type:

```c
typedef int (__cdecl* _CoreCrtSecureSearchSortCompareFunction)(void*, void const*, void const*);
```

Which seems to imply that the user data is passed as the first argument...? So I decided to give up on that.

### Another strategy

By nature of prefix matching, it is always the case that the result of a binary search for `<key>` in the form of `<common_prefix><other_chars>` will only find either:

- an exact match for `<key>`,
- or `<common_prefix>`.

It does not seem possible that the result could be `<common_prefix><some_other_chars>` where `<other_chars> != <some_other_chars>`, assuming the comparator is correct. Given this assumption, we can keep the binary search, but must linearly check elements after the returned `CompOpt*`, but until prefixes stop matching, for a longer match.

This keeps the performance of at least one binary search across all of `compopts` while ensuring correctness.

src/ccache/compopt.cpp
```diff
@@ -217,7 +217,20 @@ find_prefix(std::string_view option)
...
                          compare_prefix_compopts);
-  return static_cast<CompOpt*>(result);
+  if (!result) {
+    return {};
+  }
+  // It's possible that option starts with "-FI" and result is "-F".
+  // Search forward for a longer match until prefixes stop matching.
+  auto max_co = static_cast<const CompOpt*>(result);
+  for (auto co = max_co + 1;
+       co < std::end(compopts) && option.starts_with(co->name);
```
jrosdahl, line 227: Looking more into this, I realize that neither this nor the original algorithm is correct. For instance, both fail (at least on my platform) when adding

```c++
  CHECK(*compopt_prefix_takes_path("-Ffoo") == "foo");
```

to `TEST_CASE("prefix_takes_path")`. That's because `compare_prefix_compopts` doesn't behave like a comparison function should: it can return noncontiguous results.

For instance, searching for `-FUfoo`, it compares entries like this:

```
-F -> match
-FI -> no match
-FU -> match
```

So I don't think the `bsearch`-based method can be made to work reliably.

I've made an alternative fix in #1778 now. Would you like to review it?

jrosdahl: Thanks for reviewing it!

jrosdahl:
Superseded by #1778.

PR #1778, opened 2026-08-08
fix: Find the longest matching compiler option prefix

### LINKED ISSUE #1776: Binary search in `find_prefix` misses longer compopts

jrosdahl:
The `compare_prefix_compopts` function used with `bsearch` can produce multiple noncontiguous matches, so the lookup may miss a prefix or return a shorter one. Fix this by using `std::lower_bound` followed by searching backward to find the longest match.

Closes #1776.

### DIFF
--- a/benchmark/CMakeLists.txt
+++ b/benchmark/CMakeLists.txt
@@ -22,6 +22,7 @@ register_dependency(GoogleBenchmark DOWNLOADED "${_benchmark_version_string}")
 set(
   source_files
   main.cpp
+  benchmark_compopt.cpp
   benchmark_hashutil.cpp
 )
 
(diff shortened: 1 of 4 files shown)


trustytrojan (approved):
LGTM

PR #1779, opened 2026-08-10
feat: Support multiple -Wp macro options in direct mode

### LINKED ISSUE #1743: Support multi-arg -Wp,-U,-D in direct mode
### Problem description

Reproduced with ccache 4.10.2 (Fedora 42)

ccache with Collabora Online and LibreOffice which (optionally) uses "-Wp,-U_FORTIFY_SOURCE,-D_FORTIFY_SOURCE=2" compiled flags, which I lifted out of the default Fedora fortify flags from /usr/lib/rpm/redhat/macros which has:

```
# We use a single -Wp here to enforce order so that ccache does not ever
# reorder them.
%_fortify_level_flags   %[ 0%{?_fortify_level} > 0 ? "-Wp,-U_FORTIFY_SOURCE,-D_FORTIFY_SOURCE=%{_fortify_level}" : "" ]
```

Our configure.ac apparently defaults to using ccache's CCACHE_DEPEND=1

With that combination I found it odd that the ccache stats were very poor, some debugging with CCACHE_DEBUG=1 throws up: "Unsupported compiler option for direct mode: -Wp,-U_FORTIFY_SOURCE,-D_FORTIFY_SOURCE=2" 

I imagine this is a consequence of https://github.com/ccache/ccache/issues/1238

  ``` 
cat > hello.cpp <<'EOF'
  int main() { return 0; }
EOF

export CCACHE_DIR=/tmp/cc-repro
rm -rf $CCACHE_DIR && mkdir -p $CCACHE_DIR
ccache -z

FLAG='-Wp,-U_FORTIFY_SOURCE,-D_FORTIFY_SOURCE=2'

CCACHE_DEPEND=1 ccache g++ -c hello.cpp -o hello.o -MMD -MF hello.d -O2 "$FLAG"
CCACHE_DEPEND=1 ccache g++ -c hello.cpp -o hello.o -MMD -MF hello.d -O2 "$FLAG"

ccache -sv | grep -E 'Hits|Misses|Cacheable'
```

### Actual behavior

```
Cacheable calls:      2 /   2 (100.0%)
  Hits:               0 /   2 ( 0.00%)
  Misses:             2 /   2 (100.0%)
```

### Expected behavior

Hits instead of misses

### Ccache version

4.10.2

### Ccache configuration

CCACHE_DEPEND=1

### Operating system

Linux

### Environment

Fedora 42, but also seen on multiple ci machines of diverse Linuxes

### Logs

```shell
Unsupported compiler option for direct mode: -Wp,-U_FORTIFY_SOURCE,-D_FORTIFY_SOURCE=2
```

subotac:
Allow direct mode to handle a combined `-Wp` argument when every component is
  a macro definition or undefinition. Mixed or unknown preprocessor options still
  use the existing conservative fallback.

  The regression covers the Fedora fortify flag from #1743 with dependency mode
  enabled and verifies that the second identical compilation is a direct cache hit.

  Fixes #1743.

  ## Testing

  - `test/run direct`
  - `unittest/unittest --test-suite=argprocessing`
  - `ninja check_format`


test/suites/direct.bash
```diff
@@ -471,6 +471,21 @@ fi
...
+        -Wp,-U_FORTIFY_SOURCE,-D_FORTIFY_SOURCE=2 test.c
+    expect_stat direct_cache_hit 0
+    expect_stat preprocessed_cache_hit 0
+    expect_stat cache_miss 1
+
+    CCACHE_DEPEND=1 $CCACHE_COMPILE -c -MMD \
+        -Wp,-U_FORTIFY_SOURCE,-D_FORTIFY_SOURCE=2 test.c
+    expect_stat direct_cache_hit 1
+    expect_stat preprocessed_cache_hit 0
+    expect_stat cache_miss 1
```
jrosdahl: Since the improvement is related to direct mode, not depend mode specifically, I suggest not enabling depend mode here (there's a `depend.bash` suite if you want to test some aspect specific to depend mode):

```suggestion
    $CCACHE_COMPILE -c -Wp,-U_FORTIFY_SOURCE,-D_FORTIFY_SOURCE=2 test.c
    expect_stat direct_cache_hit 0
    expect_stat preprocessed_cache_hit 0
    expect_stat cache_miss 1

    $CCACHE_COMPILE -c -Wp,-U_FORTIFY_SOURCE,-D_FORTIFY_SOURCE=2 test.c
    expect_stat direct_cache_hit 1
    expect_stat preprocessed_cache_hit 0
    expect_stat cache_miss 1
```

jrosdahl (commented):
Nice, thanks for working on this!

test/suites/direct.bash
jrosdahl: I applied this.

jrosdahl:
Thanks, @subotac!

PR #1780, opened 2026-08-10
fix: Interpret clang-cl `-clang:` passthrough options

### LINKED ISSUE #1775: clang-cl `-clang:` passthrough option is not interpreted by ccache
The clang-cl compiler offers a way to pass options that only the Clang driver can interpret by prefixing them with [`-clang:` (or `/clang:`)](https://clang.llvm.org/docs/UsersManual.html#the-clang-option). Currently, ccache's `compopt.cpp` does not interpret any of these options, even though they may affect the compiler's behavior.

### Specific scenario where this is a problem

When CMake (4.4.2) configures a project using clang-cl and a non-MSBuild generator (either Ninja or Make), it needs to use the Clang-specific options [`-MT`](https://clang.llvm.org/docs/ClangCommandLineReference.html#cmdoption-clang-MT-arg) and [`-MF`](https://clang.llvm.org/docs/ClangCommandLineReference.html#cmdoption-clang-MT-arg) for depfile support. So it prefixes them with `-clang:`, but sometimes, for some reason, the path it passes to `-clang:-MT` contains forward-slashes, and at other times backslashes. This causes a cache miss in both direct and preprocessor mode, since `-clang:-MT` does not match any recognized option in the `compopts` table of `compopt.cpp`.

If it did, ccache could interpret it as Clang's `-MT` option and not hash the argument passed to it.

trustytrojan:
- Interpret `-clang:` options in `process_option_arg`
- Add similar interpretation to `get_option_and_value` in `ccache.cpp` for `hash_argument` (but not all of `hash_argument`)
- Add `clang_cl` subcase for `dependency_args_to_compiler` unit test
- Closes #1775

I am most likely going to need help to adjust all of `hash_argument` to handle the `-clang:` option prefix. I'll be keeping this as a draft PR until that happens. I'm also keeping this branch around to merge with others to achieve better cache hits with clang-cl.


jrosdahl:
Thanks for the PR. I'm currently (very slowly, unfortunately) working on revamping option parsing altogether. I would therefore prefer to solve the problem in a more general way.

However, I'm OK with supporting specific cases of `-clang:`, e.g. `-clang:-MF` and `-clang:-MT` to solve the specific scenario mentioned in #1775. That is, I suggest to check the `-clang:-MF` spelling (etc) next to `-MF` and not try to handle the general case yet. Does that sound reasonable to you?

Also, one complication is that _[the /clang: arguments are treated as if they were passed at the end of the clang-cl command line](https://clang.llvm.org/docs/UsersManual.html#the-clang-option)_. What's your view on that?

trustytrojan:
Now that I think about it, yes, `-clang:-MF` and `-clang:-MT` were the only options causing misses in that scenario. But in the same compile commands, `-clang:-MD` was present too, and ccache should probably be aware of the fact that the compiler is creating depfiles (so it can be stored in the manifest and whatnot). As a stopgap to fix #1775, I think it's fine to handle the depfile-related options first.

I would also like if your new compopt system's work could be in a PR I could take a look at.

As for the last sentence of documentation on `-clang:`, that's interesting, but I don't know if that really affects anything we do to track depfiles in the manifest. Do we currently have logic that ensures (GCC/Clang) options are in a certain order for caching to work properly? If so, then we might want to be concerned about this quirk of `-clang:`.

jrosdahl:
> Now that I think about it, yes, `-clang:-MF` and `-clang:-MT` were the only options causing misses in that scenario. But in the same compile commands, `-clang:-MD` was present too, and ccache should probably be aware of the fact that the compiler is creating depfiles (so it can be stored in the manifest and whatnot).

Yes.

> I would also like if your new compopt system's work could be in a PR I could take a look at.

Certainly!

> As for the last sentence of documentation on `-clang:`, that's interesting, but I don't know if that really affects anything we do to track depfiles in the manifest. Do we currently have logic that ensures (GCC/Clang) options are in a certain order for caching to work properly? If so, then we might want to be concerned about this quirk of `-clang:`.

I can't think of any example for `-clang:` specifically. It's just my gut feeling that things could be lurking here.

PR #1786, opened 2026-08-14
fix(`ignore_headers_in_manifest`): Handle directory-separators in path component comparison on Windows

### LINKED ISSUE #1785: Commit `cecb4fd` broke `ignore_headers_in_manifest` on Windows
Due to `std::filesystem::path`'s choice to sometimes represent [*directory-separators*](https://en.cppreference.com/cpp/filesystem/path#:~:text=directory%2Dseparators) on Windows as two backslashes when converting to a string, `path_components_equal_case_aware`, introduced in commit [`cecb4fd`](https://github.com/ccache/ccache/commit/cecb4fda4767614db446ca61b8fba3ce9a458550), does not account for this.

Adding the following checks to the `file_path_matches_dir_prefix_or_file` unit test in `test_ccache.cpp` fails the unit test:

```cpp
CHECK(file_path_matches_dir_prefix_or_file("\\aa", "/aa/bb"));
CHECK(file_path_matches_dir_prefix_or_file("/aa", "\\aa\\bb"));
```

This is because the [*root-directory*](https://en.cppreference.com/cpp/filesystem/path#:~:text=root%2Ddirectory) part of a `std::filesystem::path` is both a *directory-separator* **and** an iterable component when iterating the path object. This means `std::mismatch` compares the *root-directory* parts of two paths using the new `path_components_equal_case_aware` comparison function which does not handle *directory-separators*.

trustytrojan:
- Closes #1785


jrosdahl:
Thanks, @trustytrojan!

PR #1787, opened 2026-08-14
Handle `#line` directives in preprocessor output

### LINKED ISSUE #1784: ccache treats all `#line` directives as files that exist
When trying to share the cache of [qtbase v6.9.0](https://github.com/qt/qtbase) across two machines, [`hb-number-parser.hh`](https://github.com/qt/qtbase/blob/v6.9.0/src/3rdparty/harfbuzz-ng/src/hb-number-parser.hh) has `#line` directives that contain a filename, `hb-number-parser.rl`, that doesn't exist in the repository. (In [HarfBuzz v10.4.0](https://github.com/harfbuzz/harfbuzz/tree/10.4.0), which qtbase v6.9.0 uses, the file does exist. I'm guessing Qt deleted it in its copy because it isn't needed for their purposes.) The GCC and Clang preprocessors don't distinguish `#line` from `#include` directives in an obvious way. So ccache currently remembers `hb-number-parser.rl` in the manifest, and when the consuming machine reads the manifest, direct mode will miss because `hb-number-parser.rl` doesn't exist.

See #1783 for a possible ways to differentiate `#line` and `#include` expansions in preprocessor output. This is one way to solve the problem, as ccache can simply not call `remember_include_file` on filenames from `#line` directives that don't exist. Since preprocessors don't absolutize filenames in `#line` directives (they don't need to), ccache would first have to make it absolute by prepending `ctx.args_info.input_file.parent_path()`, and then check for existence.

But if #1783 is wrong, what should be recommended as a workaround? Create empty files of the same filename in `#line` directives where necessary? Since this might be a problem isolated to qtbase v6.9.0, that would be a good enough solution.

**UPDATE:** Qt 6.9.0 also bundles a library called emoji-segmenter, and it also has the same `#line N "nonexistent-file"` problem: https://github.com/qt/qtbase/blob/v6.9.0/src/3rdparty/emoji-segmenter/emoji_presentation_scanner.c

trustytrojan:
- Closes #1784
- I tested this with MSVC, and it too also does not write flags after the ending quote of a linemarker for `#line` directives. It also behaves just like HP/AIX in that instead of `# linenum ...` it writes `#line linenum ...`, so that is another avenue to check if it is a `#line` directive, but it would be specific to MSVC/HP/AIX.
- This solves all direct cache misses with qtbase v6.9.0's bundled HarfBuzz between two build areas (and two remote machines).


src/ccache/ccache.cpp
```diff
@@ -641,7 +641,12 @@ do_process_preprocessed_data(Context& ctx, Hash& hash, util::Bytes&& data)
       }
       // Look for preprocessor flags, after the "filename".
       bool system = false;
+      bool line_directive = false;
```
jrosdahl: Since the flag doesn't really determine that it's a line marker, I suggest naming it `has_linemarker_flags` and inverting the condition.

src/ccache/ccache.cpp
```diff
@@ -678,7 +683,10 @@ do_process_preprocessed_data(Context& ctx, Hash& hash, util::Bytes&& data)
         hash.hash(inc_path);
       }
 
-      TRY(remember_include_file(ctx, inc_path, hash, system, nullptr));
+      std::error_code ec;
+      if (!line_directive || std::filesystem::exists(inc_path, ec)) {
+        TRY(remember_include_file(ctx, inc_path, hash, system, nullptr));
```
jrosdahl: To avoid an extra stat, I suggest adding a `bool require_exists` or similar to `remember_include_file` instead of checking for existence here.

jrosdahl (commented):
I understand the motivation behind the change, but I'm not convinced that it's safe enough as written. The reason for letting `remember_include_file` disable direct mode if the file doesn't exist is to avoid a race:

1. ccache is called
2. ccache executes the preprocessor and saves the output
3. an included header file is removed
4. ccache tries to hash the header file, but it's gone so it won't be added to the manifest
5. when calling ccache again, ccache's direct mode won't detect changes to the header file

If we relax this for MSVC and other compilers that don't emit linemarker flags then we open up for this race.

Side note: if step 3 is not "remove file" but "change file" we detect that here: https://github.com/ccache/ccache/blob/4b11ae89cc8829e2e520701f63267f49e4395535/src/ccache/ccache.cpp#L1365-L1369

PR #1788, opened 2026-08-16
feat: Add support for caching SARIF diagnostics output

### LINKED ISSUE #1614: Add support for msvc's `/experimental:log`

jrosdahl:
Added support for GCC's `-fdiagnostics-format={text,sarif-stderr}` and `-fdiagnostics-{add,set}-output` with text output or SARIF output with an explicit file key. Also added support for MSVC's `/experimental:log` option. For now, only one diagnostics output option is supported at a time.

GCC's `-fdiagnostics-format=sarif-file` and SARIF output without an explicit file key are not supported since the implicit output location depends on the GCC version: GCC 15 uses the CWD while GCC 16 uses the output/dump directory. Ccache currently detects only the compiler type, not its version, so it cannot select the correct location.

Closes: #1614
Closes: #1706


src/ccache/argprocessing.cpp
```diff
@@ -367,6 +373,124 @@ process_profiling_option(const Context& ctx,
...
+    LOG("{} is unsupported (only text and sarif are supported)", arg);
+    return Statistic::unsupported_compiler_option;
+  }
+
+  std::string rewritten_arg{arg};
+  for (const auto key_option : util::Tokenizer(*key_options, ",")) {
+    const auto [key, value] = util::split_once_into_views(key_option, '=');
+    if (key != "file") {
+      continue;
+    }
```
kfessel, line 419: this reads like it would skip any key not 'file' e.g.:

`version=2.2`

and would not add it to the rewritten_arg

https://gcc.gnu.org/onlinedocs/gcc/Diagnostic-Message-Formatting-Options.html#index-SARIF-file-output-sink

kfessel: ahh i see `rewitten_arg` is initialized with the original `arg` and only the file key is then replaced.

jrosdahl: Yep!

PR #1794, opened 2026-08-24
fix: Find GCC 9+ profile data for object files in subdirectories

### LINKED ISSUE #1793: GCC 9+ dir-form -fprofile-use: profile data is never found when object files are in subdirectories
**Problem description**

For GCC 9+, `-fprofile-use=<dir>` compilations are never cached when the
object file path has a directory component: the gcda probe in
`hash_profile_data_file()` joins the (mangled) cwd and the (unmangled)
object base name with `/`, while GCC mangles the whole path with `#`
(documented under `-fprofile-dir`; implemented in gcc/coverage.cc,
coverage_init, which concatenates cwd + relative object path and then
calls mangle_path on the result).

**How to reproduce the problem**

    mkdir build pgo && echo 'int f(int x){return x*2;}' > t.c
    gcc -fprofile-generate=$PWD/pgo -c t.c -o build/t.o   # + link + run
    ls pgo    # -> #tmp#work#build#t.gcda
    CCACHE_LOGFILE=log ccache gcc -fprofile-use=$PWD/pgo -Wno-missing-profile -c t.c -o build/t.o
    ccache -s; grep "profile data" log

(full script: repro-gcda-lookup.sh)

**Actual behavior**

    Uncacheable calls:    2 /   2 (100.0%)      # no_input_file
    Checking for profile data file .../pgo/#tmp#work#build/t.gcda

**Expected behavior**

The probe matches GCC's naming (`#tmp#work#build#t.gcda`), the profile
data is hashed, and the second identical compilation is a cache hit.

**Ccache version:** 4.13.6 (code unchanged on current master)
**Ccache configuration:** defaults (also reproduces with compiler_check = content)
**Compiler:** gcc 13.3.0
**OS:** Linux x86_64

oraluben:
When GCC 9+ writes profile data under `-fprofile-generate=<dir>` /
`-fprofile-dir=<dir>` and the object file path is relative, the `.gcda`
file name is the whole aux path (cwd + relative object path) with every
slash replaced by `#`. This is documented for `-fprofile-dir`:

> In order to prevent the file name clashing, if the object file name is
> not an absolute path, we mangle the absolute path of the
> _sourcename_.gcda file and use it as the file name of a .gcda file.

and implemented in `gcc/coverage.cc` (`coverage_init`): the relative
object path is first concatenated to `getpwd()` and the result is mangled
as a whole (`filename = mangle_path (filename);`).

`hash_profile_data_file()` only mangles the cwd part and joins the object
base name with a real `/`:

    Checking for profile data file .../pgo/#tmp#work#build/test.gcda   <- ccache
                                   .../pgo/#tmp#work#build#test.gcda   <- what GCC wrote

so whenever the object file is not in the current directory — the normal
case for CMake/Ninja/Make builds — the probe never matches, the
compilation is counted as "No input file" and is not cached. The existing
`profiling_gcc*` tests all compile with the object file in the current
directory, where the two spellings coincide, which is why this was never
caught. Observed on a PyTorch PGO pipeline: 2900/3379 compiler calls
uncacheable in the `-fprofile-use=<dir>` build.

The first commit mangles the base name as well. When the object file is
in the current directory the candidate is byte-identical to the previous
spelling, so behavior only changes for the previously broken case. Found
profile data is content-hashed as usual, so modified training data still
invalidates the result (covered by the added test: re-running the trained
binary flips the third compilation back to a miss).

The second commit adds `-fprofile-partial-training` (GCC 10+, listed in
the option summary next to `-fprofile-use` and the already-supported
`-fprofile-values`) to the known simple profiling options. It is an
argument-less code generation modifier ("optimize uncovered parts for
speed instead of size") that does not affect where profile data is read
from or written to; previously it was classified as an unknown profiling
option, making every compilation that uses it uncacheable. It is a common
companion of `-fprofile-use` in PGO release builds.

Both changes come with test cases in `profiling_gcc_10+`; the new
subdirectory test fails without the first commit.

Closes #1793.


jrosdahl:
Thanks!

PR #1796, opened 2026-08-24
fix: Hash C++20 module files found via -fprebuilt-module-path

Ignition:
Clang resolves an import by searching the directory for <module-name>.pcm, so unlike -fmodule-file= the command line does not say which file is read. Those files were not hashed, so a consumer kept its cached object file after the module it imports changed. All module files in the searched directories are now hashed, names as well as contents, since an import resolves to a file name.

-fprebuilt-implicit-modules is counted as "Could not use modules" instead, since Clang looks up implicit modules in a subdirectory layout named by hashes of compiler internals.


Ignition:
@jrosdahl I know others are working on bigger fix #1523, but I'd would like to deliver small incremental fixes providing correctness tests as I go. Please consider this PR.

src/ccache/argprocessing.cpp
```diff
@@ -1186,6 +1187,47 @@ process_option_arg(const Context& ctx,
...
+  }
+
+  if (arg.starts_with("-fprebuilt-module-path=")) {
+    // Clang resolves an import by looking for <dir>/<module-name>.pcm, not
+    // descending into subdirectories. The command line does not say which of
+    // those files the compilation reads, so all of them are hashed. This costs
+    // a hit when an unrelated module file in the same directory changes.
+    constexpr std::string_view prebuilt_module_path_flag =
+      "-fprebuilt-module-path=";
+    const fs::path dir(arg.substr(prebuilt_module_path_flag.size()));
```
jrosdahl: Clang seems to interpret `-fprebuild-module-path=` as  `-fprebuild-module-path=.`, and since the `ec` check on lines 1216-1217 excludes missing directory I guess this could result in a false positive cache hit?

src/ccache/argprocessing.cpp
```diff
@@ -1186,6 +1187,47 @@ process_option_arg(const Context& ctx,
...
+
+    std::vector<fs::path> candidates;
+    std::error_code ec;
+    for (const auto& entry : fs::directory_iterator(dir, ec)) {
+      if (entry.path().extension() == ".pcm") {
```
jrosdahl: Also check `entry.is_regular_file()`?

src/ccache/argprocessing.cpp
```diff
@@ -521,6 +523,44 @@ is_msvc_show_includes_option(std::string_view arg)
...
+    LOG("Failed to read prebuilt module path {}: {}", dir, e.what());
+    return tl::unexpected(Statistic::could_not_use_modules);
+  }
+  // A path that is missing or is not a directory has no module files to hash.
+  // Any other error means that the files read are unknown.
+  if (ec && ec != std::errc::no_such_file_or_directory
+      && ec != std::errc::not_a_directory) {
+    LOG("Failed to read prebuilt module path {}: {}", dir, ec.message());
+    return tl::unexpected(Statistic::could_not_use_modules);
+  }
```
jrosdahl: Could this be done inside the `catch` instead (removing `ec` from `fs::directory_iterator`) to avoid duplicated error handling code?

src/ccache/argprocessing.cpp
```diff
@@ -521,6 +523,44 @@ is_msvc_show_includes_option(std::string_view arg)
...
+// that the hash does not depend on the order the directory is read in, or an
+// error if it cannot be determined which files the compilation reads.
+tl::expected<std::vector<fs::path>, Statistic>
+find_module_files(const fs::path& dir)
+{
+  std::vector<fs::path> module_files;
+  std::error_code ec;
+  try {
+    for (const auto& entry : fs::directory_iterator(dir, ec)) {
+      if (entry.path().extension() != ".pcm") {
```
jrosdahl: To support case insensitive file systems:

```suggestion
      if (util::to_lowercase(util::pstr(entry.path().extension()).str())
          != ".pcm") {
```

jrosdahl:
Thanks!

PR #1799, opened 2026-08-31
ci: Add builds and releases on s390x

luhenry:
This builds up on https://github.com/ccache/ccache/pull/1718 and https://github.com/ccache/ccache/pull/1729 for projects which uses s390x like [`ggml-org/llama.cpp`](https://github.com/ggml-org/llama.cpp/pull/28083) for example.


jrosdahl:
Thanks!

Issue #1797, opened 2026-08-28
CCACHE_BASEDIR does not normalize MSVC environment paths


whosayn:
### Problem description

On Windows, `ccache` hashes `VCToolsInstallDir`, `INCLUDE`, and `EXTERNAL_INCLUDE` using their absolute values.

This prevents cache reuse when otherwise identical builds run from different workspace locations, even when the paths are below `CCACHE_BASEDIR`. This is especially problematic for CI systems where checkout paths may differ between runners.

### How to reproduce the problem

Configure `ccache` with a shared cache and set `CCACHE_BASEDIR` to the workspace root.

Build an `MSVC` project from `workspace A`, for example: `C:\agent\a\project`

with environment variables such as:
```
VCToolsInstallDir=C:\agent\a\project\toolchain
INCLUDE=C:\agent\a\project\include;...
EXTERNAL_INCLUDE=C:\agent\a\project\external;...
```

Build the same project with identical contents and relative layout from `workspace B`: `D:\agent\b\project`

with corresponding environment variables:
```
VCToolsInstallDir=D:\agent\b\project\toolchain
INCLUDE=D:\agent\b\project\include;...
EXTERNAL_INCLUDE=D:\agent\b\project\external;...
```

Check the ccache statistics.

### Actual behavior

The second build misses the cache because the absolute environment paths are hashed verbatim.

`VCToolsInstallDir` affects the common hash, while `INCLUDE` and `EXTERNAL_INCLUDE` affect the manifest key.

### Expected behavior

Paths in `VCToolsInstallDir`, `INCLUDE`, and `EXTERNAL_INCLUDE` that are below `CCACHE_BASEDIR` should be normalized before hashing, consistently with other path-bearing compiler inputs.

### Ccache version

4.14

### Operating system

Windows

### Logs

```shell

```

trustytrojan:
This seems like a duplicate of #1772, minus the `base_dir` fix.

printfdebugging:
https://github.com/ccache/ccache/issues/1800 this one is also similar, though it's about the absolute paths in compiler generated dependency file.

Issue #1800, opened 2026-09-02
base_dir does not apply to the paths in the MSVC /sourceDependencies file


printfdebugging:
On Windows, ccache stores MSVC's `/sourceDependencies` file without applying `base_dir` to the paths inside it, and since MSVC writes them absolute, ccache has to bind the result key to the actual working directory whenever it sees the option. Two build trees on one machine then share no
results at all, even when `base_dir` makes every other input identical.

Rewriting those paths under `base_dir` before the JSON file produced by the `/sourceDependencies` flag is stored fixes that, and the working directory can then be hashed through `base_dir` like every other path. I have both working (will create a PR) on top of 4.14, tested against Collabora Online's document
engine, a LibreOffice fork built with MSVC: 79 minutes for a cold build, then 28 minutes in a second work tree with 99.2% of 10,308 compiles served from the cache the first tree built.

printfdebugging:
https://github.com/ccache/ccache/pull/1801 is the first draft of the fix, not sure what's the best way to fix this, needs review.

jrosdahl:
Not applying `base_dir` to `/sourceDependencies` is intentional since the source dependencies file [always contains absolute paths](https://learn.microsoft.com/en-us/cpp/build/reference/sourcedependencies?view=msvc-170#remarks). What could potentially be done is to rewrite matching paths in the source dep file with a **CWD placeholder** and reverse the operation when producing source dependencies on a cache hit.

For example, given `base_dir=C:\proj` and CWD=`C:\proj\workdir` we could rewrite

```json
{
    "Version": "1.2",
    "Data": {
        "Source": "C:\\proj\\workdir\\example.c",
        "Includes": [
            "C:\\proj\\workdir\\example.h",
            "C:\\proj\\dir2\\hello.h",
            "C:\\another\\file.h",
        ]
}
```

to something like

```json
{
    "Version": "1.2",
    "Data": {
        "Source": "%CCACHE_CWD%\\example.c",
        "Includes": [
            "%CCACHE_CWD%\\example.h",
            "%CCACHE_CWD%\\..\\dir2\\hello.h",
            "C:\\another\\file.h",
        ]
}
```

when stored and convert back to absolute paths for a cache hit, potentially with another CWD.

printfdebugging:
ah, that makes sense. So we still cache it without the absolute paths and write it back to how the compiler would have (or with some other CCACHE_CWD).. i think that would work, will send v2 with that soon..

PR #1801, opened 2026-09-02
fix: Apply base_dir to the paths in MSVC's /sourceDependencies file

printfdebugging:
`base_dir` works by making the compiler's command line relative, and
`depfile::make_paths_relative_in_output_dep` cleans up the absolute paths that
gcc and clang leave in the dependency file.

MSVC resolves every path in the JSON that /sourceDependencies produces to
an absolute one whatever the command line said. ccache replays that file
verbatim, so it mixes the working directory into the cache key whenever it
sees the option, and two work trees on one machine then share nothing at all.
This is the one generated output that `base_dir` was never applied to.

With `base_dir` set the file now holds paths relative to the directory the
compile ran in, so anything reading it from elsewhere has to resolve them.
That is what `base_dir` is for.

Measured on a Windows build of the LibreOffice/Collabora document engine:
79 minutes cold, then 28 minutes in a second work tree with 99.2% of 10,308
compiles served from the first tree's cache.

The rewriter is a scanner tuned to MSVC's output, not a JSON parser: it
handles the doubled backslash because that is the only escape MSVC emits, and
treats a string followed by a colon as a key. I can turn it into a tokenize
and untokenize pair like depfile.cpp if you would rather.

All of it is inert when `base_dir` is not set.

Discussion started in #1797.


jrosdahl:
Thanks for the PR.

As mentioned in https://github.com/ccache/ccache/issues/1800#issuecomment-5654003669 I don't think it's a good idea to produce relative paths in the source dependencies file.

PR #1802, opened 2026-09-04
Direct mode avoid

badumbatish:


badumbatish:
sorry, i meant to open the PR in my own fork for further self review

PR #1803, opened 2026-09-04
Add optional safe direct mode on top of direct mode

badumbatish:
This PR attempts to solve the limitation of direct mode that's been documented since https://ccache.samba.narkive.com/nqHl8wLK/direct-mode-design-bug.

More specifically, it tries to record which include directory could have include which header file (called shadow paths) and try to probe these paths again before reusing the direct key.

In summary, we're just trying to detect ahead of time which changes can influences a preprocessor output, which could make a cache stale silently in direct mode.

Some cases include:
- Ordering of -IincludeA -IincludeB where header.h exists only in includeB in the first run, but later includeA also has header.h
- Ordering of -IincludeA -IincludeB where header.h exists only in includeB in the first run and includeA doesn't exist, but later includeA also has header.h.
- header probing with __has_include(...)
- __has_embed(...)
- headers appearing in CPATH on the second run but not on the first run.

These cases can be solved by relying on:
- using -Wp,-v's result to detect such directories
- scanning the source file to see if it has __has_include() and __has_embed() or not

I'm not sure if this correctly covers all the limitation of ccache direct mode or not but would love to discuss more

Claude Code's used for this PR

### DIFF
--- a/doc/manual.adoc
+++ b/doc/manual.adoc
@@ -1819,9 +1819,13 @@ There is a catch with the direct mode: header files that were used by the
 compiler are recorded, but header files that were *not* used, but would have
 been used if they existed, are not. To mitigate this problem, ccache records
 whether directories specified with `-I` and similar exist at the time of
-compilation, which handles most cases. Still, when ccache checks if a result can
-be taken from the cache, it currently can't check with 100% accuracy if the
-existence of a new header file should invalidate the result. In practice, the
+compilation. For GCC and Clang (but not for MSVC-compatible drivers like
+clang-cl), ccache also asks the preprocessor for its header search directories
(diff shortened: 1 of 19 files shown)


jrosdahl:
Thanks for working on this!

I don't have bandwidth to have a look at this at the moment, but I'll get back to it eventually.

badumbatish:
thanks for the ack Joel, i realized this feature's quite a bit hefty. If it helps maybe I can just do some cherry-picking for the first 4 patches first after the review?

badumbatish:
hi Joel, here's my systematic analysis for covering all the safe direct mode cases: https://hackmd.io/@badumbatish/HyA_WFlYMe/edit

I hope this is helpful in reviewing

PR #1805, opened 2026-09-06
fix: Complete path mapping for relocated Clang PCH builds

janicduplessis:
Depends on #1769. This draft includes its commits unchanged; [review the follow-up diff here](https://github.com/janicduplessis/ccache/compare/052991bd7b5249dbd251eb1a1b7ca79018639c26...portable-pch-path-mapping). The fixes address cache misses, stale dependency paths, and incorrect header validation when moving a Clang PCH build between checkouts.

The setup still requires Clang relocation flags, stable generated PCH inputs, depend mode, and the existing `pch_defines,time_macros` opt-ins. It uses no `base_dir`, so #1661 is not a dependency.

With path mapping enabled:

- Paths in `-Xclang -include-pch -Xclang <path>` must map for hashing while the compiler receives the original arguments.
- Cached Makefile dependencies must map back to the receiving checkout, or Ninja sees references to the unavailable source tree.
- `symlink/../header.h` must retain its filesystem meaning. The lexical normalization added in [052991bd](https://github.com/ccache/ccache/commit/052991bd7b5249dbd251eb1a1b7ca79018639c26) can make the manifest validate a different, existing header and return stale objects after the real header changes.

The new integration suite proves PCH/object hits in B with A unavailable, compiles an edited consumer against the restored PCH, and invalidates same-size header changes, including a lookup through a symlink. Dependency escaping and Windows path comparisons have unit coverage. Prefix matching and suffix extraction both use raw path components, preserving `.` and `..` while retaining case-insensitive matching on Windows. The same integration test fails against unmodified #1769 because B adds a cache miss.

Validation: unit tests and the existing base, basedir, depend, and direct suites on macOS; the new suite passes with Android NDK 27.1 / Clang 18 targeting ARM64. Apple Clang 21 fails the compiler-only relocation probe, so the suite reports a skip there.

To run the new regression from a configured build directory:

```sh
CC="/path/to/ndk/toolchains/llvm/prebuilt/darwin-x86_64/bin/clang --target=aarch64-linux-android24" \
  ctest -R '^test.path_mapping_pch$' --output-on-failure
```

The usual direct-mode limitation for newly appearing shadow headers remains. This draft does not establish cross-machine or cross-compiler portability.


trustytrojan:
The reason I made [`052991bd`](https://github.com/ccache/ccache/commit/052991bd7b5249dbd251eb1a1b7ca79018639c26) is because [`path_starts_with` also normalizes its input](https://github.com/ccache/ccache/blob/052991bd7b5249dbd251eb1a1b7ca79018639c26/src/ccache/util/path.cpp#L255-L256), and if `remove_leading_components` receives an unnormalized version of the `path` variable, too many or too little path components get removed.

janicduplessis:
Hey @trustytrojan thanks for having a look, I mostly just wanted to see if possible to get ccache working across worktrees when using pch and found your PR, but this was the changes needed to get it working for my use case.

This is more of a proof of concept, if anyone wants to pickup the work it would be great since I don’t have much knowledge of this codebase.

PR #1810, opened 2026-09-14
fix: fix -fdump-ipa-clones for different outdir

### LINKED ISSUE #1809: Incorrect location of IPA clones file
### Problem description

I'm trying to build OpenSUSE kernel and stats shows that most calls are uncacheable:
```
Uncacheable calls:              24703 / 32805 (75.30%)
  Compiler output file missing: 20954 / 24703 (84.82%)
```

Running with debug shows that it's due to IPA clone files (most likely, since logging seems out of order)

```
[2026-09-13T06:30:20.208157 853958] Working directory: /usr/src/packages/BUILD/kernel-default-7.2.3-build/kernel-default-7.2.3/linux-7.2/linux-obj
[2026-09-13T06:30:21.809822 853958] IPA clones file ../kernel/user.c.000i.ipa-clones missing
...
Executing /usr/bin/gcc ...  -c -o kernel/user.o ../kernel/user.c
```

#1449 introduced support for `-fdump-ipa-clones` option and I think it calculates path incorrectly. The file is produced in the output directory

```
# readlink -f `fd user.c.000i.ipa-clones`
/usr/src/packages/BUILD/kernel-default-7.2.3-build/kernel-default-7.2.3/linux-7.2/linux-obj/kernel/user.c.000i.ipa-clones
```

not the source one (i.e. one level up).

GCC docs seem to agree:
> The file name is generated by appending suffix ipa-clones to the source file name, and the file is created in the same directory as the output file.

Which is very confusing, since output file and input file names may be different, not only by extension...

I tried to find how gcc produces that name 😓. It seems it is buried between different layers and can be totally overridden :/

### Ccache version

ccache version 4.14

### Operating system

Linux

jirislaby:
Fixes #1809.


jrosdahl:
Thanks!

